PR #27, opened 2019-08-30
Working on ceilDiv

### LINKED ISSUE #25: ceilDiv with non-integer types gives wrong results
It is clear both from the template parameter name and from the documentation that `IntType` must be an integral type, but it is still possible to use it with floating types with wrong results (not truncated, and even if the user truncates it, it can lead to wrong values).

e.g. `ceilDiv(3.0, 2.1)`
the expected result is 2, but it gives ~1.4, that even truncated gives 1.

albestro:
Close #25 

- Add constraint on integer types as described in issue #25.
- Enable usage of different integer types for numerator and denumerator.
- Change doxygen configuration allowing to make code easier for generating configuration (hides complex std::enable_if construct on return type).
- Update ceilDiv documentation with new features and fixes wrong precondition (den must be strictly positive)


include/dlaf/util_math.h
```diff
@@ -10,17 +10,29 @@
...
 ///
-/// @tparam IntType has to be an integer type
-/// @pre @a num >= 0 and @a den >= 0
-template <class IntType>
-constexpr IntType ceilDiv(IntType num, IntType den) {
+/// @pre @a num >= 0 and @a den > 0.
+/// @pre @a num + @a den - 1 must not overflow @a IntNumType max value
+template <typename IntNumType, typename IntDenType>
+constexpr
+RETURN_TYPE_IF(IntNumType, std::is_integral<IntNumType>::value && std::is_integral<IntDenType>::value)
```
rasolca: Why return type is `IntNumType`? What about `auto`?

include/dlaf/util_macros.h
```diff
@@ -0,0 +1,23 @@
...
+#include <type_traits>
+
+/// @file
+
+/// Allows to hide `std::enable_if` expressions from generated documentation.
+/// If `BooleanExpr` is true it results in `Type`, otherwise it is empty (useful for exploiting SFINAE).
+#ifdef DLAF_DOXYGEN
+#define RETURN_TYPE_IF(Type, BooleanExpr) Type
+#else
+#define RETURN_TYPE_IF(Type, BooleanExpr) std::enable_if_t<BooleanExpr, Type>
```
rasolca: Does not always work. E.g.
```
template <class T>
RETURN_TYPE_IF(std::array<T, 2>, std::is_integral<T>::value) foo(T a);
```
(Parenthesis are not an option here.)

include/dlaf/util_math.h
albestro: Here SFINAE is used with `std::enable_if` to constraint parameters to be integers.

We cannot use `auto` because it cannot be a template parameter, in fact this is not allowed by the language
```cpp
std::enable_if_t<std::is_integral<IntNumType>::value && std::is_integral<IntDenType>::value, auto>
```

include/dlaf/util_macros.h
albestro: As discussed, we opt for a simple solution: keep manually separated the two signatures.

Just for reference, there is a solution for this problem with comma and macros when parenthesis cannot be used. (https://stackoverflow.com/a/13842784)

albestro:
There is still an open question by @rasolca about signed vs unsigned integers and cast performance.

albestro:
> There is still an open question by @rasolca about signed vs unsigned integers and cast performance.

It has been decided with @rasolca that for the moment we opt for having the same type for both numerator and denominator; the code is constrained to not allow wrong usages (e.g. float type) and it addresses correctly the current need of having a ceiling division function for values of the same type.

include/dlaf/util_math.h
```diff
@@ -10,19 +10,34 @@
...
-/// @tparam IntType has to be an integer type
-/// @pre @a num >= 0 and @a den >= 0
-template <class IntType>
-constexpr IntType ceilDiv(IntType num, IntType den) {
+/// @pre @a num >= 0 and @a den > 0.
+/// @pre @a num + @a den - 1 must not overflow @a IntType max value
+template <typename IntType>
+constexpr auto ceilDiv(const IntType num, const IntType den)
+#ifdef DLAF_DOXYGEN
+    ;
```
rasolca: I think that for readability `#if` macros should not contain just a part of a line of code.

For doxygen I suggest to use
```
#ifdef DLAF_DOXYGEN
  full function definition (including templates) to be shown in doxygen
#else
  function definition or declaration to be compiled.
#endif
```
Also modify the doxygen section of the coding style to include this hint.

PR #28, opened 2019-08-30
Configuring deploy step

### LINKED ISSUE #24: Configure CMake install target
Configure the install/deploy step

- configure CMake install target
- set up for best interoperability with other CMake projects (modern way: export targets, ...)
- have an option for install tests

albestro:
Close #24 

- configure CMake for library installation
- configure the project to introduce CMake interoperability and usage as a dependency from other projects
- add option `DLAF_INSTALL_TESTS` for installing tests along with the library

It required a workaround to MKL and LAPACK scripts to have the same build configuration at install time (e.g. link line).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,18 +78,36 @@ target_include_directories(DLAF INTERFACE
   $<INSTALL_INTERFACE:include>
 )
 
+# ----- DEPLOY
+include(GNUInstallDirs)
+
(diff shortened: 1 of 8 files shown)


albestro:
@rasolca 
I have not managed the installation of the documentation. 

Do we want to install it along with the library if built?

rasolca:
@albestro Yes, it should be installed if built.

PR #33, opened 2019-09-03
Added format test to CI.

### LINKED ISSUE #30: Add clang-format test
Add a test to check if the format of the sources (h, ipp, cpp) is correct. (i.e. clang-format was used)

rasolca:
Added the format test.

The script re-formats the header and source files (.cpp, .h or .ipp) which has been modified
using clang-format, and removes trailing whitespaces from other modified files.

Closes #30.

### DIFF
--- a/ci/check_format.sbatch
+++ b/ci/check_format.sbatch
@@ -1,11 +1,29 @@
 #!/bin/bash -l
+#SBATCH --job-name=DLA-Future-build-doc
+#SBATCH --nodes=1
+#SBATCH --constraint=mc
+#SBATCH --partition=cscsci
+#SBATCH --time=00:05:00
+#SBATCH --output=check_format.out.txt
+#SBATCH --error=check_format.out.txt
+
+set -e
+
+module purge
+module load modules
+module load craype
+module load cray-mpich
+module load slurm
+module load xalt
+module load PrgEnv-gnu
+module load daint-mc
 module use /apps/daint/SSL/software/spack-current/share/spack/modules/cray-cnl6-broadwell
 module load llvm/8.0.0-gcc-6.2.0-cray-cnl6-broadwell-release
 
 git config --add remote.origin.fetch +refs/heads/master:refs/remotes/origin/master
 git fetch --no-tags
 
-set -ex
+set -x
 
 FILES_CHANGED=`git diff --name-only origin/master`
 
--- a/ci/check_format.sbatch
+++ b/ci/check_format.sbatch
@@ -33,9 +33,26 @@ do
     *.cpp|*.h|*.hpp)
       clang-format -i --style=file $FILE
       ;;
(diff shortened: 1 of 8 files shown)


ci/check_format.sbatch
```diff
@@ -0,0 +1,43 @@
...
+FILES_CHANGED=`git diff --name-only origin/master`
+
+for FILE in $FILES_CHANGED
+do
+  case $FILE in
+    *.cpp|*.h|*.hpp)
+      clang-format -i --style=file $FILE
+      ;;
+    *)
+      # remove trailing whitespaces
```
albestro: From the coding style guide:
> Each file should contain an empty line at the end of the file

I would check for missing newline at the end of files.

Moreover, from documentation it is not clear if multiple newlines are allowed. In case, I would clearly explicit it and here I would check about it too.

rasolca: How would you do that?

albestro:
```bash
# count number of trailing newlines
awk '$0 == "" {is = NR; while (1) { if (getline tmp == 0) { exit NR - is + 1; } else if (tmp != "") { break; } } }' $FILE
n_newlines=$?

# change the file just if it is needed
if [[ $n_newlines -eq 0 ]]
then
  echo "" >> $FILE
elif [[ $n_newlines -gt 1 ]]
then
  truncate -s -$(($n_newlines-1)) $FILE
fi
```

albestro: Due to the interpretation of \nEOF as EOF by linux tools, an hack to count the number of newlines at EOF can be to work at binary level with

```bash
cat $FILE | xxd -p | tr -d '\n' | egrep -o "(0a)+$" | grep -o 0a | wc -l
```
1. read the file
2. `xxd` with `-p` option print the binary hex of the file without addresses (but uses newline for layout purposes)
3. `tr` transforms the layout view to a single-line hex string
4. `egrep` extract a sequence of "0a" (hex ASCII code for newline) at the end of the string
5. grep extract "0a" matches from the sequence and prints one occurrence per line
6. count the occurences (lines), which is the number of newlines at the end of file

As before, it should run after the trailing spaces sanitizer and its result can be used to sanitize newlines at the end of files (just touching files that needs sanitization).

ci/check_format.sbatch
```diff
@@ -0,0 +1,60 @@
...
+module load llvm/8.0.0-gcc-6.2.0-cray-cnl6-broadwell-release
+
+git config --add remote.origin.fetch +refs/heads/master:refs/remotes/origin/master
+git fetch --no-tags
+
+set -x
+
+FILES_CHANGED=`git diff --name-only origin/master`
+
+for FILE in $FILES_CHANGED
```
albestro: Just as a reminder: every file tracked by git that falls in the `*` case will be subject to sanitization of trailing spaces and ending newlines.

PR #38, opened 2019-09-09
Tile can set make a future ready on destruction.

### LINKED ISSUE #13: Implement Tile Wrapper Object
The wrapper object is templated on the type of the object it contains.
It is used to automatically set the value of the new future of a tile of the matrix
when the operation is completed.

It contains an object of type `T` and a promise of type `hpx::promise<T>`.
On destruction the promise is set with the value of the object.

see `misc/example/m.h` for reference.

rasolca:
Instead of creating a Wrapper class to manage the setting of the Future on destruction,
this mechanism has been added to the class `Tile` directly.

Closes #13.


include/dlaf/tile.h
```diff
@@ -53,19 +59,28 @@ class Tile {
...
+      : m_(rhs.m_), n_(rhs.n_), memory_view_(std::move(rhs.memory_view_)), ld_(rhs.ld_),
+        p_(std::move(rhs.p_)) {
     rhs.m_ = 0;
     rhs.n_ = 0;
     rhs.ld_ = 1;
   }
 
+  ~Tile() {
+    if (p_) {
+      auto p = std::move(p_);
```
albestro: Is it really needed to move it?

In my knowledge it is not and, in case, I would remove it (IMHO it does not improve readability and moreover we can "save" a couple of lines).

misc/example/m.h
```diff
@@ -3,92 +3,55 @@
...
+      auto p = std::move(p_);
+      p->set_value(Tile<ncEl>(ptr_));
     }
   }
 
-  Wrapper(const Wrapper&) = delete;
-  Wrapper(Wrapper&& rhs) : t_(std::move(rhs.t_)), p_(std::move(rhs.p_)), valid_(true) {
-    rhs.valid_ = false;
+  template <class T = El>
+  std::enable_if_t<std::is_same<T, El>::value && !std::is_const<T>::value, Tile&> setPromise(
```
albestro: Do we really want to expose this in the public interface?

I think that `setPromise` should be called with caution, so I would go for managing it internally without allowing anyone who has a `Tile` to be able to call it: if it is called multiple times on the same `Tile` it is a problem.

misc/example/m.h
```diff
@@ -3,92 +3,55 @@
...
+  template <class T = El>
+  Tile(std::enable_if_t<std::is_same<T, El>::value && std::is_const<T>::value, Tile<ncEl>&&> rhs)
+      : ptr_(rhs.ptr_), p_(std::move(rhs.p_)) {}
 
-  ~Wrapper() {
-    if (valid_) {
-      p_.set_value(std::move(t_));
+  ~Tile() {
+    if (p_) {
+      auto p = std::move(p_);
```
albestro: same (minor) comment as in `dlaf/tile.h`

rasolca:
@albestro the example in the `misc` directory is just a proof of concept.
It shows with a simplified API how the matrix class should work, and therefore it doesn't require a review.
Here the example has been updated to reflect the fact that we decided not to use the `Wrapper` object.

misc/example/test.cpp
```diff
@@ -90,9 +86,9 @@ void foo() {
 
   // print the value of A and B.
   for (std::size_t i = 0; i < 4; ++i) {
-    int ai = ma(i).get().get()();
+    int ai = ma(i).get()();
```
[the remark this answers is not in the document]
rasolca: This is just used for the printout in the example with the simplified API, and will have no 1to1 match in the final API.

Anyway, in my opinion your suggested interface is more problematic since it hides the synchronization of `.get()`.

test/unit/CMakeLists.txt
```diff
@@ -9,6 +9,6 @@
 #
 
 DLAF_addTest(test_util_math SOURCES test_util_math.cpp USE_GTEST_MAIN)
-DLAF_addTest(test_tile SOURCES test_tile.cpp USE_GTEST_MAIN)
+DLAF_addTest(test_tile SOURCES test_tile.cpp LIBRARIES PRIVATE DLAF_gtest_hpx_main)
```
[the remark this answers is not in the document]
rasolca: This may be another task.

misc/example/m.h
```diff
@@ -3,92 +3,55 @@
 #include "out.h"
 
 #include <array>
+#include <cassert>
 #include <hpx/hpx.hpp>
 #include <iostream>
 #include <memory>
```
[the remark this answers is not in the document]
rasolca: Include order is one of the most customizable options in clang-format.
Current clang-format style has been designed with this constraint:
expect C / C++ headers in `<>` and other headers in "". 

I propose to open an issue to discuss about this.

include/dlaf/tile.h
rasolca: Not anymore, can be removed.

include/dlaf/tile.h
```diff
@@ -116,11 +132,23 @@ class Tile {
     return ld_;
   }
 
+  /// @brief Sets the promise to which this Tile will be moved on destruction.
+  /// @pre The promise should not be already set.
+  template <class U = T>
+  std::enable_if_t<!std::is_const<U>::value && std::is_same<T, U>::value, Tile&> setPromise(
```
rasolca: @albestro commented:
> Do we really want to expose this in the public interface?
> 
> I think that `setPromise` should be called with caution, so I would go for managing it internally without allowing anyone who has a `Tile` to be able to call it: if it is called multiple times on the same `Tile` it is a problem.

Two options are available here:
1. make it private and make the matrix class `friend`
2. make it public

Besides some small issues Option 1 has a huge problem.
It expose all the private members (which do not ensure constness of the data in case of `Tile<const T>`) to the friend classes.

To make it more safe the assertion can be replaced by a check and throw, to have the check also in Release mode.

PR #44, opened 2019-09-19
[OLD] Introduce communication module and Communicator

### LINKED ISSUE #17: Implement Communicators 1D and 2D
Design and implement communicators.

The prototype interface is basic and should be redesigned.

To create the row and column communicators it is better to use `MPI_Comm_split`. See `misc/code.(pdf|pptx)`

albestro:
A starting point for implementing communication module `dlaf::comm`.

The new module has code separation between public and private interface, which required a restructuration of the library.

Introduce:
- Communicator: an MPI-compatible wrapper for MPI_Comm with the option to automatically manage the resource;
- CommunicatorGrid: a 2D Communicator with row/column isolated Communicators, with specified rank ordering (RowMajor or ColumnMajor);

In addition:
- Index2D type for grid coordinates;
- Provide gtest_mpi_main, equivalent to gtest_main but with MPI initialization;
- Extends DLAF_AddTest CMake function to accept MPI_OVERSUBSCRIBE and allowing to use the correct gtest_main with MPI tests;
- MPI header wrapper: it helps hiding MPI library warnings and provide an `MPI_CALL` wrapper function;
- Update doxygen configuration file;

Closes #17


albestro:
@rasolca 

About warnings:
- They must be defined and managed for different compilers;
- Compiler differences must be managed also in the MPI header wrapper;

albestro:
@rasolca do you think we can delete this now that #49 has been merged?

PR #49, opened 2019-09-23
Introduce communication module and Communicator

### LINKED ISSUE #17: Implement Communicators 1D and 2D

albestro:
Closes #17 

A starting point for implementing communication module `dlaf::comm`

Introduce:
- Communicator: an MPI-compatible wrapper for MPI_Comm with the option to automatically manage the resource;
- CommunicatorGrid: a 2D Communicator with row/column isolated Communicators, with specified rank ordering (RowMajor or ColumnMajor);

In addition:
- Index2D type for grid coordinates;
- MPI header wrapper: it helps to hide MPI library warnings and provide an MPI_CALL wrapper function;


include/dlaf/communication/communicator.h
```diff
@@ -0,0 +1,81 @@
...
+  /// @brief Return the rank of the current process in the Communicator
+  int rank() const noexcept;
+  /// @brief Return the number of ranks in the Communicator
+  int size() const noexcept;
+
+protected:
+  /// @brief Release the underlying communicator
+  ///
+  /// It calls MPI_Comm_free (for additional info see MPI documentation)
+  void release() noexcept(false);
```
rasolca: This should be private. (or embedded directly in the destructor).

include/dlaf/communication/communicator_grid.h
```diff
@@ -0,0 +1,75 @@
...
+/// if a grid size bigger that overfit the available number of ranks is specified, it will raise an
+/// exception.
+///
+/// CommunicatorGrid must be destroyed before calling MPI_Finalize, to allow it releasing resources.
+class CommunicatorGrid {
+public:
+  /// @brief Create a communicator grid @p rows x @p cols with given @p ordering
+  /// @param comm must be valid during construction
+  CommunicatorGrid(Communicator comm, int rows, int cols,
+                   common::LeadingDimension ordering = common::LeadingDimension::Row) noexcept(false);
```
rasolca: I think is better not to give a default value for the communicator ordering.

include/dlaf/communication/communicator_grid.h
```diff
@@ -0,0 +1,75 @@
...
+  /// @param comm must be valid during construction
+  CommunicatorGrid(Communicator comm, const std::array<int, 2>& size,
+                   common::LeadingDimension ordering = common::LeadingDimension::Row) noexcept(false);
+
+  /// @brief Return the rank of the current process in the CommunicatorGrid
+  /// @return a common::Index2D representing the position in the grid
+  common::Index2D rank() const noexcept;
+
+  /// @brief Return the number of rows in the grid
+  int rows() const noexcept;
```
rasolca: Why the rank is returned in 2D and the size in 1D?
Moreover `row` and `rows` may not be so readable.

I would use `size()` -> 2D comm size, `rowCommunicator()` and `colCommunicator()` (to avoid the confusion with the row and col index).

test/unit/communication/test_communicator_grid.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+INSTANTIATE_TEST_CASE_P(ConstructorWithArray, CommunicatorGridTest,
+                        ::testing::Values(LeadingDimension::Row, LeadingDimension::Column));
+
+TEST_P(CommunicatorGridTest, ConstructorOverfit) {
+  Communicator world(MPI_COMM_WORLD);
+
+  EXPECT_THROW(CommunicatorGrid grid(world, NUM_MPI_RANKS, 2, GetParam()), std::invalid_argument);
+}
```
rasolca: A good test would test it with a grid of `NUM_MPI_RANKS+1`.

include/dlaf/communication/communicator.h
```diff
@@ -0,0 +1,81 @@
...
+public:
+  /// Tag to give to constructor in order to give MPI_Comm ownership to Communicator
+  struct managed {};
+
+  /// @brief Create a NULL Communicator (i.e. MPI_COMM_NULL)
+  Communicator() noexcept(false);
+
+  /// @brief Wrap an MPI_Comm into a Communicator
+  ///
+  /// The validity of the wrapped MPI_Comm must be granted and managed by the user, otherwise an UB occurs.
```
rasolca: This sentence is not clear.

include/dlaf/communication/communicator.h
```diff
@@ -0,0 +1,81 @@
...
+  /// @param mpi_communicator MPI_Comm to wrap.
+  Communicator(MPI_Comm mpi_communicator) noexcept(false);
+
+  /// @brief Wrap and manage an MPI_Comm into a Communicator
+  ///
+  /// The management of the underlying MPI_Comm is up to this Communicator and not to the user
```
rasolca: This is not clear.

include/dlaf/mpi_header.h
```diff
@@ -0,0 +1,43 @@
...
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+
+#if __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
+#pragma GCC diagnostic ignored "-Wsign-conversion"
```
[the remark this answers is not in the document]
albestro: I don't know if `-isystem` is the right option for this problem, moreover here it can not be applied easily.

In CMake the flag `-isystem` can be enabled in the directive `target_include_directories` with the option `SYSTEM`; Using FindMPI, which is highly suggested, we are not directly managing MPI include directories and so we cannot specify the SYSTEM option, since we only get a CMake target already populated.

It is possible to argue with this decision, but I think this is due to the fact that the `SYSTEM` option is just a workaround for the problem and not the right solution.

include/dlaf/mpi_header.h
```diff
@@ -0,0 +1,43 @@
...
+#include <mpi.h>
+
+#include <iostream>
+
+namespace dlaf {
+namespace comm {
+namespace mpi {
+
+constexpr MPI_Comm NULL_COMMUNICATOR = MPI_COMM_NULL;
+constexpr MPI_Comm WORLD_COMMUNICATOR = MPI_COMM_WORLD;
```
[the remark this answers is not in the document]
albestro: In this warning-free zone, some constants are re-defined in a safer way that can be used in our library code without generating warnings.

include/dlaf/communication/communicator.h
albestro: I can not get your point of having it `private` instead of `protected`, but it is true that it can be embedded in the destructor.

Moreover, the declaration you are pointing out is a dangling one, so I removed it and I used your comment in the real instance in `CommunicatorImpl`.

test/unit/communication/test_communicator.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  EXPECT_EQ(comm_null.size(), 0);
+  EXPECT_EQ(comm_null.rank(), MPI_UNDEFINED);
+}
+
+TEST_F(CommunicatorTest, Rank) {
+  if (rankInGroup()) {
+    // check new communicator size consistency and correctness
+    EXPECT_LT(odd_comm.size(), NUM_MPI_RANKS);
+    EXPECT_EQ(odd_comm.size(), (NUM_MPI_RANKS + 1) / 2);
```
rasolca: Shuldn't it be `ranks/2`

test/unit/communication/test_communicator.cpp
```diff
@@ -0,0 +1,153 @@
...
+  EXPECT_EQ(MPI_COMM_NULL, comm_null);
+
+  EXPECT_EQ(comm_null.size(), 0);
+  EXPECT_EQ(comm_null.rank(), MPI_UNDEFINED);
+}
+
+TEST_F(CommunicatorTest, Rank) {
+  if (rankInGroup()) {
+    // check new communicator size consistency and correctness
+    EXPECT_LT(odd_comm.size(), NUM_MPI_RANKS);
```
rasolca: This test is superfluous. By definition `i/2 < i` for `i>=1`

test/unit/communication/test_communicator.cpp
```diff
@@ -0,0 +1,153 @@
...
+    EXPECT_EQ(odd_comm.size(), (NUM_MPI_RANKS + 1) / 2);
+
+    // check rank consistency
+    EXPECT_LT(odd_comm.rank(), odd_comm.size());
```
rasolca: Superfluous (see above)

test/unit/communication/test_communicator.cpp
```diff
@@ -0,0 +1,153 @@
...
+    EXPECT_GE(odd_comm.rank(), 0);
+
+    // check rank correctness
+    EXPECT_NE(odd_comm.rank(), world.rank());
```
rasolca: Superfluous

test/unit/communication/test_communicator_grid.cpp
```diff
@@ -0,0 +1,209 @@
...
+    EXPECT_NE(incomplete_grid.col(), MPI_COMM_NULL);
+
+    auto coords =
+        dlaf::common::computeCoords<CommunicatorGrid::Index2D>(GetParam(), world.rank(), grid_dims);
+
+    EXPECT_EQ(coords.row(), incomplete_grid.rank().row());
+    EXPECT_EQ(coords.col(), incomplete_grid.rank().col());
+
+    EXPECT_EQ(coords.col(), incomplete_grid.row().rank());
+    EXPECT_EQ(coords.row(), incomplete_grid.col().rank());
```
rasolca: too many thing called in the same way (cfr. with ll. 158-9)

include/dlaf/mpi_header.h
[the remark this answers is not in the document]
rasolca: Anyway warnings are not set yet. This should be addressed in another PR.

include/dlaf/mpi_header.h
```diff
@@ -0,0 +1,43 @@
...
+
+}
+}
+}
+
+#define MPI_CALL(x)                                                \
+  {                                                                \
+    auto _ = x;                                                    \
```
rasolca: `_` is not a style conforming variable name

include/dlaf/mpi_header.h
```diff
@@ -0,0 +1,43 @@
...
+    if (MPI_SUCCESS != _)                                          \
+      std::cerr << "MPI ERROR [" << _ << "]: " << #x << std::endl; \
```
rasolca: Is it needed to have all the line of code which use MPI into the code? In most of the usage cases the MPI error handler will kill the application, so this piece of code would not be even called.
What about filename and line?

include/dlaf/communication/communicator_grid.h
rasolca: This has not been fully addressed.

include/dlaf/mpi_header.h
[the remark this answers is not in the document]
albestro: I'll reopen this in a separate PR.

include/dlaf/communication/communicator_grid.h
```diff
@@ -0,0 +1,80 @@
...
+/// exception.
+///
+/// CommunicatorGrid must be destroyed before calling MPI_Finalize, to allow it releasing resources.
+class CommunicatorGrid {
+  struct TAG_MPI;
+  using IndexType = int;
+
+public:
+  using Index2D = common::Index2D<IndexType, TAG_MPI>;
+  using Size2D = common::Size2D<IndexType, TAG_MPI>;
```
rasolca: I would move the index and the size out of the class.
(Index that are "shared" between different classes should be defined outside the classes, therefore, to be consistent, we define all indices outside classes)

PR #54, opened 2019-10-03
Introduce size_t arithmetic

### LINKED ISSUE #53: Need basic operations for working with indexes in safety
Working with indexes often requires to perform arithmetic operations, mainly additions and multiplications, whose results have to be stored in `std::size_t`. For instance, this result can be used to refer memory positions for which an `int` is not enough.

Mixing different integer types in arithmetic operations may lead to wrong results, e.g. due to change of sign or due to overflow. Compiler warnings may help in discovering these kinds of problems, but in order to be able to work easily with them, it is useful to have a way to safely perform this basic operations.

albestro:
Close #53 

Add `sum` and `mul` operators for `size_t` arithmetic.

### DIFF
--- a/test/unit/test_util_math.cpp
+++ b/test/unit/test_util_math.cpp
@@ -62,8 +62,19 @@ TYPED_TEST(MathUtilTest, CeilDivType) {
 TYPED_TEST(MathUtilTest, SizeTArithmetic_Sum) {
   using Type = TypeParam;
 
-  Type a = std::numeric_limits<Type>::max();
-  Type b = std::numeric_limits<Type>::max();
+  constexpr auto type_size = sizeof(Type);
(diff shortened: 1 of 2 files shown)


test/unit/test_util_math.cpp
```diff
@@ -58,3 +58,73 @@ TYPED_TEST(MathUtilTest, CeilDivType) {
...
+  constexpr auto arithmetic_size = sizeof(std::size_t);
+
+  Type a, b;
+
+  if (type_size < arithmetic_size) {
+    a = std::numeric_limits<Type>::max();
+    b = std::numeric_limits<Type>::max();
+  }
+  else {
+    a = std::numeric_limits<std::size_t>::max() / 2;
```
rasolca: What about `a = static_cast<Type>("max" / 2);`? (and similar for others)

Issue #16, opened 2019-07-24
Implement Distributed Matrix of futures of tiles.


rasolca:
See `misc/code.(pdf|pptx)` and `misc/example/m.h` for details.

PR #59, opened 2019-10-04
LayoutInfo and MatrixBase objects

rasolca:
`LayoutInfo`:
 - object which contains the information about the local matrix layout from which the matrix will be built.

`MatrixBase`:
- contains the information about the sizes of the matrix
- It will be modified later to contain also the information of the distribution.

### DIFF
--- a/include/dlaf/matrix/layout_info.h
+++ b/include/dlaf/matrix/layout_info.h
@@ -9,6 +9,7 @@
 //
 
 #pragma once
+#include <cassert>
 #include "dlaf/matrix/index.h"
 #include "dlaf/util_math.h"
 
@@ -19,22 +20,16 @@ public:
   LayoutInfo(GlobalElementSize size, TileElementSize block_size, SizeType tile_ld,
              std::size_t tile_offset_row, std::size_t tile_offset_col);
 
-  std::size_t tileOffset(LocalTileIndex index) const {
-    // TODO checks
+  /// @brief Returns the position of the first element of the @p index tile.
+  /// @pre 0 < @p index.row() < nrTiles().rows()
+  /// @pre 0 < @p index.col() < nrTiles().cols()
+  std::size_t tileOffset(const LocalTileIndex index) const {
+    assert(index.isValid() && index.isIn(nr_tiles_));
     return index.row() * tile_offset_row_ + index.col() * tile_offset_col_;
   }
 
-  std::size_t minMemSize() const noexcept {
-    if (size_.rows() == 0 || size_.cols() == 0) {
-      return 0;
-    }
-
-    SizeType last_rows = size_.rows() - block_size_.rows() * (nr_tiles_.rows() - 1);
-    SizeType last_cols = size_.cols() - block_size_.cols() * (nr_tiles_.cols() - 1);
-
-    return tileOffset({nr_tiles_.rows() - 1, nr_tiles_.cols() - 1}) +
-           static_cast<std::size_t>(ld_tile_) * static_cast<std::size_t>(last_cols - 1) + last_rows;
-  }
+  /// @brief Returns the minimum number of elements that are needed to fit a matrix with the given layout.
+  std::size_t minMemSize() const noexcept;
 
   GlobalElementSize size() const noexcept {
     return size_;
@@ -53,6 +48,7 @@ public:
   }
 
 private:
+  /// @brief Returns the minimum number of elements that are needed to fit a tile with the given size.
   std::size_t minTileSize(TileElementSize size) const noexcept;
 
   GlobalElementSize size_;
@@ -64,12 +60,14 @@ private:
   std::size_t tile_offset_col_;
 };
 
+/// @brief Returns LayoutInfo for a column major matrix.
 inline LayoutInfo ColMajorLayout(GlobalElementSize size, TileElementSize block_size, SizeType ld) {
   using util::size_t::mul;
   return LayoutInfo(size, block_size, ld, static_cast<std::size_t>(block_size.rows()),
                     mul(block_size.cols(), ld));
 }
 
+/// @brief Returns LayoutInfo for a matrix which use the tile layout.
 inline LayoutInfo TileLayout(GlobalElementSize size, TileElementSize block_size, SizeType ld_tile,
                              SizeType tiles_per_col) {
   using util::size_t::mul;
--- a/include/dlaf/matrix/layout_info.h
+++ b/include/dlaf/matrix/layout_info.h
@@ -1,0 +1,80 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2019, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+#include "dlaf/matrix/index.h"
+#include "dlaf/util_math.h"
+
+namespace dlaf {
+namespace matrix {
+class LayoutInfo {
+public:
+  LayoutInfo(GlobalElementSize size, TileElementSize block_size, SizeType tile_ld,
+             std::size_t tile_offset_row, std::size_t tile_offset_col);
+
+  std::size_t tileOffset(LocalTileIndex index) const {
+    // TODO checks
+    return index.row() * tile_offset_row_ + index.col() * tile_offset_col_;
+  }
+
+  std::size_t minMemSize() const noexcept {
+    if (size_.rows() == 0 || size_.cols() == 0) {
+      return 0;
+    }
+
+    SizeType last_rows = size_.rows() - block_size_.rows() * (nr_tiles_.rows() - 1);
+    SizeType last_cols = size_.cols() - block_size_.cols() * (nr_tiles_.cols() - 1);
+
+    return tileOffset({nr_tiles_.rows() - 1, nr_tiles_.cols() - 1}) +
+           static_cast<std::size_t>(ld_tile_) * static_cast<std::size_t>(last_cols - 1) + last_rows;
+  }
+
+  GlobalElementSize size() const noexcept {
+    return size_;
+  }
+
+  LocalTileSize nrTiles() const noexcept {
+    return nr_tiles_;
+  }
+
+  TileElementSize blockSize() const noexcept {
+    return block_size_;
+  }
+
+  SizeType ldTile() const noexcept {
+    return ld_tile_;
+  }
+
+private:
+  std::size_t minTileSize(TileElementSize size) const noexcept;
+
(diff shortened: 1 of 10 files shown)


rasolca:
Preliminary work for #16.

include/dlaf/matrix_base.h
```diff
@@ -0,0 +1,67 @@
...
+  TileElementSize blockSize() const noexcept {
+    return block_size_;
+  }
+
+private:
+  GlobalElementSize size_;
+  LocalTileSize nr_tiles_;
+  TileElementSize block_size_;
+
+  // id and size rank_src in 2D grid.
```
albestro: Is it a dangling comment? Otherwise, I'm not getting it.

src/matrix/layout_info.cpp
```diff
@@ -0,0 +1,84 @@
...
+    if (tile_offset_row_ < static_cast<std::size_t>(block_size_.rows())) {
+      throw std::invalid_argument("Error: Invalid Tile Row Offset");
+    }
+    if (tile_offset_row_ < minTileSize(block_size_) &&
+        static_cast<std::size_t>(ld_tile_) <
+            sum(mul((nr_tiles_.rows() - 1), tile_offset_row_), last_rows)) {
+      throw std::invalid_argument("Error: Invalid Leading Dimension & Tile Row Offset combination");
+    }
+    if (tile_offset_col_ <
+        mul(nr_tiles_.rows() - 1, tile_offset_row_) + minTileSize({last_rows, block_size_.cols()})) {
```
albestro: I would try to improve readability by extracting a few variables with a self-explaining name.

src/matrix/layout_info.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    if (ld_tile_ < block_size_.rows()) {
+      throw std::invalid_argument("Error: Invalid Leading Dimension");
+    }
+    if (tile_offset_row_ < static_cast<std::size_t>(block_size_.rows())) {
+      throw std::invalid_argument("Error: Invalid Tile Row Offset");
+    }
+    if (tile_offset_row_ < minTileSize(block_size_) &&
+        static_cast<std::size_t>(ld_tile_) <
+            sum(mul((nr_tiles_.rows() - 1), tile_offset_row_), last_rows)) {
```
albestro: I would try to improve readability by extracting a few variables with a self-explaining name.

include/dlaf/matrix_base.h
rasolca: It is a placeholder for distribution parameters.

include/dlaf/matrix/layout_info.h
```diff
@@ -0,0 +1,78 @@
...
+
+private:
+  /// @brief Returns the minimum number of elements that are needed to fit a tile with the given size.
+  std::size_t minTileSize(TileElementSize size) const noexcept;
+
+  GlobalElementSize size_;
+  LocalTileSize nr_tiles_;
+  TileElementSize block_size_;
+
+  SizeType ld_tile_;
```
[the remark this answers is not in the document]
rasolca: The matrix can be divided in block matrices which are called Tile.
Tile layout is a short name of the layout with the following properties:
- compact Tile storage (i.e. two different tiles are disjoint in memory)
- Tile are ordered in column major format
- Elements of the tiles are ordered in column major format.

src/matrix/layout_info.cpp
```diff
@@ -0,0 +1,84 @@
...
+LayoutInfo::LayoutInfo(GlobalElementSize size, TileElementSize block_size, SizeType tile_ld,
+                       std::size_t tile_offset_row, std::size_t tile_offset_col)
+    : size_(size), nr_tiles_(0, 0), block_size_(block_size), ld_tile_(tile_ld),
+      tile_offset_row_(tile_offset_row), tile_offset_col_(tile_offset_col) {
+  using util::size_t::sum;
+  using util::size_t::mul;
+
+  if (size_.rows() < 0 || size_.cols() < 0)
+    throw std::invalid_argument("Error: Invalid Matrix size");
+  if (block_size_.rows() < 1 || block_size_.cols() < 1)
```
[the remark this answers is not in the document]
rasolca: added `isEmpty()`

src/matrix_base.cpp
```diff
@@ -0,0 +1,48 @@
...
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#include "dlaf/matrix_base.h"
+
+namespace dlaf {
+MatrixBase::MatrixBase() noexcept : size_(0, 0), nr_tiles_(0, 0), block_size_(1, 1) {}
```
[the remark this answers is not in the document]
rasolca: I prefer to have defaults for members only when all constructors initialize them with the same value.

src/matrix_base.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+MatrixBase& MatrixBase::operator=(MatrixBase&& rhs) noexcept {
+  size_ = rhs.size_;
+  nr_tiles_ = rhs.nr_tiles_;
+  block_size_ = rhs.block_size_;
+
+  rhs.size_ = {0, 0};
+  rhs.nr_tiles_ = {0, 0};
+  rhs.block_size_ = {1, 1};
```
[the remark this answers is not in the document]
rasolca: `rhs = MatrixBase()` needs the move operator which is the operation I'm implementing.

src/matrix/layout_info.cpp
```diff
@@ -49,19 +51,18 @@ LayoutInfo::LayoutInfo(GlobalElementSize size, TileElementSize block_size, SizeT
       throw std::invalid_argument("Error: Invalid Tile Row Offset");
     }
     if (tile_offset_row_ < minTileSize(block_size_) &&
```
albestro: As far as I can tell, `tile_offset_row_ < minTileSize(block_size)` is used to check if it is a *column-major* layout.

What about making it more explicit?

src/matrix_base.cpp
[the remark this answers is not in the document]
albestro: Yeah, you are right. With the aim to follow the DRY principle, I propose to have a private function that initializes it, in order to centralize the "default initialization".

What do you think about it? I would use it in the default constructor too.

include/dlaf/matrix_base.h
```diff
@@ -0,0 +1,67 @@
...
+
+  bool operator==(const MatrixBase& rhs) const noexcept {
+    return size_ == rhs.size_ && nr_tiles_ == rhs.nr_tiles_ && block_size_ == rhs.block_size_;
+  }
+
+  bool operator!=(const MatrixBase& rhs) const noexcept {
+    return !operator==(rhs);
+  }
+
+  const GlobalElementSize& size() const noexcept {
```
albestro: I usually don't like to return `const &` in this kind of situation, because it exposes internal attributes and the lifetime of the object has to be guaranteed.
Returning by value (non-const) costs more, but I think that if it is that the reason, it may be premature optimization.

I don't think it would expose us to problems, however, I just want to point out it so you can evaluate by yourself.

rasolca: I don't see the problem:
* `auto size = matrix.size();` copies the value
* `auto& size = matrix.size();` reference the value
It is up to the user to use the correct one for his case.

include/dlaf/matrix/layout_info.h
```diff
@@ -17,29 +17,39 @@ namespace dlaf {
 namespace matrix {
 class LayoutInfo {
 public:
-  LayoutInfo(GlobalElementSize size, TileElementSize block_size, SizeType tile_ld,
+  LayoutInfo(const GlobalElementSize& size, const TileElementSize& block_size, SizeType tile_ld,
              std::size_t tile_offset_row, std::size_t tile_offset_col);
```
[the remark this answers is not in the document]
rasolca: Documentation will need to be rewritten when we distribute the matrix.
Therefore I didn't add the documentation to avoid duplicate work.

I may add a placeholder for that if needed.

src/matrix/layout_info.cpp
rasolca: Not in general. The columns may be padded.
This test checks if there is overlap between tiles in the same column and is divided in two parts:
 
if `tile_offset_row_ >= minTileSize(block_size_)` tiles are separated
otherwise the columns of the tiles are interleaved and
only if `static_cast<std::size_t>(ld_tile_) >= sum(tileOffset({nr_tiles_.rows() - 1, 0}), last_rows))`
there is enough space to store all the `i`-th columns of a tile before starting the `i+1`-th column.

Documentation (see other comment) will make this point clear.

src/matrix_base.cpp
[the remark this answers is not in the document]
rasolca: What about PR #61 and the Tile move operator. Is that OK?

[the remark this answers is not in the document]
albestro: Are you talking about this [comment](https://github.com/eth-cscs/DLA-Future/pull/61/files/916796f121cec23aa3772721966838385c706a05#r332856772)?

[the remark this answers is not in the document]
rasolca: I'm talking about the proposed solution in 8e6c88f04b8a810f4823e7abdc05877ed6b3da86.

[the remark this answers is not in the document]
albestro: I didn't see that! It looks perfect to me ;)

src/matrix/layout_info.cpp
albestro: I'm not getting your point.

Aren't **"tiles separated" === "tile-layout"** and **"tiles interleaved" === "column-layout"**?

In the comment you are saying that in case of interleaved tiles you have to check that there is enough space for storing all the columns before starting the next "tile-column". To me, this confirms that `tile_offset_row_ < minTileSize(block_size)` is there just to check that we are working with interleaved tiles, and from my assumption, a column-layout.

I may be missing something.

rasolca: Not in general (extra paddings). The relation works in a single direction:
"tiles separated" <= "tile-layout" and "tiles interleaved" <= "column-layout"
where => means implies (math symbol).

Assume a (4 x 2) matrix with (2 x 2) tiles.
```
a0  A0
a1  A1
b0  B0
b1  B1
```
E.g.
`a0 a1 b0 b1 A0 A1 B0 B1` `a0 a1 b0 b1 * A0 A1 B0 B1 *` are col major
`a0 a1 * b0 b1 * A0 A1 * B0 B1 *` is not.

albestro: After the discussion, I got your point.

Just for reference, the misunderstanding was due to my misinterpretation of the column-major layout: the column-major layout does not allow extra-padding (`*` in the example) between tiles.

PR #60, opened 2019-10-04
Implementation of broadcast functions

### LINKED ISSUE #58: Add broadcast communication function
Part of the implementation of #57 is related to communication.

It would be nice to have a wrapper around MPI calls to ease the work and make the code more clean.

For instance, working with tiles, a way to easily broadcast a `Tile` would be useful.

albestro:
Close #58 

Implement wrappers for `MPI_Bcast` and `MPI_Ibcast`.

It provides also a basic compile-time mapper between library types and MPI_Datatype. Initially it addresses just more common primitive types. In a separate PR I would add support for more comples types (e.g. tiles).

### DIFF
--- a/include/dlaf/communication/functions.h
+++ b/include/dlaf/communication/functions.h
@@ -19,16 +19,16 @@ namespace comm {
 
 /// basic wrapper for MPI_Bcast
 template <typename MessageType>
-void bcast(int broadcaster_rank, MessageType& what, Communicator where) {
-  MPI_Bcast(&what, 1, mpi_datatype<MessageType>::type, broadcaster_rank, where);
+void bcast(int broadcaster_rank, MessageType& message, Communicator communicator) {
+  MPI_Bcast(&message, 1, mpi_datatype<MessageType>::type, broadcaster_rank, communicator);
 }
 
 /// basic wrapper for MPI_Ibcast
 template <typename MessageType>
-void async_bcast(int broadcaster_rank, MessageType& what, Communicator where,
+void async_bcast(int broadcaster_rank, MessageType& message, Communicator communicator,
                  std::function<void()> action_before_retrying) {
   MPI_Request request;
-  MPI_Ibcast(&what, 1, mpi_datatype<MessageType>::type, broadcaster_rank, where, &request);
+  MPI_Ibcast(&message, 1, mpi_datatype<MessageType>::type, broadcaster_rank, communicator, &request);
 
   while (true) {
     int test_flag = 1;
@@ -43,14 +43,14 @@ namespace broadcast {
 
 /// specialized wrapper for MPI_Bcast on sender side
 template <typename MessageType>
-void send(const MessageType& what, Communicator where) {
-  bcast(where.rank(), const_cast<MessageType&>(what), where);
+void send(const MessageType& message, Communicator communicator) {
+  bcast(communicator.rank(), const_cast<MessageType&>(message), communicator);
 }
 
 /// specialized wrapper for MPI_Bcast on receiver side
 template <typename MessageType>
-void receive_from(int broadcaster_rank, MessageType& what, Communicator where) {
-  bcast(broadcaster_rank, what, where);
+void receive_from(int broadcaster_rank, MessageType& message, Communicator communicator) {
+  bcast(broadcaster_rank, message, communicator);
 }
 
 }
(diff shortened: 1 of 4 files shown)


albestro:
I opened this PR starting from a proposal of API for broadcast. Tests are there just for giving an idea of the usage of the API.

include/dlaf/communication/functions.h
```diff
@@ -0,0 +1,76 @@
...
+template <typename MessageType>
+void send(const MessageType& what, Communicator where) {
+  bcast(where.rank(), const_cast<MessageType&>(what), where);
+}
+
+/// specialized wrapper for MPI_Bcast on receiver side
+template <typename MessageType>
+void receive_from(int broadcaster_rank, MessageType& what, Communicator where) {
+  bcast(broadcaster_rank, what, where);
+}
```
rasolca: Is a separate implementation between bcast send and bcast receive needed at this level?

include/dlaf/communication/functions.h
```diff
@@ -0,0 +1,76 @@
...
+      break;
+    action_before_retrying();  // has this to be called from both sides of the communication?
+  }
+}
+
+namespace broadcast {
+
+/// specialized wrapper for MPI_Bcast on sender side
+template <typename MessageType>
+void send(const MessageType& what, Communicator where) {
```
rasolca: Did you considered to have the syntax `communicator.broadcast(...)`?

include/dlaf/communication/functions.h
albestro: I have two arguments on this:

- The first one is that this "separated API" may be useful to have better "const-correctness" control over the message passed; in fact, the receive side requires a `non-const` message, while the sender side is ok both `const` and `non-const`. It is true that it hides a `const_cast` and from the MPI standard the constness is not guaranteed, since it says *"`MPI_BCAST` broadcasts a message from the process with rank root to all processes of the group, itself included. [...] On return, the content of root’s buffer is copied to all other processes.s"*;
- The second argument is that, looking at the prototype code, there are two different code sections for the sender and receiver side (mainly due to how the HPX taks has to be created).

rasolca: I agree with the first point.
For the second point I expect to have a front-end broadcast API that deals with tiles (see previous comment)

test/unit/communication/test_broadcast.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+  dlaf::comm::async_bcast(broadcaster, message, communicator, what_to_do_before_retesting);
+  EXPECT_EQ(color, message);
+}
+
+TEST_F(BroadcastTest, AsyncBroadcast_ClassicAPI_Splitted) {
+  auto broadcaster = 0;
+  auto communicator = splitted_comm;
+
+  bool waited;
```
rasolca: waited is not initialized.

test/unit/communication/test_broadcast.cpp
```diff
@@ -0,0 +1,126 @@
...
+  auto what_to_do_before_retesting = [&waited]() { waited = true; };
+
+  if (isMasterInSplitted())
+    dlaf::comm::async_bcast(broadcaster, color, communicator, what_to_do_before_retesting);
+  else {
+    int message;
+    dlaf::comm::async_bcast(broadcaster, message, communicator, what_to_do_before_retesting);
+    EXPECT_EQ(color, message);
+  }
```
rasolca: What about a check that wait has been called?

include/dlaf/communication/mpi_datatypes.h
```diff
@@ -0,0 +1,27 @@
...
+namespace dlaf {
+namespace comm {
+
+template <typename T>
+struct mpi_datatype;
+
+template <>
+struct mpi_datatype<int> {
+  static constexpr MPI_Datatype type = MPI_INT;
+};
```
rasolca: only `int`?

include/dlaf/communication/functions.h
```diff
@@ -0,0 +1,78 @@
...
+#include "dlaf/communication/communicator.h"
+#include "dlaf/communication/mpi_datatypes.h"
+#include "dlaf/mpi_header.h"
+
+namespace dlaf {
+namespace comm {
+
+/// basic wrapper for MPI_Bcast
+template <typename MessageType>
+void bcast(int broadcaster_rank, MessageType& message, Communicator communicator) {
```
rasolca: This is a broadcast of only one element of a type.
Which use do you predict for it?

include/dlaf/communication/mpi_datatypes.h
albestro: As you have already seen, in the derived PR #70 I extend this for more types.

include/dlaf/communication/functions.h
albestro: You are right. This PR was just to give a general idea of the API, in the PR #70 you case how I propose to address this aspect.

albestro:
As you have realized, part of the problems you highlighted are addressed by #70.

There are still some work ongoing on this part, for which I'm going to open a PR after the #70 will be approved and merged.

Do you agree about merging these PR #60 and #70? Are two distinct functionalities, but the concepts are strictly related.

PR #65, opened 2019-10-14
Fix gtest_hpx_main

### LINKED ISSUE #64: HPX tests do not return the error code correctly.
HPX tests which use `gtest_hpx_main` always return 0 even on case of failure.

rasolca:
Fixes #64.

`gtest_hpx_main` now return the correct error code.
`Tile` tests which were silently failing has been fixed too.

### DIFF
--- a/test/src/gtest_hpx_main.cpp
+++ b/test/src/gtest_hpx_main.cpp
@@ -46,7 +46,9 @@
 GTEST_API_ int test_main(int argc, char** argv) {
   std::printf("Running main() from gtest_hpx_main.cpp\n");
   testing::InitGoogleTest(&argc, argv);
-  return RUN_ALL_TESTS();
+  auto ret = RUN_ALL_TESTS();
+  hpx::finalize();
+  return ret;
 }
 
 GTEST_API_ int main(int argc, char** argv) {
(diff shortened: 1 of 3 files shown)


test/src/gtest_hpx_main.cpp
```diff
@@ -46,16 +46,18 @@
...
   hpx::runtime* rt = hpx::get_runtime_ptr();
   hpx::util::yield_while([rt]() { return rt->get_state() < hpx::state_running; });
 
-  hpx::threads::run_as_hpx_thread(test_main, argc, argv);
+  auto ret = hpx::threads::run_as_hpx_thread(test_main, argc, argv);
+  hpx::stop();
 
-  hpx::apply([]() { hpx::finalize(); });
-  return hpx::stop();
+  return ret;
```
albestro: Lookin at HPX documentation [here](https://stellar-group.github.io/hpx/docs/sphinx/latest/html/manual/starting_the_hpx_runtime.html?highlight=start#supply-your-own-main-hpx-entry-point-while-avoiding-to-block-the-main-thread) it seems that we can have just as `main` code this:

```cpp
int main() {
  // rename `test_main` to `hpx_main`, otherwise pass `&test_main` as first parameter
  hpx::start(argc, argv);
  return hpx::stop();
}
```

Am I wrong? Is there any reason why you should take care of the `hpx::runtime` like you did?

rasolca: To initialize HPX in the same way we will do in the library. (We cannot force to replace main in the applications).

PR #66, opened 2019-10-15
Added (Local) Matrix of futures of tiles.

rasolca:
- Added local Matrix of futures of tiles and its tests.
- Fixed LayoutInfo which was failing for some corner cases.

### DIFF
--- a/include/dlaf/matrix.h
+++ b/include/dlaf/matrix.h
@@ -1,0 +1,132 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2019, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+#include <vector>
+#include "dlaf/matrix/layout_info.h"
+#include "dlaf/matrix_base.h"
+#include "dlaf/tile.h"
+#include "dlaf/types.h"
+
+namespace dlaf {
+
+template <class T, Device device>
+class Matrix : public MatrixBase {
+public:
+  using ElementType = T;
+  using TileType = Tile<ElementType, device>;
+  using ConstTileType = Tile<const ElementType, device>;
+  friend Matrix<const ElementType, device>;
+
+  Matrix(const GlobalElementSize& size, const TileElementSize& block_size);
+
+  Matrix(const matrix::LayoutInfo& layout, ElementType* ptr, std::size_t elements);
+
+  Matrix(const Matrix& rhs) = delete;
+  Matrix(Matrix&& rhs) = default;
+
+  Matrix& operator=(const Matrix& rhs) = delete;
+  Matrix& operator=(Matrix&& rhs) = default;
+
+  /// @brief Returns a read-only shared_future of index Tile.
+  /// TODO: Sync details.
+  /// @pre index.isValid() == true.
+  /// @pre index.isIn(nrTiles()) == true.
+  hpx::shared_future<ConstTileType> read(const LocalTileIndex& index);
+
+  /// @brief Returns a future of index Tile.
+  /// TODO: Sync details.
+  /// @pre index.isValid() == true.
+  /// @pre index.isIn(nrTiles()) == true.
+  hpx::future<TileType> operator()(const LocalTileIndex& index);
+
+protected:
+  /// @brief Returns the position in the vector of the index Tile.
+  /// @pre index.isValid() == true.
+  /// @pre index.isIn(nrTiles()) == true.
+  std::size_t tileLinearIndex(const LocalTileIndex& index) {
+    assert(index.isValid() && index.isIn(nr_tiles_));
+    using util::size_t::sum;
+    using util::size_t::mul;
+    return sum(index.row(), mul(ld_futures_, index.col()));
+  }
+
+private:
+  void setUpTiles(const memory::MemoryView<ElementType, device>& mem, const matrix::LayoutInfo& layout);
+
+  std::vector<hpx::future<TileType>> tile_futures_;
+  std::vector<hpx::shared_future<ConstTileType>> tile_shared_futures_;
+  std::size_t ld_futures_;
+};
+
+#include <dlaf/matrix.ipp>
+
+template <class T, Device device>
+class Matrix<const T, device> : public MatrixBase {
+public:
+  using ElementType = T;
+  using ConstTileType = Tile<const ElementType, device>;
+
+  Matrix(const matrix::LayoutInfo& layout, ElementType* ptr, std::size_t elements);
+
+  Matrix(const matrix::LayoutInfo& layout, const ElementType* ptr, std::size_t elements)
+      : Matrix(layout, const_cast<ElementType*>(ptr), elements) {}
+
+  Matrix(const Matrix& rhs) = delete;
+  Matrix(Matrix&& rhs) = default;
+
+  Matrix& operator=(const Matrix& rhs) = delete;
+  Matrix& operator=(Matrix&& rhs) = default;
+
+  /// @brief Returns a read-only shared_future of index Tile.
+  /// @pre index.isValid() == true.
+  /// @pre index.isIn(nrTiles()) == true.
+  hpx::shared_future<ConstTileType> read(const LocalTileIndex& index) noexcept;
+
+protected:
+  /// @brief Returns the position in the vector of the index Tile.
+  /// @pre index.isValid() == true.
+  /// @pre index.isIn(nrTiles()) == true.
+  std::size_t tileLinearIndex(const LocalTileIndex& index) noexcept {
+    assert(index.isValid() && index.isIn(nr_tiles_));
+    using util::size_t::sum;
+    using util::size_t::mul;
+    return sum(index.row(), mul(ld_futures_, index.col()));
+  }
+
+private:
+  void setUpTiles(const memory::MemoryView<ElementType, device>& mem, const matrix::LayoutInfo& layout);
+
+  std::vector<hpx::shared_future<ConstTileType>> tile_shared_futures_;
+  std::size_t ld_futures_;
+};
+
+#include <dlaf/matrix_const.ipp>
+
+template <class T, Device device>
+Matrix<T, device> createMatrixFromColMajor(const GlobalElementSize& size,
+                                           const TileElementSize& block_size, SizeType ld, T* ptr,
+                                           std::size_t elements) {
+  return Matrix<T, device>(matrix::colMajorLayout(size, block_size, ld), ptr, elements);
+}
+
+template <class T, Device device>
+Matrix<T, device> createMatrixFromTile(const GlobalElementSize& size, const TileElementSize& block_size,
+                                       T* ptr, std::size_t elements) {
+  return Matrix<T, device>(matrix::tileLayout(size, block_size), ptr, elements);
+}
+
+template <class T, Device device>
+Matrix<T, device> createMatrixFromTile(const GlobalElementSize& size, const TileElementSize& block_size,
+                                       SizeType ld_tile, SizeType tiles_per_col, T* ptr,
+                                       std::size_t elements) {
+  return Matrix<T, device>(matrix::tileLayout(size, block_size, ld_tile, tiles_per_col), ptr, elements);
+}
+}
--- a/include/dlaf/matrix.h
+++ b/include/dlaf/matrix.h
@@ -39,19 +39,19 @@ public:
   /// TODO: Sync details.
   /// @pre index.isValid() == true.
   /// @pre index.isIn(nrTiles()) == true.
-  hpx::future<TileType> operator()(const LocalTileIndex& index);
+  hpx::future<TileType> operator()(const LocalTileIndex& index) noexcept;
 
 protected:
   using Matrix<const T, device>::tileLinearIndex;
 
 private:
-  void setUpTiles(const memory::MemoryView<ElementType, device>& mem, const matrix::LayoutInfo& layout);
+  void setUpTiles(const memory::MemoryView<ElementType, device>& mem,
+                  const matrix::LayoutInfo& layout) noexcept;
 
   using Matrix<const T, device>::setUpTilesInternal;
   using Matrix<const T, device>::futureVectorSize;
   using Matrix<const T, device>::tile_futures_;
   using Matrix<const T, device>::tile_shared_futures_;
-  using Matrix<const T, device>::ld_futures_;
 };
 
 #include "dlaf/matrix.ipp"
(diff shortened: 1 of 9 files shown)


test/unit/test_matrix.cpp
```diff
@@ -0,0 +1,365 @@
...
+      set(mat, el);
+
+      CHECK_MATRIX_EQ(el, mat);
+    }
+  }
+}
+
+template <class T, Device device>
+void checkFromExisting(T* p, const matrix::LayoutInfo& layout, Matrix<T, device>& matrix) {
+  auto el = [](const GlobalElementIndex& index) {
```
albestro: Please, use better names for `el` and `el2`, e.g. `reference_matrix1`, `reference_matrix2`.

test/unit/test_matrix.cpp
```diff
@@ -0,0 +1,365 @@
...
+  EXPECT_EQ(MatrixBase(layout), matrix);
+  CHECK_MATRIX_PTR(ptr, matrix);
+  CHECK_MATRIX_EQ(el, matrix);
+
+  set(matrix, el2);
+
+  for (SizeType j = 0; j < size.cols(); ++j) {
+    for (SizeType i = 0; i < size.rows(); ++i) {
+      if (el2({i, j}) != *ptr({i, j})) {
+        FAIL() << "Error at index (" << i << ", " << j << "): "
```
albestro: In general, you can use every GoogleTest macro as a stream (see [doc](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#assertions))

In this particular case, `ASSERT_EQ(...) << "additional message"`

test/unit/test_matrix.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+    const Type* p = mem();
+    Matrix<const Type, Device::CPU> mat(layout, p, mem.size());
+
+    CHECK_FROM_EXISTING(mem(), layout, mat);
+  }
+}
+
+/// @brief Returns true if the first @p ready futures are ready.
+/// @pre Future* should be future or shared_future
```
albestro: The * stands for 1 or 2...maybe something similar to FutureN?

test/unit/test_matrix.cpp
```diff
@@ -0,0 +1,365 @@
...
+    memory::MemoryView<Type, Device::CPU> mem(layout.minMemSize());
+
+    const Type* p = mem();
+    Matrix<const Type, Device::CPU> mat(layout, p, mem.size());
+
+    CHECK_FROM_EXISTING(mem(), layout, mat);
+  }
+}
+
+/// @brief Returns true if the first @p ready futures are ready.
```
albestro: I would say _"Return turn if JUST the firsts [...]"_ and try to find a more explicit name for the function.

test/unit/test_matrix.cpp
```diff
@@ -0,0 +1,365 @@
...
+      return false;
+  }
+  return true;
+}
+
+/// @brief Checks if current[i] depends correctly on previous[i].
+/// If get_ready == true it checks if current[i] is ready after previous[i] is used.
+/// If get_ready == false it checks if current[i] is not ready after previous[i] is used.
+/// @pre Future* should be future or shared_future
+template <template <class> class Future1, template <class> class Future2, class T1, class T2,
```
albestro: You are testing futures and you are not really interested in the data itself.

I would simplify this with
```cpp
template <class Future1, class Future2>
void checkFutures(bool get_ready, const std::vector<Future1>& current, std::vector<Future2>& previous)
```

I would apply this also to the other functions whose concern is just the future (e.g. `checkFuturesStep`).

test/unit/test_matrix.cpp
```diff
@@ -0,0 +1,365 @@
...
+          Device device>
+void checkFutures(bool get_ready, const std::vector<Future1<Tile<T1, device>>>& current,
```
albestro: Can we find a better name? In this function, you are not just checking `current`, but you are also triggering `previous`. I would make it more explicit.

IMHO it would make more clear also the role of `get_ready`, it could become something like `current_ready_after_trigger`.

test/unit/test_matrix.cpp
```diff
@@ -0,0 +1,365 @@
...
+  EXPECT_TRUE(checkFuturesStep(get_ready ? current.size() : 0, current));
+}
+
+TYPED_TEST(MatrixTest, Dependencies) {
+  using Type = TypeParam;
+
+  for (const auto& size : sizes) {
+    for (const auto& block_size : block_sizes) {
+      Matrix<Type, Device::CPU> mat(size, block_size);
+
```
albestro: I would add a comment that shows the pipeline you are testing.

```bash
             shfut2a
            /       \
fut0 - fut1          fut3 - shfut4a
            \       /
             shfut2b


             shfut2a       shfut4a
            /       \     /       \
fut0 - fut1          fut3          fut5
            \       /     \       /
             shfut2b       shfut4b
```

albestro:
About my concerns about Matrix/Tile API.

I may agree on the "reference-semantic" of these objects (i.e. Tile, Matrix), but I would investigate better if there is a way to avoid this overload and the non-const-from-const inheritance "workaround".

After the discussion, we agreed that, in case, we can try refactoring this in the next early stages of the development.

src/matrix/layout_info.cpp
```diff
@@ -44,18 +45,20 @@ LayoutInfo::LayoutInfo(const GlobalElementSize& size, const TileElementSize& blo
...
+    if (ld_tile_ < tmp) {
       throw std::invalid_argument("Error: Invalid Leading Dimension");
     }
-    if (tile_offset_row_ < static_cast<std::size_t>(block_size_.rows())) {
+    if (tile_offset_row_ < static_cast<std::size_t>(tmp)) {
       throw std::invalid_argument("Error: Invalid Tile Row Offset");
     }
-    if (tile_offset_row_ < minTileSize(block_size_) &&
+    if (tile_offset_row_ < minTileMemSize(block_size_) &&
         static_cast<std::size_t>(ld_tile_) < sum(tileOffset({nr_tiles_.rows() - 1, 0}), last_rows)) {
```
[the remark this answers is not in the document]
rasolca: Debatable about readability.
Anyway you are reviewing code that didn't change.

include/dlaf/matrix.h
```diff
@@ -0,0 +1,132 @@
...
+                                       T* ptr, std::size_t elements) {
+  return Matrix<T, device>(matrix::tileLayout(size, block_size), ptr, elements);
+}
+
+template <class T, Device device>
+Matrix<T, device> createMatrixFromTile(const GlobalElementSize& size, const TileElementSize& block_size,
+                                       SizeType ld_tile, SizeType tiles_per_col, T* ptr,
+                                       std::size_t elements) {
+  return Matrix<T, device>(matrix::tileLayout(size, block_size, ld_tile, tiles_per_col), ptr, elements);
+}
```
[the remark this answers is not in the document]
rasolca: When the matrix will be distributed, the construction will not be so straightforward.

include/dlaf/matrix.h
```diff
@@ -0,0 +1,132 @@
...
+class Matrix : public MatrixBase {
+public:
+  using ElementType = T;
+  using TileType = Tile<ElementType, device>;
+  using ConstTileType = Tile<const ElementType, device>;
+  friend Matrix<const ElementType, device>;
+
+  Matrix(const GlobalElementSize& size, const TileElementSize& block_size);
+
+  Matrix(const matrix::LayoutInfo& layout, ElementType* ptr, std::size_t elements);
```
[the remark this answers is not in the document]
rasolca: I thought this was clear (first comment of the PR):
> Documentation for methods which involve a layout description will be added when the Matrix will support distributed matrices as well.

include/dlaf/matrix.ipp
```diff
@@ -0,0 +1,82 @@
...
+  return tile_shared_futures_[i];
+}
+
+template <class T, Device device>
+hpx::future<Tile<T, device>> Matrix<T, device>::operator()(const LocalTileIndex& index) {
+  std::size_t i = tileLinearIndex(index);
+  hpx::future<TileType> old_future = std::move(tile_futures_[i]);
+  hpx::promise<TileType> p;
+  tile_futures_[i] = p.get_future();
+  tile_shared_futures_[i] = {};
```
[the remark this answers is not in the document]
rasolca: I don't understand.
`shared_future` does not have a `clear()` method.

include/dlaf/matrix.h
```diff
@@ -0,0 +1,132 @@
...
+#include <vector>
+#include "dlaf/matrix/layout_info.h"
+#include "dlaf/matrix_base.h"
+#include "dlaf/tile.h"
+#include "dlaf/types.h"
+
+namespace dlaf {
+
+template <class T, Device device>
+class Matrix : public MatrixBase {
```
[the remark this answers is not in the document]
rasolca: const correctness do not always respect the DRY principle (e.g. vector operator[])
Moreover `tileLinearIndex` in an implementation detail, and, in general, the order in which the tiles are stored is indifferent.
regarding setUpTiles, not only the Tile type changes but also the future type. IMHO setUpTiles do not contradict the DRY principle.

test/include/dlaf_test/util_matrix.h
```diff
@@ -0,0 +1,150 @@
...
+/// err_message(el({i, j}), (i, j)-element) is printed for the first element which does not fulfill the comparison.
+/// @pre el argument is an index of type const GlobalElementIndex&.
+/// @pre comp should have two arguments and return true if the comparison is fulfilled and false otherwise.
+/// @pre err_message should have two arguments and return a string.
+/// @pre el return type should be the same as the type of the first argument of comp adn of err_message.
+/// @pre The second argument of comp should be either T, T& or const T&.
+/// @pre The second argument of err_message should be either T, T& or const T&.
+namespace internal {
+template <class T, class Func1, class Func2, class Func3>
+void check(Matrix<T, Device::CPU>& mat, Func1 el, Func2 comp, Func3 err_message, const char* file,
```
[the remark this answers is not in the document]
rasolca: This function do not assert. As written in the doc it checks if the elements fulfill the comparison and add a failure when it finds the first element that doesn't.
The function and macros which re supposed to be used by users (this is an internal detail) have the self explaining name.

test/unit/test_matrix.cpp
rasolca: Ok that the name should explain what the function does, but here we are going in the direction that the function name should contain the full documentation.

include/dlaf/matrix.h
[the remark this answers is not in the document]
rasolca: > As per the discussion, I'm still not convinced of this API, see review comment for additional details.
We cannot start a discussion on the API on each PR. We will never progress.
The "pointer" semantic is the only way for us to provide all the functionalities we need, therefore we use it.

test/unit/test_matrix.cpp
rasolca: FutureN is even more confusing IMO.

include/dlaf/matrix.h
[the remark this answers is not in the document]
albestro: 🙏 Sorry, I forgot it.

What about accepting just the `MemoryView&&` and not the raw pointer?

include/dlaf/matrix.h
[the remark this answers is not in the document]
albestro: Ok, I respect your point.

What about inverting the template parameters?

test/unit/test_matrix.cpp
rasolca: `reference_matrix` - > I would expect a matrix.
This is how `el` was chosen:
`matrix_elements` ~> `elements` ~>`el`
Is `elements` ok?

albestro: It was just my opinion. I agree with you that `elements` does not change so much from `el`.

include/dlaf/matrix.h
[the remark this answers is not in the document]
rasolca: Will do it

test/include/dlaf_test/util_matrix.h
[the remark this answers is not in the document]
albestro: I said it was an assert-ish thing because, looking at the code I see, as you said, that you want to stop at the first failed comparison. I would have made it clear from the call that this is going to assert and stop the test.

However, it is ok for me. 👌🏼

include/dlaf/matrix.h
```diff
@@ -0,0 +1,143 @@
...
+  /// TODO: Sync details.
+  /// @pre index.isValid() == true.
+  /// @pre index.isIn(nrTiles()) == true.
+  hpx::shared_future<ConstTileType> read(const LocalTileIndex& index) noexcept;
+
+protected:
+  /// @brief Returns the position in the vector of the index Tile.
+  /// @pre index.isValid() == true.
+  /// @pre index.isIn(nrTiles()) == true.
+  std::size_t tileLinearIndex(const LocalTileIndex& index) noexcept {
```
albestro: `const`?

include/dlaf/matrix.ipp
```diff
@@ -62,21 +48,8 @@ hpx::future<Tile<T, device>> Matrix<T, device>::operator()(const LocalTileIndex&
 template <class T, Device device>
 void Matrix<T, device>::setUpTiles(const memory::MemoryView<T, device>& mem,
                                    const matrix::LayoutInfo& layout) {
-  using util::size_t::mul;
-  const auto& nr_tiles = layout.nrTiles();
-  ld_futures_ = static_cast<std::size_t>(nr_tiles.rows());
-  std::size_t tot_tiles = mul(ld_futures_, nr_tiles.cols());
-  tile_futures_.reserve(tot_tiles);
-  tile_shared_futures_.resize(tot_tiles);
+  ld_futures_ = static_cast<std::size_t>(layout.nrTiles().rows());
```
albestro: I would initialize `ld_futures_` in the constructors.

include/dlaf/matrix.h
```diff
@@ -102,28 +97,44 @@ class Matrix<const T, device> : public MatrixBase {
...
+
+  void setUpConstTiles(const memory::MemoryView<ElementType, device>& mem,
+                       const matrix::LayoutInfo& layout);
 
+  template <template <class> class Future, class TileT>
+  void setUpTilesInternal(std::vector<Future<TileT>>& tile_futures_vector,
+                          const memory::MemoryView<ElementType, device>& mem,
+                          const matrix::LayoutInfo& layout);
+
+  std::size_t futureVectorSize(const matrix::LayoutInfo& layout);
```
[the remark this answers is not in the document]
albestro: I would also add `noexcept`.

rasolca:
- Added local Matrix of futures of tiles and its tests.
- Fixed LayoutInfo which was failing for some corner cases.

PR #70, opened 2019-10-18
Implementation of abstraction for sending data through MPI

### LINKED ISSUE #62: Add an abstraction for data to be sent/received with MPI.
I think it would be a nice thing to separate the role of communicating the data from preparing the data to be sent/received. The former one is addressed by #58, while this issue should address the latter one.

Data preparation happens on both sides of the communication:
- on the sender side, the memory has to be wrapped into an MPI Datatype;
- on the receiver side, the data value has to be extracted from the MPI Datatype;

The implementation would provide an API that maps between MPI and types used in the library, i.e. both primitive types and custom types that may require the definition of custom MPI Datatypes.

Types that should be managed are:
- primitive types (e.g. int, float, ...)
- contiguous array
- non-contiguous array

Contiguous and non-contiguous memory sequences of heterogeneous datatypes are not a requirement for now.

albestro:
Close #62

- Implement `Buffer` concept, a pointer to an underlying buffer with information about the layout, how it can be seen in terms of "one or more blocks with a fixed number of elements (optionally strided)".
- Implement `message` object that wraps data that can be sent with MPI. It uses the Buffer concept to access information and data of the buffer to be sent.
- Implement `type_handler` object that creates a custom MPI Datatype (e.g. MPI_Type_vector) based on the data given. It owns the MPI resource, so once it is deleted it calls MPI_Free_type.
- Extend `mpi_datatype` implementation with more types (even `const` one are mapped).

The broadcast API and tests have been adapted.


albestro:
It is not the final API, it is just to give an idea of what it could it be, for discussing and commenting it.

include/dlaf/util_tile.h
```diff
@@ -0,0 +1,26 @@
...
+#pragma once
+
+#include "dlaf/common/buffer.h"
+#include "dlaf/tile.h"
+
+namespace dlaf {
+
+/// @brief Create a common::Buffer from a Tile
+template <class T, Device device>
+auto create_buffer(const dlaf::Tile<T, device>& tile) {
```
albestro: @rasolca 
Do you have something against embedding this function into `tile.h`?

Since `Buffer` and `Message` concepts are separated, it wouldn't add any dependency to MPI.

Instead, having it in a separate file like this forces users to remember to include it every time they need to send a `Tile`, which IMHO is not straightforward.

include/dlaf/common/buffer.h
```diff
@@ -0,0 +1,149 @@
...
+protected:
+  T* data_;
+  std::size_t nblocks_;
+  std::size_t blocksize_;
+  std::size_t stride_;
+};
+
+/// @brief Helper class for creatig a buffer from a bounded array
+template <class T, std::size_t N>
+struct Buffer<T[N], std::enable_if_t<std::is_array<T[N]>::value>> : Buffer<std::decay_t<T[N]>> {
```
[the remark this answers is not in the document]
albestro: Yes, you're right. It was there due to an additional specialization I was developing, but it is not used anymore.

test/unit/communication/test_broadcast_tile.cpp
```diff
@@ -0,0 +1,55 @@
...
+    for (int j = 0; j < tile.size().cols(); ++j)
+      for (int i = 0; i < tile.size().rows(); ++i)
+        EXPECT_NE(message_values({i, j}), *tile.ptr({i, j}));
+
+    dlaf::comm::broadcast::receive_from(0, dlaf::comm::make_message(tile), splitted_comm);
+  }
+
+  for (int j = 0; j < tile.size().cols(); ++j)
+    for (int i = 0; i < tile.size().rows(); ++i)
+      EXPECT_EQ(message_values({i, j}), *tile.ptr({i, j}));
```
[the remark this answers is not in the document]
albestro: Good hint! 😉

Please, give a look at the code, since I added `CHECK_TILE_NE` and to do this I had to slightly change the code behind.

test/unit/communication/test_broadcast_tile.cpp
```diff
@@ -12,44 +12,42 @@
...
+    tile_test::set(tile, message_values);
 
-    dlaf::comm::broadcast::send(dlaf::comm::make_message(tile), splitted_comm);
+    broadcast::send(make_message(tile), splitted_comm);
   }
   else {
-    for (int j = 0; j < tile.size().cols(); ++j)
-      for (int i = 0; i < tile.size().rows(); ++i)
-        EXPECT_NE(message_values({i, j}), *tile.ptr({i, j}));
+    CHECK_TILE_NE(message_values, tile);
```
rasolca: Avoid checks on unitialized memory. It is not guaranteed that the memory of the tile is different from the message value.
Prefer the approach which sets the elements (e.g. set to NaN or also 0 since all your elements (real and imaginary parts for complex) are >0).

test/unit/test_tile.cpp
```diff
@@ -334,3 +334,25 @@ TYPED_TEST(TileTest, PromiseToFutureConst) {
   auto ptr = [&memory_view](const TileElementIndex& index) { return memory_view(elIndex(index, ld)); };
   CHECK_TILE_PTR(ptr, tile2);
 }
+
+// Buffer
```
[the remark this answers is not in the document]
albestro: After https://github.com/eth-cscs/DLA-Future/pull/70#discussion_r341028776 `Tile` now includes the `Buffer`.

albestro:
As discussed, I removed from this PR the async broadcast code (there will be another PR about it) and I have also reverted the wrong code change about `CHECK_TILE_NE` and replaced it with a `set` function that accepts a single element instead of a functor.

albestro:
Close #58 

I closed PR #60 since everything is in this one now.

test/include/dlaf_test/util_tile.h
```diff
@@ -38,6 +40,21 @@ void set(const Tile<T, Device::CPU>& tile, Func el) {
   }
 }
 
+/// @brief Sets all the elements of the tile with given value.
+///
+/// @pre el argument is an index of type const TileElementIndex&.
+/// @pre el return type should be T.
```
rasolca: This doc is a copy and paste from the previous. Please update it.

test/include/dlaf_test/util_tile.h
```diff
@@ -38,6 +40,21 @@ void set(const Tile<T, Device::CPU>& tile, Func el) {
...
+template <class T, class U, enable_if_convertible_t<U, T, int> = 0>
+void set(const Tile<T, Device::CPU>& tile, U el) {
+  // TODO evaluate to use the other one with a lambda
```
rasolca: IMHO both approaches are OK.
Fix this TODO before merging.

PR #73, opened 2019-10-21
Added tile operations for Cholesky

### LINKED ISSUE #23: Tile operations implementation
BLAS LAPACK wrappers for Tile objects:
templated on Type and Device.

It should checks that the matrix sizes are compatible.

required:
- gemm
- trsm

other operations as needed.

Interface should support Tile and Wrapper<Tile>. Example for gemm:
`gemm(transpose_a, transpose_b, T alpha, const Tile<T>& a, const Tile<T>& b, T beta, Tile<T>& c);`
`gemm(transpose_a, transpose_b, T alpha, const Wrapper<Tile<T>>& a, const Wrapper<Tile<T>>& b, T beta Wrapper<Tile<T>>&& c);`

rasolca:
Closes #23.

Added tile operations needed for Cholesky.
(gemm, trsm, herk, potrf)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,7 @@ endif()
 find_package(HPX REQUIRED)
 
 # ----- BLASPP/LAPACKPP
+find_package(OpenMP REQUIRED)
 find_package(blaspp REQUIRED)
 find_package(lapackpp REQUIRED)
 
--- a/include/dlaf/blas_tile.h
+++ b/include/dlaf/blas_tile.h
@@ -1,0 +1,39 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2019, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+
+#include "blas.hh"
+#include "dlaf/tile.h"
+#include "dlaf/types.h"
+
+namespace dlaf {
+namespace tile {
+
+// See BLAS documentation for more details.
+
+/// @brief Computes general matrix matrix multiplication.
+template <class T, Device device>
+void gemm(blas::Op op_a, blas::Op op_b, T alpha, const Tile<const T, device>& a,
+          const Tile<const T, device>& b, T beta, const Tile<T, device>& c);
+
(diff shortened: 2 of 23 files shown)


rasolca:
Draft discussion:
- file name: `tile_operations.h` or `tile_blas.h` + `tile_lapack.h`?
- Are tests readable?
- nested for loops (e.g trsm needs 9) or a list with some of the combinations?
- How do we document the tests?

albestro:
- FileName
Since they are going to be pretty long files, I would keep them separated, also considering that it would be simple and clear due to the "natural" separation between the libraries;

- Test readability
In my opinion no, but I do not have an alternative proposal for the moment.

- Nested loops for
It seems that GoogleTest does not provide a way to mix Type and Value parameterized tests. In this [discussion](https://groups.google.com/forum/#!topic/googletestframework/3d8H5J5Isnw) it is suggested to use loops (even if it is not so nice). Just for completeness I report here a [code](https://github.com/S3ler/GoogleTestTypeValueParameterizedTests/blob/master/TypeValueParameterizedFooTest.cpp) that should solve the problem in a different way.

- Test documentation
What do you want to document exactly? If you want to document how the test is performed, it may be necessary to have a specific document, out of the code, for each test.

rasolca:
- test readability + doc:
I tried the to use inline comments to document the gemm test. What do you think? What about readability now?
- Nested loops:
My question was not how to remove the loops, but if it was better to use nested loops (gemm1) or a single loop (gemm2). To clearify the question I added how the test looks like in the two cases for gemm.

albestro:
### Test Documentation
I think having a separate file for each test implementation is a very good thing: it avoids having very cluttered long files, which are very difficult to read.
About documentation, I think that inline comments are good, and in case of very complex tests, we can use also the function documentation.

### Nested Loops
From a readability point of view, I find `Gemm2` way more clear. We can think of generating all configurations at compile time (e.g. creating a metafunction `combinations` that computes all combinations of given tuples representing the set of values for each parameter). We could also extract configuration in a more concise way, with something like this
```cpp
blas::Op trans_a, trans_b;
SizeType m, n, k, lda, ldb, ldc;

std::tie(trans_a, trans_b, m, n, k, lda, ldb, ldc) = test;
```
Moreover, If we don't want to test all possible configurations, but just a few, it is possible.

The only downside is that with very large configuration domains, it may be problematic for the compiler to generate all instances instead of letting the program generate them at runtime one per time.

PR #74, opened 2019-10-23
Add USE_MAIN=MPIHPX option for tests

### LINKED ISSUE #63: Add HPX+MPI option for DLAF_AddTest "USE_MAIN"
At the moment it is possible to add a test that uses HPX or MPI, but not both.

It would be useful for implementing tests for #58.

albestro:
Close #63 

Implement `MPIHPX` option for `USE_MAIN`. As discussed, MPI is initialized before HPX for simulating a context similar to the one in which the library will be employed, and also trying to avoid possible problems with physical resources binding.

Minor changes to `USE_MAIN`=`HPX` to keep it aligned with the new one.


albestro:
@rasolca 
As discussed, the required threading model for MPI with HPX is set to `MPI_THREAD_SERIALIZED`.

PR #82, opened 2019-11-07
Add a template build script

teonnik:
The script doesn't cover advanced options for MKL and LAPACK.

MKL:
```
MKL_LAPACK_INCLUDE_DIR
MKL_LAPACK_LIBRARY
MKL_THREADING

MKL_SCALAPACK_INCLUDE_DIR
MKL_SCALAPACK_LIBRARY
MKL_MPI_TYPE
```

LAPACK:

```
LAPACK_TYPE
LAPACK_INCLUDE_DIR
LAPACK_LIBRARY
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 build*
 .ptp-sync-folder
 .vscode/
+!build.sh
 
 
 # ------------------- C++
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -9,8 +9,8 @@ export CC=<FIXME: path to C compiler>
 export CXX=<FIXME: path to C++ compiler>
 export MKLROOT=<FIXME: path to the root installation directory of MKL>
 HPX_DIR=<FIXME: path to the root installation directory of HPX>
-BLASPP_DIR=<FIXME: path to the root installation directory of BLASPP>
-LAPACKPP_DIR=<FIXME: path to the root installation directory of LAPACKPP>
+blaspp_DIR=<FIXME: path to the root installation directory of BLASPP>
+lapackpp_DIR=<FIXME: path to the root installation directory of LAPACKPP>
 
 # CMAKE_BUILD_TYPE := Release | RelWithDebugInfo | Debug
 #
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# Template build script
+# ---------------------
+
+rm -rf CMakeCache.txt CMakeFiles
+
+export CC=<FIXME: path to C compiler>
+export CXX=<FIXME: path to C++ compiler>
+export MKLROOT=<FIXME: path to the root installation directory of MKL>
+HPX_DIR=<FIXME: path to the root installation directory of HPX>
+BLASPP_DIR=<FIXME: path to the root installation directory of BLASPP>
+LAPACKPP_DIR=<FIXME: path to the root installation directory of LAPACKPP>
+
+# CMAKE_BUILD_TYPE := Release | RelWithDebugInfo | Debug
+#
+# DLAF_WITH_CUDA   := ON | OFF (default)
+#
+# DLAF_WITH_MKL    := ON | OFF (default)
+#
+# BUILD_DOC        := ON | OFF (default)
+#
+# DLAF_WITH_TEST   := ON (default) | OFF
+#
+cmake <FIXME: path to DLA-Future src directory> \
+  -D DLAF_WITH_MKL=ON \
+  -D CMAKE_BUILD_TYPE=RelWithDebugInfo \
+
+make <FIXME: targets to build>


.gitignore
```diff
@@ -1,7 +1,7 @@
 .DS_Store
 *~
 *.swp
-build*
+build
```
albestro: Just a suggestion:

I would keep `build*` for ignoring all `build`, `build_debug`, `build_release`, `build_test` ...

If you don't want to call the script in another way (e.g. `configure_build.sh`, `generate_build.sh`), you may add a specific rule in `.gitignore` (i.e. `!build.sh`)

scripts/build.sh
```diff
@@ -0,0 +1,28 @@
...
+# DLAF_WITH_MKL    := ON | OFF (default)
+#
+# BUILD_DOC        := ON | OFF (default)
+#
+# DLAF_WITH_TEST   := ON (default) | OFF 
+#
+cmake ~/code/DLA-Future \
+  -D CMAKE_BUILD_TYPE=RelWithDebugInfo \
+  -D CMAKE_PREFIX_PATH="$HPX_ROOT;$BLASPP_ROOT;$LAPACKPP_ROOT" \
+  -D DLAF_WITH_MKL=ON \
```
albestro: I would try to be as close as possible to the cmake command line help.

```bash
  cmake [options] <path-to-source>
  cmake [options] <path-to-existing-build>
  cmake [options] -S <path-to-source> -B <path-to-build>
```
So, I would put options first and the paths after.

scripts/build.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+# DLAF_WITH_CUDA   := ON | OFF (default)
+#
+# DLAF_WITH_MKL    := ON | OFF (default)
+#
+# BUILD_DOC        := ON | OFF (default)
+#
+# DLAF_WITH_TEST   := ON (default) | OFF 
+#
+cmake ~/code/DLA-Future \
```
albestro: I would not assume that source code will always be in `$HOME/code`.

I would replace this absolute assumption with a relative one. A relative assumption may be that the source folder is the parent one. This is not the only option, we can evaluate to assume that the build folder is next to the source folder and not nested in, but we will need to make another assumption about the name of the source folder.

I prefer the first option. What do you think?

.gitignore
rasolca: Since the script performs just the cmake configuration and does not build anything I would call it `configure.sh`.

scripts/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# Template build script
+# ---------------------
+
+rm -rf CMakeCache.txt CMakeFiles
+
+export CXX=FIXME
```
rasolca: `CC` has to be defined as well.

.gitignore
teonnik: I placed a make call in the script and preserved the current name. I also added a `!build.sh` rule in `.gitignore` and restored `build*`

scripts/build.sh
teonnik: That was an error on my part, I didn't notice that I hadn't add FIXME for the path to the source directory. I don't think we should make any assumptions on where anyone might want to place the project. I will just leave it to the user to specify the path.

scripts/build.sh
```diff
@@ -0,0 +1,31 @@
...
+# DLAF_WITH_MKL    := ON | OFF (default)
+#
+# BUILD_DOC        := ON | OFF (default)
+#
+# DLAF_WITH_TEST   := ON (default) | OFF 
+#
+cmake <FIXME: path to DLA-Future src directory> \
+  -D DLAF_WITH_MKL=ON \
+  -D CMAKE_BUILD_TYPE=RelWithDebugInfo \
+  -D CMAKE_PREFIX_PATH="$HPX_ROOT;$BLASPP_ROOT;$LAPACKPP_ROOT" \
```
albestro: If you call the variables `_DIR` instead of `_ROOT`, it is not necessary to add them neither as CMake cache variable or in `CMAKE_PREFIX_PATH`.

In particular, variables names should be:
- `HPX_DIR`
- `blaspp_DIR`
- `lapackpp_DIR`

scripts/build.sh
```diff
@@ -0,0 +1,29 @@
...
+# ---------------------
+
+rm -rf CMakeCache.txt CMakeFiles
+
+export CC=<FIXME: path to C compiler>
+export CXX=<FIXME: path to C++ compiler>
+export MKLROOT=<FIXME: path to the root installation directory of MKL>
+HPX_DIR=<FIXME: path to the root installation directory of HPX>
+BLASPP_DIR=<FIXME: path to the root installation directory of BLASPP>
+LAPACKPP_DIR=<FIXME: path to the root installation directory of LAPACKPP>
```
albestro: Have you tested it?

Because, I'm not sure that for what concerns `BLASPP` and `LAPACKPP` that uppercase variables work (as I said in my previous comment, I use `blaspp_DIR` and `lapackpp_DIR`)

teonnik: I tested it: works on my laptop.

albestro: I did a test on my configuration with CMake 3.15.2 on Mac OS.

The variable `_DIR` must be called uppercase or lowercase depending on the name used in `find_package` call, e.g.
- `find_package(BLASPP)` requires `BLASPP_DIR`
- `find_package(blaspp)` requires `blaspp_DIR`

Since the original filename of the BLASPP library CMake config file is `blasppConfig.cmake`, we should keep the lowercase call, which forces us to use the lowercase version of the `blaspp_DIR`. Exactly the same stands for `LAPACKPP`.

PR #84, opened 2019-11-08
Add HPX build type check

albestro:
Add (missing) check about build type matching between our library and HPX.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,15 @@ endif()
 # ----- HPX
 find_package(HPX REQUIRED)
 
+# In HPX 1.3.0 there is compiler matching check, but it does not check matching of build types.
+# Since it is source of headaches, waiting for HPX to add this check, we check on our own
+if (NOT "${HPX_BUILD_TYPE}" STREQUAL "${CMAKE_BUILD_TYPE}")
+  message(WARNING
+  "CMAKE_BUILD_TYPE does not match HPX_BUILD_TYPE=${HPX_BUILD_TYPE}\n"
+  "This project uses ${CMAKE_BUILD_TYPE}, while HPX uses ${HPX_BUILD_TYPE}.\n"
+  "Please add -DCMAKE_BUILD_TYPE=${HPX_BUILD_TYPE} to your cmake command or use a debug build of HPX")
+endif()
+
 # ----- BLASPP/LAPACKPP
 find_package(OpenMP REQUIRED)
 find_package(blaspp REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,11 +64,18 @@ find_package(HPX REQUIRED)
 
 # In HPX 1.3.0 there is compiler matching check, but it does not check matching of build types.
 # Since it is source of headaches, waiting for HPX to add this check, we check on our own
-if (NOT "${HPX_BUILD_TYPE}" STREQUAL "${CMAKE_BUILD_TYPE}")
-  message(WARNING
-  "CMAKE_BUILD_TYPE does not match HPX_BUILD_TYPE=${HPX_BUILD_TYPE}\n"
-  "This project uses ${CMAKE_BUILD_TYPE}, while HPX uses ${HPX_BUILD_TYPE}.\n"
-  "Please add -DCMAKE_BUILD_TYPE=${HPX_BUILD_TYPE} to your cmake command or use a debug build of HPX")
+get_property(_MULTI_CONFIG_GENERATOR GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+if (NOT _MULTI_CONFIG_GENERATOR)
+  # They may not match just if none of them is a Debug build
+  if (  (NOT HPX_BUILD_TYPE STREQUAL CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -62,6 +62,15 @@ endif()
 # ----- HPX
 find_package(HPX REQUIRED)
 
+# In HPX 1.3.0 there is compiler matching check, but it does not check matching of build types.
+# Since it is source of headaches, waiting for HPX to add this check, we check on our own
+if (NOT "${HPX_BUILD_TYPE}" STREQUAL "${CMAKE_BUILD_TYPE}")
+  message(WARNING
+  "CMAKE_BUILD_TYPE does not match HPX_BUILD_TYPE=${HPX_BUILD_TYPE}\n"
+  "This project uses ${CMAKE_BUILD_TYPE}, while HPX uses ${HPX_BUILD_TYPE}.\n"
+  "Please add -DCMAKE_BUILD_TYPE=${HPX_BUILD_TYPE} to your cmake command or use a debug build of HPX")
```
rasolca: This message is a bit strange. HPX_BUILD_TYPE is repeated multiple times.
Moreover using Release with a Debug HPX, it suggests to use a Debug HPX.
I suggest:
"The build type used (CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}) does not match with HPX build type (HPX_BUILD_TYPE=${HPX_BUILD_TYPE})."
"Please either change CMAKE_BUILD_TYPE to ${HPX_BUILD_TYPE} or use a ${CMAKE_BUILD_TYPE} build of HPX"

teonnik:
Why would we want to match the build type of HPX? If I compile HPX with `RelWithDebugInfo` but would still like to compile DLA-Future with `Debug`, would that work under your PR?

In COSMA, I used this to set the default build type and to check whether a multi-config generator is not used (if it is, CMAKE_BUILD_TYPE has not effect). Perhaps this will be helpful

```
set(default_build_type ${CMAKE_BUILD_TYPE})
if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
    message(STATUS "Setting build type to 'Release' as none was specified.")
    set(default_build_type "Release")
endif()
set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel Profile." FORCE)
```

albestro:
> Why would we want to match the build type of HPX? If I compile HPX with `RelWithDebugInfo` but would still like to compile DLA-Future with `Debug`, would that work under your PR?

The problem with build types is that it happens more often than not to mix DLAF/Release with HPX/Debug (or viceversa): it builds, it runs...but the behavior is undefined, it is generally a long wait without messages, and if you are debugging, it could take times before you realize that you mixed up build types.

About your concern, I think it would work because it is a WARNING message and not an ERROR. so it still allows you to do bad thing, but it warns you at least.

rasolca:
> Why would we want to match the build type of HPX? If I compile HPX with `RelWithDebugInfo` but would still like to compile DLA-Future with `Debug`, would that work under your PR?

This will not work as HPX will break. On the other hand mixing `Release` and `RelWithDebugInfo` should be allowed.

teonnik:
I guess you are referring to this ([HPX docs](https://stellar-group.github.io/hpx/docs/sphinx/latest/html/manual/building_hpx.html#build-types)):

> We currently don’t guarantee ABI compatibility between Debug and Release builds. Please make sure that applications built against HPX use the same build type as you used to build HPX. For CMake builds, this means that the CMAKE_BUILD_TYPE variables have to match and for projects not using CMake, the HPX_DEBUG macro has to be set in debug mode.

I agree with the PR: 

1) by default the DLA-Future build type should match the HPX build type; 
2) there should be a warning if they don't.

CMakeLists.txt
```diff
@@ -62,6 +62,15 @@ endif()
 # ----- HPX
 find_package(HPX REQUIRED)
 
+# In HPX 1.3.0 there is compiler matching check, but it does not check matching of build types.
+# Since it is source of headaches, waiting for HPX to add this check, we check on our own
+if (NOT "${HPX_BUILD_TYPE}" STREQUAL "${CMAKE_BUILD_TYPE}")
```
teonnik: Perhaps this can be improved to not be triggered for multi-config generators:

`if (NOT "${HPX_BUILD_TYPE}" STREQUAL "${CMAKE_BUILD_TYPE}" AND NOT CMAKE_CONFIGURATION_TYPES)`

albestro: What about [GENERATOR_IS_MULTI_CONFIG](https://cmake.org/cmake/help/v3.16/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html)?

albestro:
In parallel I'm trying to propose a fix in HPX.

I have changed the check from a simple exact match to a bit more advanced one that does not allow mix Release and Debug builds (e.g. it would be possible to mix Release with RelWithDebInfo but not with Debug).

Please, check if it is still ok for you.

albestro:
In the meanwhile HPX issue is going to be resolved.

See https://github.com/STEllAR-GROUP/hpx/pull/4205

PR #85, opened 2019-11-11
Use FindMKL from cmake-recipes

teonnik:
The PR replaces the existing FindMKL module with the one from `cmake-recipes` : https://github.com/eth-cscs/cmake-recipes/pull/8 . I was having linking issues on my laptop with the previous FindMKL (as discussed on Slack).

### DIFF
--- a/cmake/FindMKL.cmake
+++ b/cmake/FindMKL.cmake
@@ -1,356 +1,309 @@
-#
-# Distributed Linear Algebra with Future (DLAF)
+# Source: https://github.com/eth-cscs/cmake-recipes
 #
 # Copyright (c) 2018-2019, ETH Zurich
-# All rights reserved.
+# BSD 3-Clause License. All rights reserved.
 #
-# Please, refer to the LICENSE file in the root directory.
-# SPDX-License-Identifier: BSD-3-Clause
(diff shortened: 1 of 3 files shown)


cmake/FindMKL.cmake
```diff
@@ -1,356 +1,302 @@
-#
-# Distributed Linear Algebra with Future (DLAF)
-#
-# Copyright (c) 2018-2019, ETH Zurich
-# All rights reserved.
-#
-# Please, refer to the LICENSE file in the root directory.
-# SPDX-License-Identifier: BSD-3-Clause
-#
+#[=======================================================================[.rst:
```
rasolca: What's this line for?

teonnik: That's the usual way of writing documentation in Find modules : https://github.com/Kitware/CMake/blob/master/Modules/FindOpenMP.cmake

cmake/FindMKL.cmake
```diff
@@ -1,356 +1,309 @@
-#
-# Distributed Linear Algebra with Future (DLAF)
+# Copied from https://github.com/eth-cscs/cmake-recipes
```
rasolca: I would use `# Source: https://github.com/eth-cscs/cmake-recipes`

PR #91, opened 2019-11-21
Cholesky local 2

### LINKED ISSUE #57: Add Choleksy factorization method
Implement the first Linear Algebra algorithm of the library.

We have experience from on this, see the prototype code available in the private repository [here](https://github.com/eth-cscs/hpx_mpi_linalg/blob/master/include/hpx_mpi/cholesky2.h) for reference.

It will exploit the HPX library for implementing the task-based algorithm; it will work in-place so the input/output of the algorithm is a `MatrixRW` (#20).

Sely85:
Close #57 

First implementation of Cholesky factorization method, for CPU, using Matrix objects. 

Added "util_matrix.h" to check if Matrix and Block are squared.

Added tests for Correctness, NoSquareMatrixException and NoSquareBlockException (NonPositiveDefiniteException not done yet).


include/dlaf/util_matrix.h
```diff
@@ -0,0 +1,50 @@
...
+/// @file
+
+namespace dlaf {
+namespace util_matrix {
+
+/// @brief Verify if dlaf::Matrix is square
+///
+/// @tparam M refers to a dlaf::Matrix object
+/// @throws std::invalid_argument if the matrix is not squared
+/// @pre dlaf::Matrix sizes >= 0
```
rasolca: This condition is always fullfilled by dlaf::Matrix.

include/dlaf/util_matrix.h
```diff
@@ -0,0 +1,50 @@
...
+  assert(matrix.size().isValid());
+  if (matrix.size().rows() != matrix.size().cols())
+    throw std::invalid_argument(function + ": " + "Matrix " + mat_name + " is not square.");
+}
+
+/// @brief Verify if dlaf::Matrix tile is square
+///
+/// @tparam M refers to a dlaf::Matrix object
+/// @throws std::invalid_argument if the matrix block is not squared
+/// @pre dlaf::Matrix sizes > 0
```
rasolca: See above.

include/dlaf/mc/cholesky_local.h
```diff
@@ -0,0 +1,117 @@
+// Linear Algebra with Future (DLAF)
```
rasolca: The header is the same for all the files. It can be found in `misc/HEADER`

include/dlaf/mc/cholesky_local.h
```diff
@@ -0,0 +1,117 @@
...
+#include "dlaf/matrix/layout_info.h"
+#include "dlaf/util_blas.h"
+#include "dlaf/util_matrix.h"
+//
+#include <exception>
+//
+#include <hpx/runtime/threads/run_as_hpx_thread.hpp>
+#include "hpx/hpx.hpp"
+#include "hpx/include/parallel_executors.hpp"
+#include "hpx/include/threads.hpp"
```
rasolca: Not all headers are needed. Please check and remove non used includes

include/dlaf/mc/cholesky_local.h
```diff
@@ -0,0 +1,117 @@
...
+/// @param uplo specifies whether the matrix is \a Upper or \a Lower triangular
+/// @param side defines whether the matrix is on the \a Left or \a Right side of the unknown matrix
+/// @param op specifies the form of \p mat used in matrix multiplication: \a NoTrans, \a Trans, \a ConjTrans
+/// @param diag describes if the matrix is unit triangular (\a Unit) or not (\a NonUnit)
+///
+/// @throws std::runtime_error if \p uplo = \a Upper decomposition  is chosen (not yet implemented)
+
+template <class T>
+void cholesky_local(Matrix<T, Device::CPU>& mat, blas::Uplo uplo, blas::Side side, blas::Op op,
+                    blas::Diag diag) {
```
rasolca: The signature should be:
`void cholesky_local(blas::Uplo uplo, Matrix<T, Device::CPU>& mat)`

include/dlaf/mc/cholesky_local.h
```diff
@@ -0,0 +1,117 @@
...
+  util_matrix::check_blocksize_square(mat, "Cholesky", "mat");
+
+  // Number of tile (rows = cols)
+  SizeType nrtile = mat.nrTiles().cols();
+
+  // k-loop
+  for (SizeType k = 0; k < nrtile; ++k) {
+    /// A Cholesky decomposition with the tile::potrf function is done on the diagonal tiles.
+    // Cholesky decomposition on mat(k,k) r/w potrf (lapack operation)
+    if (uplo == blas::Uplo::Lower) {
```
rasolca: This check is for the full Cholesky execution (lines 62-114).
There will be 2 cases Upper and Lower which requires two different implementations.

test/unit/mc/test_cholesky.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+  for (const auto& size : square_sizes) {
+    for (const auto& block_size : square_block_sizes) {
+      // Matrix to undergo Cholesky decomposition
+      Matrix<Type, Device::CPU> mat(size, block_size);
+      set(mat, el);
+
+      // Matrix with analytical results
+      Matrix<Type, Device::CPU> matres(size, block_size);
+      set(matres, res);
```
rasolca: Not used.

test/unit/mc/test_cholesky.cpp
```diff
@@ -0,0 +1,152 @@
...
+  for (const auto& size : square_sizes) {
+    for (const auto& block_size : square_block_sizes) {
+      Matrix<Type, Device::CPU> mat(size, block_size);
+
+      set(mat, el);
+
+      EXPECT_NO_THROW(cholesky_local(mat, blas::Uplo::Lower, blas::Side::Right, blas::Op::ConjTrans,
+                                     blas::Diag::NonUnit));
+    }
+  }
```
rasolca: This part is not needed. It is already tested by `Correctness` test.

test/unit/mc/test_cholesky.cpp
```diff
@@ -0,0 +1,152 @@
...
```
rasolca: See above.

test/unit/mc/test_cholesky.cpp
```diff
@@ -0,0 +1,152 @@
...
+//
+// Copyright (c) 2018-2019, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+#include "dlaf/mc/cholesky_local.h"
+
+#include <typeinfo>
```
albestro: What symbol(s) do you need from `<typeinfo>`?

test/unit/mc/test_cholesky.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include "gtest/gtest.h"
+#include "dlaf/matrix/index.h"
+#include "dlaf/matrix/layout_info.h"
```
albestro: Are they really needed?

include/dlaf/mc/cholesky_local.h
```diff
@@ -0,0 +1,117 @@
...
+                    blas::Diag diag) {
+  /// First, two executors are set up on the default queue: with high and with default priority.
+
+  // Set up executor on the default queue with high priority.
+  hpx::threads::scheduled_executor matrix_HP_executor =
+      hpx::threads::executors::pool_executor("default", hpx::threads::thread_priority_high);
+
+  // Set up executor on the default queue with default priority.
+  hpx::threads::scheduled_executor matrix_normal_executor =
+      hpx::threads::executors::pool_executor("default", hpx::threads::thread_priority_default);
```
albestro: These are not used.

include/dlaf/mc/cholesky_local.h
```diff
@@ -0,0 +1,117 @@
...
+  /// Then the matrix and its sublocks are tested to confirm they are square.
+  // Check if matrix is square
+  util_matrix::check_size_square(mat, "Cholesky", "mat");
+  // Check if block matrix is square
+  util_matrix::check_blocksize_square(mat, "Cholesky", "mat");
+
+  // Number of tile (rows = cols)
+  SizeType nrtile = mat.nrTiles().cols();
+
+  // k-loop
```
albestro: IMHO not very useful comments. (see also j-loop, i-loop, internal i-loop).

test/unit/mc/test_cholesky.cpp
Sely85: Should I check square sizes in `Correctness` or is it better to check this in `NoSquareMatrixException`?

rasolca: The check that the function works for square matrices is implicit in the Correctness check.
(The correctness test checks if the function produces the correct result without throwing).

Moreover this test is likely to fail since the matrix used is not positive definite.

include/dlaf/mc/cholesky_local.h
```diff
@@ -0,0 +1,117 @@
...
+/// @param side defines whether the matrix is on the \a Left or \a Right side of the unknown matrix
+/// @param op specifies the form of \p mat used in matrix multiplication: \a NoTrans, \a Trans, \a ConjTrans
+/// @param diag describes if the matrix is unit triangular (\a Unit) or not (\a NonUnit)
+///
+/// @throws std::runtime_error if \p uplo = \a Upper decomposition  is chosen (not yet implemented)
+
+template <class T>
+void cholesky_local(Matrix<T, Device::CPU>& mat, blas::Uplo uplo, blas::Side side, blas::Op op,
+                    blas::Diag diag) {
+  /// First, two executors are set up on the default queue: with high and with default priority.
```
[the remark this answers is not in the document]
Sely85: comment

They are, but don't actually look awesome. Should I cancel them?

include/dlaf/util_matrix.h
```diff
@@ -0,0 +1,50 @@
...
+namespace dlaf {
+namespace util_matrix {
+
+/// @brief Verify if dlaf::Matrix is square
+///
+/// @tparam M refers to a dlaf::Matrix object
+/// @throws std::invalid_argument if the matrix is not squared
+/// @pre dlaf::Matrix sizes >= 0
+template <class Matrix>
+void check_size_square(const Matrix& matrix, std::string function, std::string mat_name) {
```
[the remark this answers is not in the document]
Sely85: Ok, I will use `assert` instead of `check`.

test/unit/mc/test_cholesky.cpp
Sely85: You're right. I don't need it

include/dlaf/mc/cholesky_local.h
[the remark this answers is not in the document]
albestro: IMHO those are implementation details that are not worth being mentioned in the public documentation.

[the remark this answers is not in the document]
Sely85: Ok :)

include/dlaf/mc/cholesky_local.h
```diff
@@ -0,0 +1,90 @@
...
+
+  // Number of tile (rows = cols)
+  SizeType nrtile = mat.nrTiles().cols();
+
+  if (uplo == blas::Uplo::Lower) {
+    for (SizeType k = 0; k < nrtile; ++k) {
+      // Cholesky decomposition on mat(k,k) r/w potrf (lapack operation)
+
+      hpx::dataflow(matrix_HP_executor, hpx::util::unwrapping(tile::potrf<T, Device::CPU>), uplo,
+                    std::move(mat({k, k})));
```
albestro: This is absolutely not wrong, it is just a suggestion.

Indices are strongly-typed with the aim to provide an additional implicit check of the index passed to a function , and brace initialization is a (legit) way to work around this check.

In order to profit from this check, I would suggest you specify the type of index you are creating, e.g. `LocalTileIndex{k, k}`.

About `{k, k}` in particular, you can also extract it in a variable and re-use it in the algorithm.

include/dlaf/mc/cholesky_local.h
```diff
@@ -0,0 +1,90 @@
...
+
+      for (SizeType j = k + 1; j < nrtile; ++j) {
+        // Choose queue priority
+        auto trailing_matrix_executor = matrix_normal_executor;
+        if (j == k + 1) {
+          trailing_matrix_executor = matrix_HP_executor;
+        }
+        else {
+          trailing_matrix_executor = matrix_normal_executor;
+        }
```
albestro: Am I wrong saying that the `else` branch is not very useful?

Moreover, it seems that the priority is set differently just for the first iteration, maybe this code can be changed to not use the `if`. Anyway, it is not a performance problem here.

rasolca: I would use `auto trailing_matrix_executor = j == k+1 ? matrix_HP_executor : matrix_normal_executor;`

test/unit/mc/test_cholesky.cpp
```diff
@@ -0,0 +1,105 @@
...
+  for (const auto& size : square_sizes) {
+    for (const auto& block_size : rectangular_block_sizes) {
+      Matrix<Type, Device::CPU> mat(size, block_size);
+
+      EXPECT_THROW(cholesky_local(blas::Uplo::Lower, mat), std::invalid_argument);
+    }
+  }
+}
+
+TYPED_TEST(CholeskyTest, NonPositiveDefiniteException) {}
```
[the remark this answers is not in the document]
Sely85: I use it as a reminder to add that test. Btw I can remove it for now!

include/dlaf/mc/cholesky_local.h
Sely85: I'll definitely follow your advice, to perform the additional implicit check of the index. Thanks!

include/dlaf/util_matrix.h
```diff
@@ -0,0 +1,44 @@
...
+#include <string>
+
+/// @file
+
+namespace dlaf {
+namespace util_matrix {
+
+/// @brief Verify if dlaf::Matrix is square
+///
+/// @param M refers to a dlaf::Matrix object
```
albestro: Sorry, my comment was not clear.

I think that with `M` you are referring to the template parameter `Matrix`.

i.e. `/// @tparam Matrix refers to a dlaf::Matrix object`

Same applies to the documentation of next function.

Sely85: Nope, the comment was clear. My fault, sorry.

test/unit/mc/test_cholesky.cpp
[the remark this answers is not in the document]
rasolca: Please remove it. Otherwise the test log shows the test as passed.

PR #95, opened 2019-11-25
Implementation of Pipeline object

### LINKED ISSUE #94: Serialize communications
In task-based algorithms, it would be nice to have a mechanism to "futurize" the communications.

Such a method would allow defining the order of operations in the tasks DAG, so that it would not allow executing communication operations in an order different from the one in which they are submitted.

The basic idea is to have an `hpx::future<Communicator>` so that HPX can take care of the scheduling. Moreover, it would be nice if this future becomes ready by itself, as soon as the communication is completed.

albestro:
Close #94 

Proposal for `Pipeline` object with its basic test.

`Pipeline` is initialized by moving in an object on which it creates a ready future and the only method it exposes is the call operator `Pipeline<T>::operator()()` that returns an `hpx::future`.

Internally, at every call, it chains all the futures, so that they will become ready in a serialized and ordered way. In this chain, the object given at initialization time is moved (i.e. `std::move`) from future to future as soon as they get ready, by means of the `Wrapper` object.
The `Wrapper` object is used to contain the real object (the moved one) and to make ready the next future as soon as it is destroyed. A reference to the object can be accessed through the call operator`Wrapper<T>::operator()()`.

Example of usage:
```cpp
dlaf::comm::Communicator world(MPI_COMM_WORLD);
Pipeline<dlaf::comm::Communicator> serial_comm(std::move(world));

// ...

hpx::dataflow(hpx::util::unwrapping(
  [](auto&& tile, auto&& comm_wrapper) {
    dlaf::comm::sync::broadcast::send(comm_wrapper(), tile);
  }),
  matrix.read({0, 0}), serial_comm());
```

There are a few things to point out:
- `serial_comm()` is the call to get the future from the `Pipeline`, on which the dataflow will depend on, together with the data (i.e. `matrix.read(...)`);
- The function that will be executed by the dataflow when the futures will become ready will get a `Wrapper<Communicator>`, so when the `Communicator` object is necessary, in this example for the broadcast function, `comm_wrapper()` is used;

Inside the dataflow function, the `Wrapper` will go out of scope at the end of the function, and it will trigger the next future.


rasolca:
Looks good.

albestro:
What about the specialization for `Tile`? is it ok like this for now?

PR #96, opened 2019-11-25
Implement Pool of resources

### LINKED ISSUE #93: Implement a way for workspaces re-utilization
Development of a distributed algorithm, like Cholesky factorization, may need to temporarily store data from other ranks. After usage, data are not useful anymore, but the algorithm may require other data in a next phase of the algorithm. With the term *workspace*, we refer to memory areas we use in our algorithms, but they are useful just for a limited phase of the computation. Often, working block-by-block, it comes out that workspaces are always of the same size (e.g. a Tile).

With the aim to avoid superfluous re-allocation of these memory areas, it would be nice to have a pool of pre-allocated workspaces, that can be acquired to be used in a particular phase of the algorithm. And then, as soon as they are not used anymore, they become available for the next usage.

albestro:
Close #93 

Proposal for `Pool` object and its test.

`Pool` is a container of objects that can be requested for exclusive usage and then released, allowing re-usage of the resource without concurrency.

`Pool` exposes the method `get`, which is the way to request access to the first available resource of the pool. This does not return directly the object, instead, it returns a `future` to the resource. This allows constraining the execution of a task not just on the data but also on the availability of the temporary resource.

Releasing of the resource is managed by means of a `Wrapper` object, that on destruction takes care of releasing the resource and to give back control to the `Pool`. This also means that the object needs to be retrieved by using the call operator on the `Wrapper` object.

Example of usage:
```cpp
Pool<dlaf::Tile, 2> pool;

// ...

hpx::dataflow(hpx::util::unwrapping(
  [](auto&& matrix_tile, auto&& tile_workspace_wrapper) {
    dlaf::Tile tile_workspace = tile_workspace_wrapper();
    // ... do something with the workspace
  }),
  matrix.read({0, 0}), pool.get());
```


albestro:
Currently, it is implemented as a fixed size container, with the size specified at compile time.

I would keep `Pool` a fixed size container, but it may be better to have this parameter passed as a simple argument and not as a template argument.

rasolca:
> Currently, it is implemented as a fixed size container, with the size specified at compile time.
> 
> I would keep `Pool` a fixed size container, but it may be better to have this parameter passed as a simple argument and not as a template argument.

I don't understand what you mean with fixed size container. If it means that the Pool is created with a certain size and cannot be resized I agree.

Anyway, I think that there should be the possibility to choose the size at runtime (as a parameter of the constructor).

PR #99, opened 2019-11-26
Reduce compile times and build in release mode

teonnik:
- Only compile necessary libraries for Boost
- Build HPX, lapackpp and blaspp in release mode

### DIFF
--- a/spack.yaml
+++ b/spack.yaml
@@ -1,8 +1,9 @@
 spack:
   specs:
+    - hpx build_type=Release cxxstd=14 networking=none
+    - boost cxxstd=14 ~atomic~math~iostreams~locale~graph~random~serialization~signals~timer~test~wave~mpi~python
     - mpich
     - intel-mkl
-    - hpx cxxstd=14 networking=none
-    - blaspp
-    - lapackpp
+    - blaspp build_type=Release
+    - lapackpp build_type=Release
   concretization: together


spack.yaml
```diff
@@ -1,8 +1,9 @@
 spack:
   specs:
+    - hpx build_type=Release cxxstd=14 networking=none
+    - boost cxxstd=14 ~atomic~math~iostreams~locale~graph~random~serialization~signals~timer~test~wave~mpi~python
```
albestro: I'm not sure about this.
Since `boost` is an `HPX` dependency and not our one, shouldn't we rely on `HPX` spack package for this?

teonnik: I can rewrite the above as `hpx ... ^boost ...`, however the line becomes rather long. Since we have `  concretization: together`, the approach is equivalent to the one in the PR.

albestro: Ok, with the offline chat we had I understasnd why you prefer to explicitly declare the boost dependency like this.

albestro: This dependency is maybe not enough on all systems. 

See https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_SetupBoost.cmake

From it, I was able to figure out that components
- `context`
- `thread`
- `chrono`

are required on `APPLE`, `BLUEGENEQ`, and `ARM`.

I don't think that adding them by default my affect HPX decision about which `coroutine` to use, but I would check it (or ask some HPX guy) because it may affect performances.

PR #114, opened 2019-12-11
Cholesky mc

Sely85:
First implementation of Cholesky method for CPU, using Matrix object.
Contains methods for both local and distributed memory. Files with separated implementations have been removed. 

Tests for both local and distributed memory merged in one file. Tests developed for:
- Correctness
- NoSquareMatrixException
- NoSquareBlockException
(NonPositiveDefiniteException test still to be implemented.)


include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,309 @@
...
+                    }),
+                k_rank_col, mat.tileSize(GlobalTileIndex(i, k)), serial_comm());
+          }
+        }
+      }
+
+      for (SizeType j_local = distr.nextLocalTileFromGlobalTile<Coord::Col>(k + 1);
+           j_local < localnrtile_cols; ++j_local) {
+        // Choose "high priority" for first tile of the trailing matrix
+        auto trailing_matrix_executor = (j_local == k + 1) ? matrix_HP_executor : matrix_normal_executor;
```
rasolca: This line is wrong. you are comparing a local index (`j_local`) with a global index `k+1`:
The option are two in this case:
Either only the global k+1 column is high priority, or the next column of each rank is high priority. We should check which performs best.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,309 @@
...
+        }
+      }
+      else {
+        for (SizeType i_local = distr.nextLocalTileFromGlobalTile<Coord::Row>(k + 1);
+             i_local < localnrtile_rows; ++i_local) {
+          auto i = distr.globalTileFromLocalTile<Coord::Row>(i_local);
+
+          // Avoid useless communications if one-row communicator grid
+          if (row_comm_size > 1) {
+            // Update the panel row-wise
```
rasolca: Wrong comment again... Here you are receiving the panel.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,309 @@
...
+  //    mpi_executor = mpi_exec;
+  //   }
+  //  else {
+  //    mpi_executor = matrix_HP_executor;
+  //  }
+
+  // Check if matrix is square
+  util_matrix::assert_size_square(mat, "Cholesky", "mat");
+  // Check if block matrix is square
+  util_matrix::assert_blocksize_square(mat, "Cholesky", "mat");
```
rasolca: A check about the compatibility of the communicator grid and the distribution is missing.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,309 @@
...
+///
+/// @param uplo specifies that the matrix is \a Lower triangular
+/// @tparam mat refers to a dlaf::Matrix object
+///
+/// @throws std::runtime_error if \p uplo = \a Upper decomposition is chosen (not yet implemented)
+
+template <class T>
+void cholesky(blas::Uplo uplo, Matrix<T, Device::CPU>& mat) {
+  // Check that matrix is on local memory
+  if (mat.distribution().commGridSize() == comm::Size2D{1, 1}) {
```
rasolca: Why this check has not been added as the square size checking?

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,309 @@
...
+
+  auto localnrtile_rows = distr.localNrTiles().rows();
+  auto localnrtile_cols = distr.localNrTiles().cols();
+
+  dlaf::common::Pipeline<comm::CommunicatorGrid> serial_comm(std::move(grid));
+
+  // Method only for Lower triangular matrix
+  if (uplo == blas::Uplo::Lower) {
+    for (SizeType k = 0; k < nrtile; ++k) {
+      // Create the panel as a vector of shared future
```
rasolca: The panel is not created here. The vector is just a placeholder in which the shared futures representing the panel will be added.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,309 @@
...
+      auto k_rank_row = distr.rankGlobalTile<Coord::Row>(k);
+      auto k_rank_col = distr.rankGlobalTile<Coord::Col>(k);
+
+      if (mat.rankIndex().col() == k_rank_col) {
+        auto k_local_col = distr.localTileFromGlobalTile<Coord::Col>(k);
+
+        hpx::shared_future<Tile<const T, Device::CPU>> kk_tile;
+
+        if (mat.rankIndex().row() == k_rank_row) {
+          // If the diagonal tile is on this node factorize it
```
rasolca: Which part of the code does this comment refer to?

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,309 @@
...
+        auto trailing_matrix_executor = (j_local == k + 1) ? matrix_HP_executor : matrix_normal_executor;
+
+        hpx::shared_future<Tile<const T, Device::CPU>> col_panel;
+
+        auto j = distr.globalTileFromLocalTile<Coord::Col>(j_local);
+
+        auto j_rank_row = distr.rankGlobalTile<Coord::Row>(j);
+
+        if (mat.rankIndex().row() == j_rank_row) {
+          // Check if the diagonal tile of the trailing matrix is on this node
```
rasolca: Which part of the code does this comment refer to?

include/dlaf/mc/cholesky.h
Sely85: you're right.. I'm gonna add it to `util_matrix.h`

include/dlaf/util_matrix.h
```diff
@@ -40,5 +40,31 @@ void assert_blocksize_square(const Matrix& matrix, std::string function, std::st
                                 " is not square.");
 }
 
+/// @brief Verify if dlaf::Matrix is on local memory
```
albestro: IMHO it is not clear, I had to look at the code to understand it.

You are checking if a matrix is not distributed, so it is entirely stored locally and elements are not shared with other ranks, right?

include/dlaf/util_matrix.h
```diff
@@ -40,5 +40,31 @@ void assert_blocksize_square(const Matrix& matrix, std::string function, std::st
...
+/// @tparam Matrix refers to a dlaf::Matrix object
+/// @throws std::invalid_argument if the matrix is not stored on local memory
+template <class Matrix>
+void assert_local_matrix(const Matrix& matrix, std::string function, std::string mat_name) {
+  assert(matrix.size().isValid());
+  if (matrix.distribution().commGridSize() != comm::Size2D{1, 1})
+    throw std::invalid_argument(function + ": " + "Matrix " + mat_name + " is not local.");
+}
+
+/// @brief Verify that distributed dlaf::Matrix correspond to the one in the Communicator
```
albestro: typo: *corresponds*

However, I would try to be more specific. It is not clear what `correspond` means, I would give more details, e.g. "their sizes and rank ids must match".

include/dlaf/util_matrix.h
```diff
@@ -40,5 +40,31 @@ void assert_blocksize_square(const Matrix& matrix, std::string function, std::st
...
+/// @throws std::invalid_argument if the matrix does not correspond to that of the Communicator
+template <class Matrix>
+void assert_comm_distr(const comm::CommunicatorGrid& grid, const Matrix& matrix, std::string function,
+                       std::string mat_name, std::string comm_name) {
+  assert(matrix.size().isValid());
+
+  if ((matrix.distribution().commGridSize() != grid.size()) &&
+      (matrix.distribution().rankIndex() != grid.rank()))
+    throw std::invalid_argument(function + ": " + "Distributed matrix " + mat_name +
+                                " and communicator grid " + comm_name + " are not compatibile.");
```
albestro: typo: *compatible*

include/dlaf/util_matrix.h
```diff
@@ -40,5 +40,31 @@ void assert_blocksize_square(const Matrix& matrix, std::string function, std::st
                                 " is not square.");
 }
 
+/// @brief Verify if dlaf::Matrix is on local memory
+///
+/// @tparam Matrix refers to a dlaf::Matrix object
+/// @throws std::invalid_argument if the matrix is not stored on local memory
+template <class Matrix>
+void assert_local_matrix(const Matrix& matrix, std::string function, std::string mat_name) {
+  assert(matrix.size().isValid());
```
albestro: I'm not so sure that this assert has to be here.

Why do you need this check in debug?

include/dlaf/util_matrix.h
```diff
@@ -40,5 +40,31 @@ void assert_blocksize_square(const Matrix& matrix, std::string function, std::st
...
+  if (matrix.distribution().commGridSize() != comm::Size2D{1, 1})
+    throw std::invalid_argument(function + ": " + "Matrix " + mat_name + " is not local.");
+}
+
+/// @brief Verify that distributed dlaf::Matrix correspond to the one in the Communicator
+///
+/// @tparam Matrix refers to a dlaf::Matrix object
+/// @throws std::invalid_argument if the matrix does not correspond to that of the Communicator
+template <class Matrix>
+void assert_comm_distr(const comm::CommunicatorGrid& grid, const Matrix& matrix, std::string function,
```
albestro: IMHO, this name is not so expressive. Something similar to `assert_compatible_communicator`? You can absolutely find a better name than mine! 😁

include/dlaf/matrix/distribution.h
```diff
@@ -268,10 +268,10 @@ class Distribution {
   /// whose index is the smallest index larger or equal @p global_tile
   /// and which is stored in current process.
   ///
-  /// @pre 0 <= global_tile < nrTiles().get<rc>()
+  /// @pre 0 <= global_tile <= nrTiles().get<rc>()
   template <Coord rc>
   SizeType nextLocalTileFromGlobalTile(SizeType global_tile) const noexcept {
-    assert(0 <= global_tile && global_tile < global_nr_tiles_.get<rc>());
+    assert(0 <= global_tile && global_tile <= global_nr_tiles_.get<rc>());
```
albestro: Do you need the equal because it can return an index to a tile "out of borders", right?

rasolca: This function is used to find loop limits, therefeore it should be possible to call it with also the index of the last+1 element.

include/dlaf/util_matrix.h
rasolca: Side note: The size of a matrix is always valid. The only way to have an invalid size is to manipulate the memory in which the matrix members are stored.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+///
+/// @param uplo specifies that the matrix is \a Lower triangular
+/// @tparam mat refers to a dlaf::Matrix object
+///
+/// @throws std::runtime_error if \p uplo = \a Upper decomposition is chosen (not yet implemented)
+
+template <class T>
+void cholesky(blas::Uplo uplo, Matrix<T, Device::CPU>& mat) {
+  // Set up executor on the default queue with high priority.
+  hpx::threads::scheduled_executor matrix_HP_executor =
```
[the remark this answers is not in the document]
rasolca: These names were introduced by John in the prototype.
I agree that they can be improved, but we should keep the distinction between the MPI and the other executors.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+      hpx::threads::executors::pool_executor("default", hpx::threads::thread_priority_high);
+
+  // Set up executor on the default queue with default priority.
+  hpx::threads::scheduled_executor matrix_normal_executor =
+      hpx::threads::executors::pool_executor("default", hpx::threads::thread_priority_default);
+
+  // Check if matrix is square
+  util_matrix::assert_size_square(mat, "Cholesky", "mat");
```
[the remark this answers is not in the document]
rasolca: AFAIK `__PRETTY_FUNCTION__/__FUNCTION__` are not standard, and `__func__` is hard to read.
Anyway, we can improve the exception messages later.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+          // gemm (blas operation)
+          hpx::dataflow(trailing_matrix_executor, hpx::util::unwrapping(tile::gemm<T, Device::CPU>),
+                        blas::Op::NoTrans, blas::Op::ConjTrans, -1.0, panel[i_local], col_panel, 1.0,
+                        std::move(mat(LocalTileIndex{i_local, j_local})));
+        }
+      }
+    }
+  }
+  else {
+    throw std::runtime_error("uplo = Upper not yet implemented");
```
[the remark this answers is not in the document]
rasolca: Can be a different task.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+
+  // Set up an executor on the mpi pool --> This part need to be added
+  //  hpx::threads::scheduled_executor mpi_executor;
+  //  if (use_pools && m_size > 1) {
+  //    hpx::threads::executors::pool_executor mpi_exec("mpi");
+  //    mpi_executor = mpi_exec;
+  //   }
+  //  else {
+  //    mpi_executor = matrix_HP_executor;
+  //  }
```
albestro: I'm aware this is a work in progress, but I would not add commented code.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+          // Avoid useless communication if one-column communicator and if on the last column
+          if (col_comm_size > 1 && k != (mat.size().cols() - 1)) {
+            // Broadcast the panel column-wise
+            hpx::dataflow(hpx::util::unwrapping([](auto&& tile, auto&& comm_wrapper) {
+                            dlaf::comm::sync::broadcast::send(comm_wrapper().colCommunicator(), tile);
+                          }),
+                          mat.read(kk), serial_comm());
+          }
+
+          kk_tile = mat.read(kk);
```
albestro: You can return the tile directly from the lambda.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+
+        hpx::shared_future<Tile<const T, Device::CPU>> col_panel;
+
+        auto j_rank_row = distr.rankGlobalTile<Coord::Row>(j);
+
+        if (mat.rankIndex().row() == j_rank_row) {
+          auto i_local = distr.localTileFromGlobalTile<Coord::Row>(j);
+
+          // Avoid useless communications if one-row communicator grid and if on the last panel
+          if (col_comm_size > 1 && j != (mat.size().cols() - 1)) {
```
albestro: Look similar previous comment.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+          // mat.read(j,k), using herk (blas operation)
+          hpx::dataflow(trailing_matrix_executor, hpx::util::unwrapping(tile::herk<T, Device::CPU>),
+                        uplo, blas::Op::NoTrans, -1.0, panel[i_local], 1.0,
+                        mat(LocalTileIndex{i_local, j_local}));
+
+          col_panel = panel[i_local];
+        }
+        else {
+          // Avoid useless communications if one-row communicator grid and if on the last panel
+          if (col_comm_size > 1 && j != (mat.size().cols() - 1)) {
```
albestro: Same as before.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+          }
+
+          // Check if the diagonal tile of the trailing matrix is on this node and
+          // compute first tile of the column of the trailing matrix: diagonal element mat(j,j), reading
+          // mat.read(j,k), using herk (blas operation)
+          hpx::dataflow(trailing_matrix_executor, hpx::util::unwrapping(tile::herk<T, Device::CPU>),
+                        uplo, blas::Op::NoTrans, -1.0, panel[i_local], 1.0,
+                        mat(LocalTileIndex{i_local, j_local}));
+
+          col_panel = panel[i_local];
```
albestro: I think you can return the tile directly from the "broadcast" dataflow.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+          hpx::dataflow(trailing_matrix_executor, hpx::util::unwrapping(tile::gemm<T, Device::CPU>),
+                        blas::Op::NoTrans, blas::Op::ConjTrans, -1.0, mat.read(LocalTileIndex{i, k}),
+                        mat.read(LocalTileIndex{j, k}), 1.0, std::move(mat(LocalTileIndex{i, j})));
+        }
+      }
+    }
+  }
+}
+
+/// @brief Cholesky implementation on distributed memory
```
albestro: Not just memory...also compute resources! 😁

I think Distributed Cholesky algorithm would be ok 😉

include/dlaf/mc/cholesky.h
rasolca: @albestro Why do you want to add an extra dependency on the broadcast?

include/dlaf/mc/cholesky.h
rasolca: @albestro Why do you want to add an extra dependency on the broadcast?

include/dlaf/mc/cholesky.h
rasolca: @albestro This comment is unclear, try to be more specific in these cases... (especially if the comments are from multiple reviews). Are you referring to the comment of line 165?

include/dlaf/mc/cholesky.h
albestro: You're right. My mistake.

It was misleading for me, I would reorder the code blocks in a few points...but it's ok!

include/dlaf/mc/cholesky.h
rasolca: Let us use `Distributed implementation of the Cholesky factorization`.

include/dlaf/mc/cholesky.h
albestro: Yes, exactly.

Sorry, but I usually don't care about different reviews for the same PR, I just look at the code and I see all available comments. Got it.

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+
+/// @file
+
+namespace dlaf {
+
+static bool use_pools = true;
+
+/// @brief Cholesky implementation on local memory
+///
+/// @param uplo specifies that the matrix is \a Lower triangular
```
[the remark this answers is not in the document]
Sely85: I think so, using `blas::Uplo::Upper, isn't it?

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+/// @tparam mat refers to a dlaf::Matrix object
+///
+/// @throws std::runtime_error if \p uplo = \a Upper decomposition is chosen (not yet implemented)
```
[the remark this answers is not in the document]
Sely85: Forgot it.. Thanks!

include/dlaf/mc/cholesky.h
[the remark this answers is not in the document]
Sely85: For now, I have renamed:
`matrix_HP_executor` --> `executor_hp`
`matrix_normal_executor` --> `executor_normal`
`mpi_executor` --> `executor_mpi`

include/dlaf/mc/cholesky.h
```diff
@@ -0,0 +1,306 @@
...
+
+          auto kk = LocalTileIndex{k_local_row, k_local_col};
+
+          // If the diagonal tile is on this node factorize it
+          // Cholesky decomposition on mat(k,k) r/w potrf (lapack operation)
+          hpx::dataflow(matrix_HP_executor, hpx::util::unwrapping(tile::potrf<T, Device::CPU>), uplo,
+                        std::move(mat(kk)));
+
+          // Avoid useless communication if one-column communicator and if on the last column
+          if (col_comm_size > 1 && k != (mat.size().cols() - 1)) {
```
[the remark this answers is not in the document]
Sely85: `mat.nrTiles.cols()` is not a `GlobalTileSize`? 
They should be both sizes, _i.e._ `common::Size2D`, right?

test/unit/mc/test_cholesky.cpp
```diff
@@ -7,28 +7,50 @@
...
+TYPED_TEST_SUITE(CholeskyDistributedTest, MatrixElementTypes);
 
 std::vector<LocalElementSize> square_sizes({{10, 10}, {25, 25}, {12, 12}, {0, 0}});
 std::vector<LocalElementSize> rectangular_sizes({{10, 20}, {50, 20}, {0, 10}, {20, 0}});
-std::vector<TileElementSize> square_block_sizes({{5, 5}, {20, 20}});
+std::vector<TileElementSize> square_block_sizes({{3, 3}, {5, 5}});
 std::vector<TileElementSize> rectangular_block_sizes({{10, 30}, {20, 10}});
 
-TYPED_TEST(CholeskyTest, Correctness) {
+GlobalElementSize globalTestSize(const LocalElementSize& size, const Size2D& grid_size) {
```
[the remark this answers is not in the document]
Sely85: I agree. Which one?

include/dlaf/mc/cholesky.h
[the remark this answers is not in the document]
Sely85: should I open an issue?

PR #126, opened 2020-01-24
Basic distributed Cholesky Factorization mini-app + minor fixes

albestro:
A basic working version of Cholesky factorization mini-app (distributed).

It includes:
- basic mini-app with options
- add helper class Timer
- add missing public GoogleTest dependency for DLAF_test
- fix macro parameter name in CHECK_MATRIX_NEAR

**Note:** Performance benchmarks may not be representative since the analytical matrix used as input may lead to subnormal floating-point values, which on some architectures may cause performance issues (severe slowdown, a single operation may take 200x times more).


miniapp/miniapp_cholesky.cpp
```diff
@@ -0,0 +1,241 @@
...
+    if (0 == world.rank())
+      std::cout << "[" << run_index << "]"
+                << " " << elapsed_time << "s"
+                << " " << gigaflops << "GFlop/s"
+                << " " << matrix.size() << " " << matrix.blockSize() << " " << comm_grid.size() << " "
+                << hpx::get_os_thread_count() << std::endl;
+
+    // (optional) run test
+    if (opts.do_check)
+      cholesky_check(matrix);
```
rasolca: same as for `set`...

miniapp/miniapp_cholesky.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+      MPI_Barrier(world);
+    }
+    auto elapsed_time = timeit.elapsed();
+
+    double gigaflops;
+    {
+      double n = matrix.size().rows();
+      double add_mul = n * n * n / 6;
+      static_assert(!std::is_same<T, double>::value || !std::is_same<T, float>::value, "FIX GIGAFLOPS");
```
rasolca: You should fix this before merging.

miniapp/miniapp_cholesky.cpp
```diff
@@ -0,0 +1,241 @@
...
+T analytical_input_matrix(const GlobalElementIndex& index);
+T analytical_result_matrix(const GlobalElementIndex& index);
+
+void setup_input_matrix(Matrix<T, Device::CPU>& matrix) {
+  using namespace dlaf_test::matrix_test;
+
+  util_matrix::assertSizeSquare(matrix, __PRETTY_FUNCTION__, "matrix");
+  util_matrix::assertBlocksizeSquare(matrix, __PRETTY_FUNCTION__, "matrix");
+
+  set(matrix, analytical_input_matrix);
```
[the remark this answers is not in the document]
albestro: I'm working on a separate PR where I'm implementing utils for matrix and tiles, e.g. set, random, random_positive_definite.

What do you think about opening an issue about this in order to fix it as soon as new optimized functions will be available?

miniapp/miniapp_cholesky.cpp
```diff
@@ -0,0 +1,241 @@
...
+options_t check_options(hpx::program_options::variables_map& vm);
+
+T analytical_input_matrix(const GlobalElementIndex& index);
+T analytical_result_matrix(const GlobalElementIndex& index);
+
+void setup_input_matrix(Matrix<T, Device::CPU>& matrix) {
+  using namespace dlaf_test::matrix_test;
+
+  util_matrix::assertSizeSquare(matrix, __PRETTY_FUNCTION__, "matrix");
+  util_matrix::assertBlocksizeSquare(matrix, __PRETTY_FUNCTION__, "matrix");
```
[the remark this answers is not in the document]
albestro: As soon as the work in #131 is merged, I will fix this 😉

miniapp/miniapp_cholesky.cpp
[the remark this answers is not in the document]
albestro: As soon as the work in #135 is merged, I will fix this 😉

miniapp/miniapp_cholesky.cpp
albestro: As per discussion we had, the current version is very easily optimizable, but we need to decide where to put it:
- in `dlaf_test`?
- here in the mini-app?
- or do we want to add an optimized library also for checks as we are doing for set methods in #135?

miniapp/miniapp_cholesky.cpp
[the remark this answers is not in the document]
albestro: In order to unlock this from the other PR (which is in turn dependent on another issue), for the moment we decided to use a placeholder for the macro that will be introduced.

miniapp/miniapp_cholesky.cpp
albestro: As per discussion, we decided to not implement any optimized version of this check, for now it is not needed.

miniapp/miniapp_cholesky.cpp
```diff
@@ -0,0 +1,240 @@
...
+     value<int64_t>()->default_value(1),
+     "Number of runs to compute the cholesky")
+    ("check-result",
+     bool_switch()->default_value(false),
+     "Check the cholesky factorization (for each run)")
+  ;
+  // clang-format on
+
+  // Create the resource partitioner
+  hpx::resource::partitioner rp(desc_commandline, argc, argv);
```
albestro: In HPX things on resource partitioner side are changing:

> on 1.4.0 you can only create the rp upfront and not pass it to init, and on master it can only be passed to init (just creating it does nothing)

>for the next release we will most likely change it to be so that you can't even create it yourself but you only pass resource partitioner parameters to init

Moreover, there is a problem with the management of the help message, that it does not stop the application to run. More on this at https://github.com/STEllAR-GROUP/hpx/issues/4317

PR #127, opened 2020-01-27
Basic CSV Profiler

albestro:
Basic porting of profiler from the prototype.

- `SectionScoped` helper class to profile a code section exploiting RAII
- `time_it` utility to wrap a callable and profile it

At the moment I'm not providing any way to manually manage a profile entry.


include/dlaf/profiler.h
```diff
@@ -0,0 +1,133 @@
...
+/// On destruction it dumps on a csv file all tasks
+class Manager {
+  public:
+  static Manager& get_global_profiler() {
+    static Manager global_;
+    return global_;
+  }
+
+  ~Manager() {
+    std::ofstream profiler_report("report.csv");  // TODO adapt for multiple nodes
```
rasolca: The profile name should be `<name>_<rank_id>_<num_ranks>.csv`
What about giving the possibility to set the profile name `<name>`?

include/dlaf/profiler.h
```diff
@@ -0,0 +1,133 @@
...
+  /// Container for task entries (it is NOT thread-safe)
+  struct ThreadLocalRecorder {
+    std::deque<details::task_data_t> tasks;
+  };
+
+  std::map<std::thread::id, ThreadLocalRecorder> recorders_;
+};
+
+/// Helper object to add a profile entry exploiting RAII
+struct SectionScoped {
```
rasolca: I'm not fully satisfied with the name...

include/dlaf/profiler.h
albestro: I agree with you, moreover passing the filename as a parameter would keep MPI away from the profiler.

include/dlaf/profiler.h
albestro: You're perfectly right...`ProfileScope`? `ScopedProfile`?

Do you have any suggestions?

albestro:
Self-note: thread id is implementation-dependent and usually it is not a "user-friendly" value. Moreover, for HPX would be nice to use its own thread id.

PR #128, opened 2020-01-28
Basic exception propagation for tiles

albestro:
The real exception is not propagated, but a fixed one is propagated in the tile chaining mechanism.

### DIFF
--- a/include/dlaf/tile.h
+++ b/include/dlaf/tile.h
@@ -20,6 +20,10 @@
 
 namespace dlaf {
 
+struct ContinuationException final : public std::runtime_error {
+  ContinuationException() : std::runtime_error("An exception has been thrown during the execution of the previous task") {}
+};
+
 template <class T, Device device>
 class Tile;
 
--- a/include/dlaf/tile.h
+++ b/include/dlaf/tile.h
@@ -10,6 +10,7 @@
 
 #pragma once
 
+#include <exception>
 #include <hpx/hpx.hpp>
 #include "dlaf/common/buffer.h"
 #include "dlaf/matrix/index.h"
--- a/include/dlaf/tile.h
+++ b/include/dlaf/tile.h
@@ -24,7 +24,8 @@ namespace dlaf {
 ///
 /// It is mainly used to enable exception propagation in the automatic-continuation mechanism
 struct ContinuationException final : public std::runtime_error {
-  ContinuationException() : std::runtime_error("An exception has been thrown during the execution of the previous task") {}
+  ContinuationException()
+      : std::runtime_error("An exception has been thrown during the execution of the previous task") {}
 };
 
 template <class T, Device device>
--- a/include/dlaf/tile.h
+++ b/include/dlaf/tile.h
@@ -20,6 +20,9 @@
 
 namespace dlaf {
 
+/// Exception used to notify that an exception has been thrown in a continuation task
+///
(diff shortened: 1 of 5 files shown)


albestro:
I'm going to add some basic tests.

test/unit/test_matrix.cpp
```diff
@@ -1153,3 +1153,47 @@ TEST(MatrixDestructorFutures, ConstAfterRead) {
...
+
+TEST(MatrixExceptionPropagation, PropagateOnRwWithReadAccess) {
+  auto matrix = createMatrix<TypeParam>();
+
+  auto f = matrix(LocalTileIndex(0, 0)).then(hpx::util::unwrapping([](auto&&) {
+    throw std::runtime_error("exception in task");
+  }));
+
+  EXPECT_ANY_THROW(matrix.read(LocalTileIndex(0, 0)).get());
+  EXPECT_ANY_THROW(f.get());
```
rasolca: I would check that the type and message of the exception is correct. (Same for other cases)

albestro: Currently we throw `std::runtime_error("An exception has been thrown and this tile destroyed")));`

What do you think of adding a custom exception for this?

rasolca: It can be a good option, since there is no ways to get the correct exception.

include/dlaf/tile.h
```diff
@@ -19,6 +20,14 @@
 
 namespace dlaf {
 
+/// Exception used to notify that an exception has been thrown in a continuation task
```
rasolca:
```suggestion
/// Exception used to notify a continuation task that an exception has been thrown in a dependency task.
```

include/dlaf/tile.h
```diff
@@ -19,6 +20,14 @@
...
+///
+/// It is mainly used to enable exception propagation in the automatic-continuation mechanism
```
rasolca:
```suggestion
/// It is mainly used to enable exception propagation in the automatic-continuation mechanism.
```

include/dlaf/tile.h
```diff
@@ -19,6 +20,14 @@
...
+struct ContinuationException final : public std::runtime_error {
+  ContinuationException()
+      : std::runtime_error("An exception has been thrown during the execution of the previous task") {}
```
rasolca:
```suggestion
      : std::runtime_error("An exception has been thrown during the execution of the previous task.") {}
```

test/unit/test_matrix.cpp
```diff
@@ -1153,3 +1153,47 @@ TEST(MatrixDestructorFutures, ConstAfterRead) {
...
+TEST(MatrixExceptionPropagation, NoPropagateOnReadOnlyWithReadAccess) {
+  auto matrix = createMatrix<TypeParam>();
+
+  auto f = matrix.read(LocalTileIndex(0, 0)).then(hpx::util::unwrapping([](auto&&) {
+    throw CustomException{};
+  }));
+
+  EXPECT_NO_THROW(matrix.read(LocalTileIndex(0, 0)).get());
+  EXPECT_THROW(f.get(), CustomException);
+}
```
rasolca: Please be consistent with test naming:
Rw <-> RW
ReadOnly <-> Read
What about e.g. `RWPropagatesInReadAccess`?

PR #131, opened 2020-01-30
Preconditions macros

albestro:
Move existing preconditions functions inside `dlaf::matrix::util::details` namespace and allow to use them through macro that auto-fills information about variable names and source location.

About source location, this PR adds a CMake check that defines `DLAF_SOURCE_LOCATION` based on availability of `__PRETTY_FUNCTION__` constant (compiler-dependent). In case the compiler does not provide it, the standard `__func__` is used.

### DIFF
--- a/include/dlaf/util_matrix.h
+++ b/include/dlaf/util_matrix.h
@@ -36,24 +36,25 @@ namespace internal {
 /// @brief Verify if dlaf::Matrix is square
 ///
 /// @tparam Matrix refers to a dlaf::Matrix object
-#define DLAF_PRECONDITION_SIZE_SQUARE(matrix)                                                       \
-  DLAF_ASSERT((matrix.size().rows() == matrix.size().cols()), "Matrix", #matrix, "is not square (", \
+#define DLAF_PRECONDITION_SIZE_SQUARE(matrix)                                                         \
+  DLAF_ASSERT((matrix.size().rows() == matrix.size().cols()), "Matrix ", #matrix, " is not square (", \
               matrix.size().rows(), "x", matrix.size().cols(), ").")
 
 /// @brief Verify if dlaf::Matrix tile is square
 ///
 /// @tparam Matrix refers to a dlaf::Matrix object
-#define DLAF_PRECONDITION_BLOCKSIZE_SQUARE(matrix)                                                   \
-  DLAF_ASSERT((matrix.blockSize().rows() == matrix.blockSize().cols()), "Block size in matrix",      \
-              #matrix, "is not square (", matrix.blockSize().rows(), "x", matrix.blockSize().cols(), \
+#define DLAF_PRECONDITION_BLOCKSIZE_SQUARE(matrix)                                                    \
+  DLAF_ASSERT((matrix.blockSize().rows() == matrix.blockSize().cols()), "Block size in matrix ",      \
+              #matrix, " is not square (", matrix.blockSize().rows(), "x", matrix.blockSize().cols(), \
               ").")
 
 /// @brief Verify if dlaf::Matrix is distributed on a (1x1) grid (i.e. if it is a local matrix).
 ///
 /// @tparam Matrix refers to a dlaf::Matrix object
-#define DLAF_PRECONDITION_LOCALMATRIX(matrix)                                                  \
-  DLAF_ASSERT((matrix.distribution().commGridSize() == comm::Size2D{1, 1}), "Matrix", #matrix, \
-              "is not local.")
+#define DLAF_PRECONDITION_LOCALMATRIX(matrix)                                                   \
+  DLAF_ASSERT((matrix.distribution().commGridSize() == comm::Size2D{1, 1}), "Matrix ", #matrix, \
+              " is not local (grid size: ", matrix.distribution().commGridSize().rows(), "x",   \
+              matrix.distribution().commGridSize().cols(), ").")
 
 /// @brief Verify that the matrix is distributed according to the given communicator grid.
 ///
@@ -61,8 +62,12 @@ namespace internal {
 #define DLAF_PRECONDITION_DISTRIBUTED_ON_GRID(grid, matrix)                                          \
   DLAF_ASSERT(((matrix.distribution().commGridSize() == grid.size()) &&                              \
                (matrix.distribution().rankIndex() == grid.rank())),                                  \
-              "The matrix", #matrix, "is not distributed according to the communicator grid", #grid, \
-              ".")
+              "The matrix ", #matrix, " (rank: ", matrix.distribution().rankIndex(),                 \
+              ", grid size: ", matrix.distribution().commGridSize().rows(), "x",                     \
+              matrix.distribution().commGridSize().cols(),                                           \
+              ") is not distributed according to the communicator grid ", #grid,                     \
+              " (rank: ", grid.rank(), ", grid size: ", grid.size().rows(), "x", grid.size().cols(), \
+              ").")
 
 template <class MatrixConst, class Matrix, class Location>
 void assertMultipliableMatrices(const MatrixConst& mat_a, const Matrix& mat_b, const blas::Op opA,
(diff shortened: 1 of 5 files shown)


albestro:
This PR is on hold waiting for deciding what to do with #132

albestro:
Before merging, all preconditions in the code must be replaced with the new mechanism.

(This is also the reason why checks are failing, due to preconditions in the code merged into master).

albestro:
After #163 this new PR has been refactored and adapted.

All but one precondition are now implemented using the new `DLAF_ASSERT` macro, while the remaining one, which has some more articulated checks to do, is still a function and underneath uses `DLAF_ASSERT_WITH_ORIGIN`.

The latter one has been also quite heavily refactored to make it more general.

In this PR I suggest also to improve error messages by adding values in order to better understand error conditions, otherwise we don't have any clue, apart the expression that failed.

include/dlaf/util_matrix.h
```diff
@@ -21,149 +21,111 @@ constexpr double M_PI = 3.141592;
...
 
 namespace dlaf {
-namespace util_matrix {
+namespace matrix {
+namespace util {
+namespace internal {
 
 /// @brief Verify if dlaf::Matrix is square
 ///
 /// @tparam Matrix refers to a dlaf::Matrix object
```
rasolca, line 38: Documentation has to be improved.
- failure conditions are missing
- parameters documentation has to be fixed
- what happens in the case that one of the condition is not met.

include/dlaf/util_matrix.h
```diff
@@ -21,149 +21,111 @@ constexpr double M_PI = 3.141592;
...
-/// @throws std::invalid_argument if the matrix is not squared
-template <class Matrix>
-void assertSizeSquare(const Matrix& matrix, std::string function, std::string mat_name) {
-  if (matrix.size().rows() != matrix.size().cols())
-    throw std::invalid_argument(function + ": " + "Matrix " + mat_name + " is not square.");
-}
+#define DLAF_PRECONDITION_SIZE_SQUARE(matrix)                                                         \
```
rasolca, line 39: Wouldn't `DLAF_ASSERT_SIZE_SQUARE` be a better name?
(and similar for next macros)

include/dlaf/util_matrix.h
```diff
@@ -21,149 +21,111 @@ constexpr double M_PI = 3.141592;
...
-    }
-  }
-}
-
-}
-
-namespace matrix {
-namespace util {
-namespace internal {
+#define DLAF_PRECONDITION_MULTIPLIABLE_MATRICES(a, b, opA, opB) \
```
rasolca, line 127: I would add matrix c as well to have a full check that a*b can be performed and stored to c.

include/dlaf/util_matrix.h
```diff
@@ -21,149 +21,189 @@ constexpr double M_PI = 3.141592;
...
+                              "The result of the left multiplication between ", mat_a_name, " and ",
+                              mat_b_name, " cannot be stored in ", mat_c_name, " (size of matrix ",
+                              mat_b_name, ", ", a_k(mat_b.size()), ", not equal to that of matrix ",
+                              mat_c_name, ", ", a_k(mat_c.size()), ").");
+      DLAF_ASSERT_WITH_ORIGIN(location, (a_k(mat_b.blockSize()) == a_k(mat_c.blockSize())),
+                              "The result of the left multiplication between ", mat_a_name, " and ",
+                              mat_b_name, " cannot be stored in ", mat_c_name, " (blocksize of matrix ",
+                              mat_b_name, ", ", a_k(mat_b.blockSize()), ", not equal to that of matrix ",
+                              mat_c_name, ", ", a_k(mat_c.blockSize()), ").");
+      break;
```
albestro, line 195: I would go for a simplified version, but I don't know what the others think.

It is just an example, feel free to pick something if you think it is a good idea.

@rasolca @teonnik @sanciop 

```cpp
  auto op = [](const auto& size, const blas::Op op) {
    using size2d_t = decltype(size);

    switch (op) {
      case blas::Op::NoTrans:
        return size2d_t{size.rows(), size.cols()};
      case blas::Op::Trans:
      case blas::Op::ConjTrans:
        return size2d_t{size.cols(), size.rows()};
      default:
        return size2d_t{-1, -1};
    }
  };

  bool A_is_transposed = opA != blas::Op::NoTrans;
  GlobalTileSize A_tiles = op(mat_a.nrTiles(), opA);
  GlobalElementSize A_size = op(mat_a.size(), opA);
  TileElementSize A_blocksize = op(mat_a.blockSize(), opA);

  bool B_is_transposed = opB != blas::Op::NoTrans;
  GlobalTileSize B_tiles = op(mat_b.nrTiles(), opB);
  GlobalElementSize B_size = op(mat_b.size(), opB);
  TileElementSize B_blocksize = op(mat_b.blockSize(), opB);

  GlobalTileSize C_tiles{A_tiles.rows(), B_tiles.cols()};
  GlobalElementSize C_size{A_size.rows(), B_size.cols()};
  TileElementSize C_blocksize{A_blocksize.rows(), B_blocksize.cols()};

  DLAF_ASSERT_WITH_ORIGIN(location, (A_tiles.cols() == B_tiles.rows()), "The matrices '", mat_a_name,
                          "' and '", mat_b_name, "' are not multipliable. ", A_tiles, " x ", B_tiles,
                          ". (tranpose: A=", A_is_transposed, " and B=", B_is_transposed, ")");
  DLAF_ASSERT_WITH_ORIGIN(location, (A_size.cols() == B_size.rows()), "The matrices '", mat_a_name,
                          "' and '", mat_b_name, "' are not multipliable. ", A_size, " x ", B_size,
                          ". (tranpose: A=", A_is_transposed, " and B=", B_is_transposed, ")");
  DLAF_ASSERT_WITH_ORIGIN(location, (A_blocksize.cols() == B_blocksize.rows()), "The matrices '",
                          mat_a_name, "' and '", mat_b_name, "' are not multipliable. ", A_blocksize,
                          " x ", B_blocksize, ". (tranpose: A=", A_is_transposed,
                          " and B=", B_is_transposed, ")");

  DLAF_ASSERT_WITH_ORIGIN(location, C_tiles == mat_c.nrTiles(), "The result of the multiplication '",
                          mat_a_name, "' x '", mat_b_name, "' cannot be stored in '", mat_c_name, "'. ",
                          A_tiles, " x ", B_tiles, " = ", C_tiles, " != ", mat_c.nrTiles(),
                          ". (tranpose: A=", A_is_transposed, " and B=", B_is_transposed, ")");
  DLAF_ASSERT_WITH_ORIGIN(location, C_size == mat_c.size(), "The result of the multiplication '",
                          mat_a_name, "' x '", mat_b_name, "' cannot be stored in '", mat_c_name, "'. ",
                          A_size, " x ", B_size, " = ", C_size, " != ", mat_c.size(),
                          ". (tranpose: A=", A_is_transposed, " and B=", B_is_transposed, ")");
  DLAF_ASSERT_WITH_ORIGIN(location, C_blocksize == mat_c.blockSize(),
                          "The result of the multiplication '", mat_a_name, "' x '", mat_b_name,
                          "' cannot be stored in '", mat_c_name, "'. ", A_blocksize, " x ", B_blocksize,
                          " = ", C_blocksize, " != ", mat_c.blockSize(),
                          ". (tranpose: A=", A_is_transposed, " and B=", B_is_transposed, ")");
```

include/dlaf/util_matrix.h
```diff
@@ -21,149 +21,189 @@ constexpr double M_PI = 3.141592;
...
+              matrix.distribution().commGridSize().cols(),                                           \
+              ") is not distributed according to the communicator grid ", #grid,                     \
+              " (rank: ", grid.rank(), ", grid size: ", grid.size().rows(), "x", grid.size().cols(), \
+              ").")
+
+template <class MatrixConst, class Matrix, class Mat, class Location>
+void assertMultipliableMatrices(const MatrixConst& mat_a, const Matrix& mat_b, const Mat& mat_c,
+                                const blas::Op opA, const blas::Op opB, const Location location,
+                                std::string mat_a_name, std::string mat_b_name, std::string mat_c_name) {
+  auto get_k = [](const auto& size, const blas::Op op, bool first_member) -> decltype(size.rows()) {
```
rasolca, line 80: I would go for a much simpler version:
```
{
  auto rows = [](const auto& size, const blas::Op op) -> decltype(size.rows()) {
    switch (op) {
      case blas::Op::NoTrans:
        return size.rows();
      case blas::Op::Trans:
      case blas::Op::ConjTrans:
        return size.cols();
      default:
        return {};
    }
  }
  auto cols = [](const auto& size, const blas::Op op) -> decltype(size.cols()) {
    <similar as above>
  }

  DLAF_ASSERT_WITH_ORIGIN(location, rows(mat_a, opA) == mat_c.rows() && cols(mat_a, opA) == rows(mat_b, opB) && cols(mat_b, opB) == mat_c.cols(), <error_message> );
}
```
For the error message I suggest:
Size mismatch: <size_opA_mat_a> x <size_opB_mat_B> -> <size_mat_c> cannot be performed.

Sely85: Look great. Way more simple. 
Do we need to check also `blockSize` and `nrTiles` "sizes"?

rasolca: Yes the same has to be done for the blockSize (nrTiles follows from size and blocksize).
A second assert can be added with `blockSize()` instead of `size()`.
For the error message I suggest:
Block-size mismatch: <block_size_opA_mat_a> x <block_size_opB_mat_B> -> <block_size_mat_c> cannot be performed.

Sely85: Thanks.

teonnik:
Apologies for taking a while to provide feedback on this PR.

I am a bit concerned that a number of functions are now substituted with new macros, for example:
```
 util_matrix::assertSizeSquare(mat_a, "Cholesky", "mat_a") -> DLAF_ASSERT_SIZE_SQUARE(mat_a)
```
I think we shouldn't be introducing macros unless it's absolutely necessary. If this was chosen in order to get the correct FILE and LINE information at the call site, I understand the motivation. Could you please clarify?

If we are going to keep introducing new assert macros for the reason above, we might want to go all the way and have a full stack trace reported upon error(e.g. https://github.com/boostorg/stacktrace but I hope HPX already has one so we don't need another dependency). That way we won't be discouraged from wrapping ASSERT macros in functions. In any case, this should be discussed and addressed in a separate PR.

rasolca:
The reasons for using macros are two:
1) as you mention we can get pass the information about the line function and file.
2) we can add the parameter name in the error message.

E.g.:
```
foo(Matrix bar, Matrix baz) {
  DLAF_ASSERT_SMTH(bar);
  DLAF_ASSERT_SMTH(baz);
}
```
can produce the following error message:
"<file>:<line> `foo`: Matrix `baz` doesn't satisfy smth."

If we can achieve something similar without macros, I'm for it. But let discuss it in a separate issue/PR.

Issue #20, opened 2019-07-24
Implement MatrixRW


rasolca:
See `misc/code.(pdf|pptx)` and `misc/example/m.h` for details.

rasolca:
A new Matrix API will be provided which will make this obsolete.

Issue #19, opened 2019-07-24
Implement MatrixRead


rasolca:
See `misc/code.(pdf|pptx)` and `misc/example/m.h` for details.

PR #133, opened 2020-01-31
Small refactoring pre-MatrixView

rasolca:
Matrix implementation has been reorganized to simplify (avoiding code duplication) the implementation of MatrixView (#19, #20)


include/dlaf/matrix.h
```diff
@@ -14,6 +14,7 @@
 #include "dlaf/communication/communicator_grid.h"
 #include "dlaf/matrix/distribution.h"
```
albestro: I don't know what you think about it. IMHO Matrix does not depend anymore on Distribution (directly), so I would remove it.

PR #135, opened 2020-02-03
Optimized set utilities for Matrix

### LINKED ISSUE #130: Add optimized utils for tiles and matrices values initialization
It would be useful to have basic functionalities to setup matrices and tiles.
At the moment they are available in the `DLAF_test` library, but I think it would be nice to have them in the main library, better if optimized.

The firsts I would implement are:
- set: given a constant (e.g. set all to a fixed value) or a function for computing values based on the index;
- set_random
- set_random_positive_definite

This would allow to easily initialize matrices in applications (e.g. mini-app).

albestro:
Close #130 
This implements optimized functions for `Matrix`, but not for `Tile`.

Add methods with relative tests (in `dlaf::matrix::util`):
- `set`
- `set_random`
- `set_random_positive_definite`

Modify `checkNear` behavior, instead of testing both absolute and relative difference, now it is possible to check even just one (by setting the other one to 0).

Minor fix on documentation.


test/include/dlaf_test/util_matrix.h
```diff
@@ -204,13 +204,11 @@ void checkPtr(PointerGetter exp_ptr, Matrix<T, Device::CPU>& mat, const char* fi
 /// The (i, j)-element of the matrix is compared to expected({i, j}).
 /// @pre expected argument is an index of type const GlobalElementIndex&.
 /// @pre expected return type should be T.
-/// @pre rel_err > 0.
-/// @pre abs_err > 0.
+/// @pre rel_err > 0 || abs_err > 0
```
rasolca: both has to be >= 0

albestro:
> Note: as it is now the diagonal are set incorrectly for the Cholesky miniapp.

@rasolca 
Right! As discussed offline, I'm working on a `set_positive_definite_hermitian` function (and its test) that, at least for the moment, it is what we need for going on with the Cholesky miniapp work.

albestro:
> Very handy!
> 
> Note: the header file (util_matrix.h in include/dlaf) is not in the same subdirectory of the test source code (test_util_matrix.cpp in test/unit/matrix).

@Sely85 @rasolca 
I agree with you. I think that we may decide to move everything strictly related to the matrix inside the matrix folder:
- Matrix
- Tile
- Distribution
- Matrix utils

And this should be true also for tests.

What do you think?

Sely85:
> > Very handy!
> > Note: the header file (util_matrix.h in include/dlaf) is not in the same subdirectory of the test source code (test_util_matrix.cpp in test/unit/matrix).
> 
> @Sely85 @rasolca
> I agree with you. I think that we may decide to move everything strictly related to the matrix inside the matrix folder:
> 
> * Matrix
> * Tile
> * Distribution
> * Matrix utils
> 
> And this should be true also for tests.
> 
> What do you think?

I agree with you. What about `lapack_tile` and `blas_tile`? Inside the matrix folder as well since are "tile-related", right?

rasolca:
I agree, but since I'm working on matrix views, I propose to postpone the move of the matrix and tile headers.

albestro:
I've rebased and fixed conflict with master.

Trying to resume what this PR does:
- Parallelized util functions for matrixes: `set`, `set_random`, `set_random_hermitian_positive_definite` with related tests
- Introduce `Position` for generic `Index2D` with test (**WARNING**: as it is now, it permits position comparison of indexes with different types)
- Introduce `computeLinearIndex` for `Index2D`
- `CommunicatorGrid` now has a "full" communicator, with all ranks inside the grid
- Minor fix: TypeInfo now works also with `const` types
- Helper function for complex `dlaf::conj`: differently from `std::conj` that always returns an `std::complex`, this uses as return type the same as the input parameter
- Minor fix: output messages on errors for check_near for both matrix and tile

include/dlaf/util_matrix.h
```diff
@@ -146,5 +154,207 @@ void assertMultipliableMatrices(const MatrixConst& mat_a, const Matrix& mat_b, b
...
+                    tile(TileElementIndex{i, j}) = value;
+                  else
+                    tile(TileElementIndex{j, i}) = value;
+                }
+              }
+            }
+          }),
+          matrix(tile_wrt_local));
+    }
+  }
```
rasolca: Why you check again the position in the random getter?
(See internal comments)

Moreover this approach may lead to think that value_at({i, j}) gives always the same value for a given (i, j)

include/dlaf/util_matrix.h
```diff
@@ -146,5 +154,207 @@ void assertMultipliableMatrices(const MatrixConst& mat_a, const Matrix& mat_b, b
...
+              // transposed indices
+              for (SizeType j = 0; j < tile.size().cols(); ++j) {
+                for (SizeType i = 0; i <= j; ++i) {
+                  auto value = value_at(GlobalElementIndex{tl_index.row() + i, tl_index.col() + j});
+
+                  tile(TileElementIndex{i, j}) = value;
+                  if (i != j)
+                    tile(TileElementIndex{j, i}) = dlaf::conj(value);
+                }
+              }
```
rasolca: What about the following?
```
              for (SizeType j = 0; j < tile.size().cols(); ++j) {
                for (SizeType i = 0; i < j; ++i) {
                  auto value = random();
                  tile(TileElementIndex{i, j}) = value;
                  tile(TileElementIndex{j, i}) = dlaf::conj(value);
                }
                tile({j, j}) = real(random()) + offset;
              }
```

include/dlaf/util_matrix.h
```diff
@@ -146,5 +154,207 @@ void assertMultipliableMatrices(const MatrixConst& mat_a, const Matrix& mat_b, b
...
+                for (SizeType i = 0; i < tile.size().rows(); ++i) {
+                  auto value = value_at(GlobalElementIndex{tl_index.row() + i, tl_index.col() + j});
+
+                  // but they are set row-wise in the original tile and col-wise in the transposed one
+                  if (tile_position == dlaf::common::Position::LOWER)
+                    tile(TileElementIndex{i, j}) = value;
+                  else
+                    tile(TileElementIndex{j, i}) = value;
+                }
+              }
```
rasolca: What about the following?
```
              for (SizeType j = 0; j < tile.size().cols(); ++j) {
                for (SizeType i = 0; i < tile.size().rows(); ++i) {
                  if (tile_position == dlaf::common::Position::LOWER)
                    tile(TileElementIndex{i, j}) = random();
                  else
                    tile(TileElementIndex{j, i}) = conj(random());
                }
              }
```

include/dlaf/common/index2d.h
```diff
@@ -171,6 +171,25 @@ template <class Index2DType, typename LinearIndexT>
...
+
+/// Describe the position of an index in a grid using matrix specific nomenclature
+template <class IndexT, class IndexTag>
+Position position(const dlaf::common::Index2D<IndexT, IndexTag>& index) {
+  if (index.row() > index.col())
+    return Position::LOWER;
+  else if (index.row() < index.col())
+    return Position::UPPER;
+  return Position::DIAGONAL;
+}
```
rasolca: This may may be misleading:
A tile that contains only elements of the lower triangular part but has an index (i, j) where j>i is considered UPPER.

rasolca: Moreover a simple check `if (index.row() <= index.col())`
becomes `if (position(index) >= Diagonal)` which
requires 2 condition evaluations in case of false and 3 in case of true.

include/dlaf/util_matrix.h
```diff
@@ -12,6 +12,12 @@
 #include <exception>
 #include <random>
 #include <string>
+#include <cmath>
+
+#undef M_PI
```
rasolca: First you un-define M_PI and than you check if it is defined?

include/dlaf/util_matrix.h
```diff
@@ -12,6 +12,12 @@
...
+#if !defined M_PI
```
rasolca: Personally I prefer `#ifndef`.

test/unit/communication/test_communicator_grid.cpp
```diff
@@ -187,7 +187,7 @@ TEST_P(CommunicatorGridTest, ConstructorIncomplete) {
     EXPECT_EQ(MPI_COMM_NULL, incomplete_grid.rowCommunicator());
     EXPECT_EQ(MPI_COMM_NULL, incomplete_grid.colCommunicator());
 
-    EXPECT_EQ(-1, incomplete_grid.rank_all(coords));
+    EXPECT_EQ(-1, incomplete_grid.rankFullCommunicator(coords));
```
rasolca: Shouldn't `rankFullCommunicator` be tested with different values of coords? (applies also to other places in the code)

include/dlaf/util_matrix.h
```diff
@@ -146,5 +159,179 @@ void assertMultipliableMatrices(const MatrixConst& mat_a, const Matrix& mat_b, b
...
+                      for (SizeType j = 0; j < tile.size().cols(); ++j)
+                        for (SizeType i = 0; i < tile.size().rows(); ++i)
+                          tile({i, j}) = el(GlobalElementIndex{i + tl_index.row(), j + tl_index.col()});
+                    }),
+                    matrix(tile_wrt_local));
+    }
+  }
+}
+
+/// Set the matrix with random values in the range [-1, 1]
```
rasolca: What about the complex case?

PR #138, opened 2020-02-10
Refactoring pre-MatrixView cont.

rasolca:
Refactored test utilities for tiles and matrices:
- Matrix arguments are now templated to allow the use of Matrix and MatrixView objects.
- Removed code duplication for set.
- Changed the modified files to use the new namespace structure for `dlaf_test`.

### DIFF
--- a/include/dlaf/matrix/matrix_base.h
+++ b/include/dlaf/matrix/matrix_base.h
@@ -60,6 +60,12 @@ public:
     return *distribution_;
   }
 
+  /// Returns the size of the Tile with global index @p index.
+  TileElementSize tileSize(const GlobalTileIndex& index) noexcept {
+    return {std::min(blockSize().rows(), size().rows() - index.row() * blockSize().rows()),
+            std::min(blockSize().cols(), size().cols() - index.col() * blockSize().cols())};
+  }
+
 protected:
   // move constructor and assignment are protected to avoid invalidation of objects of
   // derived classes. I.e.:
@@ -68,6 +74,10 @@ protected:
   // MatrixBase base = std::move(derived);
   MatrixBase(MatrixBase&& rhs) = default;
   MatrixBase& operator=(MatrixBase&& rhs) = default;
+
+  static std::size_t futureVectorSize(const LocalTileSize& local_nr_tiles) noexcept {
+    return util::size_t::mul(local_nr_tiles.rows(), local_nr_tiles.cols());
+  }
 
   /// Returns the position in the vector of the index Tile.
   ///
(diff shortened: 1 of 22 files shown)


test/include/dlaf_test/matrix/util_matrix_blas.h
```diff
@@ -0,0 +1,60 @@
...
+/// @pre el return type should be T.
+template <template <class, Device> class MatrixType, class T, class ElementGetter>
+void set(MatrixType<T, Device::CPU>& mat, ElementGetter el, blas::Op op) {
+  switch (op) {
+    case blas::Op::NoTrans:
+      set(mat, el);
+      break;
+
+    case blas::Op::Trans: {
+      auto op_el = [el](GlobalElementIndex i) {
```
albestro: You are passing `el` to the lambda by copy. Since there is no concurrency problem, I would pass it by reference, mainly because in this way you allow the `ElementGetter` to have its internal status and keep it for its lifetime over multiple calls (even if it is not a big deal).

In case you agree, this is not the only occurrence (see also `util_tile_blas.h`).

test/include/dlaf_test/matrix/util_matrix_blas.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include "blas.hh"
+#include "dlaf/matrix.h"
+#include "dlaf_test/matrix/util_matrix.h"
+#include "dlaf_test/util_types.h"
+
+namespace dlaf {
+namespace matrix {
+namespace test {
+using namespace dlaf_test;
```
Sely85: Do we still need the `using namespace dlaf_test` here?

albestro: For the moment yes, because of `TypeUtilities`.

I've opened issue #139, we will fix all this kind of things once a ruleset will be decided.

include/dlaf/matrix/matrix_base.h
```diff
@@ -69,6 +75,10 @@ class MatrixBase {
   MatrixBase(MatrixBase&& rhs) = default;
   MatrixBase& operator=(MatrixBase&& rhs) = default;
 
+  static std::size_t futureVectorSize(const LocalTileSize& local_nr_tiles) noexcept {
```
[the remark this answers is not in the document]
rasolca: `MatrixBase` is an implementation detail which is used to avoid code duplication among
`Matrix` and `MatrixView`.
Moreover this method is a protected method used internally to return the size needed for the future vectors which **currently** (no guarantee that we will not change that) correspond to the number of tiles.

PR #144, opened 2020-02-20
Add MPI executor for Cholesky Factorization

albestro:
Introduce executor for MPI tasks in Choleksy factorization.

MPI tasks will run on the dedicated `mpi` pool if it has been created during HPX initialization, otherwise they will run with high priority in the default pool.

Credit @rasolca


albestro:
@rasolca 

The mini-app already initializes the `mpi` pool in case there are multiple ranks.

https://github.com/eth-cscs/DLA-Future/blob/cb64e4cfcb166c59af3e718ac99634951b3b1e8f/miniapp/miniapp_cholesky.cpp#L167-L174

The only inconvenient is about HPX API: they are changing the resource partitioner initialization. In older versions, it had to be created and configured stand-alone, while in current `master` version they are proposing to pass it to the `hpx::init` function (or equivalent), otherwise the resource partitioner configuration is not considered.

PR #145, opened 2020-02-21
Fix lambda capture problem by copy on gcc 7

albestro:
From gcc 8 onward it works, but on gcc 7 it led to the error "invalidly declared function type".

### DIFF
--- a/include/dlaf/util_matrix.h
+++ b/include/dlaf/util_matrix.h
@@ -204,7 +204,7 @@ public:
 /// @pre el argument is an index of type const GlobalElementIndex&.
 /// @pre el return type should be T.
 template <class T, class ElementGetter>
-void set(Matrix<T, Device::CPU>& matrix, ElementGetter&& el) {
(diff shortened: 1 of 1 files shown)


albestro:
@Sely85 can you test it, please?
@rasolca it would be nice to document this and/or to provide something (e.g. a macro, like HPX does) to centralize these things in a single place (e.g. DLAF_CAPTURE_COPY, DLAF_CAPTURE_FORWARD, ...)

Sely85:
I've tested the branch and everything went well. Compilation and tests are both successful (using gcc 7.4.0 both on a CINECA clusters and on my virtual machine).

rasolca:
> @rasolca it would be nice to document this and/or to provide something (e.g. a macro, like HPX does) to centralize these things in a single place (e.g. DLAF_CAPTURE_COPY, DLAF_CAPTURE_FORWARD, ...)

I is a good idea to centralize it, The discussion where to centralize it is not part of this PR.

PR #154, opened 2020-03-03
Non const MatrixView 

rasolca:
Currently only a refactoring proposal which allows to avoid code duplication
for `operator()` and `read()`.


include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,79 @@
...
+#include <hpx/hpx.hpp>
+#include "dlaf/tile.h"
+
+namespace dlaf {
+namespace matrix {
+namespace internal {
+
+template <class ReturnTileType, class TileType>
+hpx::future<ReturnTileType> setPromiseTileFuture(hpx::future<TileType>& old_future,
+                                                 hpx::promise<TileType>& p) noexcept {
```
albestro: Since the promise `p` is an `In & move from` parameter, I would use an rvalue reference.

It is more informative from the user point of view since they must use `std::move` and they are aware that the promise will be moved and becomes invalid.

```suggestion
                                                 hpx::promise<TileType>&& p) noexcept {
```

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,79 @@
...
+    }
+  });
+}
+
+template <class ReturnTileType, class TileType>
+hpx::future<ReturnTileType> getTileFuture(hpx::future<TileType>& tile_future) noexcept {
+  hpx::future<TileType> old_future = std::move(tile_future);
+  hpx::promise<TileType> p;
+  tile_future = p.get_future();
+  return setPromiseTileFuture<ReturnTileType>(old_future, p);
```
albestro:
```suggestion
  return setPromiseTileFuture<ReturnTileType>(old_future, std::move(p));
```

rasolca:
Just a note about moving. I'm using the non destructive approach "change what you are already working on and fix what doesn't work", therefore the move will not be completed in this PR.

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -74,6 +78,137 @@ class TileFutureManager {
...
+class ViewTileFutureManager : private TileFutureManager<T, device> {
+  using TileManagerType = TileFutureManager<T, device>;
+
+public:
+  using typename TileManagerType::TileType;
+  using typename TileManagerType::ConstTileType;
+
+  ViewTileFutureManager() {}
+
+  ViewTileFutureManager(TileManagerType& tile_manager, bool /* force_RW */)
```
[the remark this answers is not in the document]
rasolca: To have the same signature for `TileManagerType` and `ViewTileManagerType` to simplify the templated MatrixView constructor.

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -74,6 +78,137 @@ class TileFutureManager {
...
+      : TileManagerType(), tile_promise_(), tile_shared_promise_() {
```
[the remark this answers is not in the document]
rasolca: I used them to keep in mind the order of the initializations, but I can remove them...

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -74,6 +78,137 @@ class TileFutureManager {
...
+    tile_future_ = std::move(tile_promise_.get_future());
+    tile_shared_future_ = std::move(tile_shared_promise_.get_future());
+
+    TileManagerType& base = *this;
+    std::swap(tile_manager, base);
+
+    tile_status_ = TileStatus::RW;
+    assert(tile_future_.valid());
```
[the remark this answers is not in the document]
rasolca: I will check if I can unify them...

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -74,6 +78,137 @@ class TileFutureManager {
...
+    assert(tile_status_ != TileStatus::None);
+    return TileManagerType::getReadTileSharedFuture();
+  }
+
+  hpx::future<TileType> getRWTileFuture() noexcept {
+    assert(tile_status_ == TileStatus::RW);
+    return TileManagerType::getRWTileFuture();
+  }
+
+  void makeRead() {
```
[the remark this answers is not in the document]
rasolca: I'll wait #163 to add missing assertions.

test/unit/matrix/test_matrix_view.cpp
```diff
@@ -390,9 +564,92 @@ TYPED_TEST(MatrixViewTest, DependenciesConst) {
...
+        }
+
+        auto shfut1d = getSharedFuturesUsingLocalIndex(mat);
+        EXPECT_TRUE(checkFuturesStep(0, shfut1d));
+
+        auto fut2 = getFuturesUsingLocalIndex(mat);
+        EXPECT_TRUE(checkFuturesStep(0, fut2));
+
+        CHECK_MATRIX_FUTURES(true, shfut1a, fut0);
+        EXPECT_TRUE(checkFuturesStep(shfut1b.size(), shfut1a));
```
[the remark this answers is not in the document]
rasolca: well spotted!

test/include/dlaf_test/matrix/util_matrix_futures.h
```diff
@@ -185,6 +185,35 @@ void checkFuturesDone(bool get_ready, const std::vector<Future1>& current, Matri
...
+template <class Future1, class MatrixViewType>
+void checkFuturesDoneWrite(bool get_ready, const std::vector<Future1>& current,
+                           MatrixViewType& mat_view) {
+  const auto& nr_tiles = mat_view.distribution().localNrTiles();
+  assert(current.size() == util::size_t::mul(nr_tiles.rows(), nr_tiles.cols()));
+
+  for (std::size_t index = 0; index < current.size(); ++index) {
+    EXPECT_TRUE(checkFuturesStep(get_ready ? index : 0, current));
+    LocalTileIndex tile_index(to_signed<LocalTileIndex::IndexType>(index) % nr_tiles.rows(),
+                              to_signed<LocalTileIndex::IndexType>(index) / nr_tiles.rows());
```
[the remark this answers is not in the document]
rasolca: Is there a reason why the `dims` argument of
```
 template <class Index2DType>
 typename Index2DType::IndexType computeLinearIndex(
    Ordering ordering, const Index2DType& index,
    const std::array<typename Index2DType::IndexType, 2>& dims)
```
is not a Size2DType with the same tag as the Index2DType?
(Note that an array can be still used due to implicit conversion)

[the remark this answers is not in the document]
albestro: I thought the same thing when I wrote the comment 😂

So, I've just opened a PR #183 with a patch for it ;)

include/dlaf/matrix.tpp
```diff
@@ -47,29 +50,17 @@ Matrix<T, device>::Matrix(matrix::Distribution&& distribution, const matrix::Lay
...
 template <class T, Device device>
 hpx::future<Tile<T, device>> Matrix<T, device>::operator()(const LocalTileIndex& index) noexcept {
-  std::size_t i = tileLinearIndex(index);
-  hpx::future<TileType> old_future = std::move(tile_futures_[i]);
-  hpx::promise<TileType> p;
-  tile_futures_[i] = p.get_future();
-  tile_shared_futures_[i] = {};
-  return old_future.then(hpx::launch::sync, [p = std::move(p)](hpx::future<TileType>&& fut) mutable {
-    try {
-      return std::move(fut.get().setPromise(std::move(p)));
```
teonnik, line 67: Instead of `std::move(fut.get().setPromise(std::move(p)));` why not `fut.get().setPromise(std::move(p));` the latter is already a rvalue.

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,220 @@
...
+  using TileType = Tile<T, device>;
+  using ConstTileType = Tile<const T, device>;
+
+  TileFutureManager() {}
+
+  TileFutureManager(TileType&& tile) : tile_future_(hpx::make_ready_future(std::move(tile))) {}
+
+  hpx::shared_future<ConstTileType> getReadTileSharedFuture() noexcept {
+    if (!tile_shared_future_.valid()) {
+      tile_shared_future_ = std::move(getTileFuture<ConstTileType>(tile_future_));
```
teonnik, line 54: similar here, this is already a rvalue, there is no need for `std::move`

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,220 @@
...
+namespace dlaf {
+namespace matrix {
+namespace internal {
+
+template <class ReturnTileType, class TileType>
+hpx::future<ReturnTileType> setPromiseTileFuture(hpx::future<TileType>& old_future,
+                                                 hpx::promise<TileType>& p) noexcept {
+  return old_future.then(hpx::launch::sync, [p = std::move(p)](hpx::future<TileType>&& fut) mutable {
+    try {
+      return ReturnTileType(std::move(fut.get().setPromise(std::move(p))));
```
teonnik, line 24: `fut.get().setPromise(std::move(p))` this is a rvalue, no need for the outer std::move

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,220 @@
...
+    if (tile_manager.status() == TileStatus::Read)
+      tile_shared_future_ = tile_manager.getReadTileSharedFuture();
+
+    tile_status_ = tile_manager.status();
+  }
+
+  ViewTileFutureManager(ViewTileFutureManager&& tile_manager) noexcept
+      : TileManagerType(std::move(tile_manager)), tile_status_(tile_manager.tile_status_),
+        tile_promise_(std::move(tile_manager.tile_promise_)),
+        tile_shared_promise_(std::move(tile_manager.tile_shared_promise_)) {
```
teonnik, line 117: Hmm, isn't this a bug? You have already moved `tile_manager`, using it afterwards seems like an undefined behaviour?

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,220 @@
...
+
+  // If valid, a copy of the shared future of the tile,
+  // which has the promise set to the promise for tile_future_.
+  hpx::shared_future<ConstTileType> tile_shared_future_;
+};
+
+enum class TileStatus { None = 0, Read = 1, RW = 2 };
+
+template <class T, Device device>
+class ViewTileFutureManager : private TileFutureManager<T, device> {
```
teonnik, line 85: private inheritance is the default for classes, so we can omit the qualifier... but I think it may be better leaving it as it is nice to have it explicitly written out

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+private:
+  void setUpRW(TileManagerType& tile_manager) {
+    // the futures of the tile_manager has to be moved in this object and
+    // be replaced with the futures of the promises of *this.
+    // Therefore we first set up the futures for tile_manager in *this
+    // and then swap the base class members (i.e. the futures).
+    tile_future_ = std::move(tile_promise_.get_future());
+    tile_shared_future_ = std::move(tile_shared_promise_.get_future());
```
teonnik, line 198: Is std::move needed here?

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: I prefer to have it explicitly written out...

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: The first statement moves the data members of the base class, and then the data members of the derived class are moved later.

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: setPromise() returns a reference to a Tile, so it is a lvalue.

teonnik: I meant `... std::move(fut.get() ... )`

rasolca:
bors try

PR #158, opened 2020-03-06
Fixing library deploy

### LINKED ISSUE #157: Library deployment does not work
I tried to install the library and use it from another CMake project and it does not work.

The first problem is on the dependency MKL, but it is not the only problem.

I'm going to update the problems I address in the comments below.

albestro:
Close #157 

Work in progress fixing library deployment.

I'm evaluating to add also a very simple trivial guide on how to use our library, from CMake, from another project. The output of this simple guide can become also a check to add to our CI, in order to check that our library is usable in other projects.


albestro:
It is not strictly related, but it may be worth fixing it here.

CMake allows having an empty `CMAKE_BUILD_TYPE`. Do we want to set it to some default (e.g. RelWithDebInfo)?

I noticed it because of the check about matching build_type with respect to HPX. (sidenote: we can remove the check since it has been implemented in 1.4.0, which is our minimum required version).

PR #159, opened 2020-03-09
Fix code problems and warnings

albestro:
Enabling compiler checks uncovered few problems (or potential ones) in the code. This PR aims at solving them.

- signed/unsigned conversions
- unused variables

Checks highlighted also a "minor" bug fix in the `test_gemm`.


albestro:
I'm still working on this.

In particular:
- I have test the build on daint with gcc
- Miniapp have not been fixed yet

albestro:
Highlighted fixes (just the ones that deserves more attention):
1. usage of `common::vector` in algorithms
2. `MemoryView` c'tors overloads with signed integer types
3. `M_PI` double constants cast
4. bug fix: test_gemm usage of `size_b`
5. computeGlobalSize uses grid_size just in assert. Do we want to remove it the parameter and leave the check to the caller? If we want to leave it there, do we want to sovle the unused variable for release builds like this?

include/dlaf/util_matrix.h
```diff
@@ -243,7 +243,7 @@ void set_random(Matrix<T, Device::CPU>& matrix) {
       GlobalTileIndex tile_wrt_global = dist.globalTileIndex(tile_wrt_local);
 
       auto tl_index = dist.globalElementIndex(tile_wrt_global, {0, 0});
-      auto seed = tl_index.row() * matrix.size().cols() + tl_index.col();
+      std::size_t seed = to_sizet(tl_index.row() * matrix.size().cols() + tl_index.col());
```
rasolca: What about using size_t operations in this case?
(Commented similar cases with (*))

include/dlaf/util_matrix.h
```diff
@@ -287,7 +287,7 @@ void set_random_hermitian_positive_definite(Matrix<T, Device::CPU>& matrix) {
   // Check if block matrix is square
   util_matrix::assertBlocksizeSquare(matrix, "set_hermitian_random_positive_definite", "matrix");
 
-  std::size_t offset_value = 2 * matrix.size().rows();
+  std::size_t offset_value = 2 * to_sizet(matrix.size().rows());
```
rasolca: (*)

include/dlaf/util_matrix.h
```diff
@@ -301,9 +301,9 @@ void set_random_hermitian_positive_definite(Matrix<T, Device::CPU>& matrix) {
       // of distributed matrices)
       size_t seed;
       if (tile_wrt_global.row() >= tile_wrt_global.col())  // LOWER or DIAGONAL
-        seed = tl_index.row() * matrix.size().cols() + tl_index.col();
+        seed = to_sizet(tl_index.row() * matrix.size().cols() + tl_index.col());
```
rasolca: (*)

include/dlaf/util_matrix.h
```diff
@@ -301,9 +301,9 @@ void set_random_hermitian_positive_definite(Matrix<T, Device::CPU>& matrix) {
...
       else
-        seed = tl_index.col() * matrix.size().rows() + tl_index.row();
+        seed = to_sizet(tl_index.col() * matrix.size().rows() + tl_index.row());
```
rasolca: (*)

miniapp/miniapp_cholesky.cpp
```diff
@@ -27,10 +27,10 @@ using namespace dlaf;
 using T = double;
 
 struct options_t {
-  int64_t m;
-  int64_t mb;
-  int64_t grid_rows;
-  int64_t grid_cols;
+  int m;
+  int mb;
```
rasolca: `SizeType`?

miniapp/miniapp_cholesky.cpp
```diff
@@ -99,8 +99,8 @@ int hpx_main(hpx::program_options::variables_map& vm) {
 
     double gigaflops;
     {
-      double n = matrix.size().rows();
-      double add_mul = n * n * n / 6;
+      auto n = static_cast<std::size_t>(matrix.size().rows());
+      auto add_mul = static_cast<std::size_t>(n * n * n / 6);
       gigaflops = total_ops<T>(add_mul, add_mul) / elapsed_time / 1e9;
```
rasolca: Why `int -> size_t -> double` instead of `int -> double -> double`?

miniapp/miniapp_cholesky.cpp
```diff
@@ -138,16 +138,16 @@ int main(int argc, char** argv) {
   // clang-format off
   desc_commandline.add_options()
     ("matrix-size",
-     value<int64_t>()->default_value(4096),
+     value<int>()->default_value(4096),
```
rasolca: `SizeType` ?

miniapp/miniapp_cholesky.cpp
```diff
@@ -138,16 +138,16 @@ int main(int argc, char** argv) {
...
      "Matrix size.")
     ("block-size",
-     value<int64_t>()->default_value(256),
+     value<int>()->default_value(256),
```
rasolca: `SizeType` ?

miniapp/miniapp_cholesky.cpp
```diff
@@ -182,9 +182,8 @@ int main(int argc, char** argv) {
 
 options_t check_options(hpx::program_options::variables_map& vm) {
   options_t opts = {
-      vm["matrix-size"].as<int64_t>(), vm["block-size"].as<int64_t>(),
-
-      vm["grid-rows"].as<int64_t>(),   vm["grid-cols"].as<int64_t>(),
+      vm["matrix-size"].as<int>(),   vm["block-size"].as<int>(),
```
rasolca: `SizeType`?

src/matrix/distribution.cpp
```diff
@@ -74,6 +74,7 @@ Distribution& Distribution::operator=(Distribution&& rhs) noexcept {
 void Distribution::computeGlobalSize(const LocalElementSize& local_size,
                                      const comm::Size2D& grid_size) noexcept {
   assert(grid_size == comm::Size2D(1, 1));
+  (void) grid_size;
```
rasolca: At first sight this looked strange, than I realized that you need (in release mode) to mark grid_size as used.
I think is better to add a function or macro with a self-explaining name.

test/include/dlaf_test/util_types.h
```diff
@@ -27,13 +27,13 @@ using BufferTypes =
 template <class T>
 struct TypeUtilities {
   /// @brief Returns r.
-  static T element(double r, double i) {
+  static T element(double r, double) {
```
rasolca: What about `double /*i*/)`to keep the parameter name?

test/include/dlaf_test/util_types.h
```diff
@@ -27,13 +27,13 @@ using BufferTypes =
...
     return static_cast<T>(r);
   }
 
   /// @brief Returns r.
   /// @pre r > 0
-  static T polar(double r, double i) {
+  static T polar(double r, double) {
```
rasolca: as above I suggest `double /*theta*/)`.

Please fix also the complex version below (doc says `theta`, while function signature has `i`)

src/matrix/distribution.cpp
albestro: I agree with you about creating an ad hoc macro, but I'm wondering if we really need to pass this parameter.

miniapp/miniapp_cholesky.cpp
albestro: `total_ops` accepts `std::size_t` as parameters (being number of additions and multiplications).

So, `n^3 / 6` is cast to an integer, while the output of the `total_ops` is implicitly cast to double when divided by `elapsed_time`.

src/matrix/distribution.cpp
rasolca: You are right. Maybe the name can be changed to `computeGlobalSizeForNonDistr`

albestro: I've removed the superfluous parameter and renamed the member function with the name suggested. I didn't move the assert inside the c'tor, since we would check a value hardcoded few lines above (but in case, I can quickly add it back).

miniapp/miniapp_cholesky.cpp
rasolca: That was part of the question; consider changing total_ops to double to avoid multiple conversions...
moreover max of size_t (~(2M)^3) may be a little bit tight. Currently we plan to run the eigensolver on size 100k matrices, which is just an order of magnitude away from the limit of the number whose cube can be fit in size_t.

albestro: I changed the code, the function `total_ops` now accepts two `double`. 😉

PR #162, opened 2020-03-13
Fix set_random_hermitian_positive_definite for matrix size which are not multiple of block size

### LINKED ISSUE #161: set_random_hermitian_positive_definite does not work with matrix size not multiple of block size
The utility function `set_random_hermitian_positive_definite` does not work if the matrix size is not multiple of the block size.

The test does not cover this case, and this may affect also other tests that use the same strategy for computing matrix size as a multiple of the communication grid size.

albestro:
Close #161 

This PR fixes the problem about set_random_hermitian_matrix.

It is still open to discussion how to solve the problem of testing edge configuration cases.

### DIFF
--- a/test/unit/matrix/test_util_matrix.cpp
+++ b/test/unit/matrix/test_util_matrix.cpp
@@ -163,6 +163,7 @@ TYPED_TEST(MatrixUtilsTest, SetRandomHermitianPositiveDefinite) {
   std::vector<TestSizes> square_blocks_configs({
       {{0, 0}, {13, 13}},  // square null matrix
       {{5, 5}, {26, 26}},  // square matrix single block
+      {{9, 9}, {3, 3}},    // square matrix multi block "full-tile"
       {{13, 13}, {3, 3}},  // square matrix multi block
   });
 
--- a/test/unit/matrix/test_util_matrix.cpp
+++ b/test/unit/matrix/test_util_matrix.cpp
@@ -122,16 +122,17 @@ void check_is_hermitian(dlaf::Matrix<const T, Device::CPU>& matrix,
       if (current_rank == owner_original) {
         const auto& tile_original = matrix.read(index_tile_original).get();
         hpx::shared_future<dlaf::Tile<const T, Device::CPU>> tile_transposed;
+        auto size_tile_transposed = transposed(tile_original.size());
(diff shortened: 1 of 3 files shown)


albestro:
@rasolca 
The current strategy of computing the global matrix size may lead to not test some edge cases.

The relevant one for this PR is the one in which the matrix size is not multiple of the tile size, but since the global size is computed by multiplying the tile size by the number of ranks in cols or rows, this case would not be tested.

How do we like to address this? Should we check also other tests that use the same mechanism (i.e. `dlaf_test::CommunicatorGrid6RanksEnvironment`)

rasolca:
The problem arise from the "base" sizes chosen for the tests:
```
      {{0, 0}, {13, 13}},  // square null matrix
      {{26, 26}, {2, 2}},  // square matrix multi block
      {{2, 2}, {6, 6}},    // square matrix single block
      {{4, 4}, {13, 13}},  // square matrix single block
```
The current mechanism allows to check multiple cases at once if used correctly.
E.g. for `Set` one of the size is `{{16, 24}, {3, 5}}`, which lead to (x+ y means x full tiles + y elements)
```
ranks |   1  |   2  |   3
row   |  5+1 | 10+2 | 16+0
col    |  4+4 | 9+3 | 14+2
```
i.e. in only one case the size is a multiple of the blocksize

test/unit/matrix/test_util_matrix.cpp
```diff
@@ -161,9 +162,8 @@ void check_is_hermitian(dlaf::Matrix<const T, Device::CPU>& matrix,
 TYPED_TEST(MatrixUtilsTest, SetRandomHermitianPositiveDefinite) {
   std::vector<TestSizes> square_blocks_configs({
       {{0, 0}, {13, 13}},  // square null matrix
-      {{26, 26}, {2, 2}},  // square matrix multi block
-      {{2, 2}, {6, 6}},    // square matrix single block
-      {{4, 4}, {13, 13}},  // square matrix single block
+      {{5, 5}, {26, 26}},  // square matrix single block
+      {{13, 13}, {3, 3}},  // square matrix multi block
```
rasolca: I would keep a case in which the size is multiple of block size...

PR #163, opened 2020-03-17
Proposal: how to handle errors

albestro:
This is just a proposal to work on.

Requirements:
- Lazy evaluation of messages
- 3 Levels of assertions:
  - **ASSERT**: it replaces the classic C `assert`, so it is invasive because "we don't care too much about performance, now we are debugging";
  - **DEV**: it is a tradeoff between debugging and release. It should be not so much invasive, but it should still give an extra level of assertion w.r.t. PRECONDITIONS
  - **PRECONDITIONS**: enable just preconditions, for best performances.

The idea is to have `PRECONDITIONS <= DEV <= ASSERT`, so we have that:
- PRECONDITIONS are always enabled
- DEV can be enabled with a compiler flag, both in release and debug builds, in addition to PRECONDITIONS
- ASSERT is enabled just on debug builds, together with previous ones.


albestro:
I investigated two options:
1. branch `assert_proposal` 246bc05
2. branch `assert_proposal_streaming_macros` 3d9b66e

Both solutions share these properties:
- message definition with `operator<<`;
- lazy evaluation of the message: it is evaluated iff the condition is triggered;
- no overhead in case a particular category of assert is disabled (e.g. `DLAF_ASSERT` in Release builds);

The difference between the solutions is that the first one accepts two parameters, the condition and the message, while the second tries to mimic the behavior of GoogleTest macros, accepting just the condition and the message is "streamed" to the macro.

Examples
```cpp
// solution 1: "classic macro"
DLAF_PRECONDITION(i == 0, "i should be 0, but it is " << i);

// solution 2: "streaming macros"
DLAF_PRECONDITION(i == 0) << "i should be 0, but it is " << i;
```

The second solution is a bit tricky to implement and it relies on compiler optimizations to not add overhead when the category is disabled (i.e. `if (false) { print_message() }` should not result in any instruction). The first one is more straightforward to implement.

At the moment I cannot see any benefit from using one or the other, and the implementation complexity is currently driving my decision.

I'd like to have some feedback from you about your view and preferences @rasolca @teonnik

rasolca:
I would go for the first option, but first I would try to write some complex messages to check if they can break the macro call mechanism.
For example multiple type templates break macros, but at a first sight I don't see any message that contains it.

albestro:
The basic `assert` from `#include <cassert>` does not allow a custom message. Replacing existing `assert` with custom ones would require to define a specific message for each one.

For this reason, it may be worth evaluating to add an option to not specify it getting a default message.

With "classic" solution an option may be to "overload" macros (see https://godbolt.org/z/m9Nwkh).

albestro:
I don't like very much macros, so I'm investigating other options trying to reduce the size of macros without loosing constrinats.

I'm trying to get inspiration from resources shared offline by @teonnik like https://foonathan.net/2016/09/assertions/.

teonnik:
> https://foonathan.net/2016/09/assertions/.

There is a lot of good information in the article, we can use some of the techniques outlined there.

1. We should include information about the file, line and the expression that triggered the assert.
2. Call `std::abort` or `std::terminate` if the assertion fails.
3. Make sure the asserts are noop when disabled.
4. Have two assert levels: one enabled by default for cheap checks and one enabled explicitly for expensive checks.

I suggest the following two macros based, loosely, on the C++20 contracts design:

- `DLAF_ASSERT` - cheap checks; ON by default unless the `DLAF_DISABLE_ASSERT` flag is passed
- `DLAF_AUDIT`  - expensive checks; ON if `DLAF_ENABLE_AUDIT` flag and **not** `DLAF_DISABLE_ASSERTS` flag is passed

This has some similarities to the Alberto's proposal:

1. `PRECONDITIONS` ~ `DLAF_ASSERT`, with the difference that `DLAF_ASSERT` can be disabled with the `DLAF_DISABLE_ASSERTS` flag
2. `DEV` ~ `DLAF_AUDIT`, there isn't much difference here, the flag to enable the macro is named explicitly: `DLAF_ENABLE_AUDIT`
3. `ASSERT` - there is no direct equivalent, but I don't think this is needed. If you think a check is too expensive to be enabled in Release, don't pass `DLAF_ENABLE_AUDIT`. If you don't want any checks, pass `DLAF_DISABLE_ASSERT`. If you want your debug build with expensive checks, pass `DLAF_ENABLE_AUDIT`.

Regarding composing messages, as @rasolca  mentioned, some messages may break the macros call mechanism. I suggest we handle that in a function called from the macro and demand that the extra arguments passed to the macro support `operator<<` for printing. The API will look like that for both `DLAF_ASSERT` / `DLAF_AUDIT`:

```
Tile t1 = ...
Tile t2 = ...
DLAF_ASSERT(t1 == t2, "Oops, sth is wrong!", t1, t2)
```  
the macro calls a function which calls the `operator<<` on the tile object to print:

```
[ERROR] <path_to_file> : <line> : <expression> : "Oops, sth is wrong!"
t1 : ...
t2 : ...
```

rasolca:
> 3. ASSERT - there is no direct equivalent, but I don't think this is needed. If you think a check is too expensive to be enabled in Release, don't pass DLAF_ENABLE_AUDIT. If you don't want any checks, pass DLAF_DISABLE_ASSERT. If you want your debug build with expensive checks, pass DLAF_ENABLE_AUDIT.
@teonnik I suggested the three levels (1. no / very low performance impact, 2.low / medium impact, 3. high impact) with the following situation in mind.
Imagine you encounter a bug for a specific problem, which is big and the bug cannot be reproduced in smaller cases.
In this case enabling all the checks may be very expensive (in term of time and node-hours), while it may be enough to enable only the mid level checks to identify the bug.

include/dlaf/common/assert.h
```diff
@@ -0,0 +1,38 @@
...
+              << __LINE__ << "\n"                                                            \
+              << dlaf::common::concat(#condition, ##__VA_ARGS__) << std::endl;               \
+    std::terminate();                                                                        \
+  }
+
+#ifndef NDEBUG
+#define DLAF_ASSERT(condition, ...) DLAF_TEST("ASSERT", condition, ##__VA_ARGS__)
+#else
+#define DLAF_ASSERT(condition, ...)
+#endif
```
teonnik: We might want to use a custom flag: e.g. `DLAF_DISABLE_CHECK` (or ASSERT) instead of NDEBUG here for two reasons:
- NDEBUG might affect projects we depend on if they use the `assert()` macro. We might want `DLAF_ASSERT` enabled even if assertions in depending projects are disabled.
- We might want to have some checks enabled in Release too.

include/dlaf/common/assert.h
```diff
@@ -0,0 +1,38 @@
...
+//
+
+#pragma once
+
+#include <exception>
+#include <iostream>
+
+#include "dlaf/common/utils.h"
+
+#define DLAF_TEST(category, condition, ...)                                                  \
```
teonnik: I'd call this `DLAF_CHECK` so that it's not easily confused with unit testing.

include/dlaf/common/assert.h
```diff
@@ -0,0 +1,38 @@
...
+#define DLAF_ASSERT(condition, ...)
+#endif
+
+#ifdef DLAF_ENABLE_AUDIT
+#define DLAF_AUDIT(condition, ...) DLAF_TEST("AUDIT", condition, ##__VA_ARGS__)
+#else
+#define DLAF_AUDIT(condition, ...)
+#endif
+
+#define DLAF_PRECONDITION(condition, ...) DLAF_TEST("PRECONDITION", condition, ##__VA_ARGS__)
```
teonnik: In C++20 contracts, preconditions are checks on the parameters of a function. DLAF_ASSERT/AUDIT can do that too. Do we need this?

albestro:
Thank you for your comment @teonnik!

As you said, this proposal more or less addresses the outline you gave at the beginning:
1. ✔️ It prints the information about the file, the line and the function
2. ✔️ It calls `std::terminate()`, so it by defaults calls `std::abort()` but it is also possible to call a custom terminate handler set with [`set_terminate`](https://en.cppreference.com/w/cpp/error/set_terminate) (e.g. for HPX or MPI or whatever necessary)
3. ✔️ NOP: with the classic solution, the macro is replaced with nothing, so absolutely no code is executed, nor for the condition nor for the message, and no branches are introduced.
4. The three levels instead of just two is a requirement introduced by @rasolca, and I agree with him. It may be useful during debugging to have an intermediate level of checks that does not force you to activate the more strict and deep DLAF_ASSERT that would affect dramatically the performances.

include/dlaf/common/assert.h
albestro: This macro is exactly for checking parameters of algorithms. It is clearly an assert-like thing, but as stated before in the three levels hierarchy, we want them to be always on.

include/dlaf/common/assert.h
albestro: The idea of `DLAF_ASSERT` is to have a replacement for the `assert` basic macro, mainly allowing to have a custom message.

teonnik:
I'm OK with 3 levels instead of 2 but I'm not sure about the naming of things at the moment. I would like to have a macro that is ON by default and explicitly disabled by a flag. At the moment, that macro seems to be `DLAF_PRECONDITION`. However, I would also like to use such macro inside a function, not only for checking the parameters of a function. That is why I think the `PRECONDITION` suffix is not the best name for such macro. 

Now that we have 3 levels, naming is a bit more challenging, but perhaps a simple DLAF_ASSERT_LV[1,2,3] will suffice?

`DLAF_ASSERT_LV1` - ON by default, disabled by `DLAF_DISABLE_ASSERT_LV1`
`DLAF_ASSERT_LV2` - OFF by default, enabled by `DLAF_ENABLE_ASSERT_LV2`
`DLAF_ASSERT_LV3` - OFF by default, enabled by `DLAF_ENABLE_ASSERT_LV3`

albestro:
At the moment we have `DLAF_AUDIT` (previously `DEV` check), that is by default `OFF` but you can enable with the flag `DLAF_ENABLE_AUDIT`. I think it is what we were talking about till now.

Why do you want to have something always `ON` by default that you can turn off by explicitly switching a flag?

teonnik:
> Why do you want to have something always ON by default that you can turn off by explicitly switching a flag?

Because I would like to check class invariants in constructors with such macro. Most of the time these are not expensive and are not worth switching off. The flag to switch it OFF is for very rare use: to test if our assumption that such macro is cheap is indeed true.

albestro:
Good point!

I'm not really getting the purpose of having a way to disable them, but we can manage it.

Since I would use DLAF_PRECONDITION but, as you said too, I would not like the naming in that case, I'm thinking of two options:
- make the DLAF_PRECONDITION name more generic;
- duplicate the macro and have multiple clone of the same macro with different name;

At first sight I prefer the second option, because they are clone (at least for now) but in the future we have much finer control since the metadata about the category is kept.

albestro:
UPDATE: I was working on the `DLAF_PRECONDITION` part on another branch.

Precondition need more articulated checks wrapped in a function that also manages the error message. We will talk about this in the dedicated PR #131.

For what is interesting for this one, I would say that we can skip DLAF_PRECONDITION and dedicate it to what @teonnik was talking about. Do you have any name suggestion?

include/dlaf/common/assert.h
```diff
@@ -0,0 +1,38 @@
...
+#pragma once
+
+#include <exception>
+#include <iostream>
+
+#include "dlaf/common/utils.h"
+
+#define DLAF_CHECK(category, condition, ...)                                                           \
+  if (!(condition)) {                                                                                  \
+    std::cerr << "[" category "-ERROR] " << __FILE__ << ":" << __LINE__ << " : " << DLAF_FUNCTION_NAME \
```
teonnik: Is `DLAF_FUNCTION_NAME` the name of the function calling the macro? Do we need that?

The string representation of the expression is missing from the string: `#condition`. I would put that instead of the function name.

include/dlaf/common/assert.h
```diff
@@ -0,0 +1,38 @@
...
+//
+
+#pragma once
+
+#include <exception>
+#include <iostream>
+
+#include "dlaf/common/utils.h"
+
+#define DLAF_CHECK(category, condition, ...)                                                           \
```
teonnik: Hmm, can we actually move all the logic into the lowest level assert (e.g. DLAF_PRECONDITION) and scrap `DLAF_CHECK` entirely?

include/dlaf/common/assert.h
albestro: `DLAF_FUNCTION_NAME` is set from CMake, and it is:
- `__PRETTY_FUNCTION__` if available
- `__func__` otherwise

https://github.com/eth-cscs/DLA-Future/blob/aece6d0b770f6bbaff02f9e64bbb62c781ebdad2/src/CMakeLists.txt#L47-L55

So we can use the GCC extension `__PRETTY_FUNCTION__` when available. We can decide to use just the C++ standard `__func__`. Vaguely related to this, I have on my to-do list to anticipate [`source_location`](https://en.cppreference.com/w/cpp/utility/source_location) from C++20.

About the `#condition`, it is in the next line. See next examples.

## DEFAULT MESSAGE
```cpp
DLAF_ASSERT((false == true && index != common::Index2D<int, TAG>{}));
```
```bash
[ASSERT-ERROR] /Users/ialberto/workspace/repos/DLA-Future/example_assert.cpp:11 : int main()
(false == true && index != common::Index2D<int, TAG>{})
```

## CUSTOM MESSAGE
```cpp
DLAF_ASSERT((false == true && index != common::Index2D<int, TAG>{}), "Custom message", index);
```
```bash
[ASSERT-ERROR] /Users/ialberto/workspace/repos/DLA-Future/example_assert.cpp:12 : int main()
(false == true && index != common::Index2D<int, TAG>{}) Custom message (1, 2)
```

albestro:
After a quick discussion with @teonnik, I think that we have still some open questions and the entire thing needs a recap.

### Preamble
The first problem leading to misunderstanding can be summarized as follows:

In the initial proposal `DLAF_ASSERT` is the assertion with the highest impact on performances, while @teonnik was suggesting to use it for low impact checks (after renaming it)
> - `DLAF_ASSERT` - cheap checks; ON by default unless the `DLAF_DISABLE_ASSERT` flag is passed

This basic "inverted" mapping may have lead to a lot of misunderstandings. In next paragraphs I will try to isolate the needed decisions.

### A. Macro triggers
A problem is about how to trigger the assertions.

1.  The initial idea of this proposal was to link our assertions to build types and have additional flags for turning off optional ones.
    - _HIGH impact_ - ON by default, just in debug builds (controlled by the standard `NDEBUG` flag, acting as a replacement for `<cassert>` macro)
    - _MEDIUM impact_ - OFF by default, can be enabled with every build type through a specific flag `DLAF_ENABLE_*`
    - _LOW impact_ - ON by default, cannot be disabled (e.g. class invariant)
2. @teonnik suggests to isolate our assertion system from other libraries, to let us enable our assertion independently from other libraries and build type
    - _HIGH impact_ - OFF by default, enabled by `DLAF_ENABLE_ASSERT_LV3`
    - _MEDIUM impact_ - OFF by default, enabled by `DLAF_ENABLE_ASSERT_LV2`
    - _LOW impact_ - ON by default, disabled by `DLAF_DISABLE_ASSERT_LV1`

I think that we should provide a way to address basic/classic problems without over-engineering the thing, in case of need a developer can decide to alter a bit the mechanism to match his needs of the moment (e.g. manually enabling a particular category of assertions in a particular build condition).

We may decide to get the best from both solution. I like the idea of having a debug build that automatically performs _HIGH_ impact assertions (without touching anything).

### B. Macro names
Do we want to name the macro to something related to the "weight" of the assertion or to "category" usage? e.g. `DLAF_ASSERT_LV3` vs `DLAF_ASSERT`

1. Named by-level
 About this, I find a bit risky the numbers ordering, but it may be a matter of getting used to it. I consider the "lowest **level**" assertion the equivalent of the `cassert`, instead of the one with the "lowest **impact**" on performances.
    - _HIGH impact_ `DLAF_ASSERT_LV3`
    - _MEDIUM impact_ `DLAF_ASSERT_LV2`
    - _LOW impact_ `DLAF_ASSERT_LV1`

2. Named by-type
    - _HIGH impact_ `DLAF_ASSERT`
    - _MEDIUM impact_ `DLAF_AUDIT`
    - _LOW impact_ `DLAF_?` (this is the one I was looking the name for)

teonnik:
> The initial idea of this proposal was to link our assertions to build types and have additional flags 

Note that we can still link build types and assertions. For example we can pass the ENABLE flags for LV3 (and LV2) asserts in the Debug mode from CMake:

```
target_compile_definitions(DLAF PUBLIC  $<$<CONFIG:Debug>:DLAF_ENABLE_ASSERT_LV3>)
```
That way, the HIGH IMPACT macro can still act as `<cassert>` in Debug mode. The only difference is that it's not controlled by `NDEBUG`.

Regarding naming, I suggested LV1, LV2 and LV3 because they provide some indication of lowest to highest impact. I admit this may not be very intuitive. In the initial 2-level proposal: ASSERT (low) and AUDIT (high) seemed like good names (based on C++20 contract), now that we have 3 levels, I am not entirely sure AUDIT makes sense.

Sely85:
I actually don't have an in-depth understanding of error handling, sorry.. However, I would prefer the "easiest and safer" approach. 

Concerning safety, I'd love to have the "high impact" option to be on by default in debug mode.
I would also feel kind of "safer" if the "low impact" option cannot be switched off, but I don't know if this is computationally expensive. 
Hence, in my opinion, it would be useful to link the assertions to build type. 

Regarding the ease, I would prefer macro names related to the "level". But LV1, LV2, and LV3 are a bit "confusing" to me. What about using straight the words "HIGH", "MEDIUM" and "LOW" (even if it doesn't sound right neither)?

albestro:
Thank you all for your feedbacks.

Trying to resume here the proposal, please correct me if I'm wrong or if you have any remark.

### A. Macro triggers

Default values for flags
| | Debug | ~Debug |
--|:--:|:--:
HIGH impact | ON | OFF
MED impact | ON | OFF
LOW impact | ON | ON

1. We will have a flag `DLAF_ENABLE_ASSERT_*` to turn ON/OFF the relative level;
2. These flags obviously have priority over the build type, providing flexibility to enable a custom set of assertion levels independently from the build type selected;
3. When build type is switched, flags must be reset to the default values for that particular build type;
4. NDEBUG standard define will not be taken into account for our purposes.

src/CMakeLists.txt
```diff
@@ -42,10 +42,49 @@ target_link_libraries(DLAF
...
+# In order to force a flag configuration, you have to:
+# - set DLAF_ASSERT_DEFAULT_BUILDTYPE=${CMAKE_BUILD_TYPE} (in order to bypass this check)
+# - set DLAF_ASSERT_ENABLE_* variable(s) as you prefer
+if (NOT DLAF_ASSERT_DEFAULT_BUILDTYPE STREQUAL CMAKE_BUILD_TYPE)
+  set(DLAF_ASSERT_DEFAULT_BUILDTYPE ${CMAKE_BUILD_TYPE} CACHE INTERNAL "" FORCE)
+
+  set(DLAF_ASSERT_ENABLE_LOW  ${DLAF_ASSERT_DEFAULT_LOW}  CACHE BOOL "Enable low impact assertions"    FORCE)
+  set(DLAF_ASSERT_ENABLE_MED  ${DLAF_ASSERT_DEFAULT_MED}  CACHE BOOL "Enable medium impact assertions" FORCE)
+  set(DLAF_ASSERT_ENABLE_HIGH ${DLAF_ASSERT_DEFAULT_HIGH} CACHE BOOL "Enable high impact assertions"   FORCE)
+endif()
```
albestro, line 80: @teonnik this is the solution I was talking about.

- It gives you a default value for debug or release builds
- If you do not change the build type, you are allowed to customize your flags setup
- As soon as you switch build type, flags configuration is reset to default one
- You can define a custom configuration for flags at first project configuration (e.g. for CI particular builds) as described in the documentation

albestro: Yes, it is the where I started from.

IMHO it is a trade-off between having the extra functionality and the cost of the additional variable. One may interpret the `DLAF_ASSERT_DEFAULT_BUILDTYPE` like a variable saying that you are going to specify flags setup for a particular build. However, it is true that in certain situations this auto-reset functionality may be counter-productive, but I think it may not the "main stream" situation.

About your comment, I'm absolutely not against it, we just have to decide to loose the "requirement" of not resetting flags on build type changes. In that case the default setup works just on the first configuration (with respect to the first build type chosen), then it is up to the developer to turn ON/OFF flags if he decides to switch build type. But it is completely ok for me, even if it will happen that I will forget to set assert switches 😂

src/CMakeLists.txt
```diff
@@ -42,10 +42,49 @@ target_link_libraries(DLAF
...
+# DLAF_FUNCTION_NAME
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles("int main() { const char *name = __PRETTY_FUNCTION__; }" is_pretty_function_available)
+
+target_compile_definitions(DLAF PUBLIC
+  DLAF_FUNCTION_NAME=$<IF:$<BOOL:is_pretty_function_available>,__PRETTY_FUNCTION__,__func__>
+)
+
+# Assertions
+if (CMAKE_BUILD_TYPE STREQUAL Debug)
```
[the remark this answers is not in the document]
albestro: Good point! I don't know if we want to support them.

AFAIK we can manage multi-configuration generators just through generator expressions (i.e. `<$CONFIG:Debug>`) but I think in that case we cannot go further than just the default values (assuming that in the IDE you do not add non-standard configuration names, e.g. `DEBUG`).

src/CMakeLists.txt
teonnik: After some discussion with Alberto, the following question came up:

What happens if `DLAF_ASSERT_DEFAULT_BUILDTYPE=Release` and I change from `CMAKE_BUILD_TYPE=Release` to `Debug` but forget to update the former?

The answer is that changing `DLAF_ASSERT_ENABLE` variables from the CACHE will no longer have an affect as the variables will always be overwritten to the default.

What happens if I change both `DLAF_ASSERT_DEFAULT_BUILDTYPE` and `CMAKE_BUILD_TYPE` from `Debug` to `Release`?

Unfortunately, the CACHE variables `DLAF_ASSERT_ENABLE` remain with values from `Default` as the `IF` condition was not triggered.

To avoid these traps, we decided to stick with the simple solution and just provide a single set of defaults via the `option()` clauses.

albestro:
Good point @rasolca I'm going to add some basic documentation for the macros! 😉

About the messag

albestro:
In addition to my previous comment where I highlighted the fact that category is not used for the moment, I want to list here main updates:
- documentation updated;
- I added const-correctness to the `concat` template function

PR #166, opened 2020-03-18
Bump the version to CMake 12

teonnik:
- The version sets policy CMP0074 allowing to specify a custom MPI
  directory via MPI_ROOT. The top CMake file already sets the version to
  12


teonnik:
Incorporating the FindMKL from https://github.com/eth-cscs/cmake-recipes/pull/8 will require non-trivial amount of changes and may need some more work. This PR is tiny and trivial, I suggest we merge it for the time being.

PR #170, opened 2020-03-25
Squash minor warnings and compilation errors

teonnik:
- MPI_Datatype is not guaranteed to be constexpr. it is not for OpenMPI
  where a (void*) cast is involved

- Enumerate all cases instead of reverting to default. This has the
  advantage that if the enum is extended, we won't fall into the default
  case, instead we get a clear warning for unhandled conditions. Note
  also that blas:: uses enum classes so there are no implicit
  conversions to int/char

- `memory_chunk.h` : use static cast

- `communicator_impl` : the function is not guaranteed to be constexpr
  because MPI_COMM_WORLD has different implementations in different MPI
  libraries

- Declare global variables within tests as `const`.

- Some macros automatically place `;` at the end of expressions, a
  second `;` results in a waring for empty expression. Follow the
  GoogleTest convention and leave the user of the macro to place the `;`.

### DIFF
--- a/src/communication/datatypes.cpp
+++ b/src/communication/datatypes.cpp
@@ -1,0 +1,52 @@
+#include <dlaf/communication/datatypes.h>
+
+namespace dlaf {
+namespace comm {
+
+template <>
+MPI_Datatype mpi_datatype<char>::type = MPI_CHAR;
+
+template <>
+MPI_Datatype mpi_datatype<short>::type = MPI_SHORT;
+
+template <>
+MPI_Datatype mpi_datatype<int>::type = MPI_INT;
+
+template <>
+MPI_Datatype mpi_datatype<long>::type = MPI_LONG;
+
+template <>
+MPI_Datatype mpi_datatype<long long>::type = MPI_LONG_LONG;
+
+template <>
+MPI_Datatype mpi_datatype<unsigned char>::type = MPI_UNSIGNED_CHAR;
+
+template <>
+MPI_Datatype mpi_datatype<unsigned short>::type = MPI_UNSIGNED_SHORT;
+
+template <>
+MPI_Datatype mpi_datatype<unsigned int>::type = MPI_UNSIGNED;
+
+template <>
+MPI_Datatype mpi_datatype<unsigned long>::type = MPI_UNSIGNED_LONG;
+
+template <>
+MPI_Datatype mpi_datatype<unsigned long long>::type = MPI_UNSIGNED_LONG_LONG;
+
+template <>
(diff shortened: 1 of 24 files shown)


src/communication/communicator_grid.cpp
```diff
@@ -33,9 +33,9 @@ CommunicatorGrid::CommunicatorGrid(Communicator comm, IndexT_MPI nrows, IndexT_M
   }
 
   MPI_Comm mpi_full, mpi_col, mpi_row;
-  MPI_CALL(MPI_Comm_split(comm, is_in_grid ? 0 : MPI_UNDEFINED, key_full, &mpi_full));
-  MPI_CALL(MPI_Comm_split(comm, index_row, key, &mpi_row));
-  MPI_CALL(MPI_Comm_split(comm, index_col, key, &mpi_col));
+  MPI_CALL(MPI_Comm_split(comm, is_in_grid ? 0 : MPI_UNDEFINED, key_full, &mpi_full))
```
albestro: I completely agree with your suggestion to use GoogleTest convention about semicolon and macros. What do you think about applying it here too (and everywhere else)?

rasolca: I suggest the `do{ ... } while(0)` approach as used for other macros in the code.

teonnik: The `do{ ... } while(0)` approach seems fine to me. At the moment, the macro opens a scope {}, if I place a `;` after it, I get a warning for empty expression. That won't be the case with the `do ..` approach.

teonnik:
There was a bug in the implementation of `communication/datatypes.h`.  The definitions of the static variable type caused multiple definition linker error when the header was included in two linked translation units. The definitions are now moved to a `.cpp` file which resolves the issue.

MPI_CALL now uses the `do {} while(0)` mechanism and a semicolon for consistency.

src/communication/datatypes.cpp
```diff
@@ -0,0 +1,52 @@
...
+MPI_Datatype mpi_datatype<bool>::type = MPI_CXX_BOOL;
+
+template <>
+MPI_Datatype mpi_datatype<std::complex<float>>::type = MPI_CXX_FLOAT_COMPLEX;
+
+template <>
+MPI_Datatype mpi_datatype<std::complex<double>>::type = MPI_CXX_DOUBLE_COMPLEX;
+
+} // end namespace comm
+} // end namespace dlaf
```
rasolca: Please as stated in the coding style we do not use end of namespace comments.

teonnik:
I squashed the commits and updated the commit message.

Issue #169, opened 2020-03-22
Explicit/External Template Instantiations (ETI)


teonnik:
There are two approaches outlined in the links below:

1. https://daniele77.github.io/general/2019/05/03/explicit-instantiation-reusability-modularity.html
2. https://arne-mertz.de/2019/02/extern-template-reduce-compile-times/

In approach 1. we place the definition of the template class/function inside a .cpp file and instantiate the class/function there.

In approach 2. we leave the definition of the template class/function in the header (or a file included by the header, e.g. .tpp) and place an `extern` template declaration alongside it. We then only instantiate the templates within the .cpp file (the definition doesn't go there). `extern` template is a C++11 feature. 

Approach 2. is what Kokkos is doing (https://github.com/kokkos/kokkos/blob/master/core/src/impl/Kokkos_ViewFillCopyETIDecl.hpp , note the INST and DECL macros). That way, the compile times are reduced for the most common combinations of types without preventing users from using other type combinations. Moreover, the definition and declaration can now be kept in the same file, avoiding some repetition.

There is also a trade-off that has to be acknowledged with either approach: template functions which are instantiated might not be inlined unless link-time optimization (LTO) is enabled. Now, that might be irrelevant, as most performance comes from back end libraries, but it would still be good to measure if there are noticeable performance difference with/without LTO. Note also, that LTO has an impact on the build speed.

teonnik:
#172

PR #172, opened 2020-03-25
Explicit Template Instantiations

teonnik:
This PR addresses https://github.com/eth-cscs/DLA-Future/issues/169 . ETI is implemented for `Matrix`, `Tile` and a few other types using the `extern` approach (see https://github.com/eth-cscs/DLA-Future/issues/169). 

Note that separating the implementation of some templated functions into a `tpp` file is not needed for the purposes of ETI. We can move the definitions into the `.h` file if there is consensus. That will save us some boilerplate.


teonnik:
Note that the only difference between the DECL and INST macros is an `extern` parameter. There is some room for reduction on the boilerplate.

teonnik:
> I'm wondering if ETI of the full object can influence inlining of small methods. Do you know how it works?

Yes it does, same as anything else in a cpp file, inlining is no longer possible UNLESS we use LTO (link-time optimization). An approach to alleviate the issue, is to perhaps only instantiate selected member methods of the class.

> Moreover as mentioned by you I would reduce a bit code duplication. I have a suggestion: https://godbolt.org/z/QhKzTc . What do you think?

I like the DECL/INST macros (prefixed with DLAF_) and the extra parameter on the A_ETI/B_ETI macros (I'd put KEYWORDS first) but going any further seems like too much macro magic and is a bit harder to read. I think it's also fine if we specify `extern template` and `extern` directly into the A_ETI and B_ETI macros. That way we won't have to keep including a header for DECL/INST macros. If we want to shorten the above, we can add the template keyword and pass only `extern` for DECL and nothing for INST. Apparently macros with empty parameters are allowed:

https://stackoverflow.com/questions/7666344/are-empty-macro-arguments-legal-in-c11

rasolca:
> Yes it does, same as anything else in a cpp file, inlining is no longer possible UNLESS we use LTO (link-time optimization). An approach to alleviate the issue, is to perhaps only instantiate selected member methods of the class.

This is what puzzles me, keeping in mind that all the members defined in a class are marked with inline:
In the usual .h and .cpp structure, the .h contains the class declaration + inlined members definition + other members declaration, while in the .cpp you find the other member definition.
The compiler has full control (can decide to inline or not) on members marked as inline (either implicitly in the class declaration or explicitly), while it has no control on members defined in the .cpp file.

The question is how ETI deal with member functions which as been declared as inline.

rasolca:
Apparently inline marked methods are not instantiated and can be inlined with ETI too:
https://godbolt.org/z/vTXPB9

Therefore if you want ETI to have some impact the "long" members has to be defined outside 
the class declaration and therefore I would keep the .h and .tpp distinction to underline the difference.

teonnik:
It seems that the compiler decides whether to instantiate a method or to inline it depending on whether it thinks the method is worth instantiation. For gcc and clang, defining the method outside of the class seems to encourage them to not inline it but it is not necessary as can be seen from this : https://godbolt.org/z/pd5VlE .

rasolca:
Of course, you can always try to steer compiler decisions with attributes (inline, force_inline and noinline), but I find more natural the distinction that inline methods are defined in the class declaration and non inline methods outside as specified by the standard.

teonnik:
The only reason I used `__attribute__((noinline))` was to illustrate that the definition of the method can be inside the class and still e instantiated. You don't need the attribute to achieve that if the method is deemed complicated enough not to be worth inlining. In fact, that is what I tried initially, after adding a few lines with a bunch std::string, std::vector and std::cout, it was no longer inlined and clang just referenced the instantiation (gcc stubbornly kept optimizing though).

Issue #100, opened 2019-11-27
DLAF::Init()


albestro:
Thinking about PR #97, I started wondering if it could be worth it providing a library initialization in some way, being it a simple function or a "*context-like*" object, with the aim to ease the setup and the usage.

I'm talking of something similar to what we already did for tests, for which we have implemented different `main` functions (e.g. one that setup MPI, one that setup MPI+HPX).

I don't have the full picture yet, but as far as I can tell we want to provide DLAF to be used with MPI and HPX together. But I'm willing to receive comments about this point.

At the same time, DLAF can be used as the unique library that uses MPI+HPX, or in a more complex context where MPI and/or HPX are initialized externally to be used by other libraries used by the same executable and not only by DLAF. I think this is a point we should take into account for sure.

PR #174, opened 2020-03-27
SIRIUS's tall-skinny GEMM with DLA-Future

teonnik:
- The DLAF_WITH_MPI_FUTURES macro toggles between MPI integration based
  on John's MPI futures and a manual approach using MPI_Test

### DIFF
--- a/miniapp/miniapp_tsgemm.cpp
+++ b/miniapp/miniapp_tsgemm.cpp
@@ -1,0 +1,426 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2019, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#include <dlaf/blas_tile.h>
+#include <dlaf/communication/datatypes.h>
+#include <dlaf/matrix.h>
+
+#include <hpx/dataflow.hpp>
+#include <hpx/hpx.hpp>
+#include <hpx/include/resource_partitioner.hpp>
+#include <hpx/include/threads.hpp>
+#include <hpx/mpi/mpi_executor.hpp>
+#include <hpx/mpi/mpi_future.hpp>
+#include <hpx/runtime/threads/executors/pool_executor.hpp>
+
+#include <chrono>
+#include <complex>
+#include <cstdio>
+
+// In SIRIUS, `A`, `B` and `C` are usually submatrices of bigger matrices. The
+// only difference that entails is that the `lld` for `C` might be larger than
+// assumed here. Hence writing to `C` might be slightly faster than in SIRIUS.
+//
+// Assumptions: Tall and skinny `k` >> `m` and `k` >> `n`.
+//
+// Matrices: `A` (`m x k`), `B` (`k x n`) and `C` (m x n).
+//
+// `A` is complex conjugated.
+//
+// `C` is distributed in 2D block-cyclic manner. The 2D process grid is row
+// major (the MPI default) with process 0 in the top left corner.
+//
+// All matrices are distributed in column-major order.
+//
+// Local distribution of A and B. Only the `k` dimension is split. In
+// SIRIUS, `k_loc` is approximately equally distributed. `k_loc` coincides
+// with `lld` for `A` and `B`. If there is a remainder, distributed it
+// across ranks starting from the `0`-th.
+//
+
+// Forward declations
+namespace {
+
+using dlaf::SizeType;
+using dlaf::common::Ordering;
+using dlaf::comm::Communicator;
+using dlaf::comm::CommunicatorGrid;
+using dlaf::TileElementSize;
+using dlaf::TileElementIndex;
+using dlaf::GlobalElementSize;
+using dlaf::GlobalTileSize;
+using dlaf::GlobalTileIndex;
+using dlaf::LocalElementSize;
+using dlaf::LocalTileIndex;
+using dlaf::LocalTileSize;
+using dlaf::matrix::Distribution;
+using dlaf::matrix::LayoutInfo;
+using CommSize = dlaf::comm::Size2D;
+using CommIndex = dlaf::comm::Index2D;
+using ScalarType = std::complex<double>;
+using MatrixType = dlaf::Matrix<ScalarType, dlaf::Device::CPU>;
+using ConstMatrixType = dlaf::Matrix<const ScalarType, dlaf::Device::CPU>;
+using TileType = dlaf::Tile<ScalarType, dlaf::Device::CPU>;
+using ConstTileType = dlaf::Tile<const ScalarType, dlaf::Device::CPU>;
+using hpx::program_options::variables_map;
+using hpx::program_options::options_description;
+
+// Local gemm
+//
+// - `A` is transposed and has similar layout to `B`.
+void schedule_gemm(ConstMatrixType& a_mat, ConstMatrixType& b_mat, MatrixType& cini_mat);
+
+void schedule_comm(CommunicatorGrid& comm_grid, Distribution const& cfin_dist, MatrixType& cini_mat);
+
+void schedule_offload(ConstMatrixType& cini_mat, MatrixType& cfin_mat);
+
+// Initialize matrix
+void init_matrix(MatrixType& matrix, ScalarType val);
+
+// Wait for all tiles of the matrix.
+void waitall_tiles(MatrixType& matrix);
+
+// Sum the elements of the matrix. Useful for debugging.
+ScalarType sum_matrix(Communicator const& comm, MatrixType& matrix);
+
+struct params {
+  int num_iters;
+  int len_m;
+  int len_n;
+  int len_k;
+  int tile_m;
+  int tile_n;
+  int pgrid_rows;
+  int pgrid_cols;
+};
+
+// Initialize parameters and check for consistency
+params init_params(variables_map&);
+
+// Handle input options
+options_description init_desc();
+
+}  // end namespace
+
+int hpx_main(::variables_map& vm) {
+  using clock_t = std::chrono::high_resolution_clock;
+  using seconds_t = std::chrono::duration<double>;
+
+  // Input
+  ::params ps = ::init_params(vm);
+
+  // Communicators
+  ::Communicator comm_world(MPI_COMM_WORLD);
+  ::CommunicatorGrid comm_grid(comm_world, ps.pgrid_rows, ps.pgrid_cols, ::Ordering::ColumnMajor);
+
+  // Matrices `A` and `B`
+  // The matrices are distributed only along the `k` dimension. In SIRIUS, the sections assigned to each
+  // process are not exactly equal, they differ by a little in non-trivial ways. SIRIUS's distribution
+  // for A and B is NOT a special case of block cyclic distribution. In this miniapp, the distribution is
+  // emulated by DLAF local matrices.
+  int num_procs = comm_world.size();
+  int rank = comm_world.rank();
+  int k_loc = ps.len_k / num_procs + ((rank < ps.len_k % num_procs) ? 1 : 0);
+  ::MatrixType a_mat(::LocalElementSize(k_loc, ps.len_m), ::TileElementSize(k_loc, ps.tile_m));
+  ::MatrixType b_mat(::LocalElementSize(k_loc, ps.len_n), ::TileElementSize(k_loc, ps.tile_n));
+
+  // Matrices `C`-initial and `C`-final
+  using dlaf::matrix::tileLayout;
+  ::MatrixType cini_mat(::Distribution(::LocalElementSize(ps.len_m, ps.len_n),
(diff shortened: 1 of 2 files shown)


miniapp/miniapp_tsgemm.cpp
```diff
@@ -0,0 +1,473 @@
...
+    for (SizeType tile_i = 0; tile_i < tile_size.rows(); ++tile_i) {
+      LocalTileIndex a_mat_idx(0, tile_i);
+      LocalTileIndex b_mat_idx(0, tile_j);
+      LocalTileIndex c_mat_idx(tile_i, tile_j);
+
+      auto gemm_f = hpx::util::unwrapping([](auto&& a_tile, auto&& b_tile, auto&& c_tile) {
+        dlaf::tile::gemm(blas::Op::Trans, blas::Op::NoTrans, ScalarType(1), a_tile, b_tile,
+                         ScalarType(0), c_tile);
+      });
+      hpx::dataflow(gemm_f, a_mat.read(a_mat_idx), b_mat.read(b_mat_idx), cini_mat(c_mat_idx));
```
albestro: You can also use directly the dataflow without the explicit lambda like this:

```cpp
hpx::dataflow(
  hpx::util::unwrapping(dlaf::tile::gemm),
  blas::Op::Trans, blas::Op::NoTrans, ScalarType(1),
  a_mat.read(a_mat_idx),
  b_mat.read(b_mat_idx),
  ScalarType(0), cini_mat(c_mat_idx));
```

teonnik: I prefer the lambdas as I don't like placing too many parameters in `hpx::dataflow`. I find it easier to read and reason about.

teonnik:
> Would it not be better to implement Multiplication<Backend::cpu>::tallSkinny(...) that can be used directly in SIRIUS, but also in other applications?

Yes, I'm planning to factor out the the kernel once I happy with the implementation. However, supporting a general tallSkinny multiplication is tricky. I would need to implement some aspects of COSMA unless we always assume that the tall and skinny matrices are distributed along the tall dimension (as is in SIRIUS).

miniapp/miniapp_tsgemm.cpp
```diff
@@ -0,0 +1,473 @@
...
+      TileType tile = matrix(LocalTileIndex(tile_i, tile_j)).get();
+      TileElementSize tile_size = tile.size();
+      for (SizeType j = 0; j < tile_size.cols(); ++j) {
+        for (SizeType i = 0; i < tile_size.rows(); ++i) {
+          tile(TileElementIndex(i, j)) = val;
+        }
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
teonnik: Yes, I looked at it, I will keep this one for now. I would like to change how matrices are initialized eventually and move checks into a unit test.

miniapp/miniapp_tsgemm.cpp
```diff
@@ -0,0 +1,484 @@
...
+  MPI_Finalize();
+}
+
+namespace {
+
+// Send Cini tile to the process it belongs (`tile_rank`)
+void send_tile(ConstTileType const& c_tile, SizeType tile_rank, SizeType tag, MPI_Comm comm) {
+  SizeType num_elements = c_tile.ld() * c_tile.size().cols();
+  void const* buf = c_tile.ptr(TileElementIndex(0, 0));
+  MPI_Datatype dtype = dlaf::comm::mpi_datatype<ScalarType>::type;
```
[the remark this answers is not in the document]
teonnik: I was considering using it but I wasn't sure if there are any allocations being made internally. In my version, I am technically sending more data than I need to for tiles on the edge whose size is different than the size of the tiles specified when constructing the matrix. I do that instead of allocating a contiguous buffer (staging area) that offloads only the useful part of the edge tile and sends that. Is dlaf::comm::Message going to do what I do or allocate a staging area?

miniapp/miniapp_tsgemm.cpp
```diff
@@ -0,0 +1,484 @@
...
+// Reduce into Cini tile from all other processes
+void recv_tile(TileType& c_tile, int this_rank, SizeType tag, MPI_Comm comm) {
+  int nprocs = Communicator(MPI_COMM_WORLD).size() - 1;
+  SizeType nelems = c_tile.ld() * c_tile.size().cols();
+  ScalarType* tile_buf = c_tile.ptr(TileElementIndex(0, 0));
+  MPI_Datatype dtype = dlaf::comm::mpi_datatype<ScalarType>::type;
+
+  // Note: these allocations should be better made from a pool
+  std::vector<MPI_Request> reqs(nprocs);
+  std::vector<ScalarType> staging_buf(nprocs * nelems);
```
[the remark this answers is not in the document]
teonnik: I looked at `dlaf::common::Buffer` and it is indeed a better fit for my us case since it utilizes `std::make_unique<T[]>()`. However, in the future, the allocations should ideally be made from a pool. `std::make_unique<T[]>(N)` uses `malloc()` which is often expensive.

miniapp/miniapp_tsgemm.cpp
```diff
@@ -328,28 +331,43 @@ void sirius_gemm(ExecutorType const& mpi_executor, CommunicatorGrid& comm_grid,
       int tile_tag = dlaf::common::computeLinearIndex(Ordering::ColumnMajor, c_idx,
                                                       {tile_size.rows(), tile_size.cols()});
 
+      // Orders tasks so that the number of c_tiles computed in parallel at any moment is equal to the
+      // number of cores in the `default` thread pool.
+      int c_dep_i = tile_i - dep_tile_i;
+      int c_dep_j = tile_j - dep_tile_j;
```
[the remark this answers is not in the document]
teonnik: > We may evaluate if it is reasonable to add a constraint to SizeType, e.g. it has to be a signed integral type.

I'd be happy with `SizeType` as long as it's signed. If we decide to make it unsigned however it will not be a good fit for the loop because the logic allows and depends on `c_dep_i` and `c_dep_j` to be negative.

miniapp/miniapp_tsgemm.cpp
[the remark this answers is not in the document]
albestro: Good point, when I designed `Buffer` I thought about it but I left the implementation for the future (when the Pool from #96 will be available).

Just for reference, you can give a look to the last parts of this messages, where we quickly discussed about it

https://github.com/eth-cscs/DLA-Future/pull/155#issuecomment-595769394
https://github.com/eth-cscs/DLA-Future/pull/155#discussion_r392228763

miniapp/miniapp_tsgemm.cpp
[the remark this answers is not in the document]
albestro: The basic ideas behind this are:
- The `Tile` manages the underlying allocated memory
- The `Data` concept allow the generalization of a memory layout: it does not allocate any additional memory, it just refers to the data and gives you access to the underlying layout properties in a "generic" way (in terms of number of blocks, number of elements per block, stride between blocks);
- The `Message` allows you to make a `Data` concept compatible with the MPI library, internally computing what type is the best to send that particular layout via MPI calls.

For reference, you can give a look at the recently merged PR #155.

About edge cases, where `Tile` may have a different size, I want to highlight this code snippet
https://github.com/eth-cscs/DLA-Future/blob/ec93197fa42d99bc3753eea6e0b9bd36b4828ded/include/dlaf/tile.h#L185-L190

This is the code that starting from a `Tile`, it creates a `Data` concept for it. As you can see it uses tile sizes to describe the layout of the memory. So, I would say that it should create a `Message` from it with just needed data.

I'm willing to get a feedback from you on this 😉

miniapp/miniapp_tsgemm.cpp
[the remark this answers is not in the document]
rasolca: SizeType is signed. It is meant to be the int type of BLAS and LAPACK.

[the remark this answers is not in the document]
albestro: So, I would go for adding in `dlaf/types.h` next to `SizeType` definition the following constraint

```cpp
static_assert(std::is_integral<SizeType>::value && std::is_signed<SizeType>::value, "it has to be a signed integral type.");
```

test/src/gtest_mpihpx_main.cpp
```diff
@@ -74,15 +74,9 @@ GTEST_API_ int main(int argc, char** argv) {
...
-  hpx::start(nullptr, argc, argv);
-  hpx::runtime* rt = hpx::get_runtime_ptr();
-  hpx::util::yield_while([rt]() { return rt->get_state() < hpx::state_running; });
-
-  // Run!
-  auto ret = hpx::threads::run_as_hpx_thread(test_main, argc, argv);
+  auto ret = hpx::init(test_main, argc, argv);
 
   // Tear-down
-  hpx::stop();
```
albestro, line 85: @rasolca do you agree with this change?

I remember when I developed it that you preferred the previous one, but I cannot recall the reason. Something may have changed and it may be ok now, just for checking.

miniapp/miniapp_tsgemm.cpp
```diff
@@ -10,14 +10,14 @@
 
 #include <dlaf/blas_tile.h>
 #include <dlaf/communication/datatypes.h>
+#include <dlaf/communication/message.h>
 #include <dlaf/matrix.h>
 
-#include <hpx/dataflow.hpp>
+#include <hpx/async/dataflow.hpp>
+#include <hpx/execution/executors/pool_executor.hpp>  // > HPX v.1.4.1
```
albestro: Are all the HPX executor changes compatible with 1.4.0 or just from 1.4.1+ (or even higher)?

In case, you may want to update also HPX requirements
https://github.com/eth-cscs/DLA-Future/blob/fc8429c9d2dea03fd1b9c1bf46329339b29e7ed0/CMakeLists.txt#L63

test/src/gtest_mpihpx_main.cpp
teonnik: Sorry, this will be handled in a separate PR, I just have a few branches and occasionally merge stuff into this one. This PR is dependent on a few others and will not be merged soon.

The change was suggested by @msimberg ,perhaps, he can comment?

msimberg: The only reason I suggest `hpx::init` in this case is that it's significantly simpler and does the same thing. The `yield_while` requirement was removed in 1.4.0 so that would already simplify things a bit more. You would want to use `start`, `run_as_hpx_thread`, `stop` if you need to spawn multiple functions and the start/stop and spawning of functions are separate.

include/dlaf/communication/init.h
```diff
@@ -0,0 +1,31 @@
...
+    if (thd_required != thd_provided) {
+      std::printf("Provided MPI threading model does not match the required one.\n");
+      MPI_Abort(MPI_COMM_WORLD, 1);
+    }
+  }
+
+  ~InitMPI() {
+    MPI_Finalize();
+  }
+};
```
albestro, line 28: I think it is a nice initiative that can be part of https://github.com/eth-cscs/DLA-Future/issues/100

include/dlaf/communication/init.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
```
albestro, line 1: Missing header with license.

include/dlaf/hpx_partitions.h
```diff
@@ -0,0 +1,38 @@
+
```
albestro, line 1: Missing license header.

include/dlaf/hpx_partitions.h
```diff
@@ -0,0 +1,38 @@
...
+#include <mpi.h>
+
+#include <hpx/include/resource_partitioner.hpp>
+#include <hpx/program_options/options_description.hpp>
+
+namespace dlaf {
+
+// Note: A thread pool must be declared before starting the runtime (hpx::init())
+void try_init_mpi_pool(hpx::program_options::options_description& desc, int argc, char** argv) {
```
albestro, line 14: Also this one may be part of the previously cited https://github.com/eth-cscs/DLA-Future/issues/100.

miniapp/tsgemm.cpp
```diff
@@ -155,21 +158,18 @@ int hpx_main(::variables_map& vm) {
...
   hpx::mpi::enable_user_polling enable_polling(pool_name);
   ExecutorType mpi_executor(comm_grid.fullCommunicator());
 #else
-  using hpx::threads::thread_priority;
-  std::string pool_name = (ps.setup == "mpi_pool") ? "mpi" : "default";
-  auto priority = (ps.setup == "priorities") ? thread_priority::thread_priority_high
-                                             : thread_priority::thread_priority_default;
-  ExecutorType mpi_executor(pool_name, priority);
+  using hpx::threads::thread_priority::thread_priority_high;
+  auto pool_name = (hpx::resource::pool_exists("mpi")) ? "mpi" : "default";  // New API v1.5
```
albestro, line 171: I may have already commented about this, I think it is worth updating the dependency in the `CMakeLists.txt`.

PR #175, opened 2020-03-27
Add explicit extern template for mpi_datatypes

### LINKED ISSUE #173: dlaf::comm::mpi_datatype changes introduced a warning (note) on clang comipler
The changes of `mpi_datatype` introduced in #170 addressed the compatibility problem with OpenMPI (_"MPI_Datatype is not guaranteed to be constexpr"_) and contextually moved all the template specializations to another translation unit.

This triggers a warning on clang (`Apple LLVM version 10.0.1 (clang-1001.0.46.4)`) that suggests to explicit the declaration if it is explicitly instantiated in another translation unit (as it is now for us).

```bash
include/dlaf/communication/message.h:41:52: note: add an explicit instantiation declaration to suppress this warning if
      'dlaf::comm::mpi_datatype<int>::type' is explicitly instantiated in another translation unit
      classic_type_ = dlaf::comm::mpi_datatype<T>::type;
```

albestro:
Close #173 

This is the proposal I was talking about in the issue.

Moreover, in order to reduce the number of types, I removed the helper for const types, so now the `mpi_datatype` accepts just non-const T.


teonnik:
Removing the `const` version of `mpi_datatype` puts the burden on the user of the type to remove it. Why was it necessary to remove it?

albestro:
Adding the const version means doubling the number of types (as it was before). I agree with you that it was a nice helper.

To add it back, we won't need anymore the const inherited from the non-const (that was a shortcut to automatically define the const to non-const template conversion), and we can just add the template for const types.

include/dlaf/communication/datatypes.h
```diff
@@ -10,22 +10,25 @@
...
+extern template MPI_Datatype mpi_datatype<unsigned char>::type;
+extern template MPI_Datatype mpi_datatype<unsigned short>::type;
+extern template MPI_Datatype mpi_datatype<unsigned int>::type;
+extern template MPI_Datatype mpi_datatype<unsigned long>::type;
+extern template MPI_Datatype mpi_datatype<unsigned long long>::type;
+extern template MPI_Datatype mpi_datatype<float>::type;
+extern template MPI_Datatype mpi_datatype<double>::type;
+extern template MPI_Datatype mpi_datatype<bool>::type;
+extern template MPI_Datatype mpi_datatype<std::complex<float>>::type;
+extern template MPI_Datatype mpi_datatype<std::complex<double>>::type;
```
teonnik: With clang 9.0.1 I get an error with this approach: 
```
/home/teonnik/code/dlaf/src/communication/datatypes.cpp:49:50: error: explicit
specialization of 'type' after instantiation
```
Instead, we can forward declare the specializations in the header and remove the extern stuff.

```
template<> MPI_Datatype mpi_datatype<char>::type;
template<> MPI_Datatype mpi_datatype<short>::type;
template<> MPI_Datatype mpi_datatype<int>::type;
template<> MPI_Datatype mpi_datatype<long>::type;
template<> MPI_Datatype mpi_datatype<long long>::type;
template<> MPI_Datatype mpi_datatype<unsigned char>::type;
template<> MPI_Datatype mpi_datatype<unsigned short>::type;
template<> MPI_Datatype mpi_datatype<unsigned int>::type;
template<> MPI_Datatype mpi_datatype<unsigned long>::type;
template<> MPI_Datatype mpi_datatype<unsigned long long>::type;
template<> MPI_Datatype mpi_datatype<float>::type;
template<> MPI_Datatype mpi_datatype<double>::type;
template<> MPI_Datatype mpi_datatype<bool>::type;
template<> MPI_Datatype mpi_datatype<std::complex<float>>::type;
template<> MPI_Datatype mpi_datatype<std::complex<double>>::type;
```

include/dlaf/communication/datatypes.h
```diff
@@ -10,22 +10,25 @@
...
-/// @brief mapper between language types and basic MPI_Datatype
-template <typename T>
-struct mpi_datatype {
-  static MPI_Datatype type;
-};
-
-/// helper for mapping also custom types
-template <typename T>
-struct mpi_datatype<const T> : public mpi_datatype<T> {};
-
```
teonnik: I think this shouldn't be removed. With the approach outlined below, this is still useful for removing `const`.

include/dlaf/communication/datatypes.h
```diff
@@ -10,22 +10,25 @@
 
 #pragma once
 
-#include "dlaf/mpi_header.h"
-
-#include <complex>
+#include "dlaf/communication/internal/datatypes.h"
```
teonnik: Do we really need another header? Let's just keep things in `dlaf/communication/datatypes.h`

teonnik:
Yet, another approach is to specialize on static class functions instead of the current static static class variable. That will change the calling convention as a user will have to call a functions instead of reference a type `()`. On the other hand, we can move the definitions back into the header without the multiple definitions issues that accompany static class variables defined in headers.

https://stackoverflow.com/questions/18860895/how-to-initialize-static-members-in-the-header

include/dlaf/communication/datatypes.h
```diff
@@ -10,25 +10,67 @@
...
+MPI_Datatype mpi_datatype<double>::type = MPI_DOUBLE;
+
+template <>
+MPI_Datatype mpi_datatype<bool>::type = MPI_CXX_BOOL;
+
+template <>
+MPI_Datatype mpi_datatype<std::complex<float>>::type = MPI_CXX_FLOAT_COMPLEX;
+
+template <>
+MPI_Datatype mpi_datatype<std::complex<double>>::type = MPI_CXX_DOUBLE_COMPLEX;
```
teonnik: This is the same as in the original PR : https://github.com/eth-cscs/DLA-Future/pull/170 . There are issues with it related to multiple definitions error when static class variables are defined in headers and included in multiple translation units that are linked

albestro: Right! I forgot it! Thank you ;)

PR #177, opened 2020-03-30
Adapt MPI_Reduce wrapper to the new API and enable non-contiguous data

### LINKED ISSUE #148: MPI_Reduce incompatible with Tile
`Tile` can be stored in contiguous or non-contiguous memory.

Classic `MPI_Op` functions can be used for reducing contiguous buffers, so they can be used in the former case. However, in the latter case, this is not possible anymore; in fact, non-contiguous memory is converted to an MPI_Vector by `Message`, which is not compatible with MPI_Op.

albestro:
Close #148 

This PR adapts the MPI_Reduce wrapper to the new API accepting the Data concept instead of `dlaf::comm::Message`.

Moreover, it now trivially manages reduce with non-contiguous buffers by internally using a contiguous buffer for input and/or output in case of need.

### DIFF
--- a/include/dlaf/communication/sync/reduce.h
+++ b/include/dlaf/communication/sync/reduce.h
@@ -13,6 +13,7 @@
 /// @file
 
 #include <mpi.h>
+#include "dlaf/common/data_descriptor.h"
 #include "dlaf/communication/communicator.h"
 #include "dlaf/communication/message.h"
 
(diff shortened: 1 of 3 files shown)


albestro:
As already mentioned in PR #155, at the moment it is not possible to specify a custom memory area to use as temporary buffer (e.g. preallocated workspace tiles), but it should be quite easy to add a parameter to be forwarded to the `create_temporary_buffer` function (that in turns has to take it into account).

include/dlaf/communication/sync/reduce.h
```diff
@@ -13,24 +13,90 @@
...
 
-/// MPI_Reduce wrapper
+namespace internal {
+namespace reduce {
+
+/// @brief MPI_Reduce wrapper (sender side)
+/// MPI Reduce(see MPI documentation for additional info)
+/// @param reduce_operation MPI_Op to perform on @p input data coming from ranks in @p communicator
+template <class DataIn, class DataOut>
+void collector(Communicator& communicator, MPI_Op reduce_operation, DataIn input, DataOut output) {
```
rasolca, line 31: Consider a fast return if the communicator size is one.

albestro: Good point!

I'm wondering:
- what it could be better to do in that case? Is it legit to reduce on a single rank? Or do we want to consider it an error?
- is it worth to add a test?

rasolca: - Yes it is legit.
- yes, since it is a special case with its implementation.

albestro: Changes highlights:
- Test just with const input to reduce (it was not strictly redundant, but it may be an overkill)
- Implemented wayout for single rank communicator (just copy)
- Add debug check for validity of `rank_root` parameter
- Add const correctness for Data parameters (reference semantic, it is not const the value but the Data concept)
- Add `constexpr` to `dlaf_test::TypeUtilities`

Since you already reviewed this PR, I suggest you to give a look at the changes from your last review for all files with an exception for the file containing tests, for which it will be easier to look at all changes made from the beginning.

test/unit/communication/test_reduce.cpp
```diff
@@ -22,241 +24,173 @@ using namespace dlaf::comm;
...
   MPI_Op op = MPI_SUM;
-
   sync::reduce(root, communicator, op, std::move(message_input), std::move(message_output));
+
   if (world.rank() == root) {
     for (std::size_t i_block = 0; i_block < nblocks; ++i_block)
       for (std::size_t i_element = 0; i_element < block_size; ++i_element) {
         auto mem_pos = i_block * block_distance + i_element;
-        EXPECT_EQ(value * NUM_MPI_RANKS, data_strided[mem_pos]);
+        EXPECT_EQ(value * static_cast<TypeParam>(NUM_MPI_RANKS), data_strided[mem_pos]);
```
rasolca, line 193: Are you casting an `int` into an `int`?
If you want to allow to change easily TypeParam to another type, you should consider that a floating point type can be used.

albestro: Sorry, I'm not getting your point.

`using TypeParam = std::complex<double>;`

What am I missing?

rasolca: That floating point results of two equivalent math operations may not be binary equivalent.

include/dlaf/communication/sync/reduce.h
```diff
@@ -13,24 +13,102 @@
...
+
+  common::Buffer<T> tmp_mem_input;
+  common::Buffer<T> tmp_mem_output;
+
+  common::DataDescriptor<const T> tmp_in = input;
+  common::DataDescriptor<T> tmp_out = output;
+
+  // if input is not contiguous, copy it in a contiguous temporary buffer
+  if (!input.is_contiguous()) {
+    tmp_in = tmp_mem_input = common::create_temporary_buffer(tmp_in);
```
teonnik, line 50: This is unusual, threw me off for a second, is this supposed to mean that:

```
tmp_mem_input = common::create_temporary_buffer(tmp_in);
tmp_in = tmp_mem_input
```

include/dlaf/communication/sync/reduce.h
```diff
@@ -13,24 +13,102 @@
...
+/// @brief MPI_Reduce wrapper
+/// MPI Reduce(see MPI documentation for additional info)
+/// @param rank_root the rank that will collect the result in output
+/// @param reduce_operation MPI_Op to perform on @p input data coming from ranks in @p communicator
+template <class DataIn, class DataOut>
+void reduce(const int rank_root, Communicator& communicator, MPI_Op reduce_operation, const DataIn input,
+            const DataOut output) {
+  DLAF_ASSERT_HEAVY((rank_root < communicator.size() && rank_root != MPI_UNDEFINED), rank_root,
+                    " is not a valid rank in the specified Communicator with size ",
+                    communicator.size());
```
teonnik, line 105: This is a relatively cheap check, I wouldn't mind having it as DLAF_ASSERT either.

test/unit/communication/test_reduce.cpp
```diff
@@ -8,55 +8,207 @@
...
   static_assert(NUM_MPI_RANKS >= 2, "This test requires at least two ranks");
   int root = 1;
 
-  const std::size_t N = 3;
-  int input[N] = {1, 2, 3};
-  int reduced[N];
+  constexpr std::size_t N = 3;
+  constexpr TypeParam input[N] = {dlaf_test::TypeUtilities<TypeParam>::element(0, 1),
+                                  dlaf_test::TypeUtilities<TypeParam>::element(1, 2),
+                                  dlaf_test::TypeUtilities<TypeParam>::element(2, 3)};
```
teonnik, line 102: There are few types and function used throughout the test, I'd encourage a heavier use of `using` declarations:

```
using dlaf_test::TypeUtilities<TypeParam>::element;
```

include/dlaf/communication/sync/reduce.h
albestro: Yeah, I knew that it may had been "problematic".

I would a comment, split the assignment and use more self-explaining names to make it more clear.

You have to consider the two different types and functionalities:
- `tmp_in` is a `DataDescriptor`, it does not own any memory, it just describes a memory from a `Data` concept.
- `tmp_mem_in` is a `Buffer` that implements the `Data` concept, it owns the memory.

So, it is exactly as you said:
- create a temporary buffer `Buffer` and reference to it through `tmp_mem_input`
- make `tmp_in` pointing to the just created temporary buffer in `tmp_mem_input`

test/unit/communication/test_reduce.cpp
albestro: I'm not sure that I can do what you are proposing, since `element` is a static member for the template class `TypeUtilities`.

In case you confirm that, I can propose you:
- just remove the `dlaf_test` since there is the `using dlaf_test` directive at the beginning of the file;
- for each test delcare `using TypeUtilities = typename TypeUtilities<TypeParam>` in order to create a "shortcut" alias

Feel free to suggest your solution 😉

teonnik: Both seem fine to me, it's not a big issue in the grand scheme of things :) . You can also try this:

`using TypeUtils = dlaf_test::TypeUtilities<TypeParam>` and then do `TypeUtils::element()` 

if you prefer.

PR #178, opened 2020-04-06
Introduce error handling with new tools

### LINKED ISSUE #132: Error handling / Exception usage
I'm not sure that the current error handling with exceptions is ideal.
We should discuss about error handling, in particular when we should use
- exceptions
- error codes
- termination

albestro:
Close #132.

With #163 we introduced some basic tools to use for error handling.

I started with replacing all classic <cassert> with the new `DLAF_ASSERT_HEAVY` so we have a more flexible control on them and we can also add a custom message.

The next step is to find all places where exceptions are used for notifying an error and replace them with DLAF_ASSERT_LOW.


Sely85:
Below, I sum up the other types of exceptions I've found in the code and suggest how they can be replaced. Could you please have a look at it and let me know if you agree?

-  **std::invalid_argument** replaced by **DLAF_ASSERT()**
   Example (util_matrix.h): 
      ` if (!size_.isValid)`
           `std::invalid_argument("Error: Invalid Tile sizes");`
    replaced by
     ` DLAF_ASSERT((!size_.isValid()), "Invalid Tile size.");`

- **std::logic_error**: can this be replaced by **DLAF_ASSERT_MODERATE()**?
   Example (tile.h) 
    ` if (p_)` 
    `   throw std::logic_error("setPromise has been already used on this object!");` 
  replaced by
  `  DLAF_ASSERT_MODERATE((p_), "setPromise has been already used on this object!");` 

- **std::runtime_error**: replaced by **DLAF_ASSERT()**
  Example (lapack_tile.tpp)
   example in lapack_tile.tpp
   ` if (info != 0)` 
     `  throw std::runtime_error("Error: POTRF: A is not positive definite.");` 
   replaced by
  ` DLAF_ASSERT((info != 0), "POTRF: A is not positive definite.");`  [plus more info on A]

- **std::rethrow_exception** : how can we deal with this one? Does it depend only on the "level" of exception rethrown?
  Example (matrix_const.tpp)
            ` std::rethrow_exception(current_exception_ptr); ` 

Thank you!

rasolca:
>     * **std::invalid_argument**: replaced by **DLAF_ASSERT()**
>       Example (util_matrix.h):
>       ` if (!size_.isValid)`
>       `std::invalid_argument("Error: Invalid Tile sizes");`
>       replaced by
>       ` DLAF_ASSERT((!size_.isValid()), "Invalid Tile size.");`

It is hard to generalize, I would consider DLAF_ASSERT in most of the cases, but DLAF_ASSERT_MODERATE can be better in more performance critical cases.

>     * **std::logic_error**: can this be replaced by **DLAF_ASSERT_MODERATE()**?
>       Example (tile.h)
>       ` if (p_)`
>       `   throw std::logic_error("setPromise has been already used on this object!");`
>       replaced by
>       `  DLAF_ASSERT_MODERATE((p_), "setPromise has been already used on this object!");`

For now I would use DLAF_ASSERT here, since it is critical to avoid replacing the promise.

>     * **std::runtime_error**: replaced by **DLAF_ASSERT()**
>       Example (lapack_tile.tpp)
>       example in lapack_tile.tpp
>       ` if (info != 0)`
>       `  throw std::runtime_error("Error: POTRF: A is not positive definite.");`
>       replaced by
>       ` DLAF_ASSERT((info != 0), "POTRF: A is not positive definite.");`  [plus more info on A]

Ok.

>     * **std::rethrow_exception**: how can we deal with this one? Does it depend only on the "level" of exception rethrown?
>       Example (matrix_const.tpp)
>       `std::rethrow_exception(current_exception_ptr);`

This one should not be replaced. It is needed to deal with the exception handling of hpx and the possibility that users use exceptions in their code.

albestro:
> Looks good to me! Just one comment, why surround expressions with an extra `()` in ASSERT_* ?

Good point, also @rasolca pointed out the same.

The only reason is because I replaces old `assert`s all at once (using `sed`), and to stay on the safe side with "comma" macro problems (wrong argument detection), I went for explicitly surrounding the first macro argument, even when it was not strictly necessary.

teonnik:
Perhaps we would want to remove the extra `()` before merging then.

Sely85:
I have replaced various error types with the new tools (besides cases not yet implemented in Cholesky factorization and triangular solver). I have eventually always chosen the level `DLAF_ASSERT`, but please check if this did not compromise performance. 

Regarding extra `()`, we actually needed to add them also in the `MPI_CALL` routine for a correct argument detection (thanks @albestro!). Do we want to check and remove them when not necessary?

@albestro @rasolca @teonnik @sanciop

teonnik:
@Sely85 I changed `MPI_CALL` in PR #196 , it's called DLAF_MPI_CALL and I changed the internals. Do we need the extra `()` for correctness? I thought that they were there because a bulk change was made with the `sed` command.

Sely85:
@teonnik sorry, I forgot about that change in PR #196. The `()` were needed around `SOURCE_LOCATION()` when using `DLAF_ASSERT_WITH_ORIGIN`. But I'm not sure `cuda_call` requires them as well. Let's ask @albestro

include/dlaf/lapack_tile.h
```diff
@@ -31,8 +31,10 @@ namespace tile {
 /// Compute the cholesky decomposition of a.
 
 /// Only the upper or lower triangular elements are referenced according to @p uplo.
-/// @throw std::invalid_argument if a is not square.
-/// @throw std::runtime_error if the tile was not positive definite.
+///
+/// When the assertion is enabled, terminates the program with an error message
+/// if matrix is not square or when it is not positive definite.
+/// This assertion is enabled when **DLAF_ASSERT_ENABLE** is ON.
```
rasolca, line 37: Please fix doc format if you see problems:
```
/// Compute the cholesky decomposition of a.
///
/// Only the upper or lower triangular elements are referenced according to @p uplo.
/// @pre a is a square matrix
/// @pre a is positive definite
```

include/dlaf/lapack_tile.h
```diff
@@ -42,7 +44,9 @@ void potrf(blas::Uplo uplo, const Tile<T, device>& a);
 
 /// Only the upper or lower triangular elements are referenced according to @p uplo.
 /// @returns info = 0 on success or info > 0 if the tile is not positive definite.
-/// @throw std::runtime_error if the tile was not positive definite.
+/// When the assertion is enabled, terminates the program with an error message
+/// if matrix is not positive definite.
+/// This assertion is enabled when **DLAF_ASSERT_ENABLE** is ON.
```
rasolca, line 49: This was wrong before too:
The precondition should be: a is square.

include/dlaf/tile.tpp
```diff
@@ -1,5 +1,5 @@
 //
-// Distributed Linear Algebra with Future (DLAF)
+// Distributed Algebra with Future (DLAF)
```
albestro, line 2: ?!

src/communication/communicator_impl.cpp
```diff
@@ -3,6 +3,7 @@
 #include <cassert>
```
albestro, line 3: You don't need anymore `<cassert>`

include/dlaf/mpi_header.h
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+
+#include <iostream>
+
+#include <mpi.h>
+
+#define MPI_CALL(x)                                                                         \
```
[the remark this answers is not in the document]
teonnik: This header doesn't exist anymore, it was replaced by this one : include/dlaf/communication/error.h . There shouldn't be a need to modify it.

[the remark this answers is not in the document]
albestro: So I would say it is a rebase/merge error.

teonnik:
Looks good, some minor comments:

- test_blas_tile.h : there are some comments that use the word "throw", perhaps replace by "terminate" to avoid connotations with exceptions
- test_data_descriptor.cpp : in the default clause of `create_memory()` just print the message, call `std::abort()` and mark the method with `noexcept`
- `test_matrix` : there are a number of tests for `EXPECT_THROW`
- `test_matrix_view` : there is a test with `EXPECT_THROW`
- `lapack_tile.h` : stray comment: `// Variants that throw an error on failure.`; perhaps it should be removed
- `factorization/cholesky/mc.tpp ` :  `throw std::runtime_error("uplo = Upper not yet implemented")` replace with a message, then `std::abort` and mark noexcept
- `include/dlaf/solver/triangular/mc.tpp` : similar here
- `miniapp_cholesky.cpp` : there are a bunch of `throw` which should be converted to asserts

include/dlaf/matrix.tpp
```diff
@@ -34,12 +34,12 @@ Matrix<T, device>::Matrix(matrix::Distribution&& distribution)
...
+Matrix<T, device>::Matrix(matrix::Distribution&& distribution, const matrix::LayoutInfo& layout) noexcept
     : Matrix<const T, device>(std::move(distribution), {}, {}) {
-  if (this->distribution().localSize() != layout.size())
-    throw std::invalid_argument("Error: distribution.localSize() != layout.size()");
-  if (this->blockSize() != layout.blockSize())
-    throw std::invalid_argument("Error: distribution.blockSize() != layout.blockSize()");
+  DLAF_ASSERT(this->distribution().localSize() == layout.size(), "Size of distribution ",
+              distribution.localSize(), " does not match layout size ", layout.size());
+  DLAF_ASSERT(this->blockSize() == layout.blockSize(), "Block size of distribution ",
+              distribution.blockSize(), " does not match layout size ", layout.blockSize());
```
teonnik, line 42: How about this?

```
  DLAF_ASSERT(this->distribution().localSize() == layout.size(), "Size of distribution does not match layout size!", distribution.localSize(), layout.size());
```

include/dlaf/matrix_const.tpp
```diff
@@ -17,12 +17,12 @@ Matrix<const T, device>::Matrix(const matrix::LayoutInfo& layout, ElementType* p
...
+                                ElementType* ptr) noexcept
     : MatrixBase(std::move(distribution)) {
-  if (this->distribution().localSize() != layout.size())
-    throw std::invalid_argument("Error: distribution.localSize() != layout.size()");
-  if (this->blockSize() != layout.blockSize())
-    throw std::invalid_argument("Error: distribution.blockSize() != layout.blockSize()");
+  DLAF_ASSERT(this->distribution().localSize() == layout.size(), "Size of distribution ",
+              distribution.localSize(), " does not match layout size ", layout.size());
+  DLAF_ASSERT(this->blockSize() == layout.blockSize(), "Block size of distribution ",
+              distribution.blockSize(), " does not match layout size ", layout.blockSize());
```
teonnik, line 25: Similar to previous comment^^

include/dlaf/tile.tpp
```diff
@@ -10,19 +10,16 @@
...
-    throw std::invalid_argument("Error: Invalid Tile leading dimension");
-  }
-  if (!size.isEmpty()) {
-    if (sum(size_.rows(), mul(ld_, (size_.cols() - 1))) > memory_view_.size())
-      throw std::invalid_argument("Error: Tile exceeds the MemoryView limits");
-  }
+
+  DLAF_ASSERT(size.isValid(), "Invalid Tile sizes");
+  DLAF_ASSERT(ld_ >= std::max<SizeType>(1, size_.rows()), "Invalid Tile leading dimension (", ld_, " < ",
+              std::max<SizeType>(1, size_.rows()), ")");
```
teonnik, line 20: I would simplify this to:

```
DLAF_ASSERT(ld_ >= std::max<SizeType>(1, size_.rows()), "Invalid leading dimension!", ld, size_);
```

miniapp/miniapp_cholesky.cpp
```diff
@@ -191,15 +191,12 @@ options_t check_options(hpx::program_options::variables_map& vm) {
...
-  if (opts.grid_rows <= 0)
-    throw std::runtime_error("number of grid rows must be a positive number");
-  if (opts.grid_cols <= 0)
-    throw std::runtime_error("number of grid columns must be a positive number");
+  DLAF_ASSERT(opts.m > 0, "matrix size must be a positive number (", opts.m, ")");
+  DLAF_ASSERT(opts.mb > 0, "block size must be a positive number (", opts.mb, ")");
+  DLAF_ASSERT(opts.grid_rows > 0, "number of grid rows must be a positive number (", opts.grid_rows,
+              ")");
+  DLAF_ASSERT(opts.grid_cols > 0, "number of grid columns must be a positive number (", opts.grid_cols,
+              ")");
```
teonnik, line 199: I would make the message a sentence and remove the closing bracket : ")" .

src/matrix/layout_info.cpp
```diff
@@ -22,45 +22,31 @@ LayoutInfo::LayoutInfo(const LocalElementSize& size, const TileElementSize& bloc
...
-      throw std::invalid_argument("Error: Invalid Tile Row Offset");
-    }
-    if (tile_offset_col_ < 1) {
-      throw std::invalid_argument("Error: Invalid Tile Col Offset");
-    }
-  }
-  else {
+  DLAF_ASSERT((!size_.isEmpty() || ld_tile_ >= 1), "Invalid Leading Dimension");
+  DLAF_ASSERT((!size_.isEmpty() || tile_offset_row_ >= 1), "Invalid Tile Row Offset");
+  DLAF_ASSERT((!size_.isEmpty() || tile_offset_col_ >= 1), "Invalid Tile Col Offset");
```
teonnik, line 33: Lets check `DLAF_ASSERT(!size_.isEmpty(), "Size is empty!", size)` once before these 3 and remove the check in them.

Sely85: I think that in this way when the matrix is empty, it will always result in an error. What about something like this?
```
  if (!size_.isEmpty()) {
    DLAF_ASSERT(ld_tile_ >= 1, "Invalid Leading Dimension!", size, ld_tile_);
    DLAF_ASSERT(tile_offset_row_ >= 1, "Invalid Tile Row Offset!");
    DLAF_ASSERT(tile_offset_col_ >= 1, "Invalid Tile Col Offset!");
  }
```

teonnik: Yes, this LGTM!

include/dlaf/factorization/cholesky/mc.tpp
```diff
@@ -24,8 +24,10 @@ void Factorization<Backend::MC>::cholesky(blas::Uplo uplo, Matrix<T, Device::CPU
 
   if (uplo == blas::Uplo::Lower)
     internal::mc::cholesky_L(mat_a);
-  else
-    throw std::runtime_error("uplo = Upper not yet implemented");
+  else {
+    std::cout << "uplo = Upper not yet implemented" << std::endl;
+    std::abort();
+  }
```
albestro, line 30: I forgot to open an issue about this, I've just opened it #243

include/dlaf/lapack_tile.tpp
```diff
@@ -44,20 +44,18 @@ dlaf::BaseType<T> lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag,
...
 template <class T, Device device>
 long long potrfInfo(blas::Uplo uplo, const Tile<T, device>& a) {
-  if (a.size().rows() != a.size().cols()) {
-    throw std::invalid_argument("Error: POTRF: A is not square.");
-  }
+  DLAF_ASSERT((a.size().rows() == a.size().cols()), "POTRF: `a` is not square!", a);
 
   auto info = lapack::potrf(uplo, a.size().rows(), a.ptr(), a.ld());
-  assert(info >= 0);
+  DLAF_ASSERT_HEAVY((info >= 0), "");
```
albestro, line 58: are these parentheses needed?

include/dlaf/matrix.h
```diff
@@ -181,7 +186,7 @@ class Matrix<const T, device> : public matrix::internal::MatrixBase {
 ///
 /// @param[in] ld the leading dimension of the matrix.
 /// @param[in] ptr is the pointer to the first element of the local part of the matrix.
-/// @throw std::invalid_argument if @p ld < max(1, size.row()).
+/// @pre ld > max(1, size.row())
```
albestro, line 189: `>=`?

include/dlaf/tile.h
```diff
@@ -54,10 +54,13 @@ class Tile<const T, device> {
   using ElementType = T;
 
   /// @brief Constructs a (@p size.rows() x @p size.cols()) Tile.
-  /// @throw std::invalid_argument if @p size.row() < 0, @p size.cols() < 0 or @p ld < max(1, @p size.rows()).
-  /// @throw std::invalid_argument if memory_view does not contain enough elements.
+  /// @pre size.rows() >= 0
+  /// @pre size.cols() >= 0
+  /// @pre ld > max(1, @p size.rows())
```
albestro, line 59: `>=`?

include/dlaf/tile.h
```diff
@@ -131,10 +134,13 @@ class Tile : public Tile<const T, device> {
...
```
albestro, line 139: `>=`?

src/matrix/distribution.cpp
```diff
@@ -19,10 +19,8 @@ Distribution::Distribution() noexcept
 Distribution::Distribution(const LocalElementSize& size, const TileElementSize& block_size)
     : size_(0, 0), local_size_(size), global_nr_tiles_(0, 0), local_nr_tiles_(0, 0),
       block_size_(block_size), rank_index_(0, 0), grid_size_(1, 1), source_rank_index_(0, 0) {
-  if (!local_size_.isValid())
-    throw std::invalid_argument("Error: Invalid Matrix size");
-  if (!block_size_.isValid() || block_size_.isEmpty())
-    throw std::invalid_argument("Error: Invalid Block size");
+  DLAF_ASSERT((local_size_.isValid()), "Invalid Matrix size!");
+  DLAF_ASSERT((block_size_.isValid() && !block_size_.isEmpty()), "Invalid Block size!");
```
albestro, line 23: I'm not really sure that these parentheses are mandatory

src/matrix/distribution.cpp
```diff
@@ -35,16 +33,12 @@ Distribution::Distribution(const GlobalElementSize& size, const TileElementSize&
...
-  if (!rank_index_.isValid() || !rank_index.isIn(grid_size_))
-    throw std::invalid_argument("Error: Invalid Rank Index");
-  if (!source_rank_index_.isValid() || !source_rank_index.isIn(grid_size_))
-    throw std::invalid_argument("Error: Invalid Matrix Source Rank Index");
+  DLAF_ASSERT((size_.isValid()), "Invalid Matrix size!");
+  DLAF_ASSERT((block_size_.isValid() && !block_size_.isEmpty()), "Invalid Block size!");
+  DLAF_ASSERT((grid_size_.isValid() && !grid_size_.isEmpty()), "Invalid Communicator size!");
+  DLAF_ASSERT((rank_index_.isValid() && rank_index.isIn(grid_size_)), "Invalid Rank Index!");
+  DLAF_ASSERT((source_rank_index_.isValid() && source_rank_index.isIn(grid_size_)),
+              "Invalid Matrix Source Rank Index!");
```
albestro, line 41: Also here, about redundant parentheses.

src/matrix/layout_info.cpp
```diff
@@ -22,45 +22,36 @@ LayoutInfo::LayoutInfo(const LocalElementSize& size, const TileElementSize& bloc
...
-    }
-    if (tile_offset_row_ < 1) {
-      throw std::invalid_argument("Error: Invalid Tile Row Offset");
-    }
-    if (tile_offset_col_ < 1) {
-      throw std::invalid_argument("Error: Invalid Tile Col Offset");
-    }
+  //  DLAF_ASSERT((!size_.isEmpty() || ld_tile_ >= 1), "Invalid Leading Dimension");
+  //  DLAF_ASSERT((!size_.isEmpty() || tile_offset_row_ >= 1), "Invalid Tile Row Offset");
+  //  DLAF_ASSERT((!size_.isEmpty() || tile_offset_col_ >= 1), "Invalid Tile Col Offset");
```
albestro, line 33: I would remove these

src/matrix/layout_info.cpp
```diff
@@ -22,45 +22,36 @@ LayoutInfo::LayoutInfo(const LocalElementSize& size, const TileElementSize& bloc
...
+  if (!size_.isEmpty()) {
```
albestro, line 34: I think it is the opposite check...isn't it?

include/dlaf/lapack_tile.tpp
Sely85: nope, right

src/matrix/layout_info.cpp
Sely85: true 😅

Sely85:
I think this PR is ready for another round of review. 
Now I have:
- fixed errors in preconditions,
- removed all redundant brakets (hopefully),
- made the doxygen documentation more uniform. 
By the way, I think it's worth waiting for the PR about NOT_IMPLEMENTED_YET and integrate it in this one. 

Let me know what you think, please @albestro @rasolca @teonnik.

albestro:
> Please note that the `isIn` method already include the `isValid` check. Moreover it would be good in case of failure of asserts with `isIn` to include the value of the index and the size.
> 
> On the other hand isEmpty() has isValid() as precondition and therefore I don't like to separate them as in this example:
> 
> ```
> /// @pre size.isValid()
> /// @pre !size.isEmpty()
> ```
> 
> I see 3 possibilities:
> 
>     1. Keep only `/// @pre !size.isEmpty()` (isValid() in implicit in it)
> 
>     2. combine them in `/// @pre size.isValid() && !size.isEmpty()` in this case isEmpty is evaluated only if `isValid()` is true.
> 
>     3. remove the `isValid()` precondition from `isEmpty()` in the following way:
>        
>        * size.isEmpty() returns size.isValid() && (size.row == 0 || size.col == 0)
>        * size.isNotEmpty() returns size.row > 0 && size.col > 0
> 
> 
> @albestro @teonnik what do you think?

If we agree that `isEmpty()` is a legitimate operation just on a size that `isValid()`, I would say:
- option 1 is my preferred one;
- option 2 looks "verbose", we already agreed that `isEmpty` is legitimate just on valid sizes;
- option 3, it may become dangerous to have `!isEmpty()` which is something different than `isNotEmpty().

include/dlaf/matrix/layout_info.h
```diff
@@ -35,10 +35,9 @@ class LayoutInfo {
   /// @pre tile_ld >= 1,
   /// @pre tile_offset_row >= 1,
   /// @pre tile_offset_col >= 1,
-  /// @pre !size.isEmpty(),
-  /// @pre tile_ld >= max(size.rows(), block_size.rows()),
-  /// @pre tile_row_offset >= block_size.rows(),
-  /// @pre tile_col_offset >= size of the memory (in elements, padding included)
+  /// @pre size.isEmpty() and tile_ld >= max(size.rows(), block_size.rows()),
+  /// @pre size.isEmpty() and tile_row_offset >= block_size.rows(),
+  /// @pre size.isEmpty() and tile_col_offset >= size of the memory (in elements, padding included)
```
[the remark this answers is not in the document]
Sely85: can you check if it is fine now, please?

Sely85:
I think this PR is ready for another round of review. Could you please check? @rasolca @albestro @teonnik

include/dlaf/lapack_tile.h
```diff
@@ -27,43 +27,42 @@ namespace tile {
...
-void potrf(blas::Uplo uplo, const Tile<T, device>& a);
+void potrf(blas::Uplo uplo, const Tile<T, device>& a) noexcept;
 
-/// Compute the cholesky decomposition of a. (with return code)
+/// Compute the cholesky decomposition of a (with return code).
 ///
 /// Only the upper or lower triangular elements are referenced according to @p uplo.
 /// @returns info = 0 on success or info > 0 if the tile is not positive definite.
-/// @throw std::runtime_error if the tile was not positive definite.
+/// @pre matrix @p a is positive definite.
```
rasolca, line 65: The version without return code (line 59) should have this precondition, not this one.

include/dlaf/matrix/layout_info.h
```diff
@@ -32,13 +32,17 @@ class LayoutInfo {
...
   /// See misc/matrix_distribution.md for for more detail about the parameters.
   ///
-  /// @throw std::invalid_argument if @p tile_ld < 1 or @p tile_offset_row < 1 or @p tile_offset_col < 1,
-  /// @throw std::invalid_argument if @p !size.isEmpty() and
-  ///        @p tile_ld < @c max(@c size.rows(), @c block_size.rows())
-  /// @throw std::invalid_argument if @p !size.isEmpty() and @p tile_row_offset < @c block_size.rows()
-  /// @throw std::invalid_argument if @p !size.isEmpty() and @p tile_col_offset < size of the memory
-  ///        (in elements, padding included) to store a column of tiles,
-  /// @throw std::invalid_argument if the tiles overlap (combinations of @p tile_ld, @p tile_row_offset).
+  /// If !size.isEmpty():
```
rasolca, line 35:
```suggestion
  /// If size.isEmpty():
```

include/dlaf/matrix/layout_info.h
```diff
@@ -32,13 +32,17 @@ class LayoutInfo {
...
+  /// @pre tile_ld >= 1,
+  /// @pre tile_offset_row >= 1,
+  /// @pre tile_offset_col >= 1;
+  /// while if size.isEmpty():
```
rasolca, line 39:
```suggestion
  /// If !size.isEmpty():
```

include/dlaf/matrix/layout_info.h
```diff
@@ -32,13 +32,17 @@ class LayoutInfo {
...
+  /// If size.isEmpty():
+  /// @pre tile_ld >= 1,
+  /// @pre tile_offset_row >= 1,
+  /// @pre tile_offset_col >= 1;
+  /// If !size.isEmpty():
+  /// @pre tile_ld >= max(size.rows(), block_size.rows()),
+  /// @pre tile_row_offset >= block_size.rows(),
+  /// @pre tile_col_offset >= size of the memory (in elements, padding included)
+  /// to store a column of tiles, if the tiles overlap
+  ///(combinations of @p tile_ld, @p tile_row_offset).
```
rasolca, line 44:
```suggestion
  /// @pre tile_col_offset >= size of the memory (in elements, padding included)
  ///      to store a column of tiles
  /// @pre the tiles should not overlap (combinations of @p tile_ld, @p tile_row_offset).
```

rasolca:
bors try

rasolca:
bors try

PR #179, opened 2020-04-07
Implement a version header for DLAF

teonnik:
- The version header is generated and placed in the binary directory
- The header provides information within the code about the version, git sha
  and build timestamp.
- The information can be used in assertions or to specialize parts of
  the code for different versions of DLAF (with macros)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,7 +18,7 @@ if(DEFINED GIT_EXE AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
                   ERROR_QUIET
                   OUTPUT_STRIP_TRAILING_WHITESPACE)
 else()
-  set(SIRIUS_SHA "N/A : archive")
+  set(DLAF_SHA "N/A : archive")
 endif()
 configure_file("${PROJECT_SOURCE_DIR}/include/dlaf/version.h.in"
                "${PROJECT_BINARY_DIR}/include/dlaf/version.h"
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -8,6 +8,22 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
+# Generate version header
+find_program(GIT_EXE NAMES git)
+string(TIMESTAMP DLAF_TIMESTAMP "%Y-%m-%d %H:%M:%S")
+if(DEFINED GIT_EXE AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
+  execute_process(COMMAND git rev-parse HEAD
+                  OUTPUT_VARIABLE DLAF_SHA
+                  WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
+                  ERROR_QUIET
+                  OUTPUT_STRIP_TRAILING_WHITESPACE)
+else()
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -8,6 +8,22 @@
...
+find_program(GIT_EXE NAMES git)
+string(TIMESTAMP DLAF_TIMESTAMP "%Y-%m-%d %H:%M:%S")
+if(DEFINED GIT_EXE AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
+  execute_process(COMMAND git rev-parse HEAD
+                  OUTPUT_VARIABLE DLAF_SHA
+                  WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
+                  ERROR_QUIET
+                  OUTPUT_STRIP_TRAILING_WHITESPACE)
+else()
+  set(SIRIUS_SHA "N/A : archive")
```
albestro, line 21:
```suggestion
  set(DLAF_SHA "N/A : archive")
```

teonnik: Good catch, this PR is based on work I did for header generation in SIRIUS.

src/CMakeLists.txt
```diff
@@ -8,6 +8,22 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
+# Generate version header
+find_program(GIT_EXE NAMES git)
+string(TIMESTAMP DLAF_TIMESTAMP "%Y-%m-%d %H:%M:%S")
+if(DEFINED GIT_EXE AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
+  execute_process(COMMAND git rev-parse HEAD
```
rasolca, line 15: `git rev-parse HEAD` doesn't consider local changes.
Is this behaviour wanted?
What about adding at least something if local changes are detected (e.g. `<sha> *`)?
(Note: I'm not sure which are the uses of git_hash. I'm supposing the case we need to reproduce a bug discovered in an installed instance of the library)

teonnik: > I'm not sure which are the uses of git_hash. I'm supposing the case we need to reproduce a bug discovered in an installed instance of the library

Yes, that's the main reason. I included it because SIRIUS has it and developers there have found it useful. It's handy when you build several versions of the code to get a clear picture of what state the code was at.

> doesn't consider local changes. Is this behaviour wanted?

That could be useful. We can use this `git status --porcelain`, save the result in a variable and check if the variable is not empty and if so, modify DLAF_SHA to include ` *`.

rasolca: `git status --porcelain` also includes untracked files.
What about `git diff --quiet` which returns 0 if no difference are found in tracked files and 1 otherwise?

teonnik: This answer on stack overflow touches on various approaches: https://stackoverflow.com/questions/5143795/how-can-i-check-in-a-bash-script-if-my-local-git-repository-has-changes

I can also add `--untracked-files=no` to `git status --porcelain`. I'm not sure what the trade-offs are.

teonnik:
When `clang-format` is run on `version.h.in` it messes up CMake variables by moving the `@` identified to the right and introducing a space.

PR #181, opened 2020-04-08
Refactor CUBLAS into CMake Find module

teonnik:
- Link `DLAF` to `dlaf::cublas` when the target exists


cmake/FindCUBLAS.cmake
```diff
@@ -0,0 +1,32 @@
...
+# To find CUDA:
+#   1. Use `enable_language(CUDA)`
+#   2. Set `CMAKE_CUDA_COMPILER` if cuda in custom directory
+#
+# Imported Targets: 
+#   dlaf::cublas
+#
+cmake_minimum_required(VERSION 3.12)
+
+find_path(CUDA_TOOLKIT_INCLUDE cuda.h
```
rasolca, line 11: Shouldn't be checked if `cuda_runtime.h` and `cublas_v2.h` exists in the include directory?

CMakeLists.txt
```diff
@@ -37,8 +37,7 @@ set(CMAKE_CXX_STANDARD 14)
 set(DLAF_WITH_CUDA OFF CACHE BOOL "Enable CUDA support")
 if (DLAF_WITH_CUDA)
   enable_language(CUDA)
-
-  DLAF_addCudaTargets()
+  find_package(CUBLAS REQUIRED)
```
rasolca, line 40: Since cuda runtime and cublas are orthogonal libraries (there's no dependencies among them) I think that
it is misleading that `find_package(CUBLAS REQUIRED)` finds both.
I would change the name to something like `CUDALIBS`, such that other CUDA libraries can be added as well in case of need.

cmake/FindCUBLAS.cmake
teonnik: To my knowledge, they are always in the same directory as `cuda.h`.

CMakeLists.txt
teonnik: The Find package finds `cublas` and it's dependencies, in that case `cudart`. Since we only need `cublas` for now, I don't see the point in exposing a second target or creating another Find package.

cmake/FindCUBLAS.cmake
rasolca: I agree but I was thinking that it is a check if they are available.
Anyway if the cmake practice is to just check one of the headers I'm fine with it.

CMakeLists.txt
rasolca: In my knowledge it is not a dependency (cublas only depends on the driver). We need cuda runtime to manage the devices, streams, events, ... . Moreover we use it for allocations and copies (cublas has its own methods but the methods in cuda runtime are more flexible).
I agree to avoid to have a find module fora each cuda library we use, this is the reason I proposed to rename it to CUDALIBS which currently will include cuda runtime and cublas, but can be extended to other cuda libraries as needed.

cmake/FindCUBLAS.cmake
```diff
@@ -0,0 +1,32 @@
...
+                                                     CUDA_CUBLAS_LIB
+                                                     CUDA_CUDART_LIB)
+
+if (CUBLAS_FOUND AND NOT TARGET dlaf::cublas)
+    add_library(dlaf::cublas IMPORTED INTERFACE)
+    set_target_properties(dlaf::cublas PROPERTIES
+      INTERFACE_INCLUDE_DIRECTORIES "${CUDA_TOOLKIT_INCLUDE}"
+      INTERFACE_LINK_LIBRARIES "${CUDA_CUDART_LIB};${CUDA_CUBLAS_LIB}")
+endif()
+
```
rasolca, line 32: Please note that extra new lines at the end of the file will not pass the check of the format.
(In this case your file ends with "\nendif()\n\nEOF" while it is required to end with "\nendif()\nEOF")

PR #182, opened 2020-04-08
Pass DLAF_WITH_CUDA to source code from CMake

teonnik:
Depends on  #181


teonnik:
Since #181 was merged, this is also ready to be merged.

PR #183, opened 2020-04-09
Make computeLinearIndex accepting a Size2D instead of generic std::array

albestro:
Fix `computeLinearIndex` using as parameter `Size2D` instead of a generic `std::array` without any tag.

This change imposes to use it with matching Index2D and Size2D (both the index type and the tag must match).

As @rasolca recalled correctly, it is possible to convert a `std::array` to a `Size2D` thanks to the specific constructor. The implicit conversion is not possible; you have to specify `Size2D` template parameters that you want to use, i.e. the type of the index and the tag that allows the match with the associated Index2D type.

e.g.
```cpp
struct TAG{};
Index2D<int, TAG> index; 
computeLinearIndex(Ordering::RowMajor, index, Size2D<int, TAG>({1, 2}));
```


albestro:
@rasolca 

Do we want to allow to use `Index2D` and `Size2D` with the same `TAG` but different `IndexType`?

rasolca:
> @rasolca
> 
> Do we want to allow to use `Index2D` and `Size2D` with the same `TAG` but different `IndexType`?

I would not allow a different `IndexType`.

albestro:
I think it is worth providing further details on my first comment about implicit conversions.

The implicit conversion constructor from `std::array` is available for `basic_coords`, and in turn for `Size2D` that inherits all constructors from it. It means you can do something like
```cpp
Size2D<int, struct TAG> size(std::array<int>{26, 5})
```

The `computeLinearIndex` method does not allow to use the implicit conversion, because of the template that just accepts a `Size2D`, so you have to specify it like I said in the first comment
```cpp
computeLinearIndex(..., Size2D<int, struct TAG>(std::array<int, 2>{26, 5}));
```

However, you can still use an initializer-list for `size` parameter of `computeLinearIndex`, because in that case the Size2D 2-parameters constructor will be used!
```cpp
computeLinearIndex(..., {26, 5});
```

This implies also that, even if you can't use directly the implicit conversion, you can call the conversion constructor from `std::array`, surrounding it with curly braces like this:
```cpp
computeLinearIndex(..., {std::array<int, 2>{26, 5}});
```

In addition, I'd like to point out that `Index2D` has a slightly different implementation w.r.t `Size2D`; it does *NOT* inherit constructors from `basic_coords` and currently it does not provide a conversion constructor from `std::array`.

include/dlaf/common/index2d.tpp
```diff
@@ -60,15 +60,13 @@ Index2DType computeCoords(Ordering ordering, LinearIndexT index,
...
-    const std::array<typename Index2DType::IndexType, 2>& dims) {
+template <class IndexType, class IndexTag>
+IndexType computeLinearIndex(Ordering ordering, const Index2D<IndexType, IndexTag>& index,
+                             const Size2D<IndexType, IndexTag>& dims) {
   if (!index.isIn(dims))
     return -1;
 
-  std::size_t ld_size_index = (ordering == Ordering::RowMajor) ? 1 : 0;
-  auto leading_size = dims[ld_size_index];
+  auto leading_size = (ordering == Ordering::RowMajor) ? dims.cols() : dims.rows();
```
teonnik, line 69: Relatively minor comment because the compiler probably optimizes is ti anyway but why not use `dims.cols()` and `dims.rows()` directly in the switch statement? Or just write this:

```
return (ordering == Ordering::ColumnMajor) ? dims.rows() * index.col() + index.row()
                                           : dims.cols() * index.row() + index.col();
```

Also, the method shouldn't throw, perhaps mark it with noexcept. I would also replace the `if (!index.isIn(dims))` check with DLAF_ASSERT_MODERATE or HEAVY.

albestro: Good point! I used leading size because in the previous version was more clear like this, but now it is not necessary anymore!

About the `return -1` vs `DLAF_ASSERT_MODERATE`, it seems absolutely legit, but we should first agree on the API change.

@rasolca what do you think?

albestro:
I've just gave a look at another function that may need an API adaption to `Size2D`.

https://github.com/eth-cscs/DLA-Future/blob/101737b33729dd09be6a8e67f1ae85aca9a17d9d/include/dlaf/common/index2d.tpp#L41-L57

It is the "inverse" of the one we worked on in this PR. Do we want to adapt it the same way?

@rasolca @teonnik

include/dlaf/common/index2d.tpp
rasolca: > About the `return -1` vs `DLAF_ASSERT_MODERATE`, it seems absolutely legit, but we should first agree on the API change.

I don't mind, but you should check if the `-1` return was used somewhere.

albestro:
About adapting **`computeCoords`, I will address it in a separate PR** (I've been working on it #190).

About this PR, I would address just `computeLinearIndex`, in order to unlock other PRs. **I'd like to extend a bit the tests**, moreover, the current implementation has a limitation that I think we should solve, but it needs to be discussed more generally.

**What I'd like to discuss is the relationship between**:
- `IndexType` (`IT` from now on), the `Index2D` components type (and of the matching `Size2D`);
- `LinearIndexType` (`LIT` form now on), the type of the linear index related to a particular `Index2D`.

Let's say we use an `IT=int32_t`, this would allow us to create, potentially, grids with sizes up to `{2^31, 2^31}`, so the last element would be the `2^62 - 1`-th. In this case we would be able to index the last element just using a 64bit integer, being it `int64_t` or `uint64_t`, but for sure it would not be possible to reference it with `LIT=IT`.
It is true that having an `int32_t` does not mean that it is always the case, but in general it is worth taking it into account.

My first proposal would be to remove the constraint between `LIT` and `IT`. Changing it would required to explicitly specify the template type parameter `LIT` since it is a `return type`.

It may be true that it would not be the final solution, but I would try to avoid "complex" constraints.

Looking forward to your feedback 😉 @rasolca @teonnik

include/dlaf/common/index2d.h
```diff
@@ -168,15 +178,49 @@ class Index2D : public internal::basic_coords<IndexT> {
...
+    case Ordering::RowMajor:
+      return {static_cast<IndexType>(index / dims[1]), static_cast<IndexType>(index % dims[1])};
+    case Ordering::ColumnMajor:
+      return {static_cast<IndexType>(index % dims[0]), static_cast<IndexType>(index / dims[0])};
+    default:
+      return {};
+  }
 }
+
+/// Compute linear index of an Index2D
```
rasolca, line 200: preconditions are missing.

test/unit/common/test_index2d.cpp
```diff
@@ -137,17 +142,32 @@ TYPED_TEST(Index2DTest, Print) {
...
+  }};
 
   for (const auto& ordering : orderings) {
-    EXPECT_EQ(-1, computeLinearIndex(ordering, Index2D<TypeParam>{13, 26},
-                                     {13, 26}));  // out-of-rows & out-of-cols
-    EXPECT_EQ(-1, computeLinearIndex(ordering, Index2D<TypeParam>{13, 26}, {13, 39}));  // out-of-rows
-    EXPECT_EQ(-1, computeLinearIndex(ordering, Index2D<TypeParam>{13, 26}, {26, 26}));  // out-of-cols
+    for (const Size2D<TypeParam>& size : configs) {
+      EXPECT_DEATH(computeLinearIndex(ordering, index, {size.rows(), size.cols()}), "[ERROR]");
+      EXPECT_DEATH(computeLinearIndex(ordering, index, Size2D<TypeParam>(size)), "[ERROR]");
```
rasolca, line 158: Are those tests really needed?  Moreover this test shouldn't fail in release mode (ASSERT_MODERATE).

Note that death tests are quite expensive, therefore we should use them only in special cases.

include/dlaf/common/index2d.h
```diff
@@ -139,25 +143,31 @@ class Index2D : public internal::basic_coords<IndexT> {
...
+  Index2D() noexcept : internal::basic_coords<IndexT>(-1, -1) {}
 
   /// Create a valid 2D coordinate
   /// @param row index of the row
   /// @param col index of the column
   /// @throw std::invalid_argument if row < 0 or col < 0
-  Index2D(IndexT row, IndexT col);
+  Index2D(IndexT row, IndexT col) : internal::basic_coords<IndexT>(row, col) {
+    if (!internal::basic_coords<IndexT>::isValid())
+      throw std::invalid_argument("indices are not valid (negative).");
```
teonnik, line 154: Can we make this an ASSERT_* ?

include/dlaf/common/index2d.h
albestro: Good point. Since I just moved the code to the header, and this method is going to be changed in the PR #190 , do you mind if we move this task to that PR?

test/unit/common/test_index2d.cpp
albestro: Good point. IMHO they are useful because they tests the guards and related preconditions. I know that you think that since they are preconditions, they should not be tested. But I would like to test that the "service" of the guards works as expected.

In case, we would decide to maintain them, I have two proposal:
1. forcefully define the flags `DLAF_ASSERT_...` that we need at the beginning of the file;
2. enable this thests just when the flags are enabled on build (i.e. `#ifdef DLAF_ASSERT*`)

Personally, I would go for the second.

include/dlaf/common/index2d.h
albestro: I didn't change it because it is not the main topic of this PR, and the replacement with the new tools is going on in #178

include/dlaf/common/index2d.h
albestro: Good point. Since I just moved the code to the header, and this method is going to be changed in the PR #190 , do you mind if we move this task to that PR?

test/unit/common/test_index2d.cpp
albestro: Good point. IMHO they are useful because they tests the guards and related preconditions. I know that you think that since they are preconditions, they should not be tested. But I would like to test that the "service" of the guards works as expected.

In case, we would decide to maintain them, I have two proposal:
1. forcefully define the flags `DLAF_ASSERT*` that we need at the beginning of the file;
2. enable this the tests just when the flags are enabled on build (i.e. `#ifdef DLAF_ASSERT*`)

Personally, I would go for the second.

rasolca: Second options is better.
First option may create unexpected behaviour with functions compiled in the library

test/unit/common/test_index2d.cpp
```diff
@@ -137,17 +142,35 @@ TYPED_TEST(Index2DTest, Print) {
...
-  std::array<dlaf::common::Ordering, 2> orderings{dlaf::common::Ordering::RowMajor,
-                                                  dlaf::common::Ordering::ColumnMajor};
+  using dlaf::common::Ordering;
+
+  const Index2D<TypeParam> index(13, 26);
+  std::vector<Size2D<TypeParam>> configs{{
+      Size2D<TypeParam>{13, 26},  // out-of-rows & out-of-cols
+      Size2D<TypeParam>{13, 39},  // out-of-rows
+      Size2D<TypeParam>{26, 26},  // out-of-cols
+  }};
```
rasolca, line 152: `configs` is unused if moderate asserts are disabled.
move it into ifdef.

rasolca: Some holds for `index` above...

PR #185, opened 2020-04-14
Constrain SizeType and other fixes

albestro:
This PR addresses https://github.com/eth-cscs/DLA-Future/pull/174#discussion_r404841788

In addition to:
- `dlaf::common::vector` wrapper now accepts `SizeType`
- `dlaf::memory::MemoryView`
  - bug-fix: without the explicit constructor it was possible to allocate a const matrix (fixed also where this constructor was used)
  - change: remove helper constructors for `int` type
- overload `max` function for `SizeType` and exploits ADL in the code

### DIFF
--- a/test/unit/test_lapack_tile/test_potrf.h
+++ b/test/unit/test_lapack_tile/test_potrf.h
@@ -25,6 +25,8 @@ using namespace dlaf::matrix::test;
 using namespace dlaf_test;
 using namespace testing;
 
+using dlaf::util::size_t::mul;
+
 template <class T, bool return_info>
 void testPotrf(blas::Uplo uplo, SizeType n, SizeType extra_lda) {
   TileElementSize size_a = TileElementSize(n, n);
@@ -36,7 +38,7 @@ void testPotrf(blas::Uplo uplo, SizeType n, SizeType extra_lda) {
   s << ", n = " << n << ", lda = " << lda;
   SCOPED_TRACE(s.str());
 
-  memory::MemoryView<T, Device::CPU> mem_a(util::size_t::mul(lda, size_a.cols()));
+  memory::MemoryView<T, Device::CPU> mem_a(mul(lda, size_a.cols()));
 
   // Create tiles.
   Tile<T, Device::CPU> a(size_a, std::move(mem_a), lda);
@@ -97,7 +99,7 @@ void testPotrfArgExceptions(blas::Uplo uplo, TileElementSize size_a, SizeType ex
   s << ", size_a = " << size_a << ", lda = " << lda;
   SCOPED_TRACE(s.str());
 
-  memory::MemoryView<T, Device::CPU> mem_a(util::size_t::mul(lda, size_a.cols()));
+  memory::MemoryView<T, Device::CPU> mem_a(mul(lda, size_a.cols()));
 
   // Create tiles.
   Tile<T, Device::CPU> a(size_a, std::move(mem_a), lda);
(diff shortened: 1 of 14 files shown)


teonnik:
ADL is very dangerous and tricky. The following general advice from C++ Core Guidelines:

- T.47: Avoid highly visible unconstrained templates with common names
- C.168: Define overloaded operators in the namespace of their operands
- Always qualify calls to the std algorithms

Some resources:

- https://youtu.be/u1iewQYkYGE?t=1964
- https://stackoverflow.com/questions/2958648/what-are-the-pitfalls-of-adl
- https://quuxplusone.github.io/blog/2018/06/17/std-size/

albestro:
As highlighted by @teonnik, I reverted the ADL-based work and went for solving the problem using @rasolca proposal.

This seems to work with both `32bit` and `64bit` SizeType.

I think that it may be worth adding a test for both interfaces. Do we want to open an issue about that?

Moreover, see the `dlaf::util::size_t::mul` function used in the `MemoryView` constructor.

test/unit/test_lapack_tile/test_potrf.h
```diff
@@ -90,14 +90,14 @@ void testPotrf(blas::Uplo uplo, SizeType n, SizeType extra_lda) {
...
-  SizeType lda = std::max(1, size_a.rows()) + extra_lda;
+  SizeType lda = std::max<SizeType>(1, size_a.rows()) + extra_lda;
 
   std::stringstream s;
   s << "POTRF Arguments Exceptions: " << uplo;
   s << ", size_a = " << size_a << ", lda = " << lda;
   SCOPED_TRACE(s.str());
 
-  memory::MemoryView<T, Device::CPU> mem_a(lda * size_a.cols());
+  memory::MemoryView<T, Device::CPU> mem_a(util::size_t::mul(lda, size_a.cols()));
```
teonnik, line 100: There are 3 things here:
- If `SizeType` is to have `operator*()`, I will use that instead of `util::size_t::mul()`, it's easier to read
- the namespace `size_t` is rather confusing because there is a commonly used `std` type called `std::size_t`  already
- I'd encourage heavier use of `using` declarations to shorten these expressions and improve readability: e.g. `using util::size_t::mul;` or `using MemViewT = memory::MemoryView<T, Device::CPU>`, etc.

rasolca: size_t::mul is different from operator*():
`mul` takes the operands, convert them to std::size_t and multiply them.
(The result may not fit in SizeType and the result has to be converted to size_t anyway.)

Regarding the 3rd point, for size_t add and mul `using` is generally used.
I agree we can use it more, but we should still pay attention that using it too much can be worse. I worked once with a code that had tens of using statement and identifying a type was a nightmare.

albestro: Starting from the last, I'm going to address your suggestion.

About first two points, it is worth sharing with you a bit of context.
`MemoryView` constructor accepts as parameter the dimension of the memory area, and I think is correct to use `std::size_t`.

The function `dlaf::util::size_t::mul` is an helper function that allows us to (quite) safely compute the multiplication
https://github.com/eth-cscs/DLA-Future/blob/1922140d1bca8f9448f2556393d5961cc1d098ab/include/dlaf/util_math.h#L77-L84

As you can see, the name of the namespace is strictly related to the fact that it is a group of helper functions for `std::size_t` arithmetic.

teonnik: > I agree we can use it more, but we should still pay attention that using it too much can be worse. I worked once with a code that had tens of using statement and identifying a type was a nightmare.

As most all features, it shouldn't be taken to the extreme. I presume the code you mention utilized type aliases heavily, that indeed may cause trouble with identifying types. I would say that, if you are at the point that you have so many using statements and types in the same function, it is perhaps reasonable to consider refactoring into smaller functions.

> About first two points, it is worth sharing with you a bit of context.
MemoryView constructor accepts as parameter the dimension of the memory area, and I think is correct to use std::size_t.
> 
> The function dlaf::util::size_t::mul is an helper function that allows us to (quite) safely compute the multiplication

> mul takes the operands, convert them to std::size_t and multiply them.

Sure thing! You can ignore the first and second bullet points.

PR #186, opened 2020-04-15
Proposal: Copy matrices

### LINKED ISSUE #150: Add a method to copy Matrix values
It would be useful to have a function that allows copying values between two matrices with the same distribution.

The first case can be the `miniapp_cholesky`, in which it would be nice to have something similar to:
```cpp
Matrix A_ref;
set_random_hermitian_positive_definite(A_ref);

for (auto i = 0; i < nruns; ++i) {
  Matrix input;
  copy(A_ref, input);

  cholesky(input); // compute in-place the result (input contains L)

  if (do_check) {
    Matrix A;
    copy(A_ref, A);
    check_cholesky(A, input);
  }
}
```

albestro:
Close #150 

Starting point for a method to copy values between matrices:
- non-blocking
- tile-by-tile
- element-by-element

If we want to improve this, the first way could be to optimize the copy of each tile using memory chunks (e.g. exploiting `copy` function on Data).

### DIFF
--- a/include/dlaf/lapack_tile.tpp
+++ b/include/dlaf/lapack_tile.tpp
@@ -13,13 +13,6 @@
 namespace dlaf {
 namespace tile {
 
-template <class T, Device device>
-void potrf(blas::Uplo uplo, const Tile<T, device>& a) {
-  auto info = potrfInfo(uplo, a);
-  if (info != 0)
-    throw std::runtime_error("Error: POTRF: A is not positive definite.");
-}
-
 template <class T>
 void lacpy(const Tile<const T, Device::CPU>& a, const Tile<T, Device::CPU>& b) {
   DLAF_ASSERT_HEAVY(a.size() == b.size(),
@@ -30,6 +23,13 @@ void lacpy(const Tile<const T, Device::CPU>& a, const Tile<T, Device::CPU>& b) {
   SizeType n = a.size().cols();
 
   lapack::lacpy(lapack::MatrixType::General, m, n, a.ptr(), a.ld(), b.ptr(), b.ld());
+}
+
+template <class T, Device device>
+void potrf(blas::Uplo uplo, const Tile<T, device>& a) {
+  auto info = potrfInfo(uplo, a);
+  if (info != 0)
+    throw std::runtime_error("Error: POTRF: A is not positive definite.");
 }
 
 template <class T, Device device>
--- a/include/dlaf/lapack_tile.tpp
+++ b/include/dlaf/lapack_tile.tpp
@@ -8,6 +8,28 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 
+#include "dlaf/common/assert.h"
+
+namespace dlaf {
+namespace tile {
+
+template <class T, Device device>
+void potrf(blas::Uplo uplo, const Tile<T, device>& a) {
+  auto info = potrfInfo(uplo, a);
+  if (info != 0)
+    throw std::runtime_error("Error: POTRF: A is not positive definite.");
+}
+
+template <class T>
+void lacpy(const Tile<const T, Device::CPU>& a, const Tile<T, Device::CPU>& b) {
+  DLAF_ASSERT_HEAVY(a.size() == b.size(), "Source and destination tile must have the same size. A=", a.size(), " B=", b.size());
+
+  SizeType m = a.size().rows();
+  SizeType n = a.size().cols();
+
+  lapack::lacpy(lapack::MatrixType::General, m, n, a.ptr(), a.ld(), b.ptr(), b.ld());
+}
+
 template <class T, Device device>
 long long potrfInfo(blas::Uplo uplo, const Tile<T, device>& a) {
   if (a.size().rows() != a.size().cols()) {
@@ -20,9 +42,5 @@ long long potrfInfo(blas::Uplo uplo, const Tile<T, device>& a) {
   return info;
 }
 
-template <class T, Device device>
-void potrf(blas::Uplo uplo, const Tile<T, device>& a) {
-  auto info = potrfInfo(uplo, a);
-  if (info != 0)
-    throw std::runtime_error("Error: POTRF: A is not positive definite.");
+}
 }
--- a/include/dlaf/lapack_tile.tpp
+++ b/include/dlaf/lapack_tile.tpp
@@ -22,7 +22,9 @@ void potrf(blas::Uplo uplo, const Tile<T, device>& a) {
 
 template <class T>
 void lacpy(const Tile<const T, Device::CPU>& a, const Tile<T, Device::CPU>& b) {
-  DLAF_ASSERT_HEAVY(a.size() == b.size(), "Source and destination tile must have the same size. A=", a.size(), " B=", b.size());
+  DLAF_ASSERT_HEAVY(a.size() == b.size(),
+                    "Source and destination tile must have the same size. A=", a.size(),
+                    " B=", b.size());
 
   SizeType m = a.size().rows();
   SizeType n = a.size().cols();
--- a/include/dlaf/lapack_tile.tpp
+++ b/include/dlaf/lapack_tile.tpp
@@ -15,9 +15,9 @@ namespace tile {
 
 template <class T>
 void lacpy(const Tile<const T, Device::CPU>& a, const Tile<T, Device::CPU>& b) {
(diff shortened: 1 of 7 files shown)


include/dlaf/matrix.h
```diff
@@ -103,6 +103,34 @@ class Matrix : public Matrix<const T, device> {
...
+    for (SizeType j = 0; j < local_tile_cols; ++j) {
+      for (SizeType i = 0; i < local_tile_rows; ++i) {
+        hpx::dataflow(hpx::util::unwrapping([](auto&& tile_dst, auto&& tile_src) {
+                        for (SizeType j = 0; j < tile_src.size().cols(); ++j)
+                          for (SizeType i = 0; i < tile_src.size().rows(); ++i) {
+                            TileElementIndex index(i, j);
+                            tile_dst(index) = tile_src(index);
+                          }
+                      }),
+                      this->operator()(LocalTileIndex(i, j)), source.read(LocalTileIndex(i, j)));
```
teonnik, line 129: I'm not sure if this is an unpopular opinion but nesting function calls like that seems hard to read. How about this:

```
  using hpx::util::unwrapping;
  using hpx::dataflow;
  LocalTileIndex tile_idx(i, j);
  auto copy_f = unwrapping([](auto&& tile_dst, auto&& tile_src) {
    for (SizeType j = 0; j < tile_src.size().cols(); ++j)
      for (SizeType i = 0; i < tile_src.size().rows(); ++i) {
        TileElementIndex index(i, j);
        tile_dst(index) = tile_src(index);
      }
  });
  dataflow(copy_f, this->operator()(tile_idx), source.read(tile_idx));
```

rasolca: What about using the optimized [s,d,c,z]lacpy?

rasolca: > I'm not sure if this is an unpopular opinion but nesting function calls like that seems hard to read. How about this:
> 
> ```
>   using hpx::util::unwrapping;
>   using hpx::dataflow;
>   LocalTileIndex tile_idx(i, j);
>   auto copy_f = unwrapping([](auto&& tile_dst, auto&& tile_src) {
>     for (SizeType j = 0; j < tile_src.size().cols(); ++j)
>       for (SizeType i = 0; i < tile_src.size().rows(); ++i) {
>         TileElementIndex index(i, j);
>         tile_dst(index) = tile_src(index);
>       }
>   });
>   dataflow(copy_f, this->operator()(tile_idx), source.read(tile_idx));
> ```

@teonnik In general this is a good idea to improve readability!

rasolca:
Following from my suggestion of using `*lacpy`, it could be a good idea to implement a tile copy method that can be extended to have with different device types (i.e. which can also move data between the host and the gpu memory (with cudaMemcpy2D)).

include/dlaf/matrix.h
```diff
@@ -103,6 +103,34 @@ class Matrix : public Matrix<const T, device> {
     return operator()(this->distribution().localTileIndex(index));
   }
 
+  /// Copy values from another matrix
+  ///
+  /// Given a matrix with the same geometries and distribution, this function submits tasks that will
+  /// perform the copy of each tile
+  template <class U, class = std::enable_if_t<std::is_same<T, std::remove_const_t<U>>::value>>
+  void copyFrom(Matrix<U, Device::CPU>& source) {
```
rasolca, line 111: What about making it a free function to allow usage with matrices and views (using templates)?

albestro:
@rasolca I was wondering if it is worth using `lacpy` also for `Data` concepts copy function when the type is compatible (i.e. `float`, `double`, `std::complex<float>`, `std::complex<double>`).

albestro:
I implemented your suggestion about "separation of concerns" about copy of matrices and tiles. i think it is a nice architecture.

I have just one doubt about a detail that may be relevant for future implementations.

I'm wondering what it would be better to do:  @rasolca @teonnik 
- the copy tile function takes as input the two `Tile`s (not their futures, as in the current implementation). This function is called by the `hpx::dataflow` function which schedules it on the `matrix` side, starting from the `hpx::future`s;
- the copy tile function accepts `hpx::future`, so the specific implementation will be free to decide how and if to implement the parallelization.

This doubt may be strictly related to my lack of knowledge about how to mix HPX and CUDA calls together.

Moreover, I have another question. I know it is against the concept behind this library, but I'm wondering if we will never be interested in copying the entire matrix in one shot (wait all tiles, copy the entire memory chunk), but this may be the topic for another issue.

include/dlaf/matrix/copy.h
```diff
@@ -0,0 +1,43 @@
...
+namespace dlaf {
+
+/// Copy values from another matrix
+///
+/// Given a matrix with the same geometries and distribution, this function submits tasks that will
+/// perform the copy of each tile
+template <
+    template <class, Device> class MatrixTypeSrc, template <class, Device> class MatrixTypeDst,
+    class Tsrc, class Tdst, Device device_source, Device device_dest,
+    std::enable_if_t<std::is_same<const Tsrc, const Tdst>::value && !std::is_const<Tdst>::value, int> = 0>
```
teonnik, line 27: Since things might change regarding how `Device` is handled, I suggest we assume that both matrices have the same `Device` type to reduce complexity in the meantime. The `lacpy` function defined earlier assumes the `Scalar` type (`T`) to be equal, we should either generalize lacpy for Tiles to handle different types or assume that (`Tsrc = Tdst`).

albestro: About `Device`, it's ok for me, we can extend it later.

About `Tsrc==Tdst`, I'd like to highlight the constraint specified through the `std::enable_if` directive:
- `Tsrc` and `Tdst` must be the same type, except for constness
- `Tdst` cannot be a const type
This allows to use any tile, compatible with the purpose, without having to force the conversion between const/non-const (I'm talking about the pointed type).

If you have any suggestion about this, I'll be glad to take them into account.

albestro:
Good point, in the rest of the code we do not take as input `hpx::futures`, as you correctly pointed out about blas and lapack tile wrappers.

The main point about the two options is about this: if we would decide to pass the future to the `copy_tile` function, it will be able to decide if it is ok to schedule a task for future execution (i.e. through `hpx::dataflow`), or it needs to do some blocking thing before proceeding with the next tiles (e.g. `tile_future.get()`). It's true that in this way the copy would become a synchronization point, and as I already said in a previous comment, it is against the philosophy of this library.

include/dlaf/matrix/copy_tile.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "dlaf/lapack_tile.h"
+#include "dlaf/tile.h"
+
+namespace dlaf {
+
+template <
+    class Tsrc, class Tdst,
+    std::enable_if_t<std::is_same<const Tsrc, const Tdst>::value && !std::is_const<Tdst>::value, int> = 0>
+void copy(const Tile<Tsrc, Device::CPU>& source, const Tile<Tdst, Device::CPU>& dest) {
```
teonnik, line 23: This can be simplified as `const Tile<const T, D>&` works with `const Tile<T, D>&`

include/dlaf/matrix/copy.h
albestro: I was wrong about this, as we discussed offline you are right about it.

I forgot that I can do `ConstMatrix& = Matrix&` (the problem exists with other types).

Thank you for pointing it out, I'm going to fix that!

include/dlaf/lapack_tile.tpp
```diff
@@ -8,6 +8,30 @@
...
+  SizeType n = a.size().cols();
+
+  lapack::lacpy(lapack::MatrixType::General, m, n, a.ptr(), a.ld(), b.ptr(), b.ld());
+}
+
+template <class T, Device device>
+void potrf(blas::Uplo uplo, const Tile<T, device>& a) {
+  auto info = potrfInfo(uplo, a);
+  if (info != 0)
+    throw std::runtime_error("Error: POTRF: A is not positive definite.");
```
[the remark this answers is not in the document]
albestro: This task is going on in another place. I would leave it for the dedicated PR, since here I've just re-positioned it in the file (it would ease the merge later).

Do you agree?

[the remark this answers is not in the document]
teonnik: Yeah!

rasolca:
bors try

PR #187, opened 2020-04-15
Matrix and MatrixView documentation

rasolca:
Added documentation for the tile synchronization mechanism.


misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
+# Synchronization mechanism.
+
+To simplify the dependency tracking during the generation of the task graph we introduced
+a mechanism which generates automatically the following dependencies:
+- Successive tasks that modify (write) values of a tile are ordered, therefore each write
+  task depends on the previous write task.
+- Successive tasks that only read values a tile can be executed concurrently,
```
albestro, line 7:
```suggestion
- Successive tasks that only read values of a tile can be executed concurrently,
```

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+  however they all depends on the previous write task and the next write task depends on
+  the completion of all the read tasks.
+
+## Matrix and Matrix Tiles basics
+
+The Matrix object (`dlaf::matrix::Matrix`) is the object which manage the setup of
```
albestro, line 13:
```suggestion
The Matrix object (`dlaf::matrix::Matrix`) is the object which manages the setup of
```

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+For asynchronous scheduling [matrix views](#matrix-views) has to be used.
+
+### Basic Examples
+
+(Note: this examples are used to illustrate how dependencies are determined,
+however they cannot be compiled since they are simplified to improve readability.
+(e.g. constructors and template parameters are omitted)).
+
+- Example 1: Write dependencies.
+```
```
albestro, line 36: I would add `cpp` syntax to all pseudo-code examples

```suggestion
```cpp
```

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+// Task2 - Task4
+```
+
+### Exception handling
+
+If a task throws an exception during its execution the future returned by the asynchronous call
+contains the given exception.  Moreover any tile that was supposed to be written by the Task will
+be invalidated and any successive call to `operator()` and `read()` will return a future containing
+an exception. Note that because of C++ limitations we cannot propagate the correct exception type and
+message to tiles. Therefore they will contain a generic exception.
```
albestro, line 71: It may be worth specifying the type of our custom exception.

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+Matrix Views (`dlaf::matrix::MatrixView`) are objects which allow to schedule a part of the DAG of an
+algorithm in a different task.
+
+It can be created from a Matrix or from another MatrixView which will be indicated as "parent matrix".
+Both constant and non-constant views are available. The main difference is that `operator()`
+is not allowed for constant matrices as only read-only tasks can be performed on its tiles.
+
+When a matrix view is created each successive call of `operator()` and `read()` on the parent matrix
+will return futures which depend on the tasks scheduled with the futures obtained by the matrix view.
+It can be noted that it is possible to transfer the scheduling on only a part of the tiles.
```
albestro, line 84:
```suggestion
It is possible to transfer the scheduling of all or just a subset of the tiles.
```

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+but the correct dependencies are guarenteed. An important fact to be noted is that the view needs to be
+able to notify that no other tasks will be scheduled to enable the next tasks created from the parent matrix.
+The `MatrixView` API contains 3 ways to accomplish that:
+- the destructor notifies that no new tasks will be scheduled on any of the tile,
+- `done(tile_index)` notifies that no new task will be scheduled on the tile with the given index
+(After this call any use of `operator()` or `read()` with the same tile index will return an error.),
+- `doneWrite(tile_index)` notifies that no new read-write task will be scheduled on the tile with the given index
+(After this call any use of `operator()` with the same tile index will return an error.).
+
+It has to be noted that read-only task of the parent matrix and of the View can be performed at the same time
```
albestro, line 101:
```suggestion
It has to be noted that read-only tasks of the parent matrix and of the View can be performed at the same time
```

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+however they cannot be compiled since they are simplified to improve readability.
+(e.g. constructors and template parameters are omitted)).
+
+- Example 1: Write dependencies.
+```
+  Matrix m;
+
+  hpx::dataflow(Task1, m({0, 0})));
+  hpx::dataflow(Task2, m({0, 0})));  // Depends on Task1.
+  hpx::dataflow(Task3, m({0, 1})));  // Different tile. No dependency on Task1 or Task2.
```
albestro, line 41: Please, check unbalanced parenthesis also in other code snippets.

```suggestion
  hpx::dataflow(Task1, m({0, 0}));
  hpx::dataflow(Task2, m({0, 0}));  // Depends on Task1.
  hpx::dataflow(Task3, m({0, 1}));  // Different tile. No dependency on Task1 or Task2.
```

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+  hpx::dataflow(Task2, m.read({0, 0})));  // Depends on Task1
+
+  MatrixView mv(UpLo::General, m);
+
+  hpx::dataflow(Task3, m.read({0, 0})));  // Depends on doneWrite notification and Task5
+  hpx::dataflow(Task4, mv.read({0, 0})));  // Depends on Task2.
+  hpx::dataflow(Task5, mv({0, 0})));  // Depends on Task2 and Task4.
+  hpx::dataflow(Task6, mv.read({0, 0})));  // Depends on Task5.
+
+  mv.doneWrite({0, 0});
```
albestro, line 173: I'm not sure if putting this call in the "middle" of the dag construction can be useful or misleading.

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+  Matrix m;
+
+  auto future1 = hpx::dataflow(Task1, m({0, 0})));
+  auto future2 = hpx::dataflow(Task2, m.read({0, 0}));  // Depends on Task3 (The Tile used in Task1 is still available in future1).
+
+  future2.get();  // DEADLOCK! Task2 is not ready yet, and cannot get ready because the destructor of the Tile in future1 will not be called.
+
+  hpx::dataflow(Task3, future1);  // Depends on Task1.
+```
+- Example 3: Deadlock returning Tiles (2).
```
albestro, line 243: IMHO, this is the most subtle and clear example!

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+
+### Exception handling
+
+If a task throws an exception during its execution the future returned by the asynchronous call
+contains the given exception.  Moreover any tile that was supposed to be written by the Task will
+be invalidated and any successive call to `operator()` and `read()` will return a future containing
+an exception. Note that because of C++ limitations we cannot propagate the correct exception type and
+message to tiles. Therefore they will contain a generic exception.
+
+## Matrix Views
```
albestro, line 73: _"A MatrixView can be seen as a "sub-DAG" tool in the global DAG of the Matrix. The MatrixView gives us the chance to take control of a particular asynchronous step of the Matrix, and we can continue to work on the Matrix considering that everything we will do after the MatrixView, will happen after the MatrixView will be set as `done` (or `doneWrite` if we just want to release the writeability).
This is very handy, because it allows us to schedule tasks inside the sub-DAG of the MatrixView, and it will be like inserting new dependencies in between two already existing ones."_

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+](figures/matrix_sync.png)\
+\
+![Fig. 1b DAG](figures/matrix_sync_dag.png)\
+*Fig. 1 Graphical representation of the promise-future mechanism for a tile of the matrix object.*
+
+![Fig. 2 Legend for Figure 1, 3, 4.
+](figures/matrix_sync_legend.png)\
+*Fig. 2 Legend for Figure 1, 3, 4.*
+
+Extra care has to be used to ensure correct exception handling.
```
albestro, line 328:
```suggestion
### Async Exceptions
To ensure correct exception handling, the tile destructor sets the promise value if it was called during normal execution,
while it sets a custom exception when it is called after an exception has been thrown.
In the latter case, the call to `get()` in the continuation will rethrow the exception, the promise will be destructed and the new future stored in the matrix will have an invalid shared state.
Therefore the continuations in `operator()` and `read()` have to handle it.
```

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+
+  future4.get();  // Task4 may not be ready yet, but it will get ready after the completion of Task4.
+```
+
+## Implementation details
+
+To simplify the understanding of the dependencies mechanism we explain it for simple cases and we will add
+other features one by one.
+
+The simplest case includes constant matrices with read-only tiles. In this case only a shared future
```
albestro, line 294:
```suggestion
### Matrix mechanism
The simplest case includes constant matrices with read-only tiles. In this case only a shared future
```

misc/synchronization.md
```diff
@@ -0,0 +1,368 @@
...
+and duplicates the tile included (the promise cannot be dublicated,
+therefore in the copy it is replaced by a new promise (the blue promise in Fig 3 and 4 moved))
+to set the shared future (purple and red arrow pointing to the light yellow promise in Fig 3 and 4).
+When both the blue and purple futures are ready, it means that the tiles of both the two shared futures
+went out of scope, and therefore the reading of the elements of the tiles are completed.
+At this point the main future in the parent matrix can be set. A synchrounous dataflow is used to set
+the light green future value.
+However a copy of the shared future is still included in the matrix view
+(allowing read-only operations to still be scheduled.) The shared future is cleared by the call to
+`done()` after which the tile control is fully back to the parent matrix.
```
albestro, line 353:
```suggestion
### MatrixView

`MatrixView<const T>` can be handled in a simple way. On construction it only needs to call the `read()`
method of the parent matrix to get a copy of the shared future, which is saved in the view and will be returned
each time the view is used (`read()`). To release the view the `done()` method just clears the view shared future.

Non constant `MatrixView<T>` requires a more complex handling mechanism.
On construction the two futures (future and shared_future) of the parent matrix are moved in the view and replaced by futures of the promises also stored in the view. This is important because

- on one side, the MatrixView becomes the natural continuation of the Matrix in the exact point where we created it;
- on the other side it detaches the main DAG, allowing to schedule tasks in the sub-DAG, but keeping track of where the flow has to go when the MatrixView will be marked as `done`/`doneWrite`.

For what concerns the `operator()` and `read()`, the MatrixView works exactly as the Matrix.

The new part comes when we have to deal with the extra features of the MatrixView: `doneWrite` and `done`.
These methods are the ones that allows to go back to the normal flow that was detached on construction. 

We have to keep in mind two things:
- we have to attach back two lines, future (RW) and shared_future (R) in the parent matrix
- with a call to `doneWrite`, the MatrixView should be able to continue get the read-access, but at the same time also the parent matrix should be able to read in parallel (since no other changes can be made by the subDAG of the MatrixView).

In order to achieve this, there is a articulated machinery that allows to finely control this re-attachment to the normal flow.

A call to `doneWrite` internally performs a `read()` operation, this will be the last shared_future in the sub-DAG, and it will be the one that will be returned by any next call to `read` on the MatrixView.
After that, a continuation task is executed on a copy of the shared_future. This is a crucial point of the machinery. As soon as the `shared_future` is ready, this task will create a copy of the tile (const), that will be used to trigger the shared_future of the parent matrix. This is what allows to have both the MatrixView and the parent Matrix to access in read-only mode the tile in parallel. Here there is the trick:
- the Tile in the original shared future of the MatrixView has internally a promise that triggers the future of the MatrixView
- the Tile copied, which just copies the reference and it does not copy any element, is given another promise that will set an ad-hoc temporary future.

This two tiles together provide a way to lock the future write acesses from the parent Matrix. In fact, the synchronization taks that will attach back the future of the parent matrix, is triggered by both of them. So if the MatrixView releases all shared_futures, the mechanism has to wait that read access from the MatrixView will be released too, and viceversa.
```

misc/synchronization.md
rasolca: Yes, the positions of done and doneWrite are a bit messed up.
I would put them as soon as they can be called, i.e.:

- doneWrite after creation of task5
- done after task6

What do you think?

misc/synchronization.md
```diff
@@ -72,16 +72,22 @@ message to tiles. Therefore they will contain a generic exception.
 
 ## Matrix Views
 
-Matrix Views (`dlaf::matrix::MatrixView`) are objects which allow to schedule a part of the DAG of an
-algorithm in a different task.
+A matrix view can be can be used to create "sub-DAG" in the global DAG of a matrix.
+It gives the chance to take control of a particular asynchronous step of the matrix,
+allowing to concurrently schedule the work that will bbe executed after the work
```
albestro, line 77:
```suggestion
allowing to concurrently schedule the work that will be executed after the work
```

misc/synchronization.md
```diff
@@ -72,16 +72,22 @@ message to tiles. Therefore they will contain a generic exception.
...
+scheduled by the matrix view has been executed and the matrix view is notified that
+all the tasks has been scheduled (`done()`, `doneWrite()`, destructor).
```
albestro, line 79: IMHO this sentence is quite long, you could evaluate to use two bullet points to give more structure to your thought:

>[...] allowing to concurrently schedule the work that will be executed:
>- after the work scheduled by the `MatrixView` has been executed, and
>- the `Matrix` ~view~ is notified that all the tasks has been scheduled (`done()`, `doneWrite()`, destructor).

Moreover, it is not so clear to me what you mean with the "2nd point". I think you are trying to say something like "the `MatrixView` will not schedule any additional task". In case, I would try to make it more clear.

misc/synchronization.md
```diff
@@ -98,7 +104,7 @@ The `MatrixView` API contains 3 ways to accomplish that:
 - `doneWrite(tile_index)` notifies that no new read-write task will be scheduled on the tile with the given index
 (After this call any use of `operator()` with the same tile index will return an error.).
 
-It has to be noted that read-only task of the parent matrix and of the View can be performed at the same time
+It has to be noted that read-only tasks of the parent matrix and of the View can be performed at the same time
```
albestro, line 107:
```suggestion
It has to be noted that read-only tasks of the parent matrix and of the `MatrixView` can be performed at the same time
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+To ensure correct exception handling, the tile destructor sets the promise value if it was called during normal execution,
+while it sets a custom exception when it is called after an exception has been thrown.
+In the latter case, the call to `get()` in the continuation will rethrow the exception,
+the promise will be destructed and the new future stored in the matrix will have an invalid shared state.
+Therefore the continuations in `operator()` and `read()` have to handle it.
 The tile destructor sets the promise value if it was called during normal execution, while it sets
 an exception when it is called after an exception has been thrown.
 If the future contains an exception the call to `get()` in the continuation will rethrow the exception,
 the promise will be destructed and the new future stored in the matrix will have an invalid shared state.
 Therefore the continuations in `operator()` and `read()` have to handle this case.
```
albestro: I think you wanted to remove this 😉

```suggestion
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
-(allowing read-only operations to still be scheduled.) The shared future is cleared by the call to
-`done()` after which the tile control is fully back to the parent matrix.
-
-It should be noted that the  `done()` method invokes `doneWrite()` if the tile is still in read-write mode, i.e. when `doneWrite()` hasn't be called yet.
-Figure 3 and 4 show how a non-constant matrix view works.
+### Matrix View
+
+`MatrixView<const T>` can be handled in a simple way. On construction it only needs to call the `read()`
+method of the parent matrix to get a copy of the shared future, which is saved in the view and will be returned
+each time the view is used (`read()`). To release the view the `done()` method just clears the view shared future.
```
albestro, line 356:
```suggestion
method of the parent matrix to get a copy of the shared future, which is saved internally and it will be returned
each time it is used (`read()`). To release the view, the `done()` method just clears the view shared future.
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+
+Non constant `MatrixView<T>` requires a more complex handling mechanism.
+On construction the two futures (future and shared future) of the parent matrix are moved in the view and
+replaced by futures of the promises also stored in the view. This is important because
+
+- on one side, the matrix view becomes the natural continuation of the matrix in the exact point where we created it;
```
albestro, line 362:
```suggestion
- on one side, the `MatrixView` becomes the natural continuation of the parent matrix in the exact point in the DAG where we created it;
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+- on the other side it detaches the main DAG, allowing to schedule tasks in the sub-DAG,
+but keeping track of where the flow has to go when the matrix view will be marked as `done`/`doneWrite`.
```
albestro, line 364:
```suggestion
but keeping track of where the flow has to go when the `MatrixView` will be marked as `done()`/`doneWrite()`.
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+
+For what concerns the `operator()` and `read()`, the matrix view works exactly as the matrix.
+
+The new part comes when we have to deal with the extra features of the matrix view: `doneWrite` and `done`.
```
albestro, line 368:
```suggestion
The new part comes when we have to deal with the extra features of the matrix view: `doneWrite()` and `done()`.
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+These methods are the ones that allows to go back to the normal flow that was detached on construction.
+
+We have to keep in mind two things:
+- we have to attach back two lines, future (RW) and shared future (R) in the parent matrix
+- with a call to `doneWrite`, the matrix view should be able to continue get the read-access,
+but at the same time also the parent matrix should be able to read in parallel
+(since no other changes can be made by the subDAG of the matrix view).
```
albestro, line 375:
```suggestion
(since no other changes can be made by the subDAG of the `MatrixView`).
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+
+In order to achieve this, there is a articulated machinery that allows to finely control this re-attachment to the normal flow.
+
+A call to `doneWrite` internally performs a `read()` operation, this will be the last shared future in the sub-DAG,
```
albestro, line 379:
```suggestion
A call to `doneWrite()` internally performs a `read()` operation, this will be the last shared future in the sub-DAG,
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+and it will be the one that will be returned by any next call to `read` on the matrix view.
+After that, a continuation task is executed on a copy of the shared future. This is a crucial point of the machinery.
+As soon as the `shared_future` is ready, this task will create a copy of the tile (const),
```
albestro, line 382:
```suggestion
As soon as the `shared_future` is ready, this task will create a copy of the tile (`Tile<const T>`),
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+that will be used to trigger the shared future of the parent matrix.
+This is what allows to have both the matrix view and the parent matrix to access in read-only mode the tile in parallel.
+Here there is the trick:
+- the Tile in the original shared future of the matrix view has internally a promise that triggers the future of the matrix view
+- the Tile copied, which just copies the reference and it does not copy any element,
```
albestro, line 387:
```suggestion
- the `Tile<const T>` copied, which just copies the reference and it does not copy any element,
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+  is given another promise that will set an ad-hoc temporary future.
+
+This two tiles together provide a way to lock the future write accesses from the parent matrix.
+In fact, the synchronization taks that will attach back the future of the parent matrix, is triggered by both of them.
```
albestro, line 391:
```suggestion
In fact, the synchronization task that will attach back the future of the parent matrix, is triggered by both of them.
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+So if the matrix view releases all shared futures, the mechanism has to wait that all the shared future
```
albestro, line 392:
```suggestion
So if the `MatrixView` releases all shared futures, the mechanism has to wait that all the shared futures
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+of the parent matrix are released too, and viceversa.
+
+It should be noted that the  `done()` method invokes `doneWrite()` if the tile is still in read-write mode,
```
albestro, line 395:
```suggestion
It should be noted that the `done()` method invokes `doneWrite()` if the `Tile` is still in read-write mode,
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+but keeping track of where the flow has to go when the matrix view will be marked as `done`/`doneWrite`.
+
+For what concerns the `operator()` and `read()`, the matrix view works exactly as the matrix.
+
+The new part comes when we have to deal with the extra features of the matrix view: `doneWrite` and `done`.
+These methods are the ones that allows to go back to the normal flow that was detached on construction.
+
+We have to keep in mind two things:
+- we have to attach back two lines, future (RW) and shared future (R) in the parent matrix
+- with a call to `doneWrite`, the matrix view should be able to continue get the read-access,
```
albestro, line 373:
```suggestion
- with a call to `doneWrite`, the `MatrixView` should be able to continue get the read-access,
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+but at the same time also the parent matrix should be able to read in parallel
+(since no other changes can be made by the subDAG of the matrix view).
+
+In order to achieve this, there is a articulated machinery that allows to finely control this re-attachment to the normal flow.
+
+A call to `doneWrite` internally performs a `read()` operation, this will be the last shared future in the sub-DAG,
+and it will be the one that will be returned by any next call to `read` on the matrix view.
```
albestro, line 380:
```suggestion
and it will be the one that will be returned by any next call to `read` on the `MatrixView`.
```

misc/synchronization.md
```diff
@@ -325,35 +337,63 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
+After that, a continuation task is executed on a copy of the shared future. This is a crucial point of the machinery.
+As soon as the `shared_future` is ready, this task will create a copy of the tile (const),
+that will be used to trigger the shared future of the parent matrix.
+This is what allows to have both the matrix view and the parent matrix to access in read-only mode the tile in parallel.
```
albestro, line 384:
```suggestion
This is what allows to have both the `MatrixView` and the parent matrix to access in read-only mode the tile in parallel.
```

misc/synchronization.md
rasolca: What I mean is that the work scheduled by the matrix will be executed after the work of the view only when the matrix view goes out of scope or done/doneWrite is called. Otherwise other work can be still scheduled by the view.

misc/synchronization.md
```diff
@@ -0,0 +1,413 @@
...
+to avoid deadlocks the scope of tile shared futures needs extra care.
+
+Note that returning matrix tiles from tasks may be a source of deadlocks,
+therefore it should be avoided if possible. Otherwise a carefull analysis of the dependencies is needed to avoid problems.
+
+### Examples advanced use
+
+(Note: this examples are used to illustrate how dependencies are determined,
+however they cannot be compiled since they are simplified to improve readability.
+(e.g. constructors and template parameters are omitted)).
```
rasolca, line 213:
```suggestion
however they cannot be compiled since they are simplified to improve readability
(e.g. constructors and template parameters are omitted).)
```

misc/synchronization.md
```diff
@@ -0,0 +1,413 @@
...
+(After this call any use of `operator()` with the same tile index will return an error.).
+
+It has to be noted that read-only tasks of the parent matrix and of the view can be performed at the same time
+if no read-write tasks has been scheduled in between (see [examples](#view-examples)).
+
+### View Examples
+
+(Note: this examples are used to illustrate how dependencies are determined,
+however they cannot be compiled since they are simplified to improve readability.
+(e.g. constructors and template parameters are omitted)).
```
rasolca, line 114:
```suggestion
however they cannot be compiled since they are simplified to improve readability
(e.g. constructors and template parameters are omitted).)
```

misc/synchronization.md
```diff
@@ -0,0 +1,413 @@
...
+if no `operator()` was invoked.
+
+Both the `Matrix` and the `Tile` objects are not thread safe. The future mechanism ensures that
+each tile is either accessed in read-write mode by only one task, or in read-only mode by more tasks.
+On the other hand the user has to guarantee that a `Matrix` object is used only by one task at a time.
+For asynchronous scheduling [matrix views](#matrix-views) has to be used.
+
+### Basic Examples
+
+(Note: this examples are used to illustrate how dependencies are determined,
```
rasolca, line 31:
```suggestion
(Note: these examples are used to illustrate how dependencies are determined,
```

rasolca:
@teonnik My plan is to add the markdown files in the doxygen documentation. This requires the images to be available directly in the code.
However I'm not much worried about the doc images. They are small and rarely modified after the squash merge.

(Anyway I don't understand how putting them in a different branch can speed-up the clone... by default all the branches are downloaded.)

rasolca:
@teonnik 
An example for const view is the possibility to schedule independently two algorithm which just need the matrix as input.

For non const view is a bit more complicated, but one example can be Cholesky.
Currently the execution fails if the input matrix is not positive definite.
To make it possible to return an error (e.g.like in scalapack) and allow continuation, the DAG has to be scheduled in pieces which require some intermediate results.
To avoid to stop the scheduling of everything else Cholesky DAG creation has to be scheduled on a different task and MatrixVIew is used to introduce thread safety.

misc/synchronization.md
```diff
@@ -338,62 +338,67 @@ Figure 1 shows how the mechanism works, while Figure 2 contains the legend of th
...
-So if the matrix view releases all shared futures, the mechanism has to wait that all the shared future
+However, it has to e noted that a promise cannot be copied, therefore the copy will get a new promise.
+(The copy with a new promise is represented in Figure 3 and 4
+with the purple (copy) and red (move of the new promise) arrows pointing to the light yellow promise.)
+
+The crucial point of the machinery is the correct handling of the next dependency
+that has to manage the synchronization of both the shared future of the matrix view and of the parent matrix.
+A synchronous dataflow, which sets the value of the internal future (light green future in Figure 3 and 4)
+of the parent matrix, is used in this case, whose arguments are two futures:
+- the tile in the original shared future of the matrix view has internally a promise
```
albestro, line 390: - the tile in the original shared future of the matrix view **(orange/yellow)** has internally a promise

PR #190, opened 2020-04-16
Proposal of additional checks for compute indexes functions

albestro:
This is just a "branch" proposal, because I'm not sure to be on the right path and I'm willing to receive your feedbacks. @rasolca @teonnik 

### 1. `to_signed` / `to_unsigned`
Add "same signedness" fallback functions for both `to_signed` and `to_unsigned`: these will generalize the helper functions for the `static_cast` (with DLAF_ASSERT for optional checks).

### 2. `computeCoords`
It accepts any kind of integral type as linear index, and it performs some (optional) checks.
Moreover, computations are performed by forcing the "promotion" to unsigned for both types (`IndexT` and `LinearIndexT`) and then the function `to_signed` is used for a "safe cast".

I want to highlight that the type of the `Index2D` you get as result is strictly related to the Size2D you pass, i.e. you will get the matching one.

### 3. `computeLinearIndex`
Calculation of the linear index is performed with `std::size_t` arithmetics, then it is casted using `to_signed` to the `IndexT`.

This is a limitation since it may be not possible to store the 1D linear index of a generic 2D position  (i.e. we can have a grid of more than `std::numerical_limits<IndexT>::max()` using `IndexT` as size for the two components, e.g. `IndexT = int8_t` 20x20 grid, the bottom right elements can not be calculated)

We may overcome this issue by explicitly specifying the return type of the method via the template parameter.

### DIFF
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -267,8 +267,8 @@ Index2D<IndexT, Tag> computeCoords(Ordering ordering, std::ptrdiff_t index,
 
 /// Compute linear index of an Index2D in a row-major ordered 2D grid
 ///
-/// The @tparam LinearIndexT cannot be deduced and it must be explicitly specified. It allows to internalize
-/// the casting of the value before returning it, not leaving the burden to the user.
+/// The @tparam LinearIndexT cannot be deduced and it must be explicitly specified. It allows to
+/// internalize the casting of the value before returning it, not leaving the burden to the user.
 ///
 /// @tparam LinearIndexT can be any integral type signed or unsigned
 /// @pre LinearIndexT must be able to store the result
@@ -289,8 +289,8 @@ LinearIndexT computeLinearIndexRowMajor(const Index2D<IndexT, Tag>& index,
 
 /// Compute linear index of an Index2D in a column-major ordered 2D grid
 ///
-/// The @tparam LinearIndexT cannot be deduced and it must be explicitly specified. It allows to internalize
-/// the casting of the value before returning it, not leaving the burden to the user.
+/// The @tparam LinearIndexT cannot be deduced and it must be explicitly specified. It allows to
+/// internalize the casting of the value before returning it, not leaving the burden to the user.
 ///
 /// @tparam LinearIndexT can be any integral type signed or unsigned
 /// @pre LinearIndexT must be able to store the result
@@ -314,8 +314,8 @@ LinearIndexT computeLinearIndexColMajor(const Index2D<IndexT, Tag>& index,
 /// It acts as dispatcher for computeLinearIndexColMajor() and computeLinearIndexRowMajor()
 /// depending on given @p ordering.
 ///
-/// The @tparam LinearIndexT cannot be deduced and it must be explicitly specified. It allows to internalize
-/// the casting of the value before returning it, not leaving the burden to the user.
+/// The @tparam LinearIndexT cannot be deduced and it must be explicitly specified. It allows to
+/// internalize the casting of the value before returning it, not leaving the burden to the user.
 ///
 /// @tparam LinearIndexT can be any integral type signed or unsigned (it must be explicitly specified)
 /// @pre LinearIndexT must be able to store the result
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -187,15 +187,51 @@ public:
 };
 
 template <class IndexT, class Tag, class LinearIndexT>
-Index2D<IndexT, Tag> computeCoordsRowMajor(LinearIndexT index,
+Index2D<IndexT, Tag> computeCoordsRowMajor(LinearIndexT linear_index,
                                            const Size2D<IndexT, Tag>& dims) noexcept {
-  return {static_cast<IndexT>(index / dims.cols()), static_cast<IndexT>(index % dims.cols())};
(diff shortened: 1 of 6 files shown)


albestro:
### 1. integral casting helper functions

Current status proposal provides:
- extended `to_signed` and `to_unsigned`, they now accept any integral type as input, while as output they accept just a signed or unsigned integral type, respectively;
- added `integral_cast` that acts as a dispatcher for the previous two, and can convert from/to any integral type

All these helper functions are just simple static_cast, but they have runtime checks in case DLAF_ASSERT_MODERATE level is enabled.

For these helper cast functions, the unit test `test_types` has been introduced.

albestro:
This PR evolved over time and it has been partially addressed in other PRs and merged into master.
- *"integral casting helper functions"* have been already completely addressed in another PR and merged into `master`;

This PR will now target `computeCoords` and `computeLinearIndex` (whose API have been improved in other PRs).

The main target, which is also the initial core of this PR, is _**to make functions `computeCoords` and `computeLinearIndex` safe by providing guards for numerical limits of integer types used**_.

Main steps I'm going to take:
- Evaluate to fix type for `LinearIndexT` to `std::ptrdiff_t`
- Implement guards and improve arithmetic for both functions;
- Extend and fix tests for these two functions;
- Try to reduce code duplication;

For doing this I will take into account:
- https://github.com/eth-cscs/DLA-Future/pull/183#issuecomment-619073123
- https://github.com/eth-cscs/DLA-Future/pull/199#issuecomment-621057940

albestro:
@rasolca @teonnik 

Some advancements on this side, but before a quick briefing on relevant concepts.

- By design, `Index2D` and `Size2D` use signed integral types for components;
- We have `SizeType` for BLAS, `int` for MPI, they are both signed types by requirement;

### computeCoords
Now, it uses `LIT=ptrdiff_t`, which being the counterpart of `std::size_t`, it is the biggest signed integral type. This means that we can use any signed integral type without any conversion problem, apart from the fact that by using `IT=int64_t` we may fall in the case where it is not possible to use `std::ptrdiff_t` for storing the linear index (i.e. if `dims.rows() * dims.cols() >= 2^63-1`).

It checks that the given linear index is inside the valid range `[0, rows*cols)`, then it performs the computation with `ptrdiff_t` arithmetic and convert the resulting components with `to_signed`. At this point the, potentially, narrowing conversion is safe, otherwise it would not be valid the `dims` given as input parameter. Moreover, it is just a narrowing conversion without change of signedness because we are working with signed integral types.

I used `to_signed`, but we can decide to use a simple `static_cast` without additional guards.

### computeLinearIndex
Now, it requires to explicitly specify the `LIT` as first template parameter. It is useful because this function have to deal with different interfaces (e.g. `SizeType` for BLAS and `int` for MPI), and it performs the safe cast internally without leaving the burden to the user of doing it on the output.

Internal computation of the linear index is performed in `std::size_t` arithmetic (1), and then the `integral_cast` is performed to check that the value can be stored in the chosen `LIT`.

(1) It must be noted that `size_t` arithmetic is not still fully bullet-proof, since there is no check that the result value does not exceed 64bit.

For what concerns tests, it is still not worth giving a look.

teonnik:
On general discussion on signed and unsigned types, I opened this issue : https://github.com/eth-cscs/DLA-Future/issues/212. 

Regarding the PR,

> IT=int64_t we may fall in the case where it is not possible to use std::ptrdiff_t for storing the linear index

This is incredibly unlikely and we will likely have other issues if we are at that point.

> I used to_signed, but we can decide to use a simple static_cast without additional guards.

Either way seems fine by me.

> Now, it requires to explicitly specify the LIT as first template parameter. It is useful because this function have to deal with different interfaces (e.g. SizeType for BLAS and int for MPI), and it performs the safe cast internally without leaving the burden to the user of doing it on the output.

This is a great point and it will be great if it is added as a comment at the function.

> Internal computation of the linear index is performed in std::size_t arithmetic (1), and then the integral_cast is performed to check that the value can be stored in the chosen LIT.

I would stick to std::ptrdiff_t (a signed type), to reduce casts, it's unlikely that the 2x extra size_t values will save us if we are at that extreme anyway.

albestro:
Thanks for the feedback! I agree with you.

Before passing to improve also the test part, I still have a few doubts.

>>  IT=int64_t we may fall in the case where it is not possible to use std::ptrdiff_t for storing the linear index

> This is incredibly unlikely and we will likely have other issues if we are at that point.

Do we want to add a guard on this?

>> Internal computation of the linear index is performed in std::size_t arithmetic (1), and then the integral_cast is performed to check that the value can be stored in the chosen LIT.

> I would stick to std::ptrdiff_t (a signed type), to reduce casts, it's unlikely that the 2x extra size_t values will save us if we are at that extreme anyway.

At the moment I'm using `std::size_t` arithmetic, just because of legacy reason. Would it be better to add some basic `std::ptrdiff_t` arithmetic with guards like we did for `std::size_t` and use it here?

rasolca:
I'm fine with using signed indices and introducing ptr_diff_t arithmetic.
(note that size_t arithmetic would be still needed for interfacing with STL).

I would not add guards around size_t and ptr_diff_t operations for this simple reason:
to overflow them you need a linear index of a square matrix with size O(1e9).
This means O(Exabytes) of node memory and O(1e27 flop) (with an exaflop machine this would mean O(100 years) execution time...)
I think that a larger integer will be available when it will happen to be able to deal with this numbers...

albestro:
bors try

PR #194, opened 2020-04-20
Refactor FindMKL

teonnik:
- Rename the FindMKL module to FindDLAF_MKL to avoid clashes with FindMKL modules in other projects (e.g. SIRIUS)
- Prefix CACHE variables in the module with DLAF_ to avoid clashes with other projects 
- Tailor and simplify the MKL module to our needs, namely only search for the 32bit interface and sequential MKL
- Squash a warning with CMake 3.17:
```
CMake Warning (dev) at /usr/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args`
  (MKL_BLAS_32BIT_SEQ) does not match the name of the calling package (MKL).
  This can lead to problems in calling code that expects `find_package`
  result variables (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/FindMKL.cmake:257 (find_package_handle_standard_args)
  CMakeLists.txt:54 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

- Add more search paths for MKL


rasolca:
Note that extra care should be applied to avoid that two different mkl interfaces are linked to the same app (one through DLAF and the other through the application). This is why it should be able to link with different flavours of mkl.
(I.e. 32 bit sequential is fine for developments but we will have to deal with multithreaded blas later (i.e. reducing number of threads before starting hpx and resuming later))

teonnik:
> Note that extra care should be applied to avoid that two different mkl interfaces are linked to the same app (one through DLAF and the other through the application). This is why it should be able to link with different flavours of mkl.

That is a good point. I had already stumbled on the issue when integrating COSMA in SIRIUS. From a build system perspective, I don't see an easy way to avoid conflicts unless we agree on using the same FindMKL module. That is difficult unless the Find module is part of CMake, which is, unfortunately, highly unlikely. CMake developers strongly prefer that respective projects provide Config files instead of including more Find modules in their code base.

Within CSCS, we now have https://github.com/eth-cscs/cmake-recipes . Unfortunately, there hasn't been much activity over there recently. On the upside, SIRIUS seems open to including the FindMKL module so maybe we can bring it up for discussion at the group meeting tomorrow.

teonnik:
I will close this for now, perhaps it will be better to have this as a submodule: https://github.com/eth-cscs/cmake-recipes

PR #196, opened 2020-04-20
Add a macro to check CUDA calls

teonnik:
- The macro checks for errors in CUDA calls and aborts if there are any.
  The line and filename are printed for debugging.


include/dlaf/cuda_assert.h
```diff
@@ -0,0 +1,31 @@
...
+
+namespace dlaf {
+
+#ifdef DLAF_WITH_CUDA
+inline void cuda_assert(cudaError_t err, char const* file, int line) noexcept {
+  if (err != cudaSuccess) {
+    printf("[CUDA ERROR] %s:%d: '%s'\n", file, line, cudaGetErrorString(err));
+    std::abort();
+  }
+}
```
albestro, line 25: What about using `DLAF_ASSERT` or something equivalent that it is not filtered by any flag (e.g. `DLAF_CHECK`)?

Using it we can centralize and make homogeneous the handling. It is possible to replicate the same trivial mechanism here, but it would be preferable to have it just in one place.

Moreover, `DLAF_ASSERT` uses `std::terminate` that allows to set a custom handler, that may be useful in our case where we may want to abort HPX + MPI + CUDA, depending on the context.

teonnik: > What about using DLAF_ASSERT or something equivalent that it is not filtered by any flag (e.g. DLAF_CHECK)?

`DLAF_ASSERT` and friends serve a different purpose, they allow us to check pre/postconditions and to assert statements within our functions and constructors. This is separate from error handling, it's what C++20 referred to as contracts. I made an unfortunate choice in naming the macro here. It should be DLAF_CUDA_CALL instead, mirroring the `MPI_CALL` macro (which I will also predix with DLAF_ in this PR). These macros handle errors reported by our dependencies (MPI and CUDA respectively), we have decided to just print what the errors are and abort but in principle we can make a different decision and continue executing if we know that we can recover from the error. In our case, we most likely can't, that is why we abort/terminate. HPX usually throws exceptions instead of using error codes to allow us to recover. 

As mentioned in this issue : https://github.com/eth-cscs/DLA-Future/issues/132 , there is a recent paper by Herb Sutter (one of the top C++ guys) regarding deterministic exceptions (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0709r4.pdf), the paper has the following section `4.2 Proposed cleanup: Don’t report logic errors using exceptions` (I recommend reading the section). This is what we want to do with our DLAF_ASSERT and friends, we don't want to throw exceptions or return error codes, we just want to inform the user that there is a non-recoverable bug. A lot of C++ libraries nowadays use exceptions for that but they miss out on some performance gains by not being able to mark their functions/constructors with `noexcept`.

teonnik: > Moreover, DLAF_ASSERT uses std::terminate that allows to set a custom handler, that may be useful in our case where we may want to abort HPX + MPI + CUDA, depending on the context.

Regarding `std::terminate()` vs `std::abort()`. HPX currently doesn't set a termination handler with `std::set_terminate()` except inside their unit tests. The default termination handler is just `std::abort()` (https://en.cppreference.com/w/cpp/error/terminate_handler). HPX does set a handler for `std::abort()` with `std::signal()` called `hpx::detail::on_abort()` in `src/runtime.cpp`. The handler just sets a global flag and calls `std::exit()`. HPX also sets a handler for both `std::exit()` and `std::quick_exit()` via `std::atexit()` and `std::at_quick_exit()` called `hpx::detail::on_exit()` in `src/runtime.cpp` that sets the same global flag.

I currently don't see any benefit of using `std::abort()` over `std::terminate()` so I will just use the latter as you suggested to be more uniform with other parts of the code. I am also not sure if we would want to use `std::set_terminate()` to set a handler yet. The only thing that annoys me a little bit with `std::terminate()` is that it prints a message if called directly as opposed to invoked via an uncaught exception.

teonnik: Found this gem that is worth a read on `std::terminate()` : https://akrzemi1.wordpress.com/2011/10/05/using-stdterminate/

albestro: I completely agree with you about _"DLAF_ASSERT & Friends"_ and on the separated topic about checking CUDA and MPI calls (and also their names), in addition to the performance benefits we get by being able to specify `noexcept` (that IMHO we should use it more extensively).

About `std::terminate`, I think it is a nice placeholder for now and honestly I have not thought about the exit strategy for MPI+HPX. What I have in mind for the handler, it is that at least in MPI it would be nice to call `MPI_Abort` to signal other ranks that one of them failed.

include/dlaf/communication/error.h
```diff
@@ -0,0 +1,46 @@
...
+/// handler use `MPI_Errhandler_set`.
+///
+inline void mpi_call(int err, char const* file, int line) {
+  if (err != MPI_SUCCESS) {
+    std::unique_ptr<char[]> err_buff(new char[MPI_MAX_ERROR_STRING]);
+    int err_str_len;
+    MPI_Error_string(err, err_buff.get(), &err_str_len);
+    std::printf("[MPI ERROR] %s:%d: '%s'\n", file, line, err_buff.get());
+    // MPI_Abort(MPI_COMM_WORLD, err);
+    // std::abort();
```
albestro, line 39: Are these comments to be removed? Or is it something you're willing to leave there for future reminder?

In general, I would be against having commented code.

include/dlaf/communication/error.h
```diff
@@ -0,0 +1,46 @@
...
+
+/// MPI provides two predefined error handlers
+///
+/// - `MPI_ERRORS_ARE_FATAL` (the default ?): causes MPI to abort
+/// - `MPI_ERRORS_RETURN`: causes MPI to return an error values instead of aborting
+///
+/// This function is only relevant when the error handler is `MPI_ERRORS_RETURN`. To set the error
+/// handler use `MPI_Errhandler_set`.
+///
+inline void mpi_call(int err, char const* file, int line) {
```
albestro, line 32: Now we have `SOURCE_LOCATION` and a trivial "pre-C++20" implementation of `source_location`

https://github.com/eth-cscs/DLA-Future/blob/095997946b7799179ec7526f1d4631811dffdd56/include/dlaf/common/utils.h#L23-L45

I would go for using it.

include/dlaf/cuda/error.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <cstdio>
+#include <cstdlib>
+
+namespace dlaf {
+namespace internal {
+
+#ifdef DLAF_WITH_CUDA
+
+inline void cuda_call(cudaError_t err, char const* file, int line) noexcept {
```
albestro, line 27: See `source_location` comment in `mpi_call`.

include/dlaf/communication/error.h
teonnik: I hesitated here and left this commented out for now. The issue is that if we are consistent with MPI_Abort (or set a handler using it) we have to add it to DLAF_CUDA_CALL and DLAF_ASSERT_*  . The latter will make a lot of tests require MPI initialization.

I looked at Trilinos (Tpetra and Teuchos) and Boost MPI to see whether they set a termination handler and they don't. 

Suggestions on how to proceed are welcome! :)

include/dlaf/communication/error.h
```diff
@@ -0,0 +1,46 @@
...
+#include <memory>
+
+#include <mpi.h>
+
+namespace dlaf {
+namespace internal {
+
+/// MPI provides two predefined error handlers
+///
+/// - `MPI_ERRORS_ARE_FATAL` (the default ?): causes MPI to abort
```
rasolca, line 26: Yes this is the default, and the reason why MPI calls are usually not checked for error code.
Since MPI errors are not recoverable the use of MPI_ERRORS_RETURN is useful only if the application can dump a snapshot and use it when it is restarted.
However most of the applications that use checkpoints do not use this approach to avoid to have to deal with different entry point and inconsistent data among ranks. They usually just always write a checkpoint file in specific points of the code.
Do we really want to put extra checks on MPI calls?

include/dlaf/communication/error.h
rasolca: For MPI errors I would go for MPI_Abort.

include/dlaf/cuda/error.h
```diff
@@ -0,0 +1,39 @@
...
+  if (err != cudaSuccess) {
+    std::printf("[CUDA ERROR] %s:%d: '%s'\n", file, line, cudaGetErrorString(err));
+    // std::abort();
```
rasolca, line 30: I would use std::abort or std::terminate here.

teonnik:
Changes:

- use `source_location` 
- `std::terminate()` on CUDA error
- `MPI_Abort()` on if MPI handler is set to return errors

 Note, if `DLAF_CUDA_CALL` terminates within an MPI application, MPI_Abort is not called. This is essentially the same as for our asserts. I'm not sure if that will cause issues in practice, I'm inclined to leave it as it is until/if we hit a problem.

The `DLAF_MPI_CALL` macro remains even if it's usage is quite limited. In particular, we don't have to use it for all MPI calls. It is only relevant when the handler is set to `MPI_ERRORS_RETURN`.

include/dlaf/communication/error.h
```diff
@@ -11,36 +11,36 @@
 #pragma once
 
 /// @file
+//
+#include "dlaf/common/utils.h"
+
+#include <mpi.h>
```
albestro, line 17: Following [coding style](https://github.com/eth-cscs/DLA-Future/blob/master/misc/coding.md#includes), these headers must come after the C++ standard library ones.

include/dlaf/communication/error.h
```diff
@@ -11,36 +11,36 @@
...
 ///
-/// - `MPI_ERRORS_ARE_FATAL` (the default ?): causes MPI to abort
+/// - `MPI_ERRORS_ARE_FATAL` (the default): causes MPI to abort
 /// - `MPI_ERRORS_RETURN`: causes MPI to return an error values instead of aborting
 ///
 /// This function is only relevant when the error handler is `MPI_ERRORS_RETURN`. To set the error
 /// handler use `MPI_Errhandler_set`.
 ///
-inline void mpi_call(int err, char const* file, int line) {
+inline void mpi_call(int err, common::internal::source_location const& info) {
```
albestro, line 33: I see it was already east-const, so it is not your choice. I'd like to know your opinion, @teonnik @rasolca about adding (or not) a rule to the coding style document for trying to keep uniform the codebase.

include/dlaf/cuda/error.h
```diff
@@ -0,0 +1,41 @@
...
+/// @file
+
+#include "dlaf/common/utils.h"
+
+#ifdef DLAF_WITH_CUDA
+#include <cuda_runtime.h>
+#endif
+
+#include <exception>
+#include <iostream>
```
albestro, line 22:
```suggestion
#include <exception>
#include <iostream>

#ifdef DLAF_WITH_CUDA
#include <cuda_runtime.h>
#endif

#include "dlaf/common/utils.h"
```

include/dlaf/communication/error.h
rasolca: The code base is mainly using west-const. (as all the example in the coding style even if it is not explicit.)
@albestro I think the coding style needs a full review.

teonnik: Regarding east-west const, I will use west-const as it is the style used in the rest of the code. There is a clang-format option in the works : https://reviews.llvm.org/D69764

teonnik:
bors try

Issue #160, opened 2020-03-10
Cholesky mini-app depends on code in the test folder


albestro:
Currently, the Cholesky mini-app can not be built if DLAF_WITH_TEST=off.

albestro:
This is blocking for #26.

Issue #164, opened 2020-03-17
Benchmark: Cholesky factorization mini-app (pre)


albestro:
We want to collect some preliminary results about performances of the current state of the Cholesky factorization mini-app (branch `miniapp_cholesky_random`).

It is mainly a check that we are on the right track compared to the prototype results.

The purpose of this issue is to decide and specify how the benchmark will be carried out.

ToDo & Topics:
- [x] Define configurations
- [x] Define command line
- [x] Is MPI pool enabled?
- [ ] Column vs Tile layout?
- [x] Recover build options for best HPX performances

albestro:
### Configurations
- matrix_size: [10240, 20480, 40960]
- block_size: [512]
- nodes: [1,2,4,8,16,32,64,96,128,192,256]

1 task per node
36 cores per task

Questions:
- [ ] block_size 256?

### Command Line
```bash
srun
-n ${nodes}
-c 36
miniapp_cholesky
--matrix-size {matrix_size}
--block-size {block_size}
--grid-rows {r}
--grid-cols {c}
--nruns 6
--check-result
--hpx:use_process_mask
```

Questions:
- [ ] Hyperthreading?
- [ ] Grid Size? Squarest one with priority on rows? Test all combinations?

albestro:
### HPX Build

I recovered a note that I took a few months ago talking with @biddisco about building HPX for the best performance.

```
# If you want to benchmark an application,
# it would be worth disabling some extra things
# (some features will be disabled, e.g. hpx::cout)

-DHPX_WITH_NETWORKING:BOOL=OFF
-DHPX_WITH_LOGGING:BOOL=OFF
-DHPX_WITH_IO_COUNTERS:BOOL=OFF
-DHPX_WITH_PARCELPORT_ACTION_COUNTERS:BOOL=OFF
-DHPX_WITH_PARCEL_PROFILING:BOOL=OFF
-DHPX_WITH_STACKTRACES:BOOL=OFF
-DHPX_WITH_THREAD_CUMULATIVE_COUNTS:BOOL=OFF
-DHPX_WITH_THREAD_IDLE_RATES:BOOL=OFF
-DHPX_WITH_THREAD_MANAGER_IDLE_BACKOFF:BOOL=OFF
-DHPX_WITH_THREAD_STEALING_COUNTS:BOOL=OFF
-DHPX_WITH_VERIFY_LOCKS:BOOL=OFF
[...]

Issue #117, opened 2019-12-12
Implement utility function for checking cholesky factorization in the miniAPP


albestro:
#116 requires a utility function for checking the result of the Cholesky factorization.

albestro:
This task requires a solution to #148

PR #197, opened 2020-04-20
Choleksy Factorization miniapp: migration to random values input

### LINKED ISSUE #116: Miniapp for Cholesky Factorization
Implement a miniapp useful for benchmarking of the Cholesky Factorization (using the one from the prototype as reference).

albestro:
Close #116, #117 

Migration from analytical to random values input for Cholesky Factorization Miniapp.

This opens to the benchmarks since it does not suffer of the denormalized values performance issue. Indeed, a revision of this work in progress has been used to produce benchmark results reported in #164.


albestro:
It still has some `TODO` inside and work in progress, but moreover it has the limitation that it works just with matrices which size is multiple of tile size.

miniapp/miniapp_cholesky.cpp
```diff
@@ -180,14 +194,270 @@ int main(int argc, char** argv) {
...
+      matrix(tile_wrt_global).then(hpx::util::unwrapping([tl_index](auto&& tile) {
+        for (int j = 0; j < tile.size().cols(); ++j)
+          for (int i = 0; i < tile.size().rows(); ++i) {
+            GlobalElementIndex element_wrt_global{tl_index.row() + i, tl_index.col() + j};
+            if (element_wrt_global.row() < element_wrt_global.col())
+              tile(TileElementIndex{i, j}) = 0;
+          }
+      }));
+    }
+  }
```
rasolca, line 287: I would simplify this in the following way:
```
loop over i_local
  convert i_local to i_global
  j_global = i_global
  if j_global in this rank:
    convert j_global to j_local
    schedule a task that performs lapack laset with m=row_tile_size-1 n=col_tile_size-1 and ptr=tile.ptr(0,1)
```

miniapp/miniapp_cholesky.cpp
```diff
@@ -180,14 +194,270 @@ int main(int argc, char** argv) {
...
+///
+/// 1. Compute the max norm of the original matrix
+/// 2. Compute the difference between the original and the computed matrix using the factorization
+/// 3. Compute the max norm of the difference
+/// 4. Evaluate the correctness of the result using the ratio between the two matrix max norms
+///
+/// Prints a message with the ratio and a note about the error:
+/// "":        check ok
+/// "ERROR":   error is high, there is an error in the factorization
+/// "WARNING": error is slightly high, there should be an error in the factorization
```
rasolca, line 420:
```suggestion
/// "WARNING": error is slightly high, there can be an error in the factorization
```

albestro:
Close #160 
This does not depend anymore on code from `DLAF_test`

albestro:
Last refinements undergoing I will complete soon.

Apart from them, and before opening this PR for a complete review, there are few aspects which I'd like to highlight (they may be addressed separately in the future, in case of needs).

1. About matrix initialization in the miniapp: do we prefer to initialize it with random values and then copy it when needed, or do we prefer to initialize it randomly always with the same seed?
2. There may be room for lowering the memory usage of `cholesky_diff` on `partial_result`, but it may complicate a bit the logic for reducing them over different ranks.
3. @teonnik pointed out a potential problem with wrapping `std::vector` (see issue #221)

I'm willing to get a feedback from you (@rasolca @teonnik @Sely85)

albestro:
bors try

albestro:
I merged the new master in which it is available `norm(max)` for the specific variant I need in this PR.

> [...] there are few aspects which I'd like to highlight (they may be addressed separately in the future, in case of needs).
> 
>     1. About matrix initialization in the miniapp: do we prefer to initialize it with random values and then copy it when needed, or do we prefer to initialize it randomly always with the same seed?
> 
>     2. There may be room for lowering the memory usage of `cholesky_diff` on `partial_result`, but it may complicate a bit the logic for reducing them over different ranks.
> 
>     3. @teonnik pointed out a potential problem with wrapping `std::vector` (see issue #221)
> 
> 

I'm willing to get a feedback from you @rasolca @teonnik @Sely85

teonnik:
> 1. About matrix initialization in the miniapp: do we prefer to initialize it with random values and then copy it when needed, or do we prefer to initialize it randomly always with the same seed?

Assuming the values in the matrix don't influence the performance of Cholesky much (which I'm not 100% sure about), I don't mind leaving it as it is for now... However setting random values is generally somewhat slow, it seems preferable to set them once at the start.  

> 2. There may be room for lowering the memory usage of `cholesky_diff` on `partial_result`, but it may complicate a bit the logic for reducing them over different ranks.

I think it's OK to leave that for a separate PR. I presume this is a part of the code will not be included in benchmarks.

> 3. @teonnik pointed out a potential problem with wrapping `std::vector` (see issue #221)

This is not an urgent issue, I'm OK with leaving things as they are for now in order to unblock other PRs.

albestro:
> > 1. About matrix initialization in the miniapp: do we prefer to initialize it with random values and then copy it when needed, or do we prefer to initialize it randomly always with the same seed?
> 
> Assuming the values in the matrix don't influence the performance of Cholesky much (which I'm not 100% sure about), I don't mind leaving it as it is for now... However setting random values is generally somewhat slow, it seems preferable to set them once at the start.

I've just changed this: now the miniapp initializes the matrix reference with random values (just one time) and then at each run it copies the values from it (once for the running the algorithm and an optional one if the results are checked).

> > 1. There may be room for lowering the memory usage of `cholesky_diff` on `partial_result`, but it may complicate a bit the logic for reducing them over different ranks.
> 
> I think it's OK to leave that for a separate PR. I presume this is a part of the code will not be included in benchmarks.

Yes, it does not affect the benchmark results since it is in the part for checking results.

albestro:
bors try

teonnik:
Note that this PR also fixes build issues when DLAF_WITH_TEST=OFF as the DLAF_test target used in the miniapp is no longer needed.

albestro:
> Note that this PR also fixes build issues when DLAF_WITH_TEST=OFF as the DLAF_test target used in the miniapp is no longer needed.

Yes! The issue had been already linked for being closed 😉

Sely85:
>> 1. About matrix initialization in the miniapp: do we prefer to initialize it with random values and then copy it when needed, or do we prefer to initialize it randomly always with the same seed?
>>
> I've just changed this: now the miniapp initializes the matrix reference with random values (just one time) and then at each run it copies the values from it (once for the running the algorithm and an optional one if the results are checked).

This looks great! I do prefer the first option as well.


>> 2. There may be room for lowering the memory usage of cholesky_diff on partial_result, but it may complicate a bit the logic for reducing them over different ranks.
>>
> Yes, it does not affect the benchmark results since it is in the part for checking results.

Perfect: if it does not affect benchmark results, I think it can be left as it is to avoid complicating the logic. 


>> 3. @teonnik pointed out a potential problem with wrapping std::vector (see issue #221)
>>
> This is not an urgent issue, I'm OK with leaving things as they are for now in order to unblock other PRs.

If it's not urgent, I agree not to modify anything for now.

PR #202, opened 2020-04-28
Improve and fix safe integer casting helper functions

albestro:
Highlights:

- extend `to_signed` and `to_unsigned`: they now accept any integral type as input, while as output they accept just a signed or unsigned integral type, respectively;
- add `integral_cast` that acts as a dispatcher for the previous two, and can convert from/to any integral type.

All these helper functions are just simple `static_cast`, but they have runtime checks in case DLAF_ASSERT_MODERATE level is enabled.

The unit test `test_types` has been introduced.

### DIFF
--- a/test/unit/test_types.cpp
+++ b/test/unit/test_types.cpp
@@ -1,0 +1,158 @@
+// force enable DLAF_ASSERT_MODERATE, otherwise cast checks cannot be tested
+#ifndef DLAF_ASSERT_MODERATE_ENABLE
+#define DLAF_ASSERT_MODERATE_ENABLE
+#endif
+
+#include "dlaf/types.h"
+
+#include <limits>
+
+#include <gtest/gtest.h>
+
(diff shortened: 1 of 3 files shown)


albestro:
bors try

test/unit/test_types.cpp
```diff
@@ -0,0 +1,168 @@
...
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+// force enable DLAF_ASSERT_MODERATE, otherwise cast checks cannot be tested
+#ifndef DLAF_ASSERT_MODERATE_ENABLE
+#define DLAF_ASSERT_MODERATE_ENABLE
+#endif
```
rasolca, line 14: This approach is dangerous if some of the function used are included in the compiled library.
Just enable death tests if DLAF_ASSERT_MODERATE_ENABLE is true.

PR #203, opened 2020-04-28
Improve computeCoords API

albestro:
Adapt `computeCoords` API
- accepting `Size2D` as per https://github.com/eth-cscs/DLA-Future/pull/183#issuecomment-613978093,
- plus compile-time ordering selection as per need of https://github.com/eth-cscs/DLA-Future/pull/199#discussion_r415756795.

### DIFF
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -178,6 +178,18 @@ public:
   }
 };
 
+template <class IndexT, class Tag, class LinearIndexT>
+Index2D<IndexT, Tag> computeCoordsRowMajor(LinearIndexT index,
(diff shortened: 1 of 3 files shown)


albestro:
In this PR I target just the API.

See #190 for guards and preconditions.

albestro:
About the `default` case in the `switch`.

It is useful because some compilers may complain with `warning: control reaches end of non-void function [-Wreturn-type]` if we do not add it.

Moreover, it does not hurt because recently we added the switch `-Wswitch-enum` that checks that all cases are considered, even in presence of the `default` case.

albestro:
bors try

PR #208, opened 2020-04-30
Avoid empty variadic macro (conform to the standard)

### LINKED ISSUE #200: Build problem on GCC 7.4.0
`DLAF_ASSERT*` mechanisms use variadic macro.

They are intended to support also empty messages, so with an empty variadic macro, but this is a compiler extensions.

This extension is used also in GoogleTest, and we already had to disable warnings for Clang compilers with `-Wno-gnu-zero-variadic-macro-arguments`

https://github.com/eth-cscs/DLA-Future/blob/2af76f28e3e6974feec33d7c4c020f8826afbe9a/cmake/DLAF_AddTargetWarnings.cmake#L46-L50

The problem with GCC compiler is the different management of the `-pedantic-errors` flag, since it behaves differently between versions (see https://godbolt.org/z/YDev65).

albestro:
Close #200 

Avoid empty variadic macro parameters by subsuming the first mandatory parameter into it.

Thanks @teonnik for the suggestion.

This PR is not going to be merged directly into `master`. instead it will be merged into another PR where the problem shown up. It will be part of it, and eventually merged to master.

### DIFF
--- a/include/dlaf/common/assert.h
+++ b/include/dlaf/common/assert.h
@@ -30,6 +30,7 @@
   if (!(condition)) {                                                \
     std::cerr << "[ERROR] " << origin << std::endl                   \
               << dlaf::common::concat(#condition, ' ', __VA_ARGS__); \
+    std::cerr.flush();                                               \
     std::terminate();                                                \
   }
 
--- a/include/dlaf/common/assert.h
+++ b/include/dlaf/common/assert.h
@@ -23,6 +23,8 @@
 /// by the parameter and an optional custom message composed by concatenating all extra parameters.
 /// Message composition is lazily evaluated and it will not add any overhead if the condition is true.
 ///
+/// No newline is appended to the given message, which cannot be empty.
+///
 /// **This check cannot be disabled**
 #define DLAF_CHECK_WITH_ORIGIN(category, origin, condition, ...)     \
   if (!(condition)) {                                                \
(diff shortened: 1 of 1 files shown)


albestro:
bors try

albestro:
I had to introduce a workaround, because the `DLAF_CHECK_WITH_ORIGIN` macro needs to access the condition and it is not possible to extract it from the `__VA_ARGS__`. So I need it as a plain parameter.

The workaround is: the newline character `\n` will be the last character of the message (as it was before), but now it is passed as part of the parameter message. In this way the variadic macro will always have at least this character and won't be empty in any case.

bors try

include/dlaf/common/assert.h
```diff
@@ -24,17 +24,17 @@
...
 /// **This check cannot be disabled**
-#define DLAF_CHECK_WITH_ORIGIN(category, origin, condition, ...)                    \
-  if (!(condition)) {                                                               \
-    std::cerr << "[ERROR] " << origin << std::endl                                  \
-              << dlaf::common::concat(#condition, ' ', ##__VA_ARGS__) << std::endl; \
-    std::terminate();                                                               \
+#define DLAF_CHECK_WITH_ORIGIN(category, origin, condition, ...)     \
+  if (!(condition)) {                                                \
+    std::cerr << "[ERROR] " << origin << std::endl                   \
+              << dlaf::common::concat(#condition, ' ', __VA_ARGS__); \
```
rasolca, line 30: Since you removed `endl`, I would add a `std::err.flush()` to be sure that the output is printed.

`cerr` should not be buffered, but it is better to be on the safe side if there is no performance penalty as in this case.

albestro:
This should be the last one before merging.

bors try

PR #216, opened 2020-05-06
Workaround for custom operator<< in gtest

### LINKED ISSUE #215: GoogleTest problem using some custom operator<< overload
In tests, custom `operator<<` overloading is not always selected resulting in not very useful error message in case of tests failure.

An example is with `Index2D` and `Size2D`, for which it is not able to select our custom one that prints
```
(13, 26)
```
and it selects a fallback one provided by gtest that prints
```
8-byte object <0D-00 00-00 1A-00 00-00>
```

It seems that GTest printer does not manage correctly the case where an `operator<<` overload exists for a base class.

The workaround is to not rely onto the base class one and always specify the friend function for each class.

At the moment, I don't know if we have other class using the same mechanism that may suffer from the same problem.

albestro:
Close #215 

This is the easy workaround I was talking about in the issue.

At the moment `Index2D` and `Size2D` seem to be the only ones using base class `operator<<`.

### DIFF
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -132,6 +132,10 @@ public:
     assert(internal::basic_coords<IndexT>::isValid());
     return rows() == 0 || cols() == 0;
   }
+
+  friend std::ostream& operator<<(std::ostream& out, const Size2D& index) {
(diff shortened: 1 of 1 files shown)


albestro:
bors try

PR #217, opened 2020-05-06
Basic arithmetic for ptrdiff_t

albestro:
This PR implements `sum` and `mul` for `ptrdiff_t` arithmetic.

Minor refactoring was required, `integral_cast` has been used instead of trivial `static_cast` and tests for `ptrdiff_t` have been added.

### DIFF
--- a/include/dlaf/util_math.h
+++ b/include/dlaf/util_math.h
@@ -65,7 +65,7 @@ namespace size_t {
 /// @brief Perform the sum on integer types in size_t arithmetic
 ///
 /// It casts operands to std::size_t type and then performs the sum
-/// @tparam TA and @tparam TB must be an integer types (signed or unsigned)
+/// @tparam TA and @tparam TB must be integer types (signed or unsigned)
 /// @return @p a + @p b
 /// @pre @p a >=0 and @p b >= 0
 /// @pre it must be possible to store the result in std::size_t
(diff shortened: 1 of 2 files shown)


include/dlaf/util_math.h
```diff
@@ -41,48 +44,81 @@ constexpr auto ceilDiv(const IntType num, const IntType den)
...
+/// It casts operands to @tparam ArithmeticT type and then performs the given binary operation @p op
 /// @tparam TA and @tparam TB must be an integer types (signed or unsigned)
 /// @return @p a @p op @p b
-/// @pre @p a >=0 and @p b >= 0
+/// @pre @p a and @p b can be stored in the type @tparam ArithmeticT
+/// @pre it must be possible to store the result in @tparam ArithemticT
 #ifdef DLAF_DOXYGEN
-template <typename TA, typename TB, typename BinaryOp>
+template <class ArithmeticT, class TA, class TB, class BinaryOp>
 constexpr std::size_t generic_integer_op(const TA a, const TB b, BinaryOp op);
```
rasolca, line 58: This should be updated as well, (or removed if the enable_if is integrated in the template...)

albestro: Good catch!

I opted for removing the `std::enable_if`, since the check about types is already performed on both inputs and output type by using `integral_cast`. In this way also the documentation has been unified.

albestro:
bors try

include/dlaf/util_math.h
```diff
@@ -41,48 +44,75 @@ constexpr auto ceilDiv(const IntType num, const IntType den)
...
+}
+
+}
+
+namespace ptrdiff_t {
+
+/// @brief Perform the sum on integer types in ptrdiff_t arithmetic
+///
+/// It casts operands to std::ptrdiff_t type and then performs the sum
+/// @tparam TA and @tparam TB must be an integer types (signed or unsigned)
```
rasolca, line 96:
```suggestion
/// @tparam TA and @tparam TB must be integer types (signed or unsigned)
```
same below

PR #219, opened 2020-05-07
Refactor code to use iterators

teonnik:
- Use snake_case for free function iterator_range2d


teonnik:
bors try

teonnik:
bors try

teonnik:
The standard is moving to using snake_case for pretty much everything with Concepts. Here I changed it to snake case because I noticed that free functions are in snake case in the rest of the code base.

PR #220, opened 2020-05-07
Add substraction, addition and construction utils

teonnik:
These provide us with some utilities to conveniently add, substract and cast between Index and Size types.

### DIFF
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -91,7 +91,7 @@ protected:
     return row_ == rhs.row_ && col_ == rhs.col_;
   }
 
-  /// @return true if `this` and `rhs` have the different row or column.
+  /// @return true if `this` and `rhs` have different row or column.
   bool operator!=(const basic_coords& rhs) const noexcept {
     return !operator==(rhs);
   }
@@ -137,7 +137,7 @@ public:
     return BaseT::operator==(rhs);
   }
 
-  /// @return true if `this` and `rhs` have the different row or column.
+  /// @return true if `this` and `rhs` have different row or column.
   bool operator!=(const Size2D& rhs) const noexcept {
     return BaseT::operator!=(rhs);
   }
@@ -182,7 +182,7 @@ public:
     return BaseT::operator==(rhs);
   }
 
-  /// @return true if `this` and `rhs` have the different row or column.
+  /// @return true if `this` and `rhs` have different row or column.
   bool operator!=(const Index2D& rhs) const noexcept {
     return BaseT::operator!=(rhs);
   }
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -99,6 +99,14 @@ public:
     return out << "(" << index.row_ << ", " << index.col_ << ")";
   }
 
+  friend basic_coords operator+(basic_coords lhs, basic_coords rhs) noexcept {
+    return basic_coords(lhs.row_ + rhs.row_, lhs.col_ + rhs.col_);
+  }
+
+  friend basic_coords operator-(basic_coords lhs, basic_coords rhs) noexcept {
+    return basic_coords(lhs.row_ - rhs.row_, lhs.col_ - rhs.col_);
+  }
+
 protected:
   IndexT row_;
   IndexT col_;
@@ -108,6 +116,7 @@ template <class Coords2DType>
 Coords2DType transposed(const Coords2DType& coords) {
   return {coords.col_, coords.row_};
 }
+
 }
 
 /// A strong-type for 2D sizes
@@ -115,8 +124,12 @@ Coords2DType transposed(const Coords2DType& coords) {
 /// @tparam Tag for strong-typing
 template <typename IndexT, class Tag>
 class Size2D : public internal::basic_coords<IndexT> {
+  using base = internal::basic_coords<IndexT>;
+
 public:
-  using internal::basic_coords<IndexT>::basic_coords;
+  using base::basic_coords;
+
+  Size2D(base const& coords) noexcept : base(coords) {}
 
   IndexT rows() const noexcept {
     return internal::basic_coords<IndexT>::row_;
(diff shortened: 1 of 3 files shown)


teonnik:
bors try

haampie:
sigh... a brand new slurm issue?

bors try

teonnik:
Seems like it, tests pass locally.

haampie:
bors try

teonnik:
I was considering that too but it didn't cover the following situation in `set(Matrix, ElementGetter)` :
`GlobalElementIndex{i + tl_index.row(), j + tl_index.col()}` (dlaf/util_matrix.h) where we intend to add a TileElementIndex to a GlobalElementIndex. If that's not expected to be common, I can implement the operations as you have written them. Note that we haven't completely abandoned type safety with the current implementation as we still have to explicitly construct the desired type but it's somewhat loosened.

rasolca:
Good catch, I didn't consider it. 
Anyway it is a special case in which you want to avoid each time the computation of the element index from the tile index and the tile-element index precomputing the index of the top-left element of the tile.
What about rewriting it as `tl_index + GlobalElementSize(i, j)`?

teonnik:
I made the requested changes. In the meantime, I noticed an inconsistency, we are strict about type safety for `operator+/-` but not for `operator==/!=`. Basically, we can do `LocalTileSize(4, 6) == GlobalTileIndex(4, 6)` and this is valid currently. 

Should we change `operator==/!=` too?

On another point, I don't think think asserting whether an index is positive or negative in the constructor is the right approach... negative indices are sometimes useful (e.g. your tile at (i, j) depends on tile (i - i_offset, j - j_offset)), it seems better to me if this is asserted in the algorithms themselves (i.e. DLAF_ASSERT(idx.isValid())).

rasolca:
> I made the requested changes. In the meantime, I noticed an inconsistency, we are strict about type safety for `operator+/-` but not for `operator==/!=`. Basically, we can do `LocalTileSize(4, 6) == GlobalTileIndex(4, 6)` and this is valid currently.
> 
> Should we change `operator==/!=` too?

Yes, this is a good idea.

> On another point, I don't think think asserting whether an index is positive or negative in the constructor is the right approach... negative indices are sometimes useful (e.g. your tile at `(i, j)` depends on tile `(i - i_offset, j - j_offset)`), it seems better to me if this is asserted in the algorithms themselves (i.e. `DLAF_ASSERT(idx.isValid())`) instead of in the `Index2D` constructor. This will allow us to better handle situation where an index is OK to be negative.

In principle I agree, but I don't completely understand your example...
In this case you can use operator- unless offsets can be either positive or negative. Is this the case?
Anyway this example shows a Size2D that can be negative not an Index2D (unless we allow Index + Index -> Index (which would represent the operation Index + (Index - Index(0, 0)))

teonnik:
> In principle I agree, but I don't completely understand your example...
In this case you can use operator- unless offsets can be either positive or negative. Is this the case?
Anyway this example shows a Size2D that can be negative not an Index2D (unless we allow Index + Index -> Index (which would represent the operation Index + (Index - Index(0, 0)))

I should have explained a little bit better. In TSGEMM, I have a situation where a tile with index (i, j) depends on a previous tile with index (i - i_d, j - j_d). If the latter index is negative, that means that there is no dependent tile, if it is positive, there is a dependent tile. For this particular situation, the index `(i - i_d, j - j_d)` both gives me the index of the dependent tile and the means to test if the tile exists by calling `isValid()` on it.

teonnik:
I now changed `operator==/!=` to conform to `Index2D` / `Size2D` and refactored some of the code. We need to make a decision on whether we want to assert `isValid()` within Index2D's constructor and we will be good to go with this PR.

rasolca:
YThanks iit is much clearer now!
I agree for removal, since you can already create an invalid index using the default constructor and the index is already checked in the algorithms.

teonnik:
bors try

test/unit/common/test_index2d.cpp
```diff
@@ -59,18 +59,6 @@ TYPED_TEST(Index2DTest, ConstructorFromArray) {
...
-      std::vector<std::pair<TypeParam, TypeParam>>{std::make_pair(-10, -10), std::make_pair(-1, -1),
-                                                   std::make_pair(-15, 3), std::make_pair(8, -7)};
-
-  for (auto& test : index_tests) {
-    EXPECT_THROW(Index2D<TypeParam>(test.first, test.second), std::invalid_argument);
-    EXPECT_THROW(Index2D<TypeParam>(std::array<TypeParam, 2>{test.first, test.second}),
-                 std::invalid_argument);
-  }
-}
-
```
rasolca, line 73: You should keep this tests:
You should check that you can construct objects with negative indices, that the values are correctly set, and then that is_valid() on those objects returns false.

rasolca:
bors try

include/dlaf/common/index2d.h
```diff
@@ -156,24 +152,15 @@ class Size2D : public internal::basic_coords<IndexT> {
...
-  Index2D(base const& coords) noexcept : base(coords) {}
-
-  /// Create a valid 2D coordinate
-  /// @param row index of the row
-  /// @param col index of the column
-  /// @throw std::invalid_argument if row < 0 or col < 0
-  Index2D(IndexT row, IndexT col) : internal::basic_coords<IndexT>(row, col) {
-    if (!internal::basic_coords<IndexT>::isValid())
-      throw std::invalid_argument("indices are not valid (negative).");
-  }
```
albestro, line 159: @teonnik 

I've just realized that it was possible to create an Index2D with "invalid" coords via `Index2D(base const& coords) noexcept` but it was checked to be valid via `Index2D(IndexT row, IndexT col)`.

I thought it was worth mention it, since the behavior may have changed (maybe it was neither tested).

teonnik: I removed `Index2D(base const& coords) noexcept` constructor later in the PR.

PR #224, opened 2020-05-08
Add MaxNorm

albestro:
This PR aims at adding the structure for MaxNorm function with the implementation for lower triangular matrices.


test/unit/utility/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+# Distributed Linear Algebra with Future (DLAF)
+#
+# Copyright (c) 2018-2019, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+add_subdirectory(mc)
```
teonnik, line 11: We are adding a new file, 10 lines of license information for 1 line of code. Wouldn't it be simpler to add the folder from the top CMakeLists.txt file?

test/unit/utility/mc/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+# Distributed Linear Algebra with Future (DLAF)
+#
+# Copyright (c) 2018-2019, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+DLAF_addTest(test_norm SOURCES test_norm.cpp USE_MAIN MPIHPX MPIRANKS 6)
```
teonnik, line 11: Can we add the test in the CMake above instead of having a new file?

test/unit/utility/mc/test_norm.cpp
```diff
@@ -0,0 +1,109 @@
...
+          if (blas::Uplo::Lower != uplo)
+            continue;
+
+          auto el_L = [size, value](const GlobalElementIndex& index) {
+            if (GlobalElementIndex{size.rows() - 1, 0} == index)
+              return value;
+            return TypeParam(0);
+          };
+
+          test_norm<TypeParam>("lower triangular", norm_expected, el_L, comm_grid, mat, norm, uplo);
```
rasolca, line 96:
```suggestion
          test_norm<TypeParam>("Lower", norm_expected, el_L, comm_grid, mat, norm, uplo);
```
It is a symmetric/Hermitian matrix stored in the lower triangular part...

include/dlaf/utility/mc.h
```diff
@@ -0,0 +1,62 @@
...
+  ///   - With @p uplo == blas::uplo::Lower
+  ///   @pre @p A must be square matrix, A.size().rows() == A.size().cols()
+  ///   - With @p uplo == blas::uplo::Upper
+  ///   @note not yet implemented
+  ///   - With @p uplo == blas::uplo::General
+  ///   @note not yet implemented
+  /// - With @p norm_type = lapack::Norm::{One, Two, Inf, Fro}
+  /// @note not yet implemented
+  ///
+  /// .
```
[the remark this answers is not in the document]
albestro: Good question!

https://www.doxygen.nl/manual/lists.html

> You can end a list by starting a new paragraph or by putting a dot (.) on an empty line at the same indentation level as the list you would like to end.

I used it in order to render correctly the doxygen documentation

image

albestro:
> The test coverage is not sufficient...
> I would take a random matrix, compute the norm, check that is in the range of the random numbers, set an element (you can iterate over a set of coords) to `el/|el|*(norm + delta)` compute the norm again and check that the new norm is `norm+delta` if the element was in the diagonal or lower part, or `norm` otherwise.
> This test can be then easily generalized to upper and general matrices.

I refactored a bit the test trying to address your review. I'm still not really satisfied, I look forward to get a feedback from you @rasolca if you have any suggestion on how to improve it 😉

albestro:
@teonnik I like your suggestion about improving the folder structure of the algorithms.

Would be ok for you if I open a separate PR for this so, as soon as we agree on the structure, we can fix in the same way for all existing algorithms (e.g. factorization, solver, utility)?

Instead, for what concerns the name of the category for `norm`, which is currently `utility`, do you have any suggestion?

rasolca:
> Instead, for what concerns the name of the category for `norm`, which is currently `utility`, do you have any suggestion?

Lapack calls this types of functions "auxiliary routines". We can use `auxiliary`.

test/unit/auxiliary/mc/test_norm.cpp
```diff
@@ -0,0 +1,164 @@
...
+        Index2D src_rank_index(std::max(0, comm_grid.size().rows() - 1),
+                               std::min(1, comm_grid.size().cols() - 1));
+        Distribution distribution(size, block_size, comm_grid.size(), comm_grid.rank(), src_rank_index);
+        Matrix<TypeParam, Device::CPU> matrix(std::move(distribution));
+
+        for (const auto& uplo : blas_uplos) {
+          if (blas::Uplo::Lower != uplo)
+            continue;
+
+          dlaf::matrix::util::set_random(matrix);
```
rasolca, line 110: for uplo == Lower/Upper the matrix should be either symmetric or hermitian, (i.e. for complex matrices all the diagonal elements should be either real, or purely imaginary)

test/unit/auxiliary/mc/test_norm.cpp
```diff
@@ -0,0 +1,164 @@
...
+            const GlobalElementIndex index{0, 0};
+            const NormT<TypeParam> norm_expected = matrix.size().isEmpty() ? 0 : std::abs(new_value);
+            norm_current = norm_expected;
+
+            set_and_test(comm_grid, matrix, index, new_value, norm_expected, norm_type, uplo);
+          }
+
+          // BOTTOM LEFT
+          {
+            new_value += 100;
```
rasolca, line 134: what about using a non real number for complex matrices?

include/dlaf/lapack_tile.tpp
```diff
@@ -8,6 +8,36 @@
...
+template <class T, Device device>
+dlaf::BaseType<T> lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag,
+                        const Tile<T, device>& a) noexcept {
+  switch (uplo) {
+    case blas::Uplo::Lower:
+      DLAF_ASSERT(a.size().rows() >= a.size().cols(), "Not valid ", a.size());
+      break;
+    case blas::Uplo::Upper:
+      DLAF_ASSERT(a.size().rows() <= a.size().cols(), "Not valid ", a.size());
+      break;
```
teonnik, line 27: With the new ASSERT mechanism, the tiles will be printable, we can just place them as a parameter, (e.g. `a` instead of `a.size()`). It might make sense to change the message to make it more as a sentence instead of having values in between strings, for example

"Not valid <some_val>" to just "Not valid!"

Although in this case, I wouldn't mind omitting the message altogether as the condition is quite clear.

test/unit/auxiliary/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+# Distributed Linear Algebra with Future (DLAF)
+#
+# Copyright (c) 2018-2019, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+add_subdirectory(mc)
```
teonnik, line 11: Perhaps we can move this to the top CMakeLists.txt?

test/unit/auxiliary/mc/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+# Distributed Linear Algebra with Future (DLAF)
+#
+# Copyright (c) 2018-2019, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+DLAF_addTest(test_norm SOURCES test_norm.cpp USE_MAIN MPIHPX MPIRANKS 6)
```
teonnik, line 11: same here

include/dlaf/auxiliary/norm/mc/norm_max_L.h
```diff
@@ -0,0 +1,107 @@
...
+  tiles_max.reserve(distribution.localNrTiles().rows() * distribution.localNrTiles().cols());
+
+  // for each tile in local (lower triangular), create a task that finds the max element in the tile
+  for (SizeType j_loc = 0; j_loc < distribution.localNrTiles().cols(); ++j_loc) {
+    const SizeType j = distribution.template globalTileFromLocalTile<Coord::Col>(j_loc);
+    const SizeType i_diag_loc = distribution.template nextLocalTileFromGlobalTile<Coord::Row>(j);
+
+    for (SizeType i_loc = i_diag_loc; i_loc < distribution.localNrTiles().rows(); ++i_loc) {
+      const SizeType i = distribution.template globalTileFromLocalTile<Coord::Row>(i_loc);
+
```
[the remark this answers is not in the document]
rasolca: Not really, it would be equivalent to add 
```
if (tile_wrt_global.row() < tile_wrt_global.col())
    continue;
```
but you will iterate on all the tiles of the matrix (adding also an if statement) instead of only half of them.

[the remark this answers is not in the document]
teonnik: > but you will iterate on all the tiles of the matrix (adding also an if statement) instead of only half of them.

Yes, that is the case. I was thinking more from a readability point of view as the cost of these is insignificant compared to the dataflow that follows.

> Not really, it would be equivalent to add

The comment about mentioning the `local` part threw me off a little bit. Perhaps then this would work then :

```
for (auto tile_wrt_local : iterate_range2d(dist.localNrTiles())) {
  auto tile_wrt_global = globalTileIndex(tile_wrt_local);
   // If the tile is not in the lower triangular part, skip it
   if (tile_wrt_global.row() < tile_wrt_global.col())
    continue;
  
   bool is_diag = tile_wrt_global.row() == tile_wrt_global.col();
  ...
}
```
or sth similar.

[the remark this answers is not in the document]
rasolca: Good point, the comment is incorrect... Yes this would work.

include/dlaf/auxiliary/norm/mc/norm_max_L.h
```diff
@@ -0,0 +1,107 @@
...
+  DLAF_ASSERT_SIZE_SQUARE(matrix);
+  DLAF_ASSERT_BLOCKSIZE_SQUARE(matrix);
+
+  if (GlobalElementSize{0, 0} == matrix.size())
+    return {0};
+
+  vector<hpx::future<NormT>> tiles_max;
+  tiles_max.reserve(distribution.localNrTiles().rows() * distribution.localNrTiles().cols());
+
+  // for each tile in local (lower triangular), create a task that finds the max element in the tile
```
rasolca, line 57: This comment may be misleading.
Maybe: `for each local tile in the (global) lower triangular matrix`

include/dlaf/auxiliary/norm/mc/norm_max_L.h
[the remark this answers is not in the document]
albestro: I agree with you, I didn't use it because I didn't want to iterate over all tiles and skip. It was a premature optimization.

> EDIT: We can also define separate iterators for iterating over the lower/upper triangular parts of a region that avoid the costs of the if() statement and shorten the iteration range. Perhaps that is something to consider in the long run.

It is not urgent, but I think it would be a nice thing.

> The comment above mentioning the local part threw me off a little bit.

Yes, sorry it was misleading. I fixed it as @rasolca suggested 😉

include/dlaf/lapack_tile.tpp
albestro: I'll wait that the #231 gets merged 😉

include/dlaf/auxiliary/norm/mc/norm_max_L.h
```diff
@@ -0,0 +1,97 @@
...
+  // TODO unwrapping can be skipped for optimization reasons
+  NormT local_max_value = hpx::dataflow(unwrapping([](const auto&& values) {
+                                          if (values.size() == 0)
+                                            return std::numeric_limits<NormT>::min();
+                                          return *std::max_element(values.begin(), values.end());
+                                        }),
+                                        tiles_max)
+                              .get();
+  NormT max_value;
+  dlaf::comm::sync::reduce(0, comm_grid.fullCommunicator(), MPI_MAX, make_data(&local_max_value, 1),
```
albestro, line 89: I'm wondering if it would make sense to expose as parameter where the norm result will be computed (i.e. the root rank of the reduce).

What do you think @rasolca @teonnik ?

test/unit/auxiliary/CMakeLists.txt
albestro: About this comment and your next one, see https://github.com/eth-cscs/DLA-Future/pull/224#issuecomment-633860153

I've just opened an issue about this #236 (the title of the issue is not very straightforward, but it is about this).

test/unit/auxiliary/mc/test_norm.cpp
albestro: In order to address this I've just opened #237, waiting for that before fixing this.

include/dlaf/auxiliary/norm/mc/norm_max_L.h
teonnik: Yes, I think that makes sense. It will also be nice to be able to choose the result to be available on all processes (i.e. MPI_Allreduce).

include/dlaf/lapack_tile.tpp
albestro: This should have been addressed now 😉

test/unit/auxiliary/mc/test_norm.cpp
albestro: #237 has been embedded in this PR, and now that line has been changed to `set_random_hermitian`.

include/dlaf/auxiliary/norm/mc/norm_max_L.h
albestro: @rasolca @teonnik 

Do we want to proceed in this PR or do we want to address this later? In case I will open an issue, so we can also discuss how to change the API.

albestro:
bors try

include/dlaf/auxiliary/norm/mc/norm_max_L.h
teonnik: I think it's OK if we address MPI_Allreduce in a separate PR, just adding the rank on which we want the result to the parameter list of the function and propagating it to `dlaf::comm::sync::reduce` should be sufficient for now.

albestro:
bors try

there was a problem with the "caching" of the git repository, @haampie managed to clean the directory and he is going to disable the "caching" feature.

albestro:
I fixed warnings, but before testing this, I'd like wait that #240 gets merged (currently on hold for problems with CI).

The last round of review can be done in the meanwhile. 😉

albestro:
bors try

albestro:
bors try

include/dlaf/auxiliary/norm/mc/norm_max_L.h
albestro: I've just added the parameter 😉

rasolca:
bors try

PR #225, opened 2020-05-11
Phase out scheduled_executor

teonnik:
`scheduled_executor` will not exist in the upcoming HPX 1.5 . This PR changes how we handle `try/catch` to avoid it's usage and be compatible with both current and future HPX versions.

For more details: https://github.com/STEllAR-GROUP/hpx/issues/4535

### DIFF
--- a/include/dlaf/communication/executor.h
+++ b/include/dlaf/communication/executor.h
@@ -1,0 +1,33 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2019, ETH Zurich
+// All rights reserved.
+//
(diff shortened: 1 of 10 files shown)


include/dlaf/communication/executor.h
```diff
@@ -0,0 +1,33 @@
...
+inline bool mpi_pool_exists() {
+  using hpx::threads::executors::pool_executor;
+  try {
+    pool_executor("mpi");
+    return true;
+  }
+  catch (...) {
+    return false;
+  }
+}
```
albestro, line 30: I think it is a really nice idea, but I would suggest to put it in an `internal` namespace.

teonnik:
bors try

Issue #152, opened 2020-02-27
Incomplete tiles management


albestro:
Given a square matrix `m=3` and `mb=2` distributed over a 2x1 grid of ranks, identified as A and B, the matrix globally will be:

```
1 2 | 3     A A | A
4 5 | 6     A A | A
-------     -------
7 8 | 9     B B | B
```

And locally,

A:
```
1 2     3
4 5     6
```

B:
```
7 8     9
```

This means that we have to deal with incomplete tiles, and this will affect both computation and communication. For instance, in this case, we have to deal with all tile size combinations: 2x2, 2x1, 1x2 and 1x1.

This is relevant also for preconditions, in fact, some algorithms implementations may rely on the structure of the distribution, e.g. square tiles or square matrix. But the preconditions check that "generally" matrix and tiles are square, without assuring that tiles will always be complete.

It is not a problem related just to square matrix/tile or a particular distribution grid, it may affect all combinations of geometries of matrix, tiles and distribution grids.

PR #227, opened 2020-05-11
Add lapack tile wrappers for 'lange' and 'lantr'

albestro:
Add wrappers and related tests for:
- `lange`
- `lantr`

These functions will be used starting from `miniapp_cholesky` in #197

### DIFF
--- a/include/dlaf/lapack_tile.tpp
+++ b/include/dlaf/lapack_tile.tpp
@@ -8,6 +8,8 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 
+#include "dlaf/common/assert.h"
+
 template <class T, Device device>
 dlaf::BaseType<T> lange(lapack::Norm norm, const Tile<T, device>& a) noexcept {
   return lapack::lange(norm, a.size().rows(), a.size().cols(), a.ptr(), a.ld());
@@ -16,6 +18,16 @@ dlaf::BaseType<T> lange(lapack::Norm norm, const Tile<T, device>& a) noexcept {
 template <class T, Device device>
 dlaf::BaseType<T> lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag,
                         const Tile<T, device>& a) noexcept {
+  switch (uplo) {
+    case blas::Uplo::Lower:
+      DLAF_ASSERT(a.size().rows() >= a.size().cols(), "Not valid ", a.size());
+      break;
+    case blas::Uplo::Upper:
+      DLAF_ASSERT(a.size().rows() <= a.size().cols(), "Not valid ", a.size());
+      break;
+    case blas::Uplo::General:
+      break;
+  }
   return lapack::lantr(norm, uplo, diag, a.size().rows(), a.size().cols(), a.ptr(), a.ld());
 }
 
--- a/include/dlaf/lapack_tile.tpp
+++ b/include/dlaf/lapack_tile.tpp
@@ -9,12 +9,12 @@
 //
 
 template <class T, Device device>
-dlaf::BaseType<T> lange(lapack::Norm norm, const Tile<T, device>& a) {
+dlaf::BaseType<T> lange(lapack::Norm norm, const Tile<T, device>& a) noexcept {
   return lapack::lange(norm, a.size().rows(), a.size().cols(), a.ptr(), a.ld());
 }
 
 template <class T, Device device>
-dlaf::BaseType<T> lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag, const Tile<T, device>& a) {
+dlaf::BaseType<T> lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag, const Tile<T, device>& a) noexcept {
   return lapack::lantr(norm, uplo, diag, a.size().rows(), a.size().cols(), a.ptr(), a.ld());
 }
 
--- a/include/dlaf/lapack_tile.tpp
+++ b/include/dlaf/lapack_tile.tpp
@@ -10,12 +10,12 @@
 
 template <class T, Device device>
 dlaf::BaseType<T> lange(lapack::Norm norm, const Tile<T, device>& a) {
-  return lapack::lange(norm, a.size().cols(), a.size().rows(), a.ptr(), a.ld());
+  return lapack::lange(norm, a.size().rows(), a.size().cols(), a.ptr(), a.ld());
 }
 
 template <class T, Device device>
 dlaf::BaseType<T> lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag, const Tile<T, device>& a) {
-  return lapack::lantr(norm, uplo, diag, a.size().cols(), a.size().rows(), a.ptr(), a.ld());
+  return lapack::lantr(norm, uplo, diag, a.size().rows(), a.size().cols(), a.ptr(), a.ld());
 }
 
 template <class T, Device device>
--- a/include/dlaf/lapack_tile.tpp
+++ b/include/dlaf/lapack_tile.tpp
@@ -9,6 +9,23 @@
 //
 
 template <class T, Device device>
+dlaf::BaseType<T> lange(lapack::Norm norm, const Tile<T, device>& a) {
+  return lapack::lange(norm, a.size().cols(), a.size().rows(), a.ptr(), a.ld());
+}
+
+template <class T, Device device>
+dlaf::BaseType<T> lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag, const Tile<T, device>& a) {
+  return lapack::lantr(norm, uplo, diag, a.size().cols(), a.size().rows(), a.ptr(), a.ld());
+}
+
+template <class T, Device device>
+void potrf(blas::Uplo uplo, const Tile<T, device>& a) {
+  auto info = potrfInfo(uplo, a);
+  if (info != 0)
+    throw std::runtime_error("Error: POTRF: A is not positive definite.");
+}
+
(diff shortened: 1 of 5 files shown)


include/dlaf/lapack_tile.tpp
```diff
@@ -8,6 +8,24 @@
...
+dlaf::BaseType<T> lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag,
+                        const Tile<T, device>& a) noexcept {
+  return lapack::lantr(norm, uplo, diag, a.size().rows(), a.size().cols(), a.ptr(), a.ld());
+}
+
+template <class T, Device device>
+void potrf(blas::Uplo uplo, const Tile<T, device>& a) {
+  auto info = potrfInfo(uplo, a);
+  if (info != 0)
+    throw std::runtime_error("Error: POTRF: A is not positive definite.");
```
teonnik, line 26: Are we expecting people to recover from this? If not, we should use an `DLAF_ASSERT*`.

test/unit/test_lapack_tile.cpp
```diff
@@ -22,12 +24,55 @@ using namespace dlaf_test;
...
 
 const std::vector<blas::Uplo> blas_uplos({blas::Uplo::Lower, blas::Uplo::Upper});
+const std::vector<blas::Diag> blas_diags({blas::Diag::NonUnit, blas::Diag::Unit});
 
 template <typename Type>
 class TileOperationsTest : public ::testing::Test {};
 
 TYPED_TEST_SUITE(TileOperationsTest, MatrixElementTypes);
 
+TYPED_TEST(TileOperationsTest, lange) {
```
teonnik, line 34: The testing is quite comprehensive but we only provide a thin interface over lapackpp. We essentially end-up testing lapackpp for correctness. Wouldn't it be simpler to instead compare the output from `lapackpp::lange` and `lapackpp:lantr` to our wrappers? A single test that checks parameters are passed properly should suffice. I am a bit concerned that we are adding a lot of code that we have to maintain for simple wrapper functions.

teonnik: On the other hand, if we decide to add other back ends, it might be good having the tests reference only the public API and not implementation details.

teonnik: I would leave the tests as they are, relying on the public API is IMO better. You can mark this as resolved unless you want to add something.

include/dlaf/lapack_tile.tpp
albestro: As stated in another PR, I'm leaving it for the dedicated PR #178 so it would be easier to merge the work later.

My idea about it is to not "alter" the existing code but at the same time do the right thing when adding something new.

Do you agree?

test/unit/test_lapack_tile.cpp
albestro: It's a good point and I agree with your last statement.

About tests, I'm wondering how comprehensive we want to be. For instance, do we want to test all types for norm type? I think that in theory we should, but at the same time the quantity of code, as you were concerned, may easily explode exponentially.

What about this? @rasolca @teonnik

include/dlaf/lapack_tile.tpp
teonnik: Yeah, certainly!

albestro:
**QUESTION** (somewhat related to #152)

Depending on the geometry of the Matrix, in terms of `size` and `blocksize`, an algorithm may found itself working with tiles on the borders that are not "full" size.

For instance, the `lantr` tile-wrapper implemented by this PR is employed for computing the `norm` for square tiles on the diagonal (see #224 for reference) of a distirbuted matrix.
Let's say we are computing the norm of a Lower Triangular Matrix (10, 10) with blocksize (4, 4), we will end up calling `lantr` on a (2, 4) tile (the bottom right one on the diagonal), that it is not a valid call for LAPACK `lantr` (becase with a lower triangular it is required `m >= n`).

What do we want to do in these cases? I can start listing few hypotheses which do not convince me very much, but it may be a starting point for discussion.
1. do we want to manage this at tile-wrapper level and diverging from the lapack function behavior? e.g. in the case illustrated I can think of reducing the number of number of cols to be equal to number of rows
2. do we have to manage incomplete tiles in the algorithm? in this way the burden is on the user, but this additional control does not require any hidden/different behavior in the tile-wrapper function

Do you have any idea/suggestione? @rasolca @teonnik

rasolca:
> Depending on the geometry of the Matrix, in terms of `size` and `blocksize`, an algorithm may found itself working with tiles on the borders that are not "full" size.
> 
> For instance, the `lantr` tile-wrapper implemented by this PR is employed for computing the `norm` for square tiles on the diagonal (see #224 for reference) of a distirbuted matrix.
> Let's say we are computing the norm of a Lower Triangular Matrix (10, 10) with blocksize (4, 4), we will end up calling `lantr` on a (2, 4) tile (the bottom right one on the diagonal), that it is not a valid call for LAPACK `lantr` (becase with a lower triangular it is required `m >= n`).

I don't see a problem:
- if the matrix and block size (bs) is square (sq) -> the diagonals are square
- if the matrix is non-square (n-sq) and the bs is sq -> if the full matrix has m>=n the last diagonal tile has m>=n too.
- if the bs is n-sq, extra care has to be taken anyway.
(Note that for "non diagonal" tiles the general norm has to e used.)

(Anyway there is nothing that stops you to "fix" the dimension given to lapack (a "(2x4)" lower triangular matrix is in fact a (2x2) lower triangular matrix..., looking at lapack code there is no check on this and the dimension is reduced too...) but it has to be decided if this makes sense...)

test/unit/test_lapack_tile.cpp
rasolca: I think that adding some basic tests it is always a good idea:
you check if you interface the library correctly, but also check that no strange problems are around:
e.g. header version that do not match the library...
Moreover as you already mention the interface library can be changed easily.
Of course we do not have to check all corner cases as for our implementations, but I would check that the different versions (different parameters for uplo, op, side, ...) are interfacing correctly.

albestro:
Yes, good point. I was wrong with my example.

I will add the preconditions for the wrappers as stated in the lapack documentation.

albestro:
I've just added basic tests for both `lange` and `lantr` covering all parameters combinations.

Looking forward for your feedback.

test/unit/test_lapack_tile.cpp
teonnik: Related to our discussion : https://github.com/eth-cscs/DLA-Future/issues/232 .

albestro:
I can refactor tests to reduce the number of reallocation of tiles
e.g.

```cpp
for (const auto diag : blas_diags)
  for (const auto& size : sizes)
    // allocate tile  
    for (const auto norm : lapack_norms)
      for (const auto uplo : blas_uplos)
        // set tile and test
```

instead of

```cpp
  for (const auto norm : lapack_norms)
    for (const auto uplo : blas_uplos)
      for (const auto diag : blas_diags)
        for (const auto& size : sizes)
          // allocate, set tile and then test
```

test/unit/test_lapack_tile.cpp
```diff
@@ -22,12 +24,79 @@ using namespace dlaf_test;
...
+
+    for (const auto uplo : blas_uplos) {
+      // transpose rectangular matrix to be useful for upper triangular case
+      if (blas::Uplo::Upper == uplo)
+        std::swap(m, n);
+
+      auto tile = allocate_tile<TypeParam>(TileElementSize{m, n}, extra_lda);
+
+      for (const auto norm : lapack_norms) {
+        // lange cannot be tested for norm2
```
rasolca, line 88:
```suggestion
        // lantr do not support norm2
```

test/unit/test_lapack_tile.cpp
```diff
@@ -22,12 +24,79 @@ using namespace dlaf_test;
...
+                                                                 {17, 11, 3}, {17, 11, 0}, {17, 17, 3},
+                                                                 {11, 11, 0}};
+
+  for (const auto& size : sizes) {
+    std::tie(m, n, extra_lda) = size;
+
+    auto tile = allocate_tile<TypeParam>(TileElementSize{m, n}, extra_lda);
+
+    for (const auto norm : lapack_norms) {
+      // lange cannot be tested for norm2
```
rasolca, line 61: Is it not better do not support?

test/unit/test_lapack_tile.cpp
rasolca, line 88:
```suggestion
        // lantr do not support norm2
```

test/unit/test_lapack_tile/test_lantr.h
```diff
@@ -0,0 +1,170 @@
...
+void test_lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag, const Tile<T, Device::CPU>& a,
+                NormT<T> norm_expected) {
+  set(a, TileSetter<T>{a.size(), uplo});
+
+  SCOPED_TRACE(::testing::Message() << "LANTR: " << lapack::norm2str(norm) << ", " << a.size()
+                                    << ", ld = " << a.ld() << " uplo = " << blas::uplo2str(uplo)
+                                    << " diag = " << blas::diag2str(diag));
+
+  // by LAPACK documentation, if it is an empty matrix return 0
+  norm_expected = (a.size().isEmpty()) ? 0 : norm_expected;
```
rasolca, line 115: For readability and correctness `norm_expected` should set to zero outside of this function.

rasolca:
bors try

haampie:
Interesting, it reported here twice :laughing:

rasolca:
bors try

PR #228, opened 2020-05-12
MPI executor

teonnik:
This implements a basic mechanism for handling non-blocking MPI communication within HPX tasks. The mechanism works both with and without a separate MPI pool. The PR also provides some documentation on the upcoming HPX MPI_Futures. It explains some API differences and differences in the internal communication handling mechanism.

TODO
- [x] Integration with DLAF's Data/Message interfaces.

Early feedback is welcome!


include/dlaf/communication/async.h
```diff
@@ -0,0 +1,108 @@
...
+/// dlaf::comm::async(ex, comm, MPI_Irecv, recv_ptr, num_recv_elems, MPI_DOUBLE, src_rank, tag);
+///
+/// compared to the usual MPI non-blocking call:
+///
+/// MPI_Irecv(recv_ptr, num_recv_elems, MPI_DOUBLE, src_rank, tag, comm, &request);
+///
+/// there are two differences
+///
+/// 1) MPI_Comm (comm) goes after the executor in async()
+/// 2) MPI_Requst (request) is omitted as it is handled internally in async()
```
albestro, line 46:
```suggestion
/// 2) MPI_Request (request) is omitted as it is handled internally in async()
```

test/unit/communication/test_async.cpp
```diff
@@ -0,0 +1,43 @@
...
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2019, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#include <bits/c++config.h>
```
albestro, line 11: Out of curiosity: what is this for?

test/unit/communication/test_async.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <gtest/gtest.h>
+#include <mpi.h>
+#include <hpx/lcos/wait_all.hpp>
+#include <vector>
+
+#include "dlaf/communication/async.h"
```
albestro, line 17: This header is the one under test and should be put as first, before any other header.

include/dlaf/communication/async.h
```diff
@@ -0,0 +1,108 @@
...
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+#pragma once
+
+/// @file
+
+#include <utility>
+#include <vector>
```
albestro, line 15: I don't see where you use `std::vector` here...is it needed?

include/dlaf/communication/async.h
```diff
@@ -0,0 +1,108 @@
...
+/// });
+///
+/// or don't return the future from `dlaf::comm::async()` and just do:
+///
+/// auto comm_fut = prev_fut.then([=] {
+///   dlaf::comm::async(...);
+/// });
+///
+template <class F, class... Ts>
+hpx::future<void> async(hpx::threads::executors::pool_executor& ex, MPI_Comm comm, F&& f,
```
albestro, line 64: I see you are still working on it and you have to make it compatible with the `Data`/`Message` API.

Just an additional suggestion. I would make this an `internal` function, and I would keep the same functions names used in `dlaf::comm:sync`, specifying two functions `dlaf::comm::async::send_to` and `dlaf::comm::async::receive_from`.

https://github.com/eth-cscs/DLA-Future/blob/ad7962e9073b883f8d527c34831fe0b85b27901a/include/dlaf/communication/sync/basic.h#L23-L42

include/dlaf/communication/async.h
```diff
@@ -0,0 +1,108 @@
...
+///
+/// Example usage:
+///
+/// ```
+/// // registers a `poll()` function on the thread pool with `pool_name`. This has to be in scope of all
+/// // uses of hpx::mpi.
+/// hpx::mpi::experimental::enable_user_polling enable_polling(pool_name);
+/// ...
+/// hpx::mpi::experimental::executor ex(mpi_comm);
+/// hpx::async(ex, MPI_Irecv, MPI_Irecv, recv_ptr, num_recv_elems, MPI_DOUBLE, src_rank, tag)
```
Sely85, line 89: `MPI_Irecv` is called twice: do we need to add `mpi_comm` as an argument of `async` or this is not necessary anymore since it will be forward by the executor?

test/unit/communication/test_async.cpp
teonnik: `clangd` occasionally adds some header it thinks are needed, I haven't noticed that one, I will remove it.

include/dlaf/communication/async.h
teonnik: Good catch, I was using it earlier when prototyping and forgot to remove it.

include/dlaf/communication/async.h
teonnik: Hmm, this one I will have think about for a bit longer. Note that the `dlaf::comm::async()` can schedule any non-blocking routine, not only Isend and Irecv (e.g. Ireduce etc). The difficulty with integrating the Data/Message API is deciding where to place the parameters for the message size, type and buffer encoded in the API. The current API exploits the fact that MPI_Comm and MPI_Request are always at the end, the Data/Message API parameters are somewhere in the middle.

include/dlaf/communication/async.h
teonnik: Yes, that was my mistake, I will delete one of the `MPI_Irecv`.

include/dlaf/communication/async.h
teonnik: We can, of course, just have separate function, like we do now : `send_to()` / `receive_from()`, etc. Perhaps that is what you meant. In that case, yeah, there won't be a problem.

albestro: Yes, what you are saying in the last comment it is exactly what I meant.

About your problem with `Data`/`Message` API, the additional layer of indirection (e.g. `send_to`,`receive_to`,...) before the `internal` generic function may help you in managing it. If I get correctly your point, you can leave as it is the `internal` generic function, and then you will have different "proxy" functions with different sets of parameters (potentially with a different order then the one you need for the internal one).

teonnik:
TODO: 
- [ ] API based on Message/Data
- [ ] `mpi_executor` parametrized over 
  - [ ] the yield_while approach 
  - [ ] the oversubscription model with MPI_THREAD_SERIALIZED approach 
  - [ ] array of futures approach (John's branch)

teonnik:
MPI executor now works with Cholesky! There are certain improvements that can be made, they remain WIP:

- unit tests for kernels.h
- more non-blocking kernels (I need MPI_Iallreduce for TSGEMM).
- a pool of communicators to unlock more asynchrony
- refinements/adjustments to the TaskChain mechanism

but before that, may I ask if someone will volunteer to benchmark the Cholesky implementation in this PR and compare performance with previous benchmarks? I am not familiar with the parameter set and scripts used in previous benchmarks. I would like to find out if there is a performance regression or other weird issues. In particular, this PR should also work without a separate MPI pool (i.e. just with the default pool).

teonnik:
bors try

albestro:
> may I ask if someone will volunteer to benchmark the Cholesky implementation in this PR and compare performance with previous benchmarks?

I did the benchmark for the Cholesky, so I can help/do it! Just one thing, before proceeding with the benchmark, it would be easier and better to close the #116. 😉

include/dlaf/factorization/cholesky/mc/cholesky_L.h
```diff
@@ -144,10 +147,13 @@ void cholesky_L(comm::CommunicatorGrid grid, Matrix<T, Device::CPU>& mat_a) {
         // Avoid useless communication if one-column communicator and if on the last column
         if (col_comm_size > 1 && k != (mat_a.nrTiles().cols() - 1)) {
           // Broadcast the panel column-wise
-          hpx::dataflow(hpx::util::unwrapping([](auto&& tile, auto&& comm_wrapper) {
```
rasolca, line 147: Here `executor_mpi` is missing.
It should be fixed for the comparison benchmarks.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
```diff
@@ -144,10 +147,13 @@ void cholesky_L(comm::CommunicatorGrid grid, Matrix<T, Device::CPU>& mat_a) {
...
-                          comm::sync::broadcast::send(comm_wrapper().colCommunicator(), tile);
-                        }),
-                        mat_a.read(kk), serial_comm());
+          auto send_bcast_f =
+              hpx::util::unwrapping([ex = executor_mpi_col](auto&& tile, auto&& promise) mutable {
+                auto bcast_fut = comm::bcast(ex, ex.comm().rank(), tile);
+                promise.set_value();
+                bcast_fut.get();
+              });
+          hpx::dataflow(std::move(send_bcast_f), mat_a.read(kk), col_bcast.chain());
```
rasolca, line 156: communication tasks should be high priority, or even better be synchronous (i.e. executed as soon as all the dependencies get ready). I think that by default dataflow creates a normal priority async task.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
teonnik: PR: #246

test/unit/common/CMakeLists.txt
```diff
@@ -14,3 +14,4 @@ DLAF_addTest(test_range2d SOURCES test_range2d.cpp USE_MAIN PLAIN)
 DLAF_addTest(test_data_descriptor SOURCES test_data_descriptor.cpp USE_MAIN PLAIN)
 
 DLAF_addTest(test_pipeline SOURCES test_pipeline.cpp USE_MAIN HPX)
+DLAF_addTest(test_task_chain SOURCES test_pipeline.cpp USE_MAIN HPX)
```
Sely85, line 17: should this be:
```suggestion
DLAF_addTest(test_task_chain SOURCES test_task_chain.cpp USE_MAIN HPX)
```
instead?

teonnik: Good catch! Just fixed it!

test/unit/common/test_task_chain.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "dlaf/common/task_chain.h"
+
+#include <gtest/gtest.h>
+
+#include <hpx/dataflow.hpp>
+#include <hpx/lcos/future.hpp>
+#include <hpx/lcos/wait_all.hpp>
+
+TEST(Pipeline, Basic) {
```
Sely85, line 19:
```suggestion
TEST(TaskChain, Basic) {
```
?

include/dlaf/communication/executor.h
```diff
@@ -11,25 +11,141 @@
...
+  /// or don't return the future from `dlaf::comm::async()` and just do:
+  ///
+  /// auto comm_fut = prev_fut.then([=] {
+  ///   dlaf::comm::async(...);
+  /// });
+  ///
+  template <typename F, typename... Ts>
+  hpx::future<void> async_execute(F f, Ts... ts) noexcept {
+    // TODO: docs why this is done here instead of within the task
+    // TODO: support for MPI_THREAD_SERIALIZED
```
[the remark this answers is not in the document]
teonnik: > I'm wondering if the solution you have in mind for MPI_THREAD_SERIALIZED is to serialize the communications, so that MPI_THREAD_SERIALIZE would be enough. Which sounds like, more or less, what we currently do in the algorithm with the TaskChain

To support MPI_THREAD_SERIALIZED it is sufficient to make sure that MPI calls are not submitted at the same time. This can be done via a global `hpx::lcos::mutex`. This is lot less strict then imposing a certain order on tasks which is what `TaskChain` is doing.

This was discussed with John some time ago. The upstream HPX MPI executor doesn't currently support MPI_THREAD_SERIALIZED. It should be rather straightforward to provide support though. I will implement it here for now, and bug John next week to see if he wants to add it upstream.

include/dlaf/communication/executor.h
```diff
@@ -11,25 +11,141 @@
...
+  ///
+  /// or don't return the future from `dlaf::comm::async()` and just do:
+  ///
+  /// auto comm_fut = prev_fut.then([=] {
+  ///   dlaf::comm::async(...);
+  /// });
+  ///
+  template <typename F, typename... Ts>
+  hpx::future<void> async_execute(F f, Ts... ts) noexcept {
+    // TODO: docs why this is done here instead of within the task
```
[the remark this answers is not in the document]
teonnik: The answer is that non-blocking MPI collectives on the same communicator need to issued in order. If I launch multiple such MPI calls from the **same** task, when I have the call inline, I am guaranteed to have them in order, I will not need to use `TaskChain` (or Pipeline, etc). `TaskChain` will be involved if such MPI calls are made from separate tasks. That is often the case when the call is hapenning in a continuation ( `some_fut.then([mpi_ex] {... bcast(mpi_ex,...).get(); ... })`).

Moreover, not having to pass the function and it's arguments through a capture or a function call is a modest performance win and circumvents debugging lengthy compile-time messages  that often appear when passing things into the heavily templated `hpx::async` and `hpx::dataflow`.

include/dlaf/communication/executor.h
```diff
@@ -11,25 +11,141 @@
...
+  /// 1) MPI_Comm (comm) doesn't need to be specified as it is passed by the executor
+  /// 2) MPI_Request (request) is omitted as it is handled internally in async()
+  ///
+  /// Note that if starting the communication depends on a previous task, it is often a good idea to call
+  /// the `get()` method on a returned future to avoid weird constructs like future<future<>>.
+  ///
+  /// Example:
+  ///
+  /// auto comm_fut = prev_fut.then([=] {
+  ///   return dlaf::comm::async(...).get();
```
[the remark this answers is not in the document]
teonnik: Comment is a out-of-date but still, no, it's not equivalent. With a blocking MPI call, the OS thread on which the task is executed will be blocked until the call returns. The task within which the `.get()` call is happening yields instead (internally `MPI_Test()` is used to determine whether the future is ready) so other tasks can still run on that OS thread.

include/dlaf/communication/kernels.h
```diff
@@ -0,0 +1,78 @@
...
+  using DataT = std::remove_const_t<typename common::data_traits<decltype(data)>::element_t>;
+
+  auto message = comm::make_message(common::make_data(data));
+  return ex.async_execute(MPI_Ibcast, const_cast<DataT*>(message.data()), message.count(),
+                          message.mpi_type(), root_rank);
+}
+
+/// MPI_Ireduce wrapper
+///
+/// @param reduce_operation MPI_Op to perform on @p input data coming from ranks in @p communicator
```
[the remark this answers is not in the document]
teonnik: Yes, I will place a comment there to clarify.

include/dlaf/communication/kernels.h
```diff
@@ -0,0 +1,78 @@
...
+#include <mpi.h>
+
+#include "dlaf/common/assert.h"
+#include "dlaf/common/data.h"
+#include "dlaf/communication/communicator.h"
+#include "dlaf/communication/executor.h"
+#include "dlaf/communication/message.h"
+
+namespace dlaf {
+namespace comm {
```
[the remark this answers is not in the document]
teonnik: I am not a huge fan of the `dlaf::comm::sync` namespace. There are a handful of routines that can clearly be distinguished with different function names, I don't see why we need a separate namespace here. Moreover, unless we discover any issues with the `async` routines, I don't see why we would want to keep the `sync`.

test/unit/common/test_task_chain.cpp
```diff
@@ -0,0 +1,38 @@
...
+  dlaf::common::TaskChain tc{};
+
+  int d = 0;
+  auto func_1 = [&d](hpx::future<hpx::promise<void>> fp) {
+    fp.get().set_value();
```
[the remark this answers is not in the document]
teonnik: Oops, good catch! The `set_value()` call should be after the assignment.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
teonnik: send/recv tasks now use `executor_hp`.

PR #231, opened 2020-05-15
Modify the ASSERT mechanism

teonnik:
This refactors the ASSERT implementation. 

- The origin and category parameters are removed. The origin parameter was needed because we were nesting ASSERT calls within functions. Instead, the opposite approach should be taken, move the bulk of the logic outside of the macro and into a function, then use the function within the macro. 
- ~~Each macro now only takes two parameters, the expression to evaluate and another expression for the message. This is simple and avoids variadic macros.~~
- The bulk of the ASSERT mechanism is now in `do_assert()`, the macros just reference the function.
- The documentation is updated
- `assertMultipliableMatrices()` was replaced by `multipliable_sizes`~~/multipliable_sizes_msg~~
- ~~The additional ASSERT macros (e.g. DLAF_ASSERT_DISTRIBUTED_EQ) are left unchanged for now~~

EDIT 1:
~~I tried out a few approaches but landed on this one for now. We should decide whether we want the messages and conditions in the additional assert macros (e.g. DLAF_ASSERT_DISTRIBUTED_EQ) refactored into separate functions and replaced with DLAF_ASSERT. On the upside we have less macros, on the downside, their usage is slightly more complicated as two functions have to be called.~~


teonnik:
I've made a lot of changes to what I initially proposed in the PR. The proposed macros work like this (copied relevant section from the docs I wrote in the code):

```
/// Each macro has at least one and up to seven parameter:
///
/// 1.   @Expr    (required) : an experssion that returns a bool
/// 2.   @MsgExpr (optional) : an expression that returns a `std::string`
/// 3-7. @ARG1-5  (optional) : additional arguments to print
///
/// If @Expr is false, it composes an error message and calls `std::terminate()`. The error message
/// contains information about @Expr, the origin of the error, the optional custom message if provided,
/// and any aditional arguments (ARG1-5) if proveded. @ARG1-5 need to support `operator<<()`. Parameters
/// 3-7 require that parameter 2 be present even if empty("").
///
/// A disabled ASSERT macro has no overhead, @Expr and @MsgExpr are not evaluated. Macros are
/// enabled/disabled as follows:
///
///                       Control flag                 Default
/// DLAF_CHECK            N/A                          always enabled
/// DLAF_ASSERT           DLAF_ASSERT_ENABLE           enabled
/// DLAF_ASSERT_MODERATE  DLAF_ASSERT_MODERATE_ENABLE  enabled in Debug, disabled in Release
/// DLAF_ASSERT_HEAVY     DLAF_ASSERT_HEAVY_ENABLE     enabled in Debug, disabled in Release
///
/// Examples:
///
/// ```
/// bool is_sth_true(...) {...}
/// std::string my_msg(...) {...}
/// int a = 3;
/// int b = 4;
///
/// DLAF_ASSERT(5 == 6);                     // ASSERT with no additional information
/// DLAF_ASSERT(a == b, "my msg", a, b);     // * ASSERT with additional information and argument
/// DLAF_ASSERT(is_sth_true(), my_msg());    // ASSERT calling functions
/// ```
///
/// The output from * is as follows:
//
/// ```
/// [ERROR] <file_name>:<line_number> : <function>
/// a == b
/// my msg
/// a : 3
/// b : 4
/// ```

```

Note in particular:
- The message argument is again optional, e.g. DLAF_ASSERT(5 == 6) works
- The additional arguments after the message arguments print both the variable name and the value (i.e. look at the bottom of the comment where both the variable name and values for `a` and `b` are printed). This means that there is no need to define another macro and pass it's stringified argument with `#arg` as before.

Caveats:
- The internal mechanism is essentially implementing macros overoading and is significantly more complex than before (works well though)
- The number of additional arguments is limited to 5 for now. If we need more, we can easily extend the mechanism though.
- The message argument must always precede the additional arguments even if empty. For example this `DLAF_ASSERT(a == b, a, b)` will not work, to fix add an empty string `DLAF_ASSERT(a == b, "", a, b)`. 

The last point need not be the case. I will work towards removing that issue.

include/dlaf/common/utils.h
```diff
@@ -44,6 +46,14 @@ struct source_location {
   }
 };
 
+inline void do_assert(bool expr, const source_location& loc, const char* expression,
+                      std::string const& msg) {
+  if (!expr) {
+    std::cerr << "[ERROR] " << loc << '\n' << expression << '\n' << msg << std::endl;
+    std::terminate();
```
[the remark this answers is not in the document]
teonnik: I am not sure tbh, `std::endl` already flushes but I am not sure how it interacts with `std::cerr`.

include/dlaf/util_matrix.h
```diff
@@ -38,117 +38,106 @@ namespace internal {
...
+
+template <class IndexT, class Tag>
+std::string multipliable_sizes_msg(common::Size2D<IndexT, Tag> a, common::Size2D<IndexT, Tag> b,
+                                   common::Size2D<IndexT, Tag> c, const blas::Op opA, const blas::Op opB,
+                                   const char* mat_a_name, const char* mat_b_name,
+                                   const char* mat_c_name) noexcept {
+  if (opA != blas::Op::NoTrans)
+    a.transpose();
+  if (opB != blas::Op::NoTrans)
+    b.transpose();
```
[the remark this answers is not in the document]
teonnik: I removed this method in the new version.

include/dlaf/common/index2d.h
```diff
@@ -245,10 +245,9 @@ Index2D<IndexT, Tag> computeCoordsColMajor(std::ptrdiff_t linear_index,
                                            const Size2D<IndexT, Tag>& dims) noexcept {
   using dlaf::util::ptrdiff_t::mul;
 
-  DLAF_ASSERT_MODERATE(linear_index >= 0, "The linear index cannot be negative (",
-                       std::to_string(linear_index), ")");
-  DLAF_ASSERT_MODERATE(linear_index < mul(dims.rows(), dims.cols()), "Linear index ",
-                       std::to_string(linear_index), " does not fit into grid ", dims);
+  DLAF_ASSERT_MODERATE(linear_index >= 0, "The linear index cannot be negative!", linear_index);
+  DLAF_ASSERT_MODERATE(linear_index < mul(dims.rows(), dims.cols()),
+                       "Linear index does not fit into grid!", dims);
```
Sely85, line 250: Here maybe a `linear_index` is missing
```suggestion
                       "Linear index does not fit into grid!", linear_index, dims);
```

include/dlaf/common/index2d.h
```diff
@@ -294,7 +293,7 @@ LinearIndexT computeLinearIndexRowMajor(const Index2D<IndexT, Tag>& index,
 
   static_assert(std::is_integral<LinearIndexT>::value, "LinearIndexT must be an integral type");
 
-  DLAF_ASSERT_MODERATE(index.isIn(dims), "Index ", index, " is not in the grid ", dims);
+  DLAF_ASSERT_MODERATE(index.isIn(dims), "Index is not in the grid!", dims);
```
Sely85, line 296: What about adding also the value of `index`?
```suggestion
  DLAF_ASSERT_MODERATE(index.isIn(dims), "Index is not in the grid!", index, dims);
```

include/dlaf/common/index2d.h
```diff
@@ -316,7 +315,7 @@ LinearIndexT computeLinearIndexColMajor(const Index2D<IndexT, Tag>& index,
...
```
Sely85, line 318: same comment as above

include/dlaf/solver/triangular/mc.tpp
```diff
@@ -129,7 +137,12 @@ void Solver<Backend::MC>::triangular(comm::CommunicatorGrid grid, blas::Side sid
   }
   else {
     // Check if A and B dimensions are compatible
-    DLAF_ASSERT_MULTIPLIABLE_MATRICES(mat_b, mat_a, mat_b, blas::Op::NoTrans, op);
+    DLAF_ASSERT(matrix::multipliable_sizes(mat_b.size(), mat_a.size(), mat_b.size(), blas::Op::NoTrans,
+                                           op),
+                "Sizes mismatch!", mat_a, mat_b);
+    DLAF_ASSERT(matrix::multipliable_sizes(mat_b.blockSize(), mat_a.blockSize(), mat_a.blockSize(),
```
Sely85, line 143:
```suggestion
    DLAF_ASSERT(matrix::multipliable_sizes(mat_b.blockSize(), mat_a.blockSize(), mat_b.blockSize(),
```

include/dlaf/matrix/matrix_base.h
```diff
@@ -90,6 +93,17 @@ class MatrixBase {
     return sum(index.row(), mul(distribution_->localNrTiles().rows(), index.col()));
   }
 
+  /// Prints information about the matrix.
+  friend std::ostream& operator<<(std::ostream& out, const MatrixBase& matrix) {
+    // clang-format off
+    return out << "size="         << matrix.size() 
+               << ", block_size=" << matrix.blockSize() 
+               << ", tiles_grid"  << matrix.nrTiles()
```
Sely85, line 101:
```suggestion
               << ", tiles_grid="  << matrix.nrTiles() 
```

teonnik:
bors try

teonnik:
bors try

albestro:
I try to report here a conversation we had in the last days with @teonnik about this PR.

I find the idea of switching to functions instead of macros very nice! Moreover, I understand that not using macros the same effect may come with a lot more complicated machinery, and about this I have my personal view.

I think that we can identify "must have", "nice to have" and "extra" features about ASSERT mechanism. My view is:
- must have:
  - *zero overhead*
  - *real location of the problem*: the error must point to the context where the problem happened
- nice to have:
  - *custom message*: I agree that with the new mechanism, the condition becomes way more expressive and can be enough in a lot of situations, but in other I think that the value of a particular parameter may be very useful (e.g. size of a matrix);
- extra:
  - *optional message*: I'm of the "give an explanatory message" party, so I would vote easily for not having this
  - *variable names*: I don't think it adds a lot of value to the message. As soon as I get the exact location, I will give a look at the code and quickly realize what the parameter name is.

Aiming at code simplification, I would propose to loose some features. IMHO if losing above listed extract features will help in simplifying the machinery, I will not miss them. Moreover, I find all the macro machinery an over-complication.
Last but not least, in my initial intentions DLAF_CHECK was not something to expose but just an implementation detail. I'm ok with removing it.

With my previous assumptions, a proposal might be to use variadic templates in conjunction with variadic macros and have something similar to

```cpp
template <class ... Ts>
inline void do_assert(bool expr, const common::internal::source_location& loc, const char* expression, Ts&& ... args) {
...
}

#define DLAF_CHECK(Expr, ...) dlaf::internal::do_assert(Expr, SOURCE_LOCATION(), #Expr, ##__VA_ARGS__)
```

Looking forward for your feedback and opinions!

@rasolca @Sely85 @teonnik

teonnik:
> - extra:
>    - optional message: I'm of the "give an explanatory message" party, so I would vote easily for not having this
 >   - variable names: I don't think it adds a lot of value to the message. As soon as I get the exact location, I will give a look at the code and quickly realize what the parameter name is.

I don't mind omitting these features if the current macro machinery is deemed complicated and not worth the maintenance cost. It is definitely good to provide more information in the macros, but we need to consider how likely we are that we would be able to diagnose a bug or a problem solely with the information provided in the macro. That is the situation we would find ourselves if the code is built without debug symbols, if on the other hand, the code is built with debug symbols, the core dump after the crash provides much more information that we could ever fit in a macro. 

> Last but not least, in my initial intentions DLAF_CHECK was not something to expose but just an implementation detail. I'm ok with removing it.

I also think DLAF_CHECK is superfluous. We have DLAF_ASSERT which is enabled by default and should almost never be disabled. The control flag is only there in case we do need to disable it for some temporary unforeseen situation. However, normally, it should be assumed to be enabled.

The complexity of the macro machinery is mostly due to the requirement that variable names be included in the message itself. I was trying to replace messages like these:

```
 "The matrix ", #matrixA, " and ",    \
              #matrixB, " are not distributed in the same way (rank: ", matrixA.rankIndex(), " vs ", \
              matrixB.rankIndex(), ", grid size: ", matrixA.commGridSize(), " vs ",                  \
              matrixB.commGridSize(), ")"
```

with this:

```
"The matrices are not distributed the same way!", matrixA, matrixB
```
The Matrix variables matrixA and matrixB are printable and provide all the information needed, the message is easy to read both from within the code and when reported with triggered ASSERT.

We can achieve the same with the old mechanism too, the only difference is that the variable names will not be included and should be deduced from the order of the printed arguments. We can perhaps prepend a number to make it even easier to relate the message to the variable. In any case, that is a trade-off that I find acceptable. 

Note that `##__VA_ARGS__` (as opposed to just `__VA_ARGS__`) is actually a compiler extension to C/C++ and is not technically part of the language. If we decide not to use it, we would still need at least two parameters to avoid empty variadic macro warnings/errors.

If there is support for what @albestro is proposing, I can modify the PR to align with what he suggests.

albestro:
> Note that `##__VA_ARGS__` (as opposed to just `__VA_ARGS__`) is actually a compiler extension to C/C++ and is not technically part of the language. If we decide not to use it, we would still need at least two parameters to avoid empty variadic macro warnings/errors.

Good point! I forgot to not use it in the example code but in the previous version of the mechanism I worked in order to not use it and stick to the C++ standard. About variadic templates, it is just an alternative.

Thanks for your feedback, let's see what others think!

include/dlaf/common/utils.h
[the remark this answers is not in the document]
rasolca: According to the doc it is included in endl.
https://en.cppreference.com/w/cpp/io/manip/endl

Sely85:
> I find the idea of switching to functions instead of macros very nice! 

I agree with you. 

> * nice to have:
>   
>   * _custom message_: I agree that with the new mechanism, the condition becomes way more expressive and can be enough in a lot of situations, but in other I think that the value of a particular parameter may be very useful (e.g. size of a matrix);

I would love to keep this "nice to have" feature.

> * extra:
>   
>   * _optional message_: I'm of the "give an explanatory message" party, so I would vote easily for not having this
>   * _variable names_: I don't think it adds a lot of value to the message. As soon as I get the exact location, I will give a look at the code and quickly realize what the parameter name is. 

These extra features are not that important to me as well. 

> Last but not least, in my initial intentions DLAF_CHECK was not something to expose but just an implementation detail. I'm ok with removing it.

It looks a bit redundant to me as well. I agree with @teonnik there is already DLAF_ASSERT, enabled by default.

> With my previous assumptions, a proposal might be to use variadic templates in conjunction with variadic macros and have something similar to
> 
> ```c++
> template <class ... Ts>
> inline void do_assert(bool expr, const common::internal::source_location& loc, const char* expression, Ts&& ... args) {
> ...
> }
> 
> #define DLAF_CHECK(Expr, ...) dlaf::internal::do_assert(Expr, SOURCE_LOCATION(), #Expr, ##__VA_ARGS__)
> ```
This does look good to me! So will we use `__VA_ARGS__` instead of `##_VA_ARGS__`?

rasolca:
>  must have:
        zero overhead
        real location of the problem: the error must point to the context where the problem happened
    nice to have:
        custom message: I agree that with the new mechanism, the condition becomes way more expressive and can be enough in a lot of situations, but in other I think that the value of a particular parameter may be very useful (e.g. size of a matrix);
    extra:
        optional message: I'm of the "give an explanatory message" party, so I would vote easily for not having this
        variable names: I don't think it adds a lot of value to the message. As soon as I get the exact location, I will give a look at the code and quickly realize what the parameter name is.

In general custom messages important when debug symbols are not available (or if the core file was not generated) and the problem is not always reproducible (e.g. race condition).

Variable names are sometimes helpful to localize the problem but not needed if the message contains enough details.

> I also think DLAF_CHECK is superfluous. We have DLAF_ASSERT which is enabled by default and should almost never be disabled. The control flag is only there in case we do need to disable it for some temporary unforeseen situation. However, normally, it should be assumed to be enabled.

I fully agree with it.

teonnik:
bors try

teonnik:
bors try

teonnik:
bors try

teonnik:
bors try

teonnik:
bors try

teonnik:
bors try

teonnik:
bors try

teonnik:
The macro mechanism is simplified and is quite similar to what we had before. The differences
- ASSERT accepts at least 2 arguments, the second argument can be an empty string ""
- DLAF_CHECK is no more
- Matrix and Tile are printable
- Various macros from `util_matrix.h` are refactored into functions
- The documentation has changed

This is ready for another round of reviews!

include/dlaf/common/assert.h
```diff
@@ -14,102 +14,100 @@
...
+/// bool is_sth_true(...) {...}
+/// std::string my_msg(...) {...}
+/// int a = 3;
+/// int b = 4;
 ///
-/// The error report can be extended with a custom message composed concatenating additional parameters
-/// given to the macro.
+/// DLAF_ASSERT(5 == 6, "");                 // ASSERT with no additional information
+/// DLAF_ASSERT(a == b, "my msg", a, b);     // * ASSERT with additional information and argument
+/// DLAF_ASSERT(is_sth_true(), my_msg());    // ASSERT calling functions
```
rasolca, line 81: Please add a note about the fact that if assertions are enabled `my_msg()` is evaluated even if the condition is true.

include/dlaf/common/assert.h
```diff
@@ -14,102 +14,100 @@
...
-/// be performed and it will not add any overhead, nor for the condition evaluation, nor for the message
-#define DLAF_ASSERT(...) DLAF_ASSERT_WITH_ORIGIN((SOURCE_LOCATION()), __VA_ARGS__)
+/// The output from * is as follows:
+//
+/// ```
+/// [ERROR] <file_name>:<line_number> : <function>
+/// a == b
+/// my msg
+/// a : 3
+/// b : 4
```
rasolca, line 91: I think the output is:
```
3
4
```
Am I missing something?

include/dlaf/solver/triangular/mc.tpp
```diff
@@ -93,18 +98,21 @@ void Solver<Backend::MC>::triangular(comm::CommunicatorGrid grid, blas::Side sid
...
-  // Check if block matrix A is square
-  DLAF_ASSERT_BLOCKSIZE_SQUARE(mat_a);
-  // Check compatibility of the communicator grid and the distribution of matrix A
-  DLAF_ASSERT_DISTRIBUTED_ON_GRID(grid, mat_a);
-  // Check compatibility of the communicator grid and the distribution of matrix B
-  DLAF_ASSERT_DISTRIBUTED_ON_GRID(grid, mat_b);
+  DLAF_ASSERT(matrix::square_size(mat_a), "Matrix is not square!", mat_a);
+  DLAF_ASSERT(matrix::square_blocksize(mat_a), "Matrix blocksize is not square!", mat_a);
+  DLAF_ASSERT(matrix::equal_process_grid(mat_a, grid),
+              "The matrix is not distributed according to the communicator grid!", mat_a);
```
rasolca, line 104: Ii would be a good idea to output also grid info, but I don't know if it is printable...

include/dlaf/common/assert.h
teonnik: No, you are right, I forgot to update the docs, that was what the previous mechanism printed.

include/dlaf/solver/triangular/mc.tpp
teonnik: It is not yet, but we can easily make it printable if there's consensus.

teonnik:
> It looks good to me. The only thing I don't like too much is that he error string has to be written each time (e.g. for the is square check). I thought about it but I was not able to come out with a nice solution that doesn't involve macros, but maybe someone has an idea.

If we have good enough documentation on the functions that are called in some of the macros, the message string seems a little superfluous to me. The expression is printed, people can go to the declaration of the function to see what is being checked.

include/dlaf/common/utils.h
```diff
@@ -12,6 +12,7 @@
 
 /// @file
 
+#include <iostream>
 #include <sstream>
 #include <string>
```
albestro, line 17: I think you can remove all headers now.

I see you decided to move `concat` to the `assert.h` header. I completely agree that it should be in an `internal` namespace since it is something we don't want to provide in the public API of this library. About the include file, I don't think it is worth having an additional file just for this macro with related data structure, so the the options I see are:
- move also `SOURCE_LOCATION` to the `assert.h` header file and remove the `common/utils.h`
- move back `concat` to this header file (but in the `internal` namespace as you correctly did)

I would go for the first one, if there is not any other side effect, because as you modified the `concat` function it is likley it won't be used anywhere else (considering that it concatenates using a newline).

include/dlaf/solver/triangular/mc.tpp
```diff
@@ -93,18 +98,21 @@ void Solver<Backend::MC>::triangular(comm::CommunicatorGrid grid, blas::Side sid
...
 
   if (side == blas::Side::Left) {
     // Check if A and B dimensions are compatible
-    DLAF_ASSERT_MULTIPLIABLE_MATRICES(mat_a, mat_b, mat_b, op, blas::Op::NoTrans);
+    DLAF_ASSERT(matrix::multipliable_sizes(mat_a.size(), mat_b.size(), mat_b.size(), op,
+                                           blas::Op::NoTrans),
+                "Sizes mismatch!", mat_a, mat_b);
+    DLAF_ASSERT(matrix::multipliable_sizes(mat_a.blockSize(), mat_b.blockSize(), mat_b.blockSize(), op,
+                                           blas::Op::NoTrans),
+                "Sizes mismatch!", mat_a, mat_b);
```
albestro, line 115: I think it may be worth having `matrix::multipliable` that merges these two separate checks.

rasolca:
> If we have good enough documentation on the functions that are called in some of the macros, the message string seems a little superfluous to me. The expression is printed, people can go to the declaration of the function to see what is being checked.

Good point... in fact it prints that `matrix::square_size(mat_a)` failed, and is trivial to understand that the matrix was not square. I would then just put:
```
DLAF_ASSERT(matrix::square_size(mat_a), mat_a);
```

include/dlaf/common/assert.h
albestro: Yes, true this is a downside that with macro we didn't had (it was a requirement in the beginning).

This means that as soon as the assert is enabled, in addition to the fundamental cost of the check, there is always the overhead of building the message with `concat`.

I'm wondering if using a variadic template parameter for `do_assert` cannot be an option, so we can delay the `concat` call inside it and removing the overhead (trading for an additional cost because of the templates at compilation time and in terms of binary size).

teonnik: Good points. The `do_assert()` function now uses variadic template to forward parameters to `concat()`, the function is no longer evaluated if the condition is true. However, `my_msg()` still is, I don't see a way around it unless we use a macro. Do we expect composing messages in functions to be common? If we don't we can leave the code as it is, if we do, then we should use a macro.

albestro: > However, my_msg() still is, I don't see a way around it unless we use a macro.

Yes, exactly. That was the reason I implemented it with macros in the beginning. Since you already print condition and values (with lazy evaluation) inside the default error message, there is less need of specifying "complex" custom messages, and in the 95% of time is enough just adding a string with a simple message. I think we can overlook this, I would just highlight the "downside" in the documentation as @rasolca suggested.

rasolca: A possibility may be to add an overload to `concat` for functions std::string(void) that evaluates the function at and concatenate it.
Anyway I think it is enough to just document it.

include/dlaf/common/utils.h
teonnik: > move also SOURCE_LOCATION to the assert.h header file and remove the common/utils.h

`SOURCE_LOCATION` is used in `communication/error.h` and `cuda/error.h` (also `cublas/error.h` in the future), perahps I should rename the header to `location.h` instead of `utils.h` and just keep it there?

teonnik:
bors try

teonnik:
bors try

teonnik:
This is ready for another round of reviews:

- The messages on most DLAF_ASSERT which call a function were removed as the function provides enough information to understand what went wrong.
- `blas::Op` and `CommunicationGrid` are now printable and included in the output of DLAF_ASSERT which involves them
- a separate `multipliable()` function for matrices using `multipliable_sizes` was defined
- `concat()` is no longer evaluated if the assertion is true.

include/dlaf/common/index2d.h
```diff
@@ -223,10 +224,9 @@ Index2D<IndexT, Tag> computeCoordsRowMajor(std::ptrdiff_t linear_index,
                                            const Size2D<IndexT, Tag>& dims) noexcept {
   using dlaf::util::ptrdiff_t::mul;
 
-  DLAF_ASSERT_MODERATE(linear_index >= 0, "The linear index cannot be negative (",
-                       std::to_string(linear_index), ")");
-  DLAF_ASSERT_MODERATE(linear_index < mul(dims.rows(), dims.cols()), "Linear index ",
-                       std::to_string(linear_index), " does not fit into grid ", dims);
+  DLAF_ASSERT_MODERATE(linear_index >= 0, "The linear index cannot be negative!", linear_index);
```
albestro, line 227: I wrapped `linear_index` with `std::to_string` because `uint8_t` or `int8_t` are interpreted as `char`s, so the equivalent ASCII mapping is printed instead of the numeric value (in case, see also next occurrences).

include/dlaf/common/utils.h
albestro: Good point! I like your proposal of renaming it to `location.h`.

include/dlaf/common/index2d.h
teonnik: I added a note to the code too.

teonnik:
This is ready for another round of reviews.
- `utils.h` was renamed to `source_location.h`
- `to_string` is used for `linear_index` in `index2d.h`

include/dlaf/common/index2d.h
```diff
@@ -224,9 +224,12 @@ Index2D<IndexT, Tag> computeCoordsRowMajor(std::ptrdiff_t linear_index,
                                            const Size2D<IndexT, Tag>& dims) noexcept {
   using dlaf::util::ptrdiff_t::mul;
 
-  DLAF_ASSERT_MODERATE(linear_index >= 0, "The linear index cannot be negative!", linear_index);
+  // `linear_index` is wrapped with `std::to_string` because uint8_t or int8_t are interpreted as chars,
```
albestro, line 227: If you want to leave the comment, it's ok for me, the only thing is that I don't know how useful it can be. Just to highlight that these are not the only places where there is this issue, in the past I had to fix this in another place.

teonnik: Yeah, there are indeed more than one places where this occurs. I appreciate having it once as a reminder as it is non-trivial and may come-up again. On the other hand, I didn't want to add it everywhere as it's too much bloat. Should we just leave it as it is now?

rasolca:
bors try

PR #233, opened 2020-05-26
Updating spack package

Sely85:
Proposal of spack package modification (thanks @albestro):
- renamed variant called 'gpu' to 'cuda', to align with other packages
- commented 'mpich' dependence, uncommented 'mpi'
- added '+test' variant
- added '+doc' variant
- modified blas/lapack providers, to allow both intelMKL and custom libraries
- fixed HPX version to 1.4.0 (as required at this moment)
- added CUDA_TOOLKIT_ROOT_DIR argument (required for some architectures).

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,16 +15,19 @@
 
 ## How to install DLA-Future with spack
 
-We provide a spack package DLA-Future that can be easily add to your own spack as follows:
+We provide a spack package DLA-Future that can be easily added to your own spack as follows:
 
-`spack repo add spack`
+`spack repo add <dlaf root>/spack`
 
 This will add a new repository with namespace `dlaf`.
 
-The command to install the library for example on an HPC cluster will be similar to:
+Example installation:
 
-`spack install -v --test=root dla-future ^mkl ^hpx max_cpu_count=256`
+`spack install dla-future ^intel-mkl`
 
+Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the platform, as it represents the maximum number of OS-threads. 
+
+`spack install -v dla-future ^intel-mkl ^hpx max_cpu_count=256`
 
 ## How to use the library
 
--- a/README.md
+++ b/README.md
@@ -13,6 +13,19 @@
 - [GoogleTest](https://github.com/google/googletest) (optional; bundled) - unit testing
 - Doxygen (optional) - documentation
 
+## How to install DLA-Future with spack
+
+We provide a spack package DLA-Future that can be easily add to your own spack as follows:
+
+`spack repo add spack`
+
+This will add a new repository with namespace `dlaf`.
+
+The command to install the library for example on an HPC cluster will be similar to:
+
+`spack install -v --test=root dla-future ^mkl ^hpx max_cpu_count=256`
+
+
 ## How to use the library
 
 Using DLAF in a CMake project is extremly easy!
--- a/README.md
+++ b/README.md
@@ -25,7 +25,8 @@ Example installation:
 
 `spack install dla-future ^intel-mkl`
 
-Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the platform, as it represents the maximum number of OS-threads. 
+Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the platform,
+as it represents the maximum number of OS-threads. 
 
 `spack install dla-future ^intel-mkl ^hpx max_cpu_count=256`
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Example installation:
 
 Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the platform, as it represents the maximum number of OS-threads. 
 
-`spack install -v dla-future ^intel-mkl ^hpx max_cpu_count=256`
+`spack install dla-future ^intel-mkl ^hpx max_cpu_count=256`
 
 ## How to use the library
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 
 We provide a spack package DLA-Future that can be easily added to your own spack as follows:
 
-`spack repo add <dlaf root>/spack`
+`spack repo add $DLAF_ROOT/spack`
 
 This will add a new repository with namespace `dlaf`.
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Example installation:
 `spack install dla-future ^intel-mkl`
 
 Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the platform,
-as it represents the maximum number of OS-threads. 
+as it represents the maximum number of OS-threads.
 
 `spack install dla-future ^intel-mkl ^hpx max_cpu_count=256`
 
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -39,7 +39,7 @@ class DlaFuture(CMakePackage):
 
        if '+cuda' in spec:
            args.append('-DDLAF_WITH_CUDA=ON')
-           args.append('-DCUDA_HOME={0}'.format(spec['cuda'].prefix))
+           #args.append('-DCUDA_HOME={0}'.format(spec['cuda'].prefix))
 
        if self.run_tests:
            args.append('-DDLAF_WITH_TEST=ON')
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -19,7 +19,7 @@ class DlaFuture(CMakePackage):
             description='Use the GPU/cuBLAS back end.')
     variant('doc', default=False,
             description='Build documentation.')
-    
+
     #depends_on('mpi@3:')
     depends_on('mpi')
     depends_on('blaspp')
@@ -29,24 +29,25 @@ class DlaFuture(CMakePackage):
 
     def cmake_args(self):
        spec = self.spec
-       
+
        if (spec.satisfies('^intel-mkl')):
            args = ['-DDLAF_WITH_MKL=ON']
        else:
            args = ['-DDLAF_WITH_MKL=OFF']
+           lapack_name = spec['lapack'].libs.ld_flags.split()[1]
+           blas_name = spec['blas'].libs.ld_flags.split()[1]
+           lapack_libs = spec['lapack'].prefix.lib
+           blas_libs = spec['blas'].prefix.lib
            args.append('-DLAPACK_TYPE=Custom')
-           args.append('-DLAPACK_LIBRARY=-L{0} -lblas -llapack'.format(spec['lapack'].prefix.lib))
+           args.append('-DLAPACK_LIBRARY=-L{0} {1} -L{2} {3}'.format(lapack_libs, lapack_name, blas_libs, blas_name))
 
        if '+cuda' in spec:
            args.append('-DDLAF_WITH_CUDA=ON')
-           #args.append('-DCUDA_HOME={0}'.format(spec['cuda'].prefix))
 
        if self.run_tests:
            args.append('-DDLAF_WITH_TEST=ON')
-           args.append('-DDLAF_INSTALL_TESTS=ON')
-       else:
+        else:
            args.append('-DDLAF_WITH_TEST=OFF')
-           args.append('-DDLAF_INSTALL_TESTS=OFF')
 
        if '+doc' in spec:
            args.append('-DBUILD_DOC=on')
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -30,7 +30,7 @@ class DlaFuture(CMakePackage):
     def cmake_args(self):
        spec = self.spec
 
-       if (spec.satisfies('^intel-mkl')):
+       if '^mkl' in spec:
            args = ['-DDLAF_WITH_MKL=ON']
        else:
            args = ['-DDLAF_WITH_MKL=OFF']
@@ -42,7 +42,7 @@ class DlaFuture(CMakePackage):
 
        if self.run_tests:
            args.append('-DDLAF_WITH_TEST=ON')
-        else:
+       else:
            args.append('-DDLAF_WITH_TEST=OFF')
 
        if '+doc' in spec:
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -25,7 +25,7 @@ class DlaFuture(CMakePackage):
     depends_on('blaspp')
     depends_on('lapackpp')
     depends_on('hpx@1.4.0: cxxstd=14 networking=none')
-    depends_on('cuda', when='cuda=True')
+    depends_on('cuda', when='+cuda')
 
     def cmake_args(self):
        spec = self.spec
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -16,23 +16,41 @@ class DlaFuture(CMakePackage):
 
     version('develop', branch='master')
 
-    variant('gpu', default=False,
+    variant('cuda', default=False,
             description='Use the GPU/cuBLAS back end.')
-
-    # Until mpich is default comment this out
+    variant('test', default=False,
+            description='Run unit tests.')
+    variant('doc', default=False,
+            description='Build documentation.')
+    
     #depends_on('mpi@3:')
-    depends_on('mpich')
-    depends_on('mkl')
+    depends_on('mpi')
+    depends_on('blas')
+    depends_on('lapack')
     depends_on('blaspp')
     depends_on('lapackpp')
-    depends_on('hpx cxxstd=14 networking=none')
-    depends_on('cuda', when='gpu=True')
+    depends_on('hpx@1.4.0 cxxstd=14 networking=none')
+    depends_on('cuda', when='cuda=True')
 
     def cmake_args(self):
        spec = self.spec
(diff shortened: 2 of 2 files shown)


spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,41 @@ class DlaFuture(CMakePackage):
...
-    depends_on('mpich')
-    depends_on('mkl')
+    depends_on('mpi')
+    depends_on('blas')
+    depends_on('lapack')
     depends_on('blaspp')
     depends_on('lapackpp')
-    depends_on('hpx cxxstd=14 networking=none')
-    depends_on('cuda', when='gpu=True')
+    depends_on('hpx@1.4.0 cxxstd=14 networking=none')
```
albestro, line 31: This stick us to `1.4.0`, but I would say that we require **AT LEAST** 1.4.0.

I know that something is going on with the HPX API, so maybe the check may become a bit more complicated than simply 1.4.0+.

@teonnik have a better overview than me on this.

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,41 @@ class DlaFuture(CMakePackage):
 
     version('develop', branch='master')
 
-    variant('gpu', default=False,
+    variant('cuda', default=False,
             description='Use the GPU/cuBLAS back end.')
-
-    # Until mpich is default comment this out
+    variant('test', default=False,
+            description='Run unit tests.')
```
albestro, line 21: I'm not really sure it is the right way of doing it.

See:
- [Build-time tests doc](https://spack.readthedocs.io/en/latest/packaging_guide.html#build-time-tests)
- [CMakePackage doc](https://spack.readthedocs.io/en/latest/build_systems/cmakepackage.html?highlight=cmakepackage#testing)

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,41 @@ class DlaFuture(CMakePackage):
...
+    variant('doc', default=False,
+            description='Build documentation.')
+    
     #depends_on('mpi@3:')
-    depends_on('mpich')
-    depends_on('mkl')
+    depends_on('mpi')
+    depends_on('blas')
+    depends_on('lapack')
```
albestro, line 28: I'm a bit concerned on this, I think we should depend just in blaspp/lapackpp and use the one used by them.

spack/packages/dla-future/package.py
teonnik: > I know that something is going on with the HPX API, so maybe the check may become a bit more complicated than simply 1.4.0+.

1.4.0+ should be OK for now. When HPX 1.5 is released, we will bump that to 1.5.0. So the line can be changed to :
`depends_on('hpx@1.4.0: cxxstd=14 networking=none')`

spack/packages/dla-future/package.py
teonnik: Yes, this variant is not needed, users can control testing with `spack install --test=root`. We can remove the variant and the `+test` check in `cmake_args()` and replace it with this:

```
args = [-DDLAF_WITH_TEST'={0}'.format('ON' if self.run_tests else 'OFF'), -DDLAF_INSTALL_TESTS'={0}'.format('ON' if self.run_tests else 'OFF')]
```

spack/packages/dla-future/package.py
teonnik: Depending on `blaspp` and `lapackpp` for now is probably sufficient. We can add explicit blas and lapack dependence if we start depending directly on a BLAS/LAPACK library directly. However, I'm OK with leaving them too if the package still works, the concretizer should be able to provide a consistent dependency tree regardless.

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,41 @@ class DlaFuture(CMakePackage):
...
+       if (spec.satisfies('^intel-mkl')):
+           args = ['-DDLAF_WITH_MKL=ON']
+       else:
+           args = ['-DDLAF_WITH_MKL=OFF']
+           args.append('-DLAPACK_TYPE=Custom')
+           args.append('-DLAPACK_LIBRARY=-L{0} -lblas -llapack'.format(spec['lapack'].prefix.lib))
+
+       if '+cuda' in spec:
            args.append('-DDLAF_WITH_CUDA=ON')
+           args.append('-DCUDA_TOOLKIT_ROOT_DIR:STRING={0}'.format(spec['cuda'].prefix))
```
teonnik, line 46: Are we sure we need this line? Recent versions of cuda's spack package set the `CUDA_HOME` variable when building, would that be sufficient?

Sely85: You're right, I've tried to substitute the -DCUDA_TOOLKIT_ROOT_DIR with -DCUDA_HOME and it does compile (it does not pass tests but maybe we can think about that later). I'm going to fix it, thanks!

spack/packages/dla-future/package.py
Sely85: I have tried and it works perfectly without these dependencies, as you both suggested. I will remove them. Thanks.

spack/packages/dla-future/package.py
Sely85: true, so that we can run the install with `--run_tests`. @sanciop and I remove the variant `+test` and modified the argument accordingly. is it ok?

teonnik: Yes, this is OK!

spack/packages/dla-future/package.py
teonnik: Here I was actually wondering whether it will work without either `CUDA_HOME` or `CUDA_TOOLKIT_ROOT_DIR`? I meant to say that with recent spack versions, the `cuda` package sets the `CUDA_HOME` variable at build time, if that is the case, there shouldn't be a need to set it in our `package.py`. Would you be able to confirm if this is true?

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,40 @@ class DlaFuture(CMakePackage):
...
     #depends_on('mpi@3:')
-    depends_on('mpich')
-    depends_on('mkl')
+    depends_on('mpi')
     depends_on('blaspp')
     depends_on('lapackpp')
-    depends_on('hpx cxxstd=14 networking=none')
-    depends_on('cuda', when='gpu=True')
+    depends_on('hpx@1.4.0: cxxstd=14 networking=none')
+    depends_on('cuda', when='cuda=True')
```
albestro, line 28: Can you confirm that it is possible to check boolean variants like this?

From the CLI documentation I would say no, but maybe it is different with `depends_on`
```
variants:
      +variant                      enable <variant>
      -variant or ~variant          disable <variant>
      variant=value                 set non-boolean <variant> to <value>
      variant=value1,value2,value3  set multi-value <variant> values
```

Sely85: You're right, sorry. I forgot to try again the `+cuda` option, after moving to variant. I'm going to fix it, thanks.

spack/packages/dla-future/package.py
Sely85: I'm still not sure it will work on Power cluster, but as soon as we close the PR #197 I will try it.

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,40 @@ class DlaFuture(CMakePackage):
...
+       if (spec.satisfies('^intel-mkl')):
+           args = ['-DDLAF_WITH_MKL=ON']
+       else:
+           args = ['-DDLAF_WITH_MKL=OFF']
+           args.append('-DLAPACK_TYPE=Custom')
+           args.append('-DLAPACK_LIBRARY=-L{0} -lblas -llapack'.format(spec['lapack'].prefix.lib))
+
+       if '+cuda' in spec:
            args.append('-DDLAF_WITH_CUDA=ON')
+           #args.append('-DCUDA_HOME={0}'.format(spec['cuda'].prefix))
```
albestro, line 42: This comment seem superfluous.

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,40 @@ class DlaFuture(CMakePackage):
...
+
+       if self.run_tests:
+           args.append('-DDLAF_WITH_TEST=ON')
+           args.append('-DDLAF_INSTALL_TESTS=ON')
```
albestro, line 46: I think it would be enough to just generate the tests with `-DDLAF_WITH_TEST=on` and it should not be necessary to also install them.

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,40 @@ class DlaFuture(CMakePackage):
...
+    
     #depends_on('mpi@3:')
-    depends_on('mpich')
-    depends_on('mkl')
+    depends_on('mpi')
     depends_on('blaspp')
     depends_on('lapackpp')
-    depends_on('hpx cxxstd=14 networking=none')
-    depends_on('cuda', when='gpu=True')
+    depends_on('hpx@1.4.1: max_cpu_count=128 cxxstd=14 networking=none')
```
albestro, line 27: I see that you opened the issue #248 about HPX version. I'm going to look into it. 😉

spack/packages/dla-future/package.py
```diff
@@ -29,24 +29,25 @@ class DlaFuture(CMakePackage):
...
            args = ['-DDLAF_WITH_MKL=ON']
        else:
            args = ['-DDLAF_WITH_MKL=OFF']
+           lapack_name = spec['lapack'].libs.ld_flags.split()[1]
+           blas_name = spec['blas'].libs.ld_flags.split()[1]
+           lapack_libs = spec['lapack'].prefix.lib
+           blas_libs = spec['blas'].prefix.lib
            args.append('-DLAPACK_TYPE=Custom')
-           args.append('-DLAPACK_LIBRARY=-L{0} -lblas -llapack'.format(spec['lapack'].prefix.lib))
+           args.append('-DLAPACK_LIBRARY=-L{0} {1} -L{2} {3}'.format(lapack_libs, lapack_name, blas_libs, blas_name))
```
albestro, line 42: Wouldn't be possible to have something like this?

```
-DLAPACK_LIBRARY='{} {}'".format(spec['lapack'].libs.ld_flags, spec['blas'].libs.ld_flags)
```

Sely85: sure

spack/packages/dla-future/package.py
albestro: I'm still not able to reproduce the problem in #248, so currently HPX 1.4.0 seems to work correctly with our code.
On the other side, HPX 1.5.0 requires some changes (see #249) so I would add the constraint on the upper version.

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,37 @@ class DlaFuture(CMakePackage):
...
-    depends_on('cuda', when='gpu=True')
+    depends_on('hpx@1.4.0:1.4.1 max_cpu_count=128 cxxstd=14 networking=none')
+    depends_on('cuda', when='+cuda')
 
     def cmake_args(self):
        spec = self.spec
-       args = ['-DDLAF_WITH_MKL=ON']
 
-       if '+gpu' in spec:
+       if (spec.satisfies('^intel-mkl')):
```
teonnik, line 33: This is better written as `if '^mkl' in spec:`

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,37 @@ class DlaFuture(CMakePackage):
...
-    # Until mpich is default comment this out
     #depends_on('mpi@3:')
-    depends_on('mpich')
-    depends_on('mkl')
+    depends_on('mpi')
     depends_on('blaspp')
     depends_on('lapackpp')
-    depends_on('hpx cxxstd=14 networking=none')
-    depends_on('cuda', when='gpu=True')
+    depends_on('hpx@1.4.0:1.4.1 max_cpu_count=128 cxxstd=14 networking=none')
```
albestro, line 27: I think `max_cpu_count` constraint is too strict.

This would not allow to depend on an hpx installation that has been compiled with any other value for this variant. 

`spack spec dla-future ^ hpx@1.4.1 max_cpu_count=256`

Let's say that on my particular platform I need to build hpx with `max_cpu_count=256`, this constraint would not allow me to do it.

albestro:
An additional note, since we are providing the spack package, which I think it is really nice, it may be worth mentioning it in the README with a very quick start introduction on how to install our library with it.

e.g.
```
spack repo add .
spack install dla-future
```

spack/packages/dla-future/package.py
Sely85: Right.

Sely85:
> An additional note, since we are providing the spack package, which I think it is really nice, it may be worth mentioning it in the README with a very quick start introduction on how to install our library with it.
> 
> e.g.
> 
> ```
> spack repo add .
> spack install dla-future
> ```

Good point. I have added a few sentences. Would you mind checking them, please?

README.md
```diff
@@ -13,6 +13,19 @@
 - [GoogleTest](https://github.com/google/googletest) (optional; bundled) - unit testing
 - Doxygen (optional) - documentation
 
+## How to install DLA-Future with spack
+
+We provide a spack package DLA-Future that can be easily add to your own spack as follows:
```
teonnik, line 18: add -> added

README.md
```diff
@@ -13,6 +13,19 @@
...
+
+`spack repo add spack`
```
teonnik, line 20: `spack repo add <dlaf_root>/spack`

README.md
```diff
@@ -13,6 +13,19 @@
...
+
+This will add a new repository with namespace `dlaf`.
+
+The command to install the library for example on an HPC cluster will be similar to:
+
+`spack install -v --test=root dla-future ^mkl ^hpx max_cpu_count=256`
```
teonnik, line 26: I would rephrase this to just :

Example installation: `spack install -v dla-future ^intel-mkl`.

Note that `^mkl` doesn't work because `mkl` is a virtual package.

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,37 @@ class DlaFuture(CMakePackage):
...
     version('develop', branch='master')
 
-    variant('gpu', default=False,
+    variant('cuda', default=False,
             description='Use the GPU/cuBLAS back end.')
+    variant('doc', default=False,
+            description='Build documentation.')
 
-    # Until mpich is default comment this out
     #depends_on('mpi@3:')
```
teonnik, line 23: I think this comment can be removed.

README.md
albestro, line 26:
```suggestion
`spack install -v --test=root dla-future ^intel-mkl ^hpx max_cpu_count=256`
```

Are you sure this command works? AFAIK `mkl` is a virtual dependency, so you cannot use it as dependency, instead you have to specify a package providing it.

I think it would be nice to give the user the most basic way to start using our library, and I think this looks like the most basic one (I'm not really sure about `-v`). But, since it is not just a simple `spack install dla-future`, I would evaluate specifying that the `HPX` option `max_cpu_count` must be set accordingly to the platform quickly explaining how.

spack/packages/dla-future/package.py
```diff
@@ -16,23 +15,37 @@ class DlaFuture(CMakePackage):
...
-    depends_on('mpich')
-    depends_on('mkl')
+    depends_on('mpi')
```
albestro, line 24: Do we still need the previous commented line?

README.md
albestro: I didn't get any notification about your comments while I was commenting. We commented on the same things.

About this, in addition to the virtual dependency problem, I quickly thought about having the simplest form `spack install dla-future`. The problem with this, IMHO, is that hpx `max_cpu_count` option is quite important for making it work. At the same time, I know that it should be up to HPX.

Sely85: Your both right, sorry. I tried yesterday and notice that it doesn't work since, as you said, it's only a virtual package, but I forgot to fix it here. Thanks. 

Regarding the max_cpu_count I have added another line with a small explanation below. What do you think?

spack/packages/dla-future/package.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2013-2020 Lawrence Livermore National Security, LLC and other
+#Copyright 2013-2020 Lawrence Livermore National Security, LLC and other
```
albestro, line 1: was it intended?

README.md
```diff
@@ -15,16 +15,19 @@
...
 
-The command to install the library for example on an HPC cluster will be similar to:
+Example installation:
 
-`spack install -v --test=root dla-future ^mkl ^hpx max_cpu_count=256`
+`spack install dla-future ^intel-mkl`
 
+Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the platform, as it represents the maximum number of OS-threads. 
+
+`spack install -v dla-future ^intel-mkl ^hpx max_cpu_count=256`
```
albestro, line 30: `-v` option looks superfluous

README.md
```diff
@@ -13,6 +13,22 @@
 - [GoogleTest](https://github.com/google/googletest) (optional; bundled) - unit testing
 - Doxygen (optional) - documentation
 
+## How to install DLA-Future with spack
+
+We provide a spack package DLA-Future that can be easily added to your own spack as follows:
+
+`spack repo add <dlaf root>/spack`
```
albestro, line 20: I would standardize the `<dlaf root>`: here we are using it like this, while later in the document we use `$DLAF_ROOT`.

teonnik: Alright, lets stick to `$DLAF_ROOT` then.

README.md
teonnik: LGTM

PR #234, opened 2020-05-26
cuBLAS executor

teonnik:
This PR integrates cuBLAS and HPX. A small miniapp illustrates the usage of the API. There is a decent amount of documentation on various design decisions and potential improvements.

This PR needs more testing before it can be merged but preliminary tests of the miniapp produce correct results. However, feedback is much appreciated!


teonnik:
TODO:
- [ ] Complex number handling
- [ ] Convenient API based on MemoryView
- [ ] HPX / CUDA back-end based on the oversubscription model
- [ ] HPX / CUDA back-end based on the array of futures model
- [ ] Refactor out a `cuda_executor` from the `cublas_executor`

include/dlaf/cuda/executor.h
```diff
@@ -0,0 +1,220 @@
...
+///
+/// It is unclear at the moment which approach is the best. According to John 1) and 3) ought to have
+/// very similar performance. The two approaches are also available to MPI, performance comparisons
+/// there did not produce a clear winner which appears to confirm John's statement.
+///
+/// Most CUDA functions are thread safe, that includes Streams and CUBLAS handles. The executor accepts a
+/// `pool_executor` which specifies the host threads on which the CUBLAS calls are scheduled. There is a
+/// price to pay for potentially sharing the CUBLAS handles between multiple threads:
+///
+///  - `cudaSetDevice()` needs to precede the CUBLAS call to make it's handle valid on that thread
```
rasolca, line 139: That's not a problem and anyway always needed in hpx since the device set is thread local.

include/dlaf/cuda/executor.h
```diff
@@ -0,0 +1,220 @@
...
+///  - a mutex is needed to ensure the event is scheduled after the call and there is nothing in-between
+///
+/// We can eliminate both of these if we can dedicate a host thread to a particular device and only
+/// execute CUBLAS / CUDA calls there. There are two approaches to achieve that with HPX:
+///
+/// a) separate pools with a single host threads per device
```
rasolca, line 145: Problem: requires a specific pool coonfiguration...

include/dlaf/cuda/executor.h
```diff
@@ -0,0 +1,220 @@
...
+/// b) force bind tasks to OS threads
```
rasolca, line 146: Requires a specific scheduler...

include/dlaf/cuda/executor.h
```diff
@@ -0,0 +1,220 @@
...
+///
```
rasolca, line 147: option c) if this in incorporated in HPX, an helper thread (as the one used for hpx::cout) can be used.

include/dlaf/cuda/executor.h
teonnik: Did you read through the end of the comment? I mention that...

include/dlaf/cuda/executor.h
teonnik: What do you mean by a helper thread? Do you mean a HPX task or a separate OS thread?

include/dlaf/cuda/executor.h
teonnik: I discuss that further down in the comment.

include/dlaf/cuda/executor.h
teonnik: With the oversubscription approach, we will be able to call `cudaSetDevice()` just once for the dedicate OS thread, instead of calling it every time a kernel is launched (like now). That's the difference.

include/dlaf/cuda/pool.h
```diff
@@ -0,0 +1,131 @@
...
+    //
+    // [1]: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#stream-and-event-behavior
+    cudaSetDevice(device);
+    for (int i_stream = 0; i_stream < num_streams; ++i_stream) {
+      cudaStream_t stream;
+      DLAF_CUDA_CALL(cudaStreamCreateWithFlags(&stream, cudaStreamNonBlocking));
+      cublasHandle_t handle;
+      DLAF_CUBLAS_CALL(cublasCreate(&handle));
+      DLAF_CUBLAS_CALL(cublasSetStream(handle, stream));
+      handles_arr_.push_back(handle);
```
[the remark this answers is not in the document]
teonnik: Having a separate handle per stream avoids this:

> - set handle stream

which I think is a win.

teonnik:
> The main drawback of this approach is that it introduce some latency:
GPU finish computation -> CPU discover -> launch new kernel -> GPU execute kernel (compared to GPU finish computation -> GPU execute new kernel, if the new kernel is launched on the same stream or using events to sync the streams.)

To which approach are you comparing, there are a few? Do you mean compare to the oversubscription approach with a separate pool for each device?

> Anyway I have a question:
GPU management and executors is something that HPX should provide. If this is a test to experiment various approach and to use it until HPX directly provides them it is fine, on the other hand I would like to avoid to have to maintain an extension of HPX for the time being...

Ideally yes, but we don't know when that's going to happen, if we are to make progress on our GPU algorithms, we need this. When/If HPX provides the executors, we can remove our implementations. The PR strives to have an API compatible with what a future HPX implementation might look like.

> However a few thoughts and ideas that came to my mind:
>
>    have a single executor for cuda and cublas (I don't see the reason to have two different set of streams)

What I was planning to do is to have a separate `cuda_pool` and `cublas_pool` as well as a `cuda_executor` and a `cublas_executor`. The `cublas_*` variants are built using the `cuda_` variants. My reasoning for the separation is that the function actually launching the kernels will be different. Also, the data stored in the executor will be different. For example a `cuda_executor` doesn't have to store a pointer mode.

> how does this mechanism work with the tile mechanism? 

Tighter integration is on the TODO list but overall, it works as any other `hpx::async()` call

> what would be the preferred way to schedule two kernels on the same stream without getting back to the host?

A `cublas_executor` stores a handle which is associated to a particular stream. To schedule two kernels on the same stream, just use the same executor. To schedule on different streams, use different executors.

include/dlaf/cuda/executor.h
```diff
@@ -0,0 +1,220 @@
...
+  //
+  // The event indicates the the function `f` has completed. The handle may be shared by mutliple host
+  // threads, the mutex is here to make sure no other CUBLAS calls or events are scheduled between the
+  // call to `f` and it's corresponding event.
+  {
+    static hpx::lcos::local::mutex mt;
+    std::lock_guard<hpx::lcos::local::mutex> lk(mt);
+    DLAF_CUBLAS_CALL(cublasSetPointerMode(handle, pointer_mode));
+    DLAF_CUBLAS_CALL(f(handle, std::forward<Ts>(ts)...));
+    DLAF_CUDA_CALL(cudaEventRecord(event, stream));
```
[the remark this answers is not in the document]
teonnik: Yes, that would be better in principle. However, I'd rather not have the mutex at all by implementing the oversubscription approach I mention in the docs.

include/dlaf/cuda/executor.h
rasolca: Yes, I used these two comments as notes, but I forgot to remove them before submitting the review...
Anyway what I mean is that both approaches require a specific HPX setup, which may not be optimal for the application using DLAF.

include/dlaf/cuda/executor.h
rasolca: These are extra OS threads that HPX uses for specific operations.

teonnik:
After further discussion with John, we've decided to implement further back-ends at his upstream HPX branch. This branch will remain with a compatible API but other back-end implementations will not be added. The current implementation is sufficient for the time being and will allow us to write GPU algorithms and build APIs compatible with future upstream implementations. However we do not commit to maintain CUDA executors forever, once we switch to a version of HPX that provides executors, we can remove our implementation.

include/dlaf/cuda/pool.h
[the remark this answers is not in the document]
rasolca: cublasSetStream adds no overhead; it is just setting the id of stream that will be used for next cublas calls.
On the other hand "cublasCreate allocates some internal resources" https://docs.nvidia.com/cuda/cublas/index.html#cublascreate

include/dlaf/cuda/executor.h
rasolca: I would not consider a problem anyway:
cudaSetDevice() "should be considered a very low overhead call."
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g159587909ffa0791bbe4b40187a4c6bb

rasolca:
> > The main drawback of this approach is that it introduce some latency:
> > GPU finish computation -> CPU discover -> launch new kernel -> GPU execute kernel (compared to GPU finish computation -> GPU execute new kernel, if the new kernel is launched on the same stream or using events to sync the streams.)
> 
> To which approach are you comparing, there are a few? Do you mean compare to the oversubscription approach with a separate pool for each device?

Sorry here the question was missing...I wanted to know if you checked if it was considered. 
I know that with HPX is not trivial, but John had some ideas at some point and maybe he shared them with you. (I'm comparing to static scheduling.)

> > Anyway I have a question:
> > GPU management and executors is something that HPX should provide. If this is a test to experiment various approach and to use it until HPX directly provides them it is fine, on the other hand I would like to avoid to have to maintain an extension of HPX for the time being...
> 
> Ideally yes, but we don't know when that's going to happen, if we are to make progress on our GPU algorithms, we need this. When/If HPX provides the executors, we can remove our implementations. The PR strives to have an API compatible with what a future HPX implementation might look like.

As discussed, as long as the plan is to go back in HPX is fine.

> > However a few thoughts and ideas that came to my mind:
> > have a single executor for cuda and cublas (I don't see the reason to have two different set of streams)
> 
> What I was planning to do is to have a separate `cuda_pool` and `cublas_pool` as well as a `cuda_executor` and a `cublas_executor`. The `cublas_*` variants are built using the `cuda_` variants. My reasoning for the separation is that the function actually launching the kernels will be different. Also, the data stored in the executor will be different. For example a `cuda_executor` doesn't have to store a pointer mode.
> 
> > how does this mechanism work with the tile mechanism?
> 
> Tighter integration is on the TODO list but overall, it works as any other `hpx::async()` call

I disagree: with CPU code is easy to keep the tile alive until the end of the tasks, how do you plan to keep the tile alive until the end of the GPU kernel?

> > what would be the preferred way to schedule two kernels on the same stream without getting back to the host?
> 
> A `cublas_executor` stores a handle which is associated to a particular stream. To schedule two kernels on the same stream, just use the same executor. To schedule on different streams, use different executors.

Ok, I missed the part in which the handle was assigned to the executor when created and not on execution.
Anyway if cuda and cublas executors are different how can I get the same stream?

include/dlaf/cuda/pool.h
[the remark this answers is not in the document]
teonnik: `cublasCreate()` is only called at the creation of the pool. Usually, a single pool suffices for the entire duration of the program and can be used by multiple executors. We are essentially trading, memory overhead for some execution (perhaps minimal) overhead. I haven't benchmarked `cublasSetStream()` and I am not confident whether it "adds no overhead"... I'd expect there is at least an atomic operation or some mutex involved because multiple threads may call the function.

biddisco:
Just to add a few thoughts. Yes - we do want to support all this in HPX and I don't really see the need for separate cuda/cublas pools or executors since we can use overloading of some functions to distinguish between a cuda kernel launch and a cublas call. (We kind of do that already in the cuda_future_helper code in the  cuda branch in hpx). Converting the cuda_future_helper into an executor and adding support for cuda_event types (rather than callbacks) gets us half way there, then copying the hpx::mpi polling method to set cuda futures means that you can do away with custom pools and reuse the mechanism we have in place - just create a cuda/cublas pool at startup and enable polling. Teodor and I discussed oversubscription and future setting/polling etc and we'll work on this stuff over the next week or two.

include/dlaf/cuda/executor.h
teonnik: This is similar to my concern with `cublasSetStream()`, I am aware that the docs consider it a very low overhead call, but I haven't benchmarked it and I'd expect at least an atomic or a mutex involved.

include/dlaf/cuda/executor.h
biddisco: Currently at startup, hpx creates 2 threads for IO (if HPX_WITH_IO_POOL is ON which it is by default) and 2 threads for timing if (HPX_WITH_TIMER_POOL is on, which it is by default). These threads mostly sleep, but occasionally wake up. The IO threads are for hpx::cout as raffaele states and the timer threads are there to support the `std::` calls like invoke after or invoke when or something (I forget the API), but some tasks can be set to run only after some time has elapsed. Nobody ever uses them.

In fact we have resolved internally at some point to remove these 'hardcoded' threads and instead make them integrated with the resource_partitioner so that the user can enable/disable them at startup or even on demand, but instantiating an io pool or timer pools. This would then be identical to the oversubscription method discussed and could be mimicked for the cuda/mpi polling features.

include/dlaf/cuda/executor.h
rasolca: cudaSetDevice() is thread local, why should a mutex or an atomic operation be needed?

teonnik: https://github.com/illuhad/hipSYCL/issues/87

rasolca: They have 7-10 calls per kernel launch, in our case it will be just once per kernel.
Moreover one of the developer stated:
> While that many calls to cudaSetDevice are certainly a little bizarre, I actually don't think they are related to your observed kernel launch latency."

include/dlaf/cuda/executor.h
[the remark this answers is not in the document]
biddisco: "The main drawback of this approach is that it introduce some latency:
GPU finish computation -> CPU discover -> launch new kernel -> GPU execute kernel (compared to GPU finish computation -> GPU execute new kernel, if the new kernel is launched on the same stream or using events to sync the streams.)"

Note that in the hpx cuda_helper and cublas helper code, we provide both an `hpx::async(...)` call that inserts a callback (or in future an event) into the stream, and an `hpx::apply(...)` that implements the one-sided executor model where you fire and forget and do not get a future back from it. These can be mixed freely, so you can invoke N kernels using apply and then get a future back from the last one (or some other significant one), this allows you to avoid any cpu synchronization and leave it to the cuda stream management for kernels or cublas calls that don't need any cpu synchronization.

Anyway, I'm upgrading all the code to use an executor and adding @teonnik 's cuda event handling in too. I'll keep you updated with progress.

misc/runtimes.md
```diff
@@ -0,0 +1,61 @@
+Design rationale:
+
+There are a few alternative HPX/CUDA implementation options:
+
+1) The approach taken here inserts an event just after a CUBLAS function is called and uses it to
+   track when the function has finished. Each task calls `hpx::util::yield_while()` and
+   `cudaEventQuery()` to infrom HPX's scheduler when it's is ready.
```
Sely85, line 7:
```suggestion
   `cudaEventQuery()` to infrom HPX's scheduler when it is ready.
```

misc/runtimes.md
```diff
@@ -0,0 +1,61 @@
...
+
+ - `cudaSetDevice()` needs to precede the CUBLAS call to make it's handle valid on that thread
+ - a mutex is needed to ensure the event is scheduled after the call and there is nothing in-between
+
+We can eliminate both of these if we can dedicate a host thread to a particular device and only
+execute CUBLAS / CUDA calls there. There are two approaches to achieve that with HPX:
+
+a) separate pools with a single host threads per device
+b) force bind tasks to OS threads
+c) TLS (thread local storage)-like approach (TODO)
```
msimberg, line 43: I'm just catching up with the discussions here (even though they are a few months old already, I hope there isn't too much missing from this document).

I think this third option may be the easiest. This can easily be achieved with a pool of streams/handles per worker thread even without HPX TLS by using `hpx::get_os_thread_count`/`hpx::get_worker_thread_num` (or the pool-specific equivalents). The downside is that this leads to a lot of streams (num worker threads * num streams per worker thread). and while I know that in practice there's no advantage to have more than some unspecified number of streams (the CUDA runtime doesn't give you more concurrency than that), I don't know if having far too many streams actually makes things worse.

include/dlaf/memory/memory_chunk.h
```diff
@@ -131,18 +127,15 @@ class MemoryChunk {
 private:
   void deallocate() {
     if (allocated_) {
-#ifdef DLAF_WITH_CUDA
       if (device == Device::CPU) {
-        DLAF_CUDA_CALL(cudaFreeHost(ptr_));
+        // FIXME: Pinned host memory DLAF_CUDA_CALL(cudaFreeHost(ptr_));
```
msimberg, line 131: Yes! This is a must eventually. Has there been any discussion about passing something context/executor/device-like explicitly to the various solvers and copy?

msimberg:
I've pushed quite a significant cleanup. Quite a few of the points discussed here are not really finalized and I'll try to comment on some of them here.

- As mentioned earlier this should 100% go into HPX eventually. It's however a lot easier to experiment and figure out the requirements in one repo, and I'd like to upstream this only when we're sure about the requirements.
- This now uses the event polling built into HPX's schedulers. I don't think there's a good reason to support different mechanisms just for the sake of it, but if someone thinks this is important that's possible as well. I find it unlikely that one mechanism is going to perform significantly better in one situation but significantly worse in another.
- Regarding the current approach requiring synchronization via the host: see previous comment. I don't believe the current executors/futures are appropriate for doing that and I don't think the benefit would be worth the effort. If that's something we want to explore it should be done based on senders/receivers.
- Tile lifetime management is a big issue. I see two potential approaches: 1. make cublas wrappers blocking (does not require a special executor), or 2. extend the lifetime of the tiles by returning them from the cublas wrappers. I've gone for approach 2. (not in this PR, but in separate changes for the GPU trsm) since it avoids having to yield in the wrappers. It also lets most of the future logic be hidden in the executor.

In general I'd like to move forward with this PR even if it's not 100% there yet, and keep iterating afterwards. That would let me build on the triangular solver off master, instead of in branches.

This currently relies on https://github.com/STEllAR-GROUP/hpx/pull/5003 in HPX.

I've so far benchmarked this comparing the `yield_while` based approach with different numbers of streams against the scheduler-based polling. I did not run the full range of benchmarks with the scheduler-based polling.

image

There is practically no difference between the different approaches. The only blip is that for block size 128 using the default stream and no locking performs slightly worse. This may be due to internal (OS) locks in the CUDA runtime, whereas using HPX locks may avoid contention there and allow other threads to run. The difference is very small though. I've excluded results using the callback approach, since they were terrible. I'm still unsure if that's inherent to the method or if I've made a mistake somewhere. Given that it's deprecated don't think it's worth pursuing that approach further.

This PR still needs a cleanup of unused functionality. I've left it there for the time being until we're on the same page about how to continue with this.

Things I still want to try out (preferably in separate PRs):
- Priorities: I've added support for priorities but not benchmarked it. My first try lead to problems with cublas handles. They're potentially not as thread safe as they should be.
- Thread local streams/handles/locks.
- Moving stream pool/handle pool creation and handling out of the executors. Having it in the executor is not practical for real applications. We should not be creating handles and streams for each invocation of an algorithm. I'll open a separate issue about this.

rasolca:
@msimberg I'm a bit puzzled about the results. How can you reach >6 Tflop/s per node on daint? Is it single precision?

msimberg:
> @msimberg I'm a bit puzzled about the results. How can you reach >6 Tflop/s per node on daint? Is it single precision?

Indeed. I only ran the benchmarks for one precision. Sorry I didn't specify that upfront.

include/dlaf/memory/memory_chunk.h
msimberg: I've removed these changes from this PR.

misc/runtimes.md
msimberg: I'll fix typos in this document later on. Do we still want to document all the potential approaches, or should I update this to describe only what is currently done?

include/dlaf/cuda/executor.h
msimberg: I think the having specific pools for this is something that we don't want to look into any further.

teonnik:
@msimberg Thanks for pushing this PR over the finish line! :) 

> This now uses the event polling built into HPX's schedulers. I don't think there's a good reason to support different mechanisms just for the sake of it, but if someone thinks this is important that's possible as well. I find it unlikely that one mechanism is going to perform significantly better in one situation but significantly worse in another.

Based on your data, I agree! The event polling mechanism is more complex internally than the `yield_while()` mechanism but since John has already implemented it and there is no performance difference, we should just use the upstream approach. We should update the `spack` recipe to require building `hpx` with `async_cuda` when we are built with cuda as the component is now required. Also, I added some temporary branches/versions in the recipe that you may remove (`cublas`).

> I've excluded results using the callback approach, since they were terrible. I'm still unsure if that's inherent to the method or if I've made a mistake somewhere. Given that it's deprecated don't think it's worth pursuing that approach further.

Agreed, although there is a version of this with a non-deprecated API (`cudaLaunchHostFunc()`), but perhaps we can leave that out for now.

misc/runtimes.md
teonnik: I don't have a strong preference either way. Perhaps you can elaborate on the approach we chose and just mention the rest as bullet-points without going into details?

include/dlaf/cuda/executor.h
teonnik: Yes, the docs can be fixed here. The TLS approach seems more reasonable to me anyway, if we do decide to explore new approaches in future.

msimberg:
> @msimberg Thanks for pushing this PR over the finish line! :)
> 
> > This now uses the event polling built into HPX's schedulers. I don't think there's a good reason to support different mechanisms just for the sake of it, but if someone thinks this is important that's possible as well. I find it unlikely that one mechanism is going to perform significantly better in one situation but significantly worse in another.
> 
> Based on your data, I agree! The event polling mechanism is more complex internally than the `yield_while()` mechanism but since John has already implemented it and there is no performance difference, we should just use the upstream approach. We should update the `spack` recipe to require building `hpx` with `async_cuda` when we are built with cuda as the component is now required. Also, I added some temporary branches/versions in the recipe that you may remove (`cublas`).

Agreed that it's more complex, but it's less code in DLAF ;) Especially if we get rid of the RAII struct for enabling polling there's not much code needed in DLAF for synchronizing (well, none once the executor moves as well). Yep, I'll update the recipe as necessary. I still want https://github.com/STEllAR-GROUP/hpx/pull/5003 in HPX master before this goes in so that your CI doesn't have to track a random feature branch.
 
> > I've excluded results using the callback approach, since they were terrible. I'm still unsure if that's inherent to the method or if I've made a mistake somewhere. Given that it's deprecated don't think it's worth pursuing that approach further.
> 
> Agreed, although there is a version of this with a non-deprecated API (`cudaLaunchHostFunc()`), but perhaps we can leave that out for now.

Indeed. I'd still ike to try that out, but I want to make that change in HPX and then we can simply swap `get_future_with_event` with `get_future_with_callback`.

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

@haampie I've fixed the commit used on the CUDA builds to a specific hash from HPX master. I suppose that's preferable to just specifying literally `master`, or how does the CI deal with rebuilds of the base images?

haampie:
It should automatically rebuild the base image automatically :)

msimberg:
bors try

msimberg:
bors try

msimberg:
I've done a few more benchmarks. The plot is getting a bit unwieldy but hopefully the main points can be seen. The second plot is just zoomed in on the small block sizes.

I've added three new lines:
- `blocking-10-streams`: This does essentially `hpx::cuda::experimental::detail::get_future_with_event(stream).wait()` in the cuBLAS wrapper. Uses the normal host executors. 10 total streams with round robin and locking. This is worse than the previous results.
- `blocking-10-streams-prio`: Same as above, but with high priority executors as for the CPU version. This is worse than the previous results.
- `scheduler-polling-10-streams-per-thread`: 10 thread local streams used in round robin (one cublas handle per thread). This is better than the previous results, including the first results in https://github.com/eth-cscs/DLA-Future/pull/234#issuecomment-732944597.

The thread local streams seem to make the biggest difference. The results in https://github.com/eth-cscs/DLA-Future/pull/234#issuecomment-732944597 are also from using thread local streams, but host executors and blocking in the wrappers. Unfortunately I don't seem to have the raw data from this one anymore.

image
image

I'll clean this up to use thread local streams (I have that only locally right now), and do some more detailed benchmarks on that specific variation (different numbers of streams, priorities).

msimberg:
Two more plots. The first is a comparison of different numbers of thread local streams, with and without priorities. `noprio-normal-n` means one executor with n normal priority streams per HPX worker thread. ` prio-normal-n-high-m` means two executors, one with n normal priority streams per HPX worker thread and m high priority streams per HPX worker thread. The normal and high priority executors are used as in the CPU version. In general the differences are not big. Anything that has more than one normal priority and more than one high priority stream performs better.

image

Second is the same plot as earlier, but with 3 normal priority/3 high priority streams per HPX worker thread. The approach from above is clearly best. However, the majority of the performance difference does not actually come from optimizing those parameters, but simply from not creating/destroying the streams and handles in the algorithm (I emulate this for the benchmarks with function local static executors, but resolving #306 would mean doing this completely outside the algorithms). Not having to do the initialization seems to save around 10 ms.

image

With this I think this is ready for review. What's clearly missing are thorough tests, but I'd leave that for when this gets upstreamed to HPX, and trust that tests for the algorithms that will be built on top of this will test the executors properly. Please let me know if you think this is a bad idea.

msimberg:
Two small additions:

1. There was discussion on whether the CUDA and cuBLAS executors should be separate or the same. I also came to the conclusion that they should be separate since a CUDA executor shouldn't require creating cuBLAS handles (even if this could be hidden from the user).
2. I've had two slightly different variations of the stream and handle pools. The first had the handle pool take and hold a stream pool. This is slightly more encapsulated and easier for the user. However, I settled on a second approach in the end where the two are completely independent, and the user instead supplies a stream (from anywhere) when asking for a handle from the pool. This adds slightly more responsibility to the user, but makes the two pools more independent and allows using the same handle pool with different stream pools. I'd appreciate feedback and preferences on this.

msimberg:
bors try

msimberg:
bors try

miniapp/miniapp_cublas.cpp
```diff
@@ -0,0 +1,72 @@
...
+        // Note: This doesn't lead to a race condition because this executes
+        // after the `cublasDaxpy()`.
+        thrust::host_vector<double> y_h = y;
+        double sum_of_elems = 0;
+        for (double e : y_h)
+          sum_of_elems += e;
+
+        std::cout << "result : " << sum_of_elems << std::endl;
+      })
+      .get();
```
rasolca, line 65: I think that it is better to separate this statement to improve readability.

rasolca (approved):
LGTM

miniapp/miniapp_cublas.cpp
msimberg: Good point. Will update.

msimberg: Done.

msimberg:
bors try

msimberg:
bors try

msimberg:
By the way, is this waiting for more reviews (@albestro, @teonnik?) or for me to merge it?

albestro:
Sorry, I missed it.

I started giving a quick look, the first thing that caught my attention is HPX version. I see that both in CMake and in spack package the requirement is 1.5.0, while there are changes in the CI that asks for 1.5.1.

Is 1.5.1 required for your work with cublas?

msimberg:
> Sorry, I missed it.
> 
> I started giving a quick look, the first thing that caught my attention is HPX version. I see that both in CMake and in spack package the requirement is 1.5.0, while there are changes in the CI that asks for 1.5.1.
> 
> Is 1.5.1 required for your work with cublas?

Good questions. For cuBLAS HPX master is required. The changes to the dockerfiles for using 1.5.1 are strictly unnecessary, and I can remove them if you'd like. I just updated them since 1.5.1 is available. Note that it's only the non-CUDA builds that use 1.5.1. The CUDA builds specify a commit from master.

The basic functionality for CUDA polling is already in 1.5.0, but the change required to make dataflow is only on master.

albestro:
> > Sorry, I missed it.
> > I started giving a quick look, the first thing that caught my attention is HPX version. I see that both in CMake and in spack package the requirement is 1.5.0, while there are changes in the CI that asks for 1.5.1.
> > Is 1.5.1 required for your work with cublas?
> 
> Good questions. For cuBLAS HPX master is required. The changes to the dockerfiles for using 1.5.1 are strictly unnecessary, and I can remove them if you'd like. I just updated them since 1.5.1 is available. Note that it's only the non-CUDA builds that use 1.5.1. The CUDA builds specify a commit from master.
> 
> The basic functionality for CUDA polling is already in 1.5.0, but the change required to make dataflow work with CUDA/cuBLAS-like executors is only on master.

I don't know exactly how to deal with it. In general I'm ok updating for the latest HPX version available, but I would be for testing the entire library with a specific version. Even multiple ones, e.g.  both 1.5.0 and 1.5.1, but both tested on the entire library assuring that given an HPX version everything works correctly). Let's see also what others think about it.

msimberg:
> > > Sorry, I missed it.
> > > I started giving a quick look, the first thing that caught my attention is HPX version. I see that both in CMake and in spack package the requirement is 1.5.0, while there are changes in the CI that asks for 1.5.1.
> > > Is 1.5.1 required for your work with cublas?
> > 
> > 
> > Good questions. For cuBLAS HPX master is required. The changes to the dockerfiles for using 1.5.1 are strictly unnecessary, and I can remove them if you'd like. I just updated them since 1.5.1 is available. Note that it's only the non-CUDA builds that use 1.5.1. The CUDA builds specify a commit from master.
> > The basic functionality for CUDA polling is already in 1.5.0, but the change required to make dataflow work with CUDA/cuBLAS-like executors is only on master.
> 
> I don't know exactly how to deal with it. In general I'm ok updating for the latest HPX version available, but I would be for testing the entire library with a specific version. Even multiple ones, e.g. both 1.5.0 and 1.5.1, but both tested on the entire library assuring that given an HPX version everything works correctly). Let's see also what others think about it.

That's a good point. In this case I think the difference between 1.5.0 and 1.5.1 is not really worth testing. In general it would be nice to be able to test multiple versions easily. I suppose it's a question for @haampie how much of the dockerfiles is reasonable and possible to further abstract away. Once HPX 1.6.0 is out that'll be the required version at least when CUDA is enabled, and I doubt that you'll want different minimum version requirements depending on which features are enabled.

haampie:
It's not hard to add an additional build job for HPX at a fixed commit, assuming the cmake flags and everything don't depend on the version. If so, feel free to add new build + run steps with the env variables changed a bit :). But if you want it dynamic where HPX is built at the latest master before building DLA-Future, that would require some changes to the dockerfiles.

include/dlaf/cuda/executor.h
```diff
@@ -0,0 +1,195 @@
...
+  bool operator!=(StreamPool const& rhs) const noexcept {
+    return !(*this == rhs);
+  }
+};
+
+/// An executor for cuBLAS functions. Uses handles and streams from the given
+/// HandlePool and StreamPool. A cuBLAS function is defined as any function that
+/// takes a cuBLAS handle as the first argument. The executor inserts a cuBLAS
+/// handle into the argument list, i.e. a handle should not be provided at the
+/// apply/async/dataflow invocation site.
```
teonnik, line 137: This comment seems misplaced. This is the CUDA executor.

teonnik (commented):
LGTM! Some minor comments.

include/dlaf/cuda/executor.h
msimberg: Indeed, good catch!

msimberg:
bors try

albestro (approved):

rasolca:
I think that testing with the last HPX release (1.5.1) is OK even if DLAF works even with previous release (>= 1.5.0).
We can introduce more tests with different versions of the dependencies for DLAF releases (or weekly based on master if needed).

PR #235, opened 2020-05-26
Doxygen refactoring for GitHub Action

albestro:
Allow to generate the documentation:
- as part of the cmake project
- independently without having to configure the project

The idea is to add in this PR the related GitHub action.


albestro:
I've fixed some deprecation that I found using Doxygen 1.8.6 on my machine, but there are more.

Moreover, I'd like to highlight that a positive feedback by `Docs` does not really mean anything for now (we have a lot of warnings and it does not return an error in any case).

We may evaluate to parse and filter the output of doxygen to check if there is something weird going on. But, I think it may be the topic for another PR 😉

rasolca:
@albestro Note that the scope of this test is to check if the doc can be built. Its content cannot be automatically checked (E.g. some functions may be without comments on purpose).

Issue #226, opened 2020-05-11
Proposal: Split DLAF into `core` and `kernels`


teonnik:
Since we introduced ETI, contrary to the initial goal, some of our incremental compile times have suffered. This is mostly due to `  factorization/mc.cpp` and `solver/mc.cpp`. For example, working on TSGEMM, if I modify the `index2d.h`, `matrix.h` or `tile.h` headers, I trigger recompilation for the `*mc.cpp` files too because they are part of the DLAF target and depend on the mentioned headers. As we add more and more kernels, the problem will become worse as we will have to compile all other kernels even if they aren't used or mentioned in the kernel we are working on. That is why I suggest we split DLAF into two parts: `core` and `kernels`. `kernels` contains `[factorization/solver]/mc.cpp` (and `tsgemm` in the future, etc) and their headers, core contains everything else (Matrix, MatrixView, Tile, Index2D, Size2D, etc : the fundamental types).

An outline for the proposed file structure could look like this:

core/
    |-- include/
    |-- src/
    |-- CMakeLists.txt

kernels/
   |-- include/
   |-- src/
   |-- CMakeLists.txt

We can expose two targets: dlaf.core and dlaf.kernels. Tests that need only link with dlaf.core will benefit greatly from incremental compile times.

Note that we don't have to change the code structure (i.e. namespace, names, etc), just the physical structure of the project.

What do you think?

teonnik:
An alternative, more modest proposal is to keep the current project layout but still define two different CMake targets: `dlaf.core` and `dlaf.kernels` and link our tests / miniapps with them.

rasolca:
I don't understand how the split solves the problem:
if you modify the core, and then you want to compile a test which involves one of the kernels, the full kernel library has to be recompiled too to be able to link the tests.

teonnik:
> if you modify the core, and then you want to compile a test which involves one of the kernels, the full kernel library has to be recompiled too to be able to link the tests.

Yes, in that situation, there isn't any benefit, however in the following is:

1. A test that only depends on core
2. A miniapp that only depends on core  (this may be the case when prototyping a kernel)

rasolca:
Ok, but in this case I would go for the light version (i.e. structure as it is now, multiple targets).
I would anyway still create the target dlaf with everything for external usage.
What about separating the kernels as well? This would help even when developing a kernel.
- dlaf.core
- dlaf.factorization
- dlaf....
@albestro What do you think?

[...]

PR #239, opened 2020-06-04
Split DLAF into OBJECT libraries

teonnik:
This PR addresses : https://github.com/eth-cscs/DLA-Future/issues/226 . In particular, (incremental) compile-times for tests, miniapps and future algorithms depending only on `dlaf.core` is significantly improved. 

DLAF is split into object libraries:
- dlaf.core
- dlaf.factorization
- dlaf.solver

The combined library is still under the `DLAF` target.

DLAF_AddTest was refactored a little bit. In particular the dlaf object library is specified via the `LIBRARIES` parameter of the `DLAF_addTest()` function.

Suggestion:
- Move the CACHE varibales from `DLAF_addTest()` to the top CMakeLists.txt file?

Note: The CMake version was bumped to 3.14 as linking object libraries is supported starting 3.13.

EDIT: The PR also adds an option to build DLAF as a shared library. However, there seems to be a problem with `lapackpp`  (the problem is unrelated to this PR, it happens on master too):

```
/usr/bin/ld: /home/teonnik/software/spack/opt/spack/linux-arch-broadwell/gcc-8.4.0/lapackpp-develop-e5l4clzhnoqu5tmrpydlu65owgzy53jo/lib/liblapackpp.a(potrf.cc.o): relocation R_X86_64_PC32 against symbol `_ZTVN6lapack5ErrorE' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
make[3]: *** [src/CMakeFiles/DLAF.dir/build.make:270: src/libDLAF.so] Error 1
make[2]: *** [CMakeFiles/Makefile2:1299: src/CMakeFiles/DLAF.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:1306: src/CMakeFiles/DLAF.dir/rule] Error 2
make: *** [Makefile:585: DLAF] Error 2
```

### DIFF
--- a/cmake/DLAF_AddTest.cmake
+++ b/cmake/DLAF_AddTest.cmake
@@ -57,72 +57,46 @@ function(DLAF_addTest test_target_name)
     message(FATAL_ERROR "No sources specified for this test")
   endif()
 
-  ### Check if the test use MPI
-  set(_DLAF_MPI_MAINS "MPI;MPIHPX")
   set(IS_AN_MPI_TEST FALSE)
-  if (DLAF_AT_USE_MAIN IN_LIST _DLAF_MPI_MAINS)
+  set(IS_AN_HPX_TEST FALSE)
+  if (NOT DLAF_AT_USE_MAIN)
+    set(_gtest_tgt gtest)
+  elseif (DLAF_AT_USE_MAIN STREQUAL PLAIN)
+    set(_gtest_tgt gtest_main)
+  elseif (DLAF_AT_USE_MAIN STREQUAL HPX)
+    set(_gtest_tgt DLAF_gtest_hpx_main)
+    set(IS_AN_HPX_TEST TRUE)
+  elseif (DLAF_AT_USE_MAIN STREQUAL MPI)
+    set(_gtest_tgt DLAF_gtest_mpi_main)
     set(IS_AN_MPI_TEST TRUE)
+  elseif (DLAF_AT_USE_MAIN STREQUAL MPIHPX)
+    set(_gtest_tgt DLAF_gtest_mpihpx_main)
+    set(IS_AN_MPI_TEST TRUE)
+    set(IS_AN_HPX_TEST TRUE)
+  else()
+    message(FATAL_ERROR "USE_MAIN=${DLAF_AT_USE_MAIN} is not a supported option")
   endif()
 
-  if (DLAF_AT_USE_MAIN)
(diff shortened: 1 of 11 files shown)


teonnik:
bors try

albestro:
General question (more a curiosity of mine) about target naming.

When DLAF project is exported, IIRC, CMake exports it in a "namespace" (which is more an unwritten convention) with name `dlaf::<target>`. The current proposal is to have `<target>` to be:
- `dlaf` the all-in-one
- `dlaf.*` for specific ones, e.g. `dlaf.core` 

I'm just wondering if there is any "standard" guideline about naming of these targets. If I'm not wrong about this, with the current proposal the user of the library will have to depend on a cmake target named `dlaf::dlaf` or `dlaf::dlaf.core`. About the former one, I'm pretty sure to have seen it in various projects, I don't have any experience about the latter one.

teonnik:
> I'm just wondering if there is any "standard" guideline about naming of these targets.

I'm not aware of a standard guideline. Using `dlaf::` seems fine to me.
> If I'm not wrong about this, with the current proposal the user of the library will have to depend on a cmake target named dlaf::dlaf or dlaf::dlaf.core

Note that we don't export OBJECT libraries, users will only see the complete library : dlaf::DLAF. I would suggest we keep it that way as external users will not benefit from separate OBJECT libraries.

cmake/DLAF_AddTest.cmake
```diff
@@ -194,10 +147,25 @@ function(DLAF_addTest test_target_name)
     list(APPEND _TEST_ARGUMENTS ${_HPX_EXTRA_ARGS_LIST})
   endif()
 
+  ### Test executable target
+  add_executable(${test_target_name} ${DLAF_AT_SOURCES})
+  target_link_libraries(${test_target_name}
+    PRIVATE
+      ${_gtest_tgt}
+      DLAF_test
```
albestro, line 155: Shouldn't you add something like this?

`$<$<BOOL:${IS_AN_MPI_TEST}>:MPI::MPI_CXX>`

teonnik: MPI is always linked because of `dlaf.prop`, we shouldn't need this.

teonnik:
I've added an auxiliary object library. Looking forward to feedback.

rasolca:
bors try

PR #241, opened 2020-06-09
Silence pushd/popd

haampie:
Didn't realize pushd and popd always output the stack, so better silence them to reduce noise in CI


haampie:
bors try

haampie:
bors try

PR #244, opened 2020-06-15
Add NOT_YET_IMPLEMENTED assertion

### LINKED ISSUE #243: Common way to manage NOT IMPLEMENTED cases
I think it is worth adding a common way to manage all the cases that are not yet implemented.

albestro:
Close #243 

The starting point is a very trivial implementation.

As soon as we agree on a solution, I will add few lines of documentation and I will employ it in the codebase (e.g. `Auxiliary::norm`).

### DIFF
--- a/include/dlaf/auxiliary/norm/mc.tpp
+++ b/include/dlaf/auxiliary/norm/mc.tpp
@@ -39,7 +39,7 @@ dlaf::BaseType<T> Auxiliary<Backend::MC>::norm(comm::CommunicatorGrid grid, comm
     case lapack::Norm::Two:
     case lapack::Norm::Inf:
     case lapack::Norm::Fro:
-      DLAF_ASSERT(false, "not yet implemented", norm_type);
+      DLAF_UNIMPLEMENTED(norm_type);
       return {};
     case lapack::Norm::Max:
       switch (uplo) {
@@ -47,7 +47,7 @@ dlaf::BaseType<T> Auxiliary<Backend::MC>::norm(comm::CommunicatorGrid grid, comm
           return internal::mc::norm_max_L(grid, rank, A);
         case blas::Uplo::Upper:
         case blas::Uplo::General:
-          DLAF_ASSERT(false, "not yet implemented", norm_type, uplo);
+          DLAF_UNIMPLEMENTED(norm_type, uplo);
           return {};
         default:
           return {};
--- a/include/dlaf/common/assert.h
+++ b/include/dlaf/common/assert.h
@@ -114,4 +114,4 @@ inline void do_assert(bool expr, const common::internal::source_location& loc, c
 #define DLAF_ASSERT_HEAVY(Expr, ...)
 #endif
 
-#define DLAF_NOT_YET_IMPLEMENTED(...) DLAF_ASSERT(dlaf::internal::not_yet_implemented, __VA_ARGS__)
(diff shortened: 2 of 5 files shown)


teonnik:
This looks good to me but I don't understand what the problem with `DLAF_ASSERT(false, "Not implemented sth sth ...")` is?

albestro:
> but I don't understand what the problem with
> 
> ```
> DLAF_ASSERT(false, "Not implemented sth sth ...")
> ```
> 
> is?

I thought the usage was a bit "forced". But, if we agree that it is preferable to use a `DLAF_ASSERT` instead of a dedicated one for the `NOT_YET_IMPLEMENTED`, we can just skip this. No problem.

What do you think @rasolca @Sely85 ?

teonnik:
If you are referring to this:

```
switch (norm_type) {
    case lapack::Norm::One:
    case lapack::Norm::Two:
    case lapack::Norm::Inf:
    case lapack::Norm::Fro:
      DLAF_ASSERT(false, "not yet implemented", lapack::norm2str(norm_type));
      return {};
    case lapack::Norm::Max:
      switch (uplo) {
        case blas::Uplo::Lower:
          return internal::mc::norm_max_L(grid, rank, A);
        case blas::Uplo::Upper:
        case blas::Uplo::General:
          DLAF_ASSERT(false, "not yet implemented", lapack::norm2str(norm_type), blas::uplo2str(uplo));
          return {};
        default:
          return {};
      }
    default:
      return {};
  }
```

this can also be written as 

DLAF_ASSERT(norm_type == lapack::Norm::Max, "Not yet implemented!", norm_type);
DLAF_ASSERT(uplo == blas::Uplo::Lower, "Not yet implemented!", uplo);

Sely85:
> > but I don't understand what the problem with
> > ```
> > DLAF_ASSERT(false, "Not implemented sth sth ...")
> > ```
> > is?
> 
> I thought the usage was a bit "forced". But, if we agree that it is preferable to use a `DLAF_ASSERT` instead of a dedicated one for the `NOT_YET_IMPLEMENTED`, we can just skip this. No problem.
> 
> What do you think @rasolca @Sely85 ?

I would prefer to use a dedicated one because it would be easier to "spot", to avoid confusing preconditions check with features that we still need to implement for example. 
By the way, both approaches work well.

albestro:
@rasolca @msimberg what do you think about this? should we close it or merge it?

msimberg:
IMO this is quite nice, since it's not a lot of code but it adds quite a bit of clarity. The macro should probably add the message "Not yet implemented" for the macro to make sense. `UNIMPLEMENTED` instead of `NOT_YET_IMPLEMENTED`?

I think using `DLAF_ASSERT` is also fine, this is not such a big deal.

albestro:
This should sum up all our point of views.

I also applied it in the code, since I think that in one way or another, at the end of this PR, these code lines should be unified.

teonnik (approved):

rasolca:
bors try

rasolca (approved):

Sely85 (approved):
LGTM

rasolca (changes requested):
It is not needed to use blaspp utilities to print the enums. We provide output operators for them (currently even two copies (see #281), which will be fixed by #282).

albestro:
@rasolca as soon as we merge #282 I will migrate completely to them 😉

albestro:
bors try

after rebase where operator<< for blas and lapack enums is available

rasolca (approved):

rasolca:
bors try

PR #245, opened 2020-06-16
Miniapp triangular solver

albestro:
Introduce miniapp for benchmarking the triangular solver.

The starting point is a
- left side
- lower
- non transpose
- non unit diagonal

and if not necessary, I would not expose these options.

It's just a starting point for now, I have to work on it.

### DIFF
--- a/miniapp/miniapp_triangular_solver.cpp
+++ b/miniapp/miniapp_triangular_solver.cpp
@@ -77,6 +77,11 @@ int hpx_main(hpx::program_options::variables_map& vm) {
   const auto diag = blas::Diag::NonUnit;
   const T alpha = 2.0;
 
+  double m = A.size().rows();
+  double n = b.size().cols();
+  auto add_mul = n * m * m / 2;
+  const double total_ops = dlaf::total_ops<T>(add_mul, add_mul);
+
   using dlaf::matrix::test::getLeftTriangularSystem;
   std::function<T(const GlobalElementIndex&)> setter_A, setter_b, expected_b;
   std::tie(setter_A, setter_b, expected_b) =
@@ -106,12 +111,14 @@ int hpx_main(hpx::program_options::variables_map& vm) {
 
     auto elapsed_time = timeit.elapsed();
 
-    // TODO compute gigaflops
+    // compute gigaflops
+    double gigaflops = total_ops / elapsed_time / 1e9;
 
-    // TODO print benchmark results
+    // print benchmark results
     if (0 == world.rank())
       std::cout << "[" << run_index << "]"
                 << " " << elapsed_time << "s"
+                << " " << gigaflops << "GFlop/s"
                 << " " << A.size() << " " << A.blockSize() << " " << comm_grid.size() << " " << b.size()
                 << " " << b.blockSize() << " " << hpx::get_os_thread_count() << std::endl;
 
--- a/miniapp/miniapp_triangular_solver.cpp
+++ b/miniapp/miniapp_triangular_solver.cpp
@@ -71,6 +71,12 @@ int hpx_main(hpx::program_options::variables_map& vm) {
   MatrixType a(GlobalElementSize{opts.m, opts.m}, TileElementSize{opts.mb, opts.mb}, comm_grid);
   MatrixType b(GlobalElementSize{opts.m, opts.n}, TileElementSize{opts.mb, opts.nb}, comm_grid);
 
+  auto sync_barrier = [&] () {
+    ::waitall_tiles(a);
+    ::waitall_tiles(b);
+    MPI_Barrier(world);
+  };
+
   const auto side = blas::Side::Left;
   const auto uplo = blas::Uplo::Lower;
   const auto op = blas::Op::NoTrans;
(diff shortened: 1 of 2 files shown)


miniapp/miniapp_triangular_solver.cpp
```diff
@@ -0,0 +1,210 @@
...
+  // options
+  using namespace hpx::program_options;
+  options_description desc_commandline("Usage: " HPX_APPLICATION_STRING " [options]");
+
+  // clang-format off
+  desc_commandline.add_options()
+    ("matrix-size",       value<SizeType>()->default_value(4096),  "Matrix size.")
+    ("block-size",        value<SizeType>()->default_value(256),   "Block cyclic distribution size.")
+    ("result-cols",       value<SizeType>()->default_value(512),   "Matrix size.")
+    ("result-block-cols", value<SizeType>()->default_value(512),   "Block cyclic distribution size.")
```
Sely85, line 148: What about adding the matrix the values referred to? For example "Matrix A size" and "Matrix b size", or "Input matrix size" and "Output matrix size".

teonnik (commented):
LGTM, except:

```
add_executable(miniapp_triangular_solver miniapp_triangular_solver.cpp)
target_link_libraries(miniapp_triangular_solver PRIVATE DLAF DLAF_test)
target_add_warnings(miniapp_triangular_solver)
```

the dependency on `DLAF_test` should be remove. The function `getLeftTriangularSystem()` can be moved to a header in `include/dlaf`.

A question I have is whether we would want to extend this miniapp to support other combinations of Side, Uplo and Transpose in the future? Currently, only LLN is distributed, once we implement the other variants, would we want separate miniapps for them or would we want to integrate them in this one? If it is the former, I suggest that we append the file name with  `_LLN`.

miniapp/miniapp_triangular_solver.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+  double m = A.size().rows();
+  double n = b.size().cols();
+  auto add_mul = n * m * m / 2;
+  const double total_ops = dlaf::total_ops<T>(add_mul, add_mul);
+
+  using dlaf::matrix::test::getLeftTriangularSystem;
+  std::function<T(const GlobalElementIndex&)> setter_A, setter_b, expected_b;
+  std::tie(setter_A, setter_b, expected_b) =
+      getLeftTriangularSystem<GlobalElementIndex, T>(uplo, op, diag, alpha, A.size().rows());
```
rasolca, line 88: I suspect that setup will hit the subnormal problem as well.

miniapp/miniapp_triangular_solver.cpp
```diff
@@ -0,0 +1,210 @@
...
+}
+
+int hpx_main(hpx::program_options::variables_map& vm) {
+  options_t opts = check_options(vm);
+
+  Communicator world(MPI_COMM_WORLD);
+  CommunicatorGrid comm_grid(world, opts.grid_rows, opts.grid_cols, Ordering::ColumnMajor);
+
+  // Allocate memory for the matrices
+  MatrixType A(GlobalElementSize{opts.m, opts.m}, TileElementSize{opts.mb, opts.mb}, comm_grid);
```
rasolca, line 71: naming conventions: `a`

rasolca (changes requested):
Looks good as a first attempt, but we should pay attention to subnormal ops.

miniapp/miniapp_triangular_solver.cpp
albestro: I renamed the parameters according to the `trsm` function, so all the sizes refers to the B matrix, from which it is possible to infer all other sizes.

albestro:
> the dependency on `DLAF_test` should be remove. The function `getLeftTriangularSystem()` can be moved to a header in `include/dlaf`.

As discussed separately, I reduced at maximum the dependency with `DLAF_test`:
- copied a slightly less general version of the function `getLeftTriangularSystem()` which generates the sample triangular system (now it works just with floating point values and not complex, so it does not depend anymore on `TypeUtilities` for generalized value type management)
- copied in-place the computation of mul/add error (from `TypeUtilties`)

Unfortunately, it was not easily possible to remove the last dependency because checking results requires `CHECK_MATRIX_NEAR` (that it is quite trivial and slack, it may be improved in the future). I added a comment as a reminder.

> A question I have is whether we would want to extend this miniapp to support other combinations of Side, Uplo and Transpose in the future? Currently, only LLN is distributed, once we implement the other variants, would we want separate miniapps for them or would we want to integrate them in this one? If it is the former, I suggest that we append the file name with `_LLN`.

If you agree, I would consider this in the future as soon as we will implement other variants.

miniapp/miniapp_triangular_solver.cpp
albestro: Experimentally, checking with `perf` for subnormal interrupts and by looking at single node performances, it does not suffer (with m=10240, n=10240). I'm checking how values of the reference linear systems are generated (I will share with you the details) and we should be on the safe side.

rasolca (approved):

Issue #212, opened 2020-05-05
Signed and unsigned integers


teonnik:
Various parts of the code mix signed and unsigned integers. Mixing signed and unsigned integer arithmetic leads to lots of warnings or verbosity as casts have to be used. 

An excellent discussion on the issue can be found here: https://github.com/ericniebler/stl2/issues/182 

In particular, this comment: https://github.com/ericniebler/stl2/issues/182#issuecomment-287683189 provides a link to a panel of C++ committee members discussion the issue.

Would we be in favour of using signed types for everything?

teonnik:
Unsigned types:

Pros:

- 1 more bit to play around (2x) more values
- clearer semantics (i.e. we want this to be positive)

Cons:

- mixes poorly with signed types
- wraps around (i.e. -1 is perfectly fine input which gives you the largest unsigned int...) so it doesn't really enforce positive input
- count of things need not be positive
- 2x size only really matters for bitfields (really hard to believe you need the extra bit.... if that's the case, you more likely need a bigger integer type altogether)

rasolca:
I prefer signed integers too if STL is not involved (e.g. when accessing a vector a unsigned index has to be used).
If STL moves to signed integers, it will be the time to stop using unsigned (It will be needed only for bit manipulation (see behaviour of shift `operator>>` for negative numbers...)...)

PR #247, opened 2020-06-23
Reduce usage of unsigned types

teonnik:
This PR attempts to address #212. Uses of unsigned types are limited to interactions with the STL.


teonnik:
bors try

teonnik:
`std::ptrdiff_t` is now aliased to `ssize` (signed size) which is easier to type and understand.  

> In some cases overflow may occur when multiplying 2 ints. (I commented the first I saw, ut there are other cases.)

This is fixed now, if you notice other overflows do let me know.

> Moreover for the sizes in our tests using int might look fine, but in my opinion it introduces a sort of "bad documentation" (in the tests it works so I use this approach...).
E.g.
for (int i = 0; i < mem.size(); ++i) is something I would avoid...

The tests are now using `ssize` instead of `int` to address your comment.

teonnik:
bors try

teonnik:
bors try

rasolca:
bors try

teonnik:
@rasolca the majority of the overflows can be addressed if instead of using `SizeType (= int)` we use `ssize`. The fact that it's aliased to `int` at the moment has a few pros and cons:

pros:

- integrates well with MPI
- a smaller type, this may provide a performance boost if there are multiple nested loops (>5) and there are index-heavy manipulations (we don't have these at the moment and I am not sure if we will)

cons:

- workarounds needed if a certain dimension has >4billion valus: for example a tall-skinny matrix use-case
- potential for overflows as you have pointed out in several cases

Is there a reason why we would want to keep `SizeType`? @rasolca  @albestro @Sely85

rasolca:
> cons:
> 
>     * workarounds needed if a certain dimension has >4billion valus: for example a tall-skinny matrix use-case
This is the reason SizeType has been used, allowing the easy change of the type.

>     * potential for overflows as you have pointed out in several cases
The primary scope of add and mul was to solve this problem. The suppression of warning was only a secondary feature.
 
> Is there a reason why we would want to keep `SizeType`? @rasolca @albestro @Sely85
The reason behind this choice is to be able to differentiate between 2D and 1D sizes.

Now if you want to change SizeType to 64 bits, it can be done; I only ask to:
- run some benchmarks to exclude any loss of performance,
- find a way to handle type conversion for MPI.

Just note that the type name ssize is not conform to the coding style (it should be SSize) (it is not conform to C++ naming conventions either (xyz_t)).

rasolca:
Please note that some changes of other PRs were reverted during the merge (e.g. in the documentation).
We should try to merge this PR soon to avoid to introduce a large number of conflicts again.

teonnik:
This PR is mostly ready. If you notice something out of order, do let me know.

@rasolca  There is no performance degradation when it comes to running all tests on `master` vs `this branch` to address your request for some benchmarks. In case you prefer a benchmark on Piz Daint, what parameters and algorithms do you want tested?

rasolca (approved):

rasolca:
It's fine and almost ready to merge.
Please fix the small formatting issues and then we should be ready to go.

Sely85 (approved):
Looks good

teonnik:
@rasolca I ran clang-format on the files but nothing changed. Perhaps there is a difference between the clang-format version I am using and the one used by the "Check format" service. Mine is `10.0.1`

rasolca:
@teonnik We are using clang-format 8. Unfortunately clang-format is not consistent among versions, so we never updated to avoid problems. If you agree I can commit the fixes.

teonnik:
@rasolca you can go ahead and commit the fixes. I have to be careful to not reformat the files in my other PRs because the discrepancy will keep showing up.

albestro:
@teonnik Does the `SizeType` type will be the unique one to be used for interoperability with both BLAS/LAPACK and MPI?

teonnik:
@albestro Yes, we can cast to int before calling MPI/BLAS/LAPACK wherever necessary. There is a `to_int()` in `types.h`.

rasolca:
Please note that BLASPP and LAPACKPP use a int64 API (internally converted to the underlying library precision).
bors try

PR #251, opened 2020-06-25
Test CI failure in case of error.

rasolca:
Discovered in DLA-Interface that if tests fail the CI doesn't discover them. Let me start with a simple test.


rasolca:
bors try

rasolca:
bors try

rasolca:
bors try

PR #254, opened 2020-06-29
Make CI run in parallel again

haampie:
Removes [resource_group](https://docs.gitlab.com/ee/ci/yaml/#resource_group) s.t. ci runs jobs in parallel again.

### DIFF
--- a/ci/.gitlab-ci.yml
+++ b/ci/.gitlab-ci.yml
@@ -31,7 +31,8 @@ cpu debug build:
     BUILD_IMAGE: $CI_REGISTRY_IMAGE/debug/build:20200625-lapackpp.f878fa
     IMAGE: $CI_REGISTRY_IMAGE/debug/deploy:$CI_COMMIT_SHA
   script:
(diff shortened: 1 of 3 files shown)


haampie:
bors try

haampie:
bors try

albestro:
From what I understand, we are going to loose the caching of intermediate layers because these layers may get updated in parallel by the different slurm jobs that now are not serialized anymore.

It looks good to me.

rasolca:
Are the resource group change and docker cache changes correlated?
As far as I can see only the child pipeline are affected by the removal of the resource group, and I don't understand why the inlined docker cache has been disabled.

haampie:
It's orthogonal. Inline cache requires builds to run on the same node, which used to be the case in the Kubernetes cluster, but now isn't anymore due to different configuration. On Daint there is no caching of docker images, except it downloads the image once for a complete pipeline.

rasolca:
Ok, I didn't get the part in which the kubernetes config was changed as well.

PR #255, opened 2020-06-30
API to Initialize/Finalize and Query MPI thread level

teonnik:
`mpi_serialized()` is useful in `mpi_invoke()` which is useful for #228 .

The PR takes care of the MPI part of : #100

### DIFF
--- a/miniapp/miniapp_cholesky.cpp
+++ b/miniapp/miniapp_cholesky.cpp
@@ -15,6 +15,8 @@
 
 #include "dlaf/auxiliary/mc.h"
 #include "dlaf/communication/communicator_grid.h"
+#include "dlaf/communication/functions_sync.h"
+#include "dlaf/communication/init.h"
 #include "dlaf/factorization/mc.h"
 #include "dlaf/matrix.h"
 #include "dlaf/matrix/copy.h"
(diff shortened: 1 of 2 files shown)


teonnik:
bors try

teonnik:
bors try

miniapp/miniapp_cholesky.cpp
```diff
@@ -148,15 +150,7 @@ int hpx_main(hpx::program_options::variables_map& vm) {
...
-  // Initialize MPI
-  int threading_required = MPI_THREAD_SERIALIZED;
-  int threading_provided;
-  MPI_Init_thread(&argc, &argv, threading_required, &threading_provided);
-
-  if (threading_provided != threading_required) {
-    std::fprintf(stderr, "Provided MPI threading model does not match the required one.\n");
-    MPI_Abort(MPI_COMM_WORLD, 1);
-  }
+  dlaf::comm::mpi_init(argc, argv, dlaf::comm::mpi_thread_level::serialized);
```
albestro, line 153: What makes you prefer a C-like interface over a RAII object?

teonnik: I've been going back and forth on this... I don't mind implementing it as a RAII, the only downside I can think of is forgetting to name the variable: `mpi_raii()`; -> nothing happens here, neither the destructor nor the constructor is called.

albestro: I see your point.

With the C-interface you have to remember to call the "destructor", while with the RAII you have to remember to give it a name, as you said.

I asked just for understanding your design choice 😉

teonnik: I just changed to a RAII type, I don't think the issue I mentioned is likely to happen as we need this just once in main().

teonnik:
bors try

PR #257, opened 2020-07-09
Fixes and changes for dla-future spack package

albestro:
Add missing build dependencies/constraints:
- Add the missing version constraint about `cmake` version ([reference](https://spack.readthedocs.io/en/latest/build_systems/cmakepackage.html#build-system-dependencies))
- Add `doxygen` build requirement when building documentation

Additionally:
- the package file was bad indented, so I quickly fixed it (it should be compliant also to spack rules, I didn't check thoroughly)
- fix project urls (wiki link is empty anyway, but it was not working before)
- use `self.define` and `self.define_from_variant` helpers

### DIFF
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -35,17 +35,17 @@ class DlaFuture(CMakePackage):
 
         # BLAS/LAPACK
         if '^mkl' in spec:
-            args = ['-DDLAF_WITH_MKL=ON']
+            args = [ self.define('DLAF_WITH_MKL', True) ]
         else:
             args = [
-                    '-DDLAF_WITH_MKL=OFF',
-                    '-DLAPACK_TYPE=Custom',
-                    '-DLAPACK_LIBRARY={} {}'.format(
-                        spec['lapack'].libs.ld_flags, spec['blas'].libs.ld_flags)
+                    self.define('DLAF_WITH_MKL', False),
+                    self.define('LAPACK_TYPE', 'Custom'),
+                    self.define('LAPACK_LIBRARY',
+                        ' '.join([spec[dep].libs.ld_flags for dep in ['blas', 'lapack']]))
                    ]
 
         # CUDA
-        args.append(self.define_from_variant("DLAF_WITH_CUDA", 'cuda'))
+        args.append(self.define_from_variant('DLAF_WITH_CUDA', 'cuda'))
 
         # DOC
         args.append(self.define_from_variant('BUILD_DOC', 'doc'))
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -20,6 +20,8 @@ class DlaFuture(CMakePackage):
     variant('doc', default=False,
             description='Build documentation.')
 
+    depends_on('cmake@3.14:', type='build')
+
     depends_on('mpi')
     depends_on('blaspp')
     depends_on('lapackpp')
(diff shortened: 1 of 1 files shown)


spack/packages/dla-future/package.py
```diff
@@ -20,31 +21,36 @@ class DlaFuture(CMakePackage):
...
-           args.append('-DLAPACK_LIBRARY={} {}'.format(spec['lapack'].libs.ld_flags, spec['blas'].libs.ld_flags))
+        # BLAS/LAPACK
+        if '^mkl' in spec:
+            args = ['-DDLAF_WITH_MKL=ON']
+        else:
+            args = [
+                    '-DDLAF_WITH_MKL=OFF',
+                    '-DLAPACK_TYPE=Custom',
+                    '-DLAPACK_LIBRARY={} {}'.format(
+                        spec['lapack'].libs.ld_flags, spec['blas'].libs.ld_flags)
```
albestro, line 44: I checked this because I was concerned about having a not escaped string, but it seems to work correctly.

Do someone of you what spack does?

albestro: This is what `spack-configure-args.txt` contains.

```
-DDLAF_WITH_MKL=OFF -DLAPACK_TYPE=Custom -DLAPACK_LIBRARY=-L/home/ubuntu/spack/opt/spack/linux-ubuntu18.04-broadwell/gcc-7.5.0/openblas-0.3.10-brsannfkoozhl4kjblqysvqgrhuedi7a/lib -lopenblas -L/home/ubuntu/spack/opt/spack/linux-ubuntu18.04-broadwell/gcc-7.5.0/openblas-0.3.10-brsannfkoozhl4kjblqysvqgrhuedi7a/lib -lopenblas -DDLAF_WITH_CUDA:BOOL=OFF -DBUILD_DOC:BOOL=OFF -DDLAF_WITH_TEST:BOOL=OFF
```

If I use it has it is with cmake, I expect (for what concerns `-DLAPACK_LIBRARY`) it ignores everything but the first string without spaces, i.e. `-L/home/ubuntu/spack/opt/spack/linux-ubuntu18.04-broadwell/gcc-7.5.0/openblas-0.3.10-brsannfkoozhl4kjblqysvqgrhuedi7a/lib`

Could it be that it works because this first part of the string is enough and everything else is discovered through other dependencies?

Sely85: What about adding extra quote marks?
Writing in the spack package
`'-DLAPACK_LIBRARY=\'{} {}\''.format(spec['lapack'].libs.ld_flags, spec['blas'].libs.ld_flags`
leads to the following arguments (in the spack-configure-args.txt)
`-DDLAF_WITH_MKL=OFF -DLAPACK_TYPE=Custom -DLAPACK_LIBRARY='-L/m100_work/PROJECTS/spack/spack-0.14/install/linux-rhel7-power9le/gcc-8.4.0/netlib-lapack-3.8.0-wm3og2ntx7xwa7nk4ri6wjsjn23v7x4v/lib64 -llapack -L/m100_work/PROJECTS/spack/spack-0.14/install/linux-rhel7-power9le/gcc-8.4.0/netlib-lapack-3.8.0-wm3og2ntx7xwa7nk4ri6wjsjn23v7x4v/lib64 -lblas' -DDLAF_WITH_TEST=ON`
Will this solve the problem?

albestro: Ok, I did some test.

I tested what it does using `self.define('LAPACK_LIBRARY', ...)` and it does not put any extra quote.

Moreover, I checked and It works correctly because when spack calls cmake, it puts quote around each argument, e.g. `'cmake' '-DVAR=a b c' ..`.

So, everything is correct and works in the right way.

I would just propose to use `self.define`, because it produces a more complete cmake syntax (e.g. `-DLAPACK_LIBRARY:STRING=value`)

PR #261, opened 2020-07-13
Algorithm gen2std eigenproblem (lower case)

Sely85:
Implementation of the algorithm for the transformation from generalized to standard eigenproblem. The algorithm is implemented for CPU, only on local memory so far. This method solves the equation: B= L^(-1) A L^(-H), where L is a lower triangular matrix (obtained with a Cholesky decomposition and A is a Hermitian square matrix).

Tests, using analytical expressions, are implemented both for this algorithm and for the tile wrapper of the HEGST LAPACK function (for this specific case). 

Also, HEMM and HERK BLAS wrapper functions have been added along with their tests. 

Two functions for printing elements of matrices or tiles have been implemented.
The documentation has been fixed in a few files.


include/dlaf/lapack_tile.h
```diff
@@ -66,6 +66,15 @@ void potrf(blas::Uplo uplo, const Tile<T, device>& a) noexcept;
...
 
+/// Reduce a Hermitian definite generalized eigenproblem to standard form
+/// (solving inv(U**H)*A*inv(U) or inv(L)*A*inv(L**H))
+///
+/// @pre a must be a square Hermitian matrix
+/// @pre b must be the triangular factor from the Cholesky factorization of B
+/// @throw std::runtime_error if the tile was not positive definite.
+template <class T, Device device>
+void hegst(int itype, blas::Uplo uplo, const Tile<T, device>& a, const Tile<T, device>& b);
+
```
albestro, line 77: Please, move this (see general comment).

include/dlaf/lapack_tile.tpp
```diff
@@ -43,6 +43,18 @@ dlaf::BaseType<T> lantr(lapack::Norm norm, blas::Uplo uplo, blas::Diag diag,
...
+  if (a.size().rows() != a.size().cols()) {
+    throw std::invalid_argument("Error: POTRF: A is not square.");
+  }
+
+  auto info = lapack::potrf(uplo, a.size().rows(), a.ptr(), a.ld());
+  assert(info >= 0);
+
+  return info;
+}
+
```
albestro, line 57: We want to keep info/non-info variants together, but I would suggest to keep the alphabetic order. So, I would keep this after the non-info variant, but please let me know if you moved it for any specific reason.

include/dlaf/lapack_tile.tpp
```diff
@@ -51,11 +63,13 @@ void potrf(blas::Uplo uplo, const Tile<T, device>& a) noexcept {
 }
 
 template <class T, Device device>
-long long potrfInfo(blas::Uplo uplo, const Tile<T, device>& a) {
-  DLAF_ASSERT(a.size().rows() == a.size().cols(), "POTRF: `a` is not square!", a);
+void hegst(int itype, blas::Uplo uplo, const Tile<T, device>& a, const Tile<T, device>& b) {
```
albestro, line 66: Please, move this (see general comment).

test/include/dlaf_test/util_types.h
```diff
@@ -18,6 +18,7 @@ namespace dlaf_test {
 using ElementTypes =
     ::testing::Types<int, long long, float, double, std::complex<float>, std::complex<double>>;
 using MatrixElementTypes = ::testing::Types<float, double, std::complex<float>, std::complex<double>>;
+// using MatrixElementTypes = ::testing::Types<float, double>;
```
albestro, line 21: spurious (commented) line of code.

test/unit/CMakeLists.txt
```diff
@@ -1,11 +1,11 @@
...
-# Copyright (c) 2018-2019, ETH Zurich
-# All rights reserved.
+#Copyright(c) 2018 - 2019, ETH Zurich
+#All rights reserved.
 #
-# Please, refer to the LICENSE file in the root directory.
-# SPDX-License-Identifier: BSD-3-Clause
+#Please, refer to the LICENSE file in the root directory.
+#SPDX - License - Identifier : BSD - 3 - Clause
 #
```
albestro, line 9: Messed up license header

test/unit/CMakeLists.txt
```diff
@@ -43,13 +43,14 @@ DLAF_addTest(test_matrix
   MPIRANKS 6
 )
 
-# Generic libraries
+#Generic libraries
```
albestro, line 46: I think you have a problem in your way of checking file styles...please, give it a check.

test/unit/eigensolver/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+#
+#Distributed Linear Algebra with Future(DLAF)
+#
+#Copyright(c) 2018 - 2019, ETH Zurich
+#All rights reserved.
+#
+#Please, refer to the LICENSE file in the root directory.
+#SPDX - License - Identifier : BSD - 3 - Clause
+#
```
albestro, line 9: Messed up license header.

test/unit/eigensolver/mc/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+#
+#Distributed Linear Algebra with Future(DLAF)
+#
+#Copyright(c) 2018 - 2019, ETH Zurich
+#All rights reserved.
+#
+#Please, refer to the LICENSE file in the root directory.
+#SPDX - License - Identifier : BSD - 3 - Clause
+#
```
albestro, line 9: here, too.

test/unit/test_lapack_tile.cpp
```diff
@@ -137,3 +138,17 @@ TYPED_TEST(TileOperationsTest, PotrfNonPositiveDefinite) {
     }
   }
 }
+
+TYPED_TEST(TileOperationsTest, Hegst) {
```
albestro, line 142: Please, move this too in the right alphabetic order.

include/dlaf/blas_tile.h
```diff
@@ -26,12 +26,22 @@ template <class T, Device device>
...
+
+/// Performs a rank k update of hermitian (symmetric if T is real) tile @p a.
 template <class T, Device device>
 void herk(blas::Uplo uplo, blas::Op op, BaseType<T> alpha, const Tile<const T, device>& a,
           BaseType<T> beta, const Tile<T, device>& c) noexcept;
 
-/// Performs a triangular solve.
+/// @brief Performs a rank 2k update of hermitian (symmetric if T is real) tile a.
+template <class T, Device device>
+void her2k(blas::Uplo uplo, blas::Op op, BaseType<T> alpha, const Tile<const T, device>& a,
```
albestro, line 41: Please, move this in the right alphabetic order.

teonnik (commented):
I get a few warnings with clang:

```
/home/teonnik/code/dlaf/include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h:54:12: warning: unused variable 'tilerow' [-Wunused-variable]
  SizeType tilerow = mat_a.blockSize().rows();
           ^
/home/teonnik/code/dlaf/include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h:55:12: warning: unused variable 'tilecol' [-Wunused-variable]
  SizeType tilecol = mat_a.blockSize().cols();
```

Other than that LGTM!

A request, can you add a reference to implemented algorithm. For example, if the implementation is based on an algorithm from Raffaele's thesis, it would be nice to have a reference to his thesis, page number, algorithm number, etc. in a comment.

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -0,0 +1,140 @@
...
+      for (SizeType j = k + 1; j < n; ++j) {
+        auto jj = LocalTileIndex{j, j};
+        auto jk = LocalTileIndex{j, k};
+
+        // TRSM
+        hpx::dataflow(executor_hp, hpx::util::unwrapping(tile::trsm<T, Device::CPU>), Left, Lower,
+                      NoTrans, NonUnit, 1.0, mat_l.read(jj), std::move(mat_a(jk)));
+
+        for (SizeType i = j + 1; i < m; ++i) {
+          // Working on trailing matrix...
```
rasolca, line 125: It still working on the panel...

include/dlaf/tile_output.h
```diff
@@ -0,0 +1,30 @@
...
+#pragma once
+
+/// @file
+
+#include "dlaf/tile.h"
+
+namespace dlaf {
+
+template <class T>
+void printElementTile(const Tile<T, Device::CPU>& tile) {
```
rasolca, line 20: I would call it `printElements` as well.

include/dlaf/matrix_output.h
```diff
@@ -0,0 +1,43 @@
...
+      SizeType elcol = icol % blockcol;
+
+      const LocalTileIndex idx = {tilerow, tilecol};
+      auto tile = mat(idx).get();
+
+      std::cout << tile({elrow, elcol}) << " ";
+    }
+    std::cout << " " << std::endl;
+  }
+  std::cout << " " << std::endl;
```
rasolca, line 40: This works only on local matrices.

test/unit/eigensolver/mc/test_genToStd.cpp
```diff
@@ -0,0 +1,75 @@
+//
```
rasolca, line 1: please use `test_gen_to_std.cpp` as filename

test/unit/eigensolver/mc/test_genToStd.cpp
```diff
@@ -0,0 +1,75 @@
...
+using namespace testing;
+
+template <typename Type>
+class EigensolverGenToStdLocalTest : public ::testing::Test {};
+
+TYPED_TEST_SUITE(EigensolverGenToStdLocalTest, MatrixElementTypes);
+
+const std::vector<std::tuple<SizeType, SizeType>> sizes = {
+    {0, 2},                              // m = 0
+    {2, 2}, {5, 5}, {13, 13}, {34, 34},  // m = mb
```
rasolca, line 39: It is not neede to test all these cases with a single tile.

rasolca (changes requested):

test/unit/eigensolver/mc/test_gen_to_std.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+template <typename Type>
+class EigensolverGenToStdLocalTest : public ::testing::Test {};
+
+TYPED_TEST_SUITE(EigensolverGenToStdLocalTest, MatrixElementTypes);
+
+const std::vector<std::tuple<SizeType, SizeType>> sizes = {
+    {0, 2},                     // m = 0
+    {5, 5}, {34, 34},           // m = mb
+    {4, 3}, {16, 10}, {34, 13}  // m != mb
```
rasolca, line 40: I would add e.g. {32, 5} to have a case with more tiles.

rasolca (changes requested):

rasolca:
bors try

test/include/dlaf_test/matrix/util_generic_lapack.h
```diff
@@ -0,0 +1,129 @@
...
+/// A_ij = (i+1)(j+1) (beta*beta*gamma)/(2^(i+j))*exp(I*alpha*(i-j))
+/// or for @p itype = 2 and @p itype = 3:
+/// A_ij = gamma/(2^(i+j))*exp(I*alpha*(i-j))
+/// where alpha = 0 and I = 0 for real types or I is the complex unit for complex types.
+///
+/// The elements of T (@p el_T), a triangular matrix, are computed, for @p itype = 1, as
+/// T_ij = beta/(2^(i-j)) * exp(I*alpha*(i-j))
+/// and, for @p itype = 2 or @p itype = 3, as
+/// T_ij = beta/(2^(-i-j)) * exp(I*alpha*(i-j))
+/// When the matrix is @p Upper row and column indices are swapped.
```
rasolca, line 40: Warning: for complex types you cannot simply swap the indices.

test/include/dlaf_test/matrix/util_generic_lapack.h
```diff
@@ -0,0 +1,129 @@
...
+
+    double i, j;
+    if (uplo == blas::Uplo::Lower) {
+      i = index.row();
+      j = index.col();
+    }
+    else if (uplo == blas::Uplo::Upper) {
+      j = index.row();
+      i = index.col();
+    }
```
rasolca, line 63: In this way you simply return the transposed of T, which is not correct for complex types. (PART 1)
```suggestion
```

test/include/dlaf_test/matrix/util_generic_lapack.h
```diff
@@ -0,0 +1,129 @@
...
+    else if (uplo == blas::Uplo::General) {
+      std::cout << "uplo == General not allowed" << std::endl;
+      std::abort;
+    }
+
+    if (itype == 1) {
+      return TypeUtilities<T>::polar(beta / std::exp2(i - j), alpha * (i - j));
+    }
```
rasolca, line 71: In this way you simply return the transposed of T, which is not correct for complex types. (PART 2)
```suggestion
    if (itype == 1) {
      return TypeUtilities<T>::polar(beta / std::exp2(std::abs(i - j)), alpha * (i - j));
    }
```

test/include/dlaf_test/matrix/util_generic_lapack.h
```diff
@@ -0,0 +1,129 @@
...
+
+/// @file
+
+namespace dlaf {
+namespace matrix {
+namespace test {
+
+using namespace dlaf_test;
+
+/// Returns a tuple of element generators of three matrices T (m x m), A(m x m) and B (m x m).
```
rasolca, line 28: I would add also the condition that A,B and T fulfill.

test/include/dlaf_test/matrix/util_generic_lapack.h
```diff
@@ -0,0 +1,129 @@
...
+/// The elements of T (@p el_T), a triangular matrix, are computed, for @p itype = 1, as
+/// T_ij = beta/(2^(i-j)) * exp(I*alpha*(i-j))
+/// and, for @p itype = 2 or @p itype = 3, as
+/// T_ij = beta/(2^(-i-j)) * exp(I*alpha*(i-j))
+/// When the matrix is @p Upper row and column indices are swapped.
+///
+/// Finally, the elements of B (@p el_b) should be, for @p itype = 1:
+/// B_ij = (gamma)/(2^(i+j))*exp(I*alpha*(i-j))
+/// and for @p itype = 2 or @p itype = 3:
+/// B_ij = ((n-i)*(n-j)*(beta*beta*gamma)/(2^(-i-j))*exp(I*alpha*(i-j))
```
rasolca, line 45: Try to reduce the chance of overflow: (PART 2)
```suggestion
/// B_ij = ((n-i)*(n-j)*(beta*beta*gamma)/(2^(i+j))*exp(I*alpha*(i-j))
```

test/include/dlaf_test/matrix/util_generic_lapack.h
```diff
@@ -0,0 +1,129 @@
...
+    else if (uplo == blas::Uplo::General) {
+      std::cout << "uplo == General not allowed" << std::endl;
+      std::abort;
+    }
+
+    if (itype == 1) {
+      return TypeUtilities<T>::polar(beta / std::exp2(i - j), alpha * (i - j));
+    }
+    else if (itype == 2 || itype == 3) {
+      return TypeUtilities<T>::polar(beta / std::exp2(-i - j), alpha * (i - j));
```
rasolca, line 73: In this way you simply return the transposed of T, which is not correct for complex types. (PART 3)
Try to reduce the chance of overflow: (PART 3)
```suggestion
      return TypeUtilities<T>::polar(beta * std::exp2(std::abs(i - j)), alpha * (i - j));
```

test/include/dlaf_test/matrix/util_generic_lapack.h
```diff
@@ -0,0 +1,129 @@
...
+      return TypeUtilities<T>::element(-9.9, 0);
+
+    double i = index.row();
+    double j = index.col();
+
+    if (itype == 1) {
+      return TypeUtilities<T>::polar(gamma / std::exp2(i + j), alpha * (i - j));
+    }
+    else if (itype == 2 || itype == 3) {
+      return TypeUtilities<T>::polar((n - i) * (n - j) * (beta * beta * gamma) / std::exp2(-i - j),
```
rasolca, line 115: Try to reduce the chance of overflow: (PART 4)
```suggestion
      return TypeUtilities<T>::polar((n - i) * (n - j) * (beta * beta * gamma) / std::exp2(i + j),
```

rasolca (changes requested):

rasolca:
bors try

Sely85:
bors try

Sely85:
bors try

rasolca:
bors try

Sely85:
bors try

rasolca:
I restored the status before the rebase, because something went wrong. (comments were messed up).

test/include/dlaf_test/matrix/util_generic_lapack.h
```diff
@@ -0,0 +1,108 @@
...
+/// @file
+
+namespace dlaf {
+namespace matrix {
+namespace test {
+
+using namespace dlaf_test;
+
+/// Returns a tuple of element generators of three matrices T (n x n), A(n x n) and B (n x n).
+/// The equations solved by this routine are for @p itype = 1
```
rasolca, line 29:
```suggestion
/// It holds, for @p itype == 1:
```

rasolca (changes requested):

include/dlaf/blas_tile.h
```diff
@@ -26,7 +26,17 @@ template <class T, Device device>
 void gemm(blas::Op op_a, blas::Op op_b, T alpha, const Tile<const T, device>& a,
           const Tile<const T, device>& b, T beta, const Tile<T, device>& c) noexcept;
 
-/// Performs a rank k update of hermitian (symmetric if T is real) tile a.
+/// Computes matrix matrix multiplication where matrix @p a is Hermitian.
+template <class T, Device device>
+void hemm(blas::Side side, blas::Uplo uplo, T alpha, const Tile<const T, device>& a,
+          const Tile<const T, device>& b, T beta, const Tile<T, device>& c);
+
+/// @brief Performs a rank 2k update of hermitian (symmetric if T is real) tile a.
```
albestro, line 34: For documentation uniformity, remove `@brief`

include/dlaf/blas_tile.tpp
```diff
@@ -40,6 +40,55 @@ void gemm(blas::Op op_a, blas::Op op_b, T alpha, const Tile<const T, device>& a,
...
+template <class T, Device device>
+void hemm(blas::Side side, blas::Uplo uplo, T alpha, const Tile<const T, device>& a,
+          const Tile<const T, device>& b, T beta, const Tile<T, device>& c) {
+  SizeType m;
+  SizeType n;
+  m = c.size().rows();
+  n = c.size().cols();
+
+  if (side == blas::Side::Left) {
+    DLAF_ASSERT(m == a.size().rows(), "`m` cannot be determined!", m, a);
```
albestro, line 52: Commenting this line for all similar ones.

I'm wondering if it wouldn't be easier to use
`DLAF_ASSERT(m == a.size().rows(), m, a.size());`

My points are:
1. I don't remember what printing a tile means (maybe just shape related properties), but I think it would be easier to read if we report in the message just the values we are checking.
2. Generally, I would not add a "human" message like "`m` cannot be determined!", hoping that the condition, which is reported in the error literally, is clear enough.

include/dlaf/blas_tile.tpp
```diff
@@ -40,6 +40,55 @@ void gemm(blas::Op op_a, blas::Op op_b, T alpha, const Tile<const T, device>& a,
              c.ptr(), c.ld());
 }
 
+template <class T, Device device>
+void hemm(blas::Side side, blas::Uplo uplo, T alpha, const Tile<const T, device>& a,
```
albestro, line 44: I have some concerns about not having `const` all parameters (i.e. `side`, `uplo`, `alpha`, `beta`)...maybe there is a reason not to make them `const`, but I don't see it.

This is valid for all functions in this file, so it is something that may be addressed (if needed) in another PR.

include/dlaf/eigensolver/gen_to_std/mc.tpp
```diff
@@ -0,0 +1,37 @@
...
+  DLAF_ASSERT(matrix::square_blocksize(mat_l), mat_l);
+  DLAF_ASSERT(matrix::local_matrix(mat_a), mat_a);
+  DLAF_ASSERT(matrix::local_matrix(mat_l), mat_l);
+
+  if (uplo == blas::Uplo::Lower)
+    internal::mc::genToStd_L(mat_a, mat_l);
+  else {
+    std::cout << "uplo = Upper not yet implemented" << std::endl;
+    std::abort();
+  }
```
albestro, line 34: Instead of using an if, I think it is generally safer to use a `switch` with enums. In fact, in case you miss any value of the enum, you will get a warning.

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -0,0 +1,121 @@
...
+
+    // Direct transformation to standard eigenvalue problem of the diagonal tile
+    hpx::dataflow(executor_hp, unwrapping(tile::hegst<T, Device::CPU>), 1, Lower, std::move(mat_a(kk)),
+                  std::move(mat_l(kk)));
+
+    if (k != (n - 1)) {
+      for (SizeType i = k + 1; i < m; ++i) {
+        // Working on panel...
+        auto ik = LocalTileIndex{i, k};
+        hpx::dataflow(executor_normal, hpx::util::unwrapping(tile::trsm<T, Device::CPU>), Right, Lower,
```
albestro, line 72: Thanks to the `using` statement at the beginning of the function, you can simply refer to `unwrapping` without fully qualifying it (here and in all other places in this function).

include/dlaf/eigensolver/mc.h
```diff
@@ -0,0 +1,51 @@
...
+
+namespace dlaf {
+
+template <>
+struct Eigensolver<Backend::MC> {
+  /// Reduce a Hermitian definite generalized eigenproblem to standard form, using the factorization
+  /// obtained from potrf (lower matrix from Cholesky decomposition), solving B=inv(L)*A*inv(L**H),
+  /// implementation on local memory.
+  ///
+  /// @param mat_a on entry it contains the square and Hermitian matrix A, on exit the matrix elements
```
albestro, line 26: An hermitian matrix is square by definition.

include/dlaf/lapack_tile.tpp
```diff
@@ -10,6 +10,17 @@
 
 #include "dlaf/common/assert.h"
 
+template <class T, Device device>
+void hegst(int itype, blas::Uplo uplo, const Tile<T, device>& a, const Tile<T, device>& b) {
```
albestro, line 14: Would we want to provide a more self-explaining name and have a more "safe" control logic for this parameter?

I'm thinking of something similar to the other enumeration types used for other BLAS/LAPACK parameters.

If not, I would at least add a `DLAF_ASSERT`/`DLAF_ASSERT_MODERATE` check for validating the input.

include/dlaf/lapack_tile.tpp
```diff
@@ -10,6 +10,17 @@
...
+  DLAF_ASSERT(a.size().rows() == a.size().cols(), "Error: HEGST: matrix A is not square.");
+
+  // itype = 1 to solve inv(U**H)*A*inv(U) or inv(L)*A*inv(L**H))
+  // itype = 2, 3 to solve U*A*U**H or L**H*A*L
+  auto info = lapack::hegst(itype, uplo, a.size().cols(), a.ptr(), a.ld(), b.ptr(), b.ld());
+
+  DLAF_ASSERT(info == 0, "");
```
albestro, line 21: Instead of an empty message, I would report the `info` parameter, otherwise we don't have any information on fault.

include/dlaf/matrix_output.h
```diff
@@ -0,0 +1,46 @@
...
+
+/// @file
+
+#include "dlaf/matrix.h"
+#include "dlaf/util_matrix.h"
+
+namespace dlaf {
+
+template <class T>
+void printElements(Matrix<T, Device::CPU>& mat) {
```
albestro, line 20: I'm not really sure this is a needed function. If you all think we want this, I would propose to at least put it in the `matrix` namespace (and maybe the name can be improved, or better I would go for making it `std::ostream& operator<<`).

IMHO, it is something we don't want to provide in the library, even if it can be useful during debugging (in case I would move it DLAF_test library).

include/dlaf/matrix_output.h
```diff
@@ -0,0 +1,46 @@
...
+  DLAF_ASSERT(matrix::local_matrix(mat), mat);
+
+  SizeType nrow = mat.size().rows();
+  SizeType ncol = mat.size().cols();
+  SizeType blockrow = mat.blockSize().rows();
+  SizeType blockcol = mat.blockSize().cols();
+
+  for (SizeType irow = 0; irow < nrow; ++irow) {
```
albestro, line 28: I think you can use `iterate_range2d` function available in the library.

include/dlaf/tile_output.h
```diff
@@ -0,0 +1,30 @@
...
+#pragma once
+
+/// @file
+
+#include "dlaf/tile.h"
+
+namespace dlaf {
+
+template <class T>
+void printElements(const Tile<T, Device::CPU>& tile) {
```
albestro, line 20: Same thoughts I shared for `printElements` on Matrix.

test/unit/test_blas_tile/test_hemm.h
```diff
@@ -0,0 +1,150 @@
...
+
+  std::stringstream s;
+  s << "HEMM: " << side << ", " << uplo;
+  s << ", m = " << m << ", n = " << n;
+  s << ", lda = " << lda << ", ldb = " << ldb << ", ldc = " << ldc;
+  SCOPED_TRACE(s.str());
+
+  T alpha = TypeUtilities<T>::element(1.2, .7);
+  T beta = TypeUtilities<T>::element(1.1, .4);
+  BaseType<T> gamma = 1.3f;
```
albestro, line 62: `const`?

test/unit/test_blas_tile/test_hemm.h
```diff
@@ -0,0 +1,150 @@
...
+  };
+
+  // Set tile elements.
+  set(a0, el_a);
+  set(b0, el_b);
+  set(c, el_c);
+
+  // Read-only tiles become constant if CT is const T.
+  const Tile<CT, Device::CPU> a(std::move(a0));
+  const Tile<CT, Device::CPU> b(std::move(b0));
```
albestro, line 143: FYI
From what I see, you'd like to initialize readonly `Tile<const T>`, but you have to use a "support" `Tile<T>` (e.g. `a0`), so you can initialize it with the functions above and then you move the support tile to the final read-only one.

An alternative can be doing something like this:

```cpp
auto setup_readonly_tile = [](auto&& values, const auto size, const SizeType ld){
  memory::MemoryView<T, Device::CPU> support_mem(mul(ld, size.cols()));
  Tile<T, Device::CPU> support(size, std::move(support_mem), ld);
  set(support, values);
  return Tile<CT, Device::CPU>(std::move(support));
};

const Tile<CT, Device::CPU> a = setup_readonly_tile(el_a);
const Tile<CT, Device::CPU> b = setup_readonly_tile(el_b);
const Tile<CT, Device::CPU> c = setup_readonly_tile(el_c);
```

*DISCLAIMER* I didn't test it, so I'm not sure it works. But I hope you get the idea.

albestro (commented):
I've given a quick overall view, I didn't go into numeric and algorithmic details yet.

include/dlaf/matrix_output.h
rasolca: ostream operator is already used for matrix info (shape, ...)

test/unit/test_blas_tile/test_hemm.h
rasolca: This is used everywhere in the tile blas and lapack tests. Should be addressed in a new PR...

include/dlaf/matrix_output.h
rasolca: Anyway, this is a utility you want to have in the library, since you don't want to link DLAF_test to the main application.

include/dlaf/blas_tile.tpp
rasolca: This was used previously. It should be addressed in another PR.

Sely85: This is the error printed without the "human" message:
```
[ERROR] /home/lara/Documents/DLAF/DLA-Future_trans2stdeigen_4rebase/include/dlaf/blas_tile.tpp:35 : void dlaf::tile::gemm(blas::Op, blas::Op, T, const dlaf::Tile<const T, device>&, const dlaf::Tile<const T, device>&, T, const dlaf::Tile<T, device>&) [with T = float; dlaf::Device device = (dlaf::Device)0]
m != c.size().rows()
0
size=(0, 0), ld=1
```
What do you think, @albestro? Let's ask @rasolca and @teonnik, as well.
If you agree that this is easier to read, I can modify all similar ASSERT (here or in a different PR?).

include/dlaf/blas_tile.tpp
Sely85: I agree with you: having all parameters `const` looks safer. 
I have modified all the parameters in `blas_tile.tpp` , `lapack_tile.tpp` and related `.h` and it works well. I can move these changes into a new PR if you agree.

include/dlaf/blas_tile.tpp
rasolca: It was already used in many place so it should be updated everywhere. It should be adressed elsewhere to avoid to add clutters to this PR.

include/dlaf/lapack_tile.tpp
rasolca: I would just add a `DLAF_ASSERT` for itype.

include/dlaf/matrix_output.h
```diff
@@ -0,0 +1,46 @@
...
+    SizeType tilerow = irow / blockrow;
+    SizeType elrow = irow % blockrow;
+
+    for (SizeType icol = 0; icol < ncol; ++icol) {
+      SizeType tilecol = icol / blockcol;
+      SizeType elcol = icol % blockcol;
+
+      const LocalTileIndex idx = {tilerow, tilecol};
+      auto tile = mat(idx).get();
```
rasolca, line 37:
```suggestion
      auto& tile = mat.read(idx).get();
```

test/include/dlaf_test/matrix/util_generic_lapack.h
```diff
@@ -0,0 +1,108 @@
...
+/// B_ij = gamma / 2^(i+j) exp(I alpha (i-j))
+/// and for @p itype = 2 or @p itype = 3:
+/// B_ij = ((n-i) (n-j) (beta^2 gamma)) / 2^(i+j) exp(I alpha (i-j))
+/// where @p n is the size of the matrix.
+///
+template <class ElementIndex, class T>
+auto getGenToStdElementSetters(SizeType n, int itype, blas::Uplo uplo, T alpha, T beta, T gamma) {
+  DLAF_ASSERT(uplo == blas::Uplo::Lower || uplo == blas::Uplo::Upper, "Only Upper and Lower supported",
+              uplo);
+  DLAF_ASSERT(itype < 4, "Only itype = 1, 2, 3 allowed", itype);
```
rasolca, line 57: itype should also be > 0

include/dlaf/matrix_output.h
albestro: Ok, I propose to move this work about printing tiles and matrices to a dedicated PR. It is already a quite big PR and there is no need of it here. Do you agree?

Sely85:
bors try

include/dlaf/lapack_tile.tpp
```diff
@@ -13,6 +13,7 @@
 template <class T, Device device>
 void hegst(int itype, blas::Uplo uplo, const Tile<T, device>& a, const Tile<T, device>& b) {
   DLAF_ASSERT(a.size().rows() == a.size().cols(), "Error: HEGST: matrix A is not square.");
+  DLAF_ASSERT(itype > 0 && itype < 4, "Only itype = 1, 2, 3 allowed", itype);
```
albestro, line 16: Just remarking that `DLAF_ASSERT(itype >= 0 && itype <=3, itype);`, IMHO, it would have been more concise and clear.

Sely85: My only concern is that in the description of hegst parameters of the error message, only `int` appears, not also `itype`:
```
[ERROR] /home/lara/Documents/DLAF/DLA-Future-trans2stdeigen/include/dlaf/lapack_tile.tpp:16 : void dlaf::tile::hegst(int, blas::Uplo, const dlaf::Tile<T, device>&, const dlaf::Tile<T, device>&) [with T = float; dlaf::Device device = (dlaf::Device)0]
itype < 0 && itype <= 3
5
```
Is it clear enough in this case?

albestro (changes requested):
In general LGTM.

I gave a quick overview, and the more invasive thing is that I would use more extensively `const` modifier (both variables and parameters).

I haven't gone through all the analytic details of the tests yet.

include/dlaf/blas_tile.tpp
```diff
@@ -40,6 +40,55 @@ void gemm(blas::Op op_a, blas::Op op_b, T alpha, const Tile<const T, device>& a,
...
+          const Tile<const T, device>& b, T beta, const Tile<T, device>& c) {
+  SizeType m;
+  SizeType n;
+  m = c.size().rows();
+  n = c.size().cols();
+
+  if (side == blas::Side::Left) {
```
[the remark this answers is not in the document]
teonnik: I don't like switches: 1) awkward syntax; 2) if you forget a `break;` all hell breaks loose. Why would you prefer them?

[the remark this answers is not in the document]
albestro: I completely agree with your points. What I found safer is that the compiler can warn you about a missing case for the enum.

The alternative I see is `if + [else if] + else`, with the "mandatory" `else` as a catch all. Maybe with `enum class` is even safe to just address all `if/else if cases` (but you may still missing some).

In this case, it is used `if (left) + if (right)` for `enum class Side   : char { Left     = 'L', Right    = 'R' };`, and I would at least change it to `if + else if`.

But well, I'm open to changing my mind 😁 I may be over-concerned about this, and the their awkwardness may easily overcome the added utility. Even more now that we use `enum class` instead of basic `enum`.

Looking forward to read your thoughts about this.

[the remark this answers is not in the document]
teonnik: > What I found safer is that the compiler can warn you about a missing case for the enum.

Yes, that's a good point. I find it less problematic in our case because we have few values/cases per enum (2-3), so I'm less concerned of missing a case.

Whatever is decided, I'm onboard. IMO we shouldn't spend too much time deciding between switch / if-else.

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -0,0 +1,120 @@
...
+          const auto ik = LocalTileIndex{i, k};
+          const auto ij = LocalTileIndex{i, j};
+          hpx::dataflow(executor_normal, unwrapping(tile::gemm<T, Device::CPU>), NoTrans, ConjTrans,
+                        -1.0, mat_a.read(ik), mat_l.read(jk), 1.0, std::move(mat_a(ij)));
+          hpx::dataflow(executor_normal, unwrapping(tile::gemm<T, Device::CPU>), NoTrans, ConjTrans,
+                        -1.0, mat_l.read(ik), mat_a.read(jk), 1.0, std::move(mat_a(ij)));
+        }
+      }
+
+      for (SizeType i = k + 1; i < m; ++i) {
```
albestro, line 94: Here you can easily use `common::iterate_range2d`

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -0,0 +1,120 @@
...
+
+  for (SizeType k = 0; k < n; ++k) {
+    const auto kk = LocalTileIndex{k, k};
+
+    // Direct transformation to standard eigenvalue problem of the diagonal tile
+    hpx::dataflow(executor_hp, unwrapping(tile::hegst<T, Device::CPU>), 1, Lower, std::move(mat_a(kk)),
+                  std::move(mat_l(kk)));
+
+    if (k != (n - 1)) {
+      for (SizeType i = k + 1; i < m; ++i) {
```
albestro, line 68: Here you can easily use `common::iterate_range2d`

include/dlaf/lapack_tile.h
```diff
@@ -27,6 +27,15 @@ namespace tile {
 
 // See LAPACK documentation for more details.
 
+/// Reduce a Hermitian definite generalized eigenproblem to standard form
+/// (solving inv(U**H)*A*inv(U) or inv(L)*A*inv(L**H))
+///
+/// @pre a must be a square Hermitian matrix
```
albestro, line 33: hermitian matrix is enough, it is square by definition.

include/dlaf/lapack_tile.h
```diff
@@ -27,6 +27,15 @@ namespace tile {
 
 // See LAPACK documentation for more details.
 
+/// Reduce a Hermitian definite generalized eigenproblem to standard form
+/// (solving inv(U**H)*A*inv(U) or inv(L)*A*inv(L**H))
```
albestro, line 31: This should cover all cases. Report here an extent from LAPACK documentation

```
 If ITYPE = 1, the problem is A*x = lambda*B*x,
 and A is overwritten by inv(U**H)*A*inv(U) or inv(L)*A*inv(L**H)

 If ITYPE = 2 or 3, the problem is A*B*x = lambda*x or
 B*A*x = lambda*x, and A is overwritten by U*A*U**H or L**H*A*L.

 B must have been previously factorized as U**H*U or L*L**H by ZPOTRF.
```

include/dlaf/lapack_tile.tpp
```diff
@@ -10,6 +10,18 @@
 
 #include "dlaf/common/assert.h"
 
+template <class T, Device device>
+void hegst(const int itype, const blas::Uplo uplo, const Tile<T, device>& a, const Tile<T, device>& b) {
+  DLAF_ASSERT(a.size().rows() == a.size().cols(), a.size());
+  DLAF_ASSERT(itype >= 1 && itype <= 3, itype);
+
+  // itype = 1 to solve inv(U**H)*A*inv(U) or inv(L)*A*inv(L**H))
+  // itype = 2, 3 to solve U*A*U**H or L**H*A*L
```
albestro, line 19: IMHO this comment is superfluous, since it should be available in the documentation of the function.

albestro (changes requested):
Another round of review, I will look at the tests soon.

You will see my comments in the code, but I'd like to point out something more about documentation.

Tile wrappers for blas/lapack referring to Hermitian matrices, AFAICT, are valid also for Symmetric real matrices, but in the documentation we don't say this or at least it is implicit that Hermitian is "extended" also for real values. There may be discussion about how to address this (e.g. having just the hermitian call addressing both complex and real, having two separate calls, ...), but I'm quite sure it is something we don't want to address in this PR. Do you @rasolca @teonnik think it is worth opening an issue about this?

test/unit/test_lapack_tile/test_hegst.h
```diff
@@ -0,0 +1,67 @@
...
+  std::tie(el_t, el_a, res_a) =
+      dlaf::matrix::test::getGenToStdElementSetters<ElementIndex, BaseType<T>>(m, itype, uplo, alpha,
+                                                                               beta, gamma);
+
+  set(t, el_t);
+  set(a, el_a);
+
+  tile::hegst(itype, uplo, a, t);
+
+  CHECK_TILE_NEAR(res_a, a, 10 * (m + 1) * TypeUtilities<T>::error,
```
albestro, line 65: See my previous comment about `CHECK_MATRIX_NEAR`. This macro would not work, but I would ask you to open an issue about it in case.

albestro (changes requested):

test/include/dlaf_test/matrix/util_generic_blas.h
```diff
@@ -60,15 +60,15 @@ auto getLeftTriangularSystem(blas::Uplo uplo, blas::Op op, blas::Diag diag, T al
         (diag == blas::Diag::Unit && index.row() == index.col()))
       return TypeUtilities<T>::element(-9.9, 0);
 
-    double i = index.row();
-    double k = index.col();
+    const double i = index.row();
```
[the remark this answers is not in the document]
Sely85: true

test/unit/test_lapack_tile/test_hegst.h
Sely85: I admit I just stuck to what was already written. Probably only checking absolute value would be enough. 
What do you think, @rasolca?

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
Sely85: @albestro would you mind checking if I implemented it correctly, please? Thanks!

test/unit/eigensolver/mc/test_gen_to_std.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+  std::tie(el_l, el_a, res_b) =
+      test::getGenToStdElementSetters<GlobalElementIndex, T>(m, 1, uplo, alpha, beta, gamma);
+
+  set(mat_a, el_a);
+  set(mat_l, el_l);
+
+  Eigensolver<Backend::MC>::genToStd(uplo, mat_a, mat_l);
+
+  CHECK_MATRIX_NEAR(res_b, mat_a, 10 * (mat_a.size().rows() + 1) * TypeUtilities<T>::error,
```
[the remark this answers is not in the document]
rasolca: Relative errors are not good to check small numbers.
Eg. if one element is expected to be 1e-20 and it is found to be 2e-20, the relative error is huge. However the result can be totally fine if other elements of the matrix have larger order of magnitude.

test/unit/test_lapack_tile/test_hegst.h
rasolca: See my other comment above.

include/dlaf/eigensolver/gen_to_std/mc.tpp
```diff
@@ -0,0 +1,39 @@
...
+  DLAF_ASSERT(matrix::square_blocksize(mat_l), mat_l);
+  DLAF_ASSERT(matrix::local_matrix(mat_a), mat_a);
+  DLAF_ASSERT(matrix::local_matrix(mat_l), mat_l);
+
+  switch (uplo) {
+    case blas::Uplo::Lower:
+      internal::mc::genToStd_L(mat_a, mat_l);
+      break;
+    case blas::Uplo::Upper:
+      DLAF_ASSERT(false, "not yet implemented", blas::uplo2str(uplo));
```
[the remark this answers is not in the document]
Sely85: I will!

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -80,21 +84,29 @@ void genToStd_L(Matrix<T, Device::CPU>& mat_a, Matrix<T, Device::CPU>& mat_l) {
...
-        // Working on panel...
-        const auto ik = LocalTileIndex{i, k};
-        hpx::dataflow(executor_hp, unwrapping(tile::hemm<T, Device::CPU>), Right, Lower, -0.5,
-                      mat_a.read(kk), mat_l.read(ik), 1.0, mat_a(ik));
+      if (k + 1 < m) {
+        LocalTileIndex istart(k + 1, 0);
+        LocalTileIndex iend(m, 1);
+        for (auto i : dlaf::common::iterate_range2d(istart, iend)) {
+          // Working on panel...
+          const auto ik = LocalTileIndex{i.row(), k};
```
albestro, line 106: See my previous comment.

Moreover, I'd like to point out that you can re-use the same `iterate_range2d` iterator, since you have already used it. Just to make it more clear, I'll try to give share with you my idea.

In this algorithm you are going to work twice over the same panel, for which you added the comment `// Working on panel...`. IMHO, It would have been more clear to have

```cpp
LocalTileIndex ai_start(k + 1, k);
LocalTileSize ai_size(m - (k + 1), 1);
const auto ai_panel = dlaf::common::iterate_range2d(ai_start, ai_size);

// ...

for (const auto ik : ai_panel) {
}

// ...

for (const auto ik : ai_panel) {
}
```

and even the comment would look superfluous once the intentions are quite clear directly from `for (const auto ij : ai_panel)` since it already says that you want to iterate over the `ai_panel`.

@rasolca what do you think?

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -80,21 +84,29 @@ void genToStd_L(Matrix<T, Device::CPU>& mat_a, Matrix<T, Device::CPU>& mat_l) {
...
+          LocalTileIndex istart(j + 1, 0);
+          LocalTileIndex iend(m, 1);
+          for (auto i : dlaf::common::iterate_range2d(istart, iend)) {
+            const auto ik = LocalTileIndex{i.row(), k};
+            const auto ij = LocalTileIndex{i.row(), j};
+            hpx::dataflow(executor_normal, unwrapping(tile::gemm<T, Device::CPU>), NoTrans, ConjTrans,
+                          -1.0, mat_a.read(ik), mat_l.read(jk), 1.0, mat_a(ij));
+            hpx::dataflow(executor_normal, unwrapping(tile::gemm<T, Device::CPU>), NoTrans, ConjTrans,
+                          -1.0, mat_l.read(ik), mat_a.read(jk), 1.0, mat_a(ij));
+          }
```
albestro, line 97: IMHO in this kind of situations, the `iterate_range2d` is not the best choice. To me it reduces drastically the readability of the code.

albestro (changes requested):

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
rasolca: I agree.

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
rasolca: The correct way to use `iterate_range_2d` here is to merge the j (line 80) and i loop (line 90) and select with a if statement when to perform either her2k, gemm or nothing.
Anyway I'm not sure if this improves the readability.

include/dlaf/eigensolver/mc.h
```diff
@@ -0,0 +1,52 @@
...
+namespace dlaf {
+
+template <>
+struct Eigensolver<Backend::MC> {
+  /// Reduce a Hermitian definite generalized eigenproblem to standard form, using the factorization
+  /// obtained from potrf (Cholesky factorization). It solves B=inv(L)*A*inv(L**H)
+  /// or B=inv(U**H)*A*inv(U)
+  /// Implementation on local memory.
+  ///
+  /// @param mat_a on entry it contains the Hermitian matrix A, on exit the matrix elements
```
albestro, line 27: I think the "symmetric" part is missing. Please, check all occurrences.

include/dlaf/eigensolver/mc.h
```diff
@@ -0,0 +1,51 @@
...
+#include "dlaf/matrix.h"
+#include "dlaf/types.h"
+
+namespace dlaf {
+
+template <>
+struct Eigensolver<Backend::MC> {
+  /// Reduce a Hermitian definite generalized eigenproblem to standard form, using the factorization
+  /// obtained from potrf (lower matrix from Cholesky decomposition), solving B=inv(L)*A*inv(L**H),
+  /// implementation on local memory.
```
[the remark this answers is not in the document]
albestro: @rasolca what do you think about the lapack signature (mainly parameter names) vs this one?

include/dlaf/lapack_tile.h
albestro: @rasolca I was talking about this in my main comment.

> Tile wrappers for blas/lapack referring to Hermitian matrices, AFAICT, are valid also for Symmetric real matrices, but in the documentation we don't say this or at least it is implicit that Hermitian is "extended" also for real values. There may be discussion about how to address this (e.g. having just the hermitian call addressing both complex and real, having two separate calls, ...), but I'm quite sure it is something we don't want to address in this PR. Do you @rasolca @teonnik think it is worth opening an issue about this?

albestro (commented):
Just cleaned up solved comments. I still have a few comments.

include/dlaf/eigensolver/gen_to_std/mc.tpp
```diff
@@ -0,0 +1,42 @@
...
+
+  switch (uplo) {
+    case blas::Uplo::Lower:
+      internal::mc::genToStd_L(mat_a, mat_t);
+      break;
+    case blas::Uplo::Upper:
+      DLAF_ASSERT(false, "not yet implemented", blas::uplo2str(uplo));
+      break;
+    default:
+      DLAF_ASSERT(false, "not yet implemented", blas::uplo2str(uplo));
```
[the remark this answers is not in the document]
rasolca: It can make sense: It can represent the case in which all the elements of the input and output are set.

include/dlaf/eigensolver/mc.h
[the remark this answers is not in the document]
rasolca: I think that we should pay attention also on where the results is stored:
So in this case `A <- f(A, <triangular>)` or a similar notation is better.
Regarding the name of the tridiagonal matrix, one option is to use the lapack name `B`.
A second option can be `L` which is stored in Hermitian form if uplo==Upper...

include/dlaf/lapack_tile.h
rasolca: @albestro, a real symmetric matrix is an Hermitian matrix, (and complex transposition is equivalent to transposition for real matrices) so I don't see any reason why we should separate the calls.

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -64,13 +64,17 @@ void genToStd_L(Matrix<T, Device::CPU>& mat_a, Matrix<T, Device::CPU>& mat_l) {
...
-        // Working on panel...
-        const auto ik = LocalTileIndex{i, k};
-        hpx::dataflow(executor_normal, unwrapping(tile::trsm<T, Device::CPU>), Right, Lower, ConjTrans,
-                      NonUnit, 1.0, mat_l.read(kk), mat_a(ik));
-        hpx::dataflow(executor_normal, unwrapping(tile::hemm<T, Device::CPU>), Right, Lower, -0.5,
-                      mat_a.read(kk), mat_l.read(ik), 1.0, mat_a(ik));
+      if (k + 1 < m) {
+        LocalTileIndex istart(k + 1, 0);
+        LocalTileIndex iend(m, 1);
+        for (auto i : dlaf::common::iterate_range2d(istart, iend)) {
```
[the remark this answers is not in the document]
Sely85: That's true. I probably made a mistake in a previous attempt that forced me to add this `if`. But of course you're right. It is definitely not needed.

include/dlaf/lapack_tile.h
albestro: I agree with you, but I'm not getting what has to be done in the documentation.

Do we want to explicit that it works for real symmetric like we did in other places (because also LAPACK make this distinction) or for us it is enough saying Hermitian to address both complex and real (symmetric) case?

It's not a blocking thing, it's just an genuine interest about what do in the documentation to make it uniform.

include/dlaf/eigensolver/gen_to_std/mc.tpp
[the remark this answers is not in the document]
albestro: Likely this is going to be merged before the #244. I will adapt this later.

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -0,0 +1,125 @@
...
+                      NonUnit, 1.0, mat_l.read(kk), mat_a(ik));
+        hpx::dataflow(executor_normal, unwrapping(tile::hemm<T, Device::CPU>), Right, Lower, -0.5,
+                      mat_a.read(kk), mat_l.read(ik), 1.0, mat_a(ik));
+      }
+
+      const LocalTileIndex ti_start(k + 1, k + 1);
+      const LocalTileIndex ti_end(n, m);
+      const auto ti_trailing = dlaf::common::iterate_range2d(ti_start, ti_end);
+      for (auto ji : ti_trailing) {
+        const auto jj = LocalTileIndex{ji.row(), ji.row()};
```
rasolca, line 82: no need for jj. When `ij.row() == ij.col()` you can use `ij`.

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -0,0 +1,125 @@
...
+        hpx::dataflow(executor_normal, unwrapping(tile::trsm<T, Device::CPU>), Right, Lower, ConjTrans,
+                      NonUnit, 1.0, mat_l.read(kk), mat_a(ik));
+        hpx::dataflow(executor_normal, unwrapping(tile::hemm<T, Device::CPU>), Right, Lower, -0.5,
+                      mat_a.read(kk), mat_l.read(ik), 1.0, mat_a(ik));
+      }
+
+      const LocalTileIndex ti_start(k + 1, k + 1);
+      const LocalTileIndex ti_end(n, m);
+      const auto ti_trailing = dlaf::common::iterate_range2d(ti_start, ti_end);
+      for (const auto& ji : ti_trailing) {
```
rasolca, line 81: I don't understand why you iterate over `ji` and then use only the transposed `ij`.
Iterate directly on `ij`.

include/dlaf/eigensolver/gen_to_std/mc/gen_to_std_L.h
```diff
@@ -0,0 +1,125 @@
...
+                        mat_a.read(jk), mat_l.read(jk), 1.0, mat_a(jj));
+        }
+        else if (ji.row() < ji.col()) {
+          hpx::dataflow(executor_normal, unwrapping(tile::gemm<T, Device::CPU>), NoTrans, ConjTrans,
+                        -1.0, mat_a.read(ik), mat_l.read(jk), 1.0, mat_a(ij));
+          hpx::dataflow(executor_normal, unwrapping(tile::gemm<T, Device::CPU>), NoTrans, ConjTrans,
+                        -1.0, mat_l.read(ik), mat_a.read(jk), 1.0, mat_a(ij));
+        }
+        else {
+        }
```
rasolca, line 98: The empty else is not needed.

rasolca (changes requested):

rasolca:
bors try

rasolca (approved):

rasolca:
bors try

albestro (approved):

PR #262, opened 2020-07-14
Extend README with build instructions

albestro:
Extend the README with some basic instructions on how to build our library and dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,11 @@ Example installation:
 
 `spack install dla-future ^intel-mkl`
 
-Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the architecture, as it represents the maximum number of OS-threads.
+Or you can go even further with a more detailed spec like this one, which builds dla-future in debug mode, using the clang compiler, specifying that the HPX on which it depends has to be built
+in debug mode too, with APEX instrumentation enabled, and that we want to use MPICH as MPI implementation, without fortran support (because clang does not support it).
+`spack install dla-future %clang build_type=Debug ^hpx build_type=Debug instrumentation=apex ^mpich ~fortran`
+
+Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the architecture, as it represents the size of the bitmask to interface with hardware threads.
 
 `spack install dla-future ^intel-mkl ^hpx max_cpu_count=256`
 
@@ -46,6 +50,7 @@ Notice that, for the package to work correctly, the HPX option `max_cpu_count` m
 You can build all the dependencies by yourself, but you have to ensure that:
 - BLAS/LAPACK implementation is not multithreaded
 - HPX: `HPX_WITH_NETWORKING=none` + `HPX_WITH_MAX_CPU_COUNT=n` (according to number of cores in the architecture, suggested the next closest power of 2)
+- HPX and DLAF must have a compatible `CMAKE_BUILD_TYPE`: they must be built both in Debug, or with any combination of release types (Release, RelWithDebInfo or MinSizeRel)
 
 TODO configure cmake script for DLAF (with/without MKL)
 
@@ -56,11 +61,11 @@ Using DLAF in a CMake project is extremely easy!
 Configure your project with:
 
 ```bash
-# DLAF_INSTALL_PREFIX must be set to where you installed DLAF (i.e. CMAKE_INSTALL_PREFIX)
+# DLAF_INSTALL_PREFIX must be set to where you installed DLAF
 cmake -DDLAF_DIR="$DLAF_INSTALL_PREFIX/lib/cmake" ..
 
 # ... or in case you used spack
-# cmake -DDLAF_DIR=`spack location -i dla-future` ..
+cmake -DDLAF_DIR=$(spack location -i dla-future) ..
 ```
 
 Then, it is just as simple as adding these directives in your `CMakeLists.txt`:
--- a/README.md
+++ b/README.md
@@ -57,18 +57,22 @@ And here the main CMake options for DLAF build customization:
 
 CMake option | Values | Note
 :---|:---|:---
-`DLAF_ASSERT_ENABLE` | `{ON,OFF}` (default: `ON`) | to enable/disable cheap assertions
-`DLAF_ASSERT_MODERATE_ENABLE` | `{ON,OFF}` (default: `ON` in Debug, `OFF` otherwise) | to enable/disable moderate assertions
-`DLAF_ASSERT_HEAVY_ENABLE` | `{ON,OFF}` (default: `ON` in Debug, `OFF` otherwise) | to enable/disable heavy assertions
+`HPX_DIR` | CMAKE:PATH | Location of the HPX CMake-config file
+`blaspp_DIR` | CMAKE:PATH | Location of the blaspp CMake-config file
+`lapackpp_DIR` | CMAKE:PATH | Location of the lapackpp CMake-config file
+`DLAF_WITH_MKL` | `{ON,OFF}` (default: `OFF`) | if blaspp/lapackpp is built with MKL
+`MKL_ROOT` | CMAKE:PATH | Location of the MKL library
+`DLAF_ASSERT_ENABLE` | `{ON,OFF}` (default: `ON`) | enable/disable cheap assertions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,34 +37,41 @@ Example installation:
...
 `spack install dla-future ^intel-mkl ^hpx max_cpu_count=256`
 
-## How to use the library
+#### Build the old good way
 
-Using DLAF in a CMake project is extremely easy!
+You can build all the dependencies by yourself, but you have to ensure that:
+- BLAS/LAPACK implementation is not multithreaded
+- HPX: `HPX_WITH_NETWORKING=none` + `HPX_WITH_MAX_CPU_COUNT=n` (according to number of cores in the architecture, suggested the next closest power of 2)
+
```
teonnik, line 49: I would add that:

- HPX and DLAF need to have the same `CMAKE_BUILD_TYPE`, i.e. if HPX is built with `Debug`, then DLAF should also be built with `Debug`.

README.md
```diff
@@ -25,34 +37,41 @@ Example installation:
...
+### Link your program/library with DLAF
+
+Using DLAF in a CMake project is extremely easy!
 
 Configure your project with:
 
 ```bash
-cmake -DDLAF_DIR="$DLAF_ROOT/lib/cmake" ..
+# DLAF_INSTALL_PREFIX must be set to where you installed DLAF (i.e. CMAKE_INSTALL_PREFIX)
+cmake -DDLAF_DIR="$DLAF_INSTALL_PREFIX/lib/cmake" ..
```
teonnik, line 60: or to add `$DLAF_INSTALL_PREFIX` to `CMAKE_PREFIX_PATH`

teonnik (commented):
I would also add spack instruction on how to build with clang, mpich and APEX support:

`spack install dla-future %clang build_type=Debug ^hpx build_type=Debug instrumentation=apex ^mpich ~fortran`

The `~fortran` option on `mpich` is important because clang doesn't provide a fortran compiler by default.

haampie:
Probably you mean `~fortran` @teonnik

README.md
```diff
@@ -25,34 +37,41 @@ Example installation:
 
 `spack install dla-future ^intel-mkl`
 
-Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the platform,
-as it represents the maximum number of OS-threads.
+Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the architecture, as it represents the maximum number of OS-threads.
```
rasolca, line 40: It it the length of the mask used to interface hardware threads.
having a too small `max_cpu_count` might create problems even with `--hpx-threads=1`

rasolca (changes requested):

README.md
```diff
@@ -37,7 +37,11 @@ Example installation:
 
 `spack install dla-future ^intel-mkl`
 
-Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the architecture, as it represents the maximum number of OS-threads.
+Or you can go even further with a more detailed spec like this one, which builds dla-future in debug mode, using the clang compiler, specifying that the HPX on which it depends has to be built
+in debug mode too, with APEX instrumentation enabled, and that we want to use MPICH as MPI implementation, without fortran support (because clang does not support it).
+`spack install dla-future %clang build_type=Debug ^hpx build_type=Debug instrumentation=apex ^mpich ~fortran`
+
+Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the architecture, as it represents the size of the bitmask to interface with hardware threads.
```
albestro, line 44: @rasolca I tried to address your previous comment about the `max_cpu_count`. Do you want to improve it further?

README.md
```diff
@@ -37,7 +37,11 @@ Example installation:
 
 `spack install dla-future ^intel-mkl`
 
-Notice that, for the package to work correctly, the HPX option `max_cpu_count` must be set accordingly to the architecture, as it represents the maximum number of OS-threads.
+Or you can go even further with a more detailed spec like this one, which builds dla-future in debug mode, using the clang compiler, specifying that the HPX on which it depends has to be built
+in debug mode too, with APEX instrumentation enabled, and that we want to use MPICH as MPI implementation, without fortran support (because clang does not support it).
+`spack install dla-future %clang build_type=Debug ^hpx build_type=Debug instrumentation=apex ^mpich ~fortran`
```
albestro, line 42: @teonnik does it satisfy your previous comment? do you want to change it/improve it further?

README.md
albestro: I tried to give more details about how it works, I can revert and use your more straightforward one if you prefer. Let me know @rasolca @teonnik @msimberg

README.md
albestro: I'm not used to changing `CMAKE_PREFIX_PATH`, because I feel like changing a "system" thing. What would you want to add exactly. Feel free to propose an actual suggestion for the documentation 👍

README.md
```diff
@@ -25,34 +37,47 @@ Example installation:
...
-## How to use the library
+#### Build the old good way
 
-Using DLAF in a CMake project is extremely easy!
+You can build all the dependencies by yourself, but you have to ensure that:
+- BLAS/LAPACK implementation is not multithreaded
+- HPX: `HPX_WITH_NETWORKING=none` + `HPX_WITH_MAX_CPU_COUNT=n` (according to number of cores in the architecture, suggested the next closest power of 2)
+- HPX and DLAF must have a compatible `CMAKE_BUILD_TYPE`: they must be built both in Debug, or with any combination of release types (Release, RelWithDebInfo or MinSizeRel)
+
+TODO configure cmake script for DLAF (with/without MKL)
```
albestro, line 56: @rasolca @teonnik 
About this TODO, do you have any suggestion on how to deal with it? Do we want to explain all CMake variables of our project? Do we want to address it here or in another dedicated PR (in case I could open an issue about this)?

README.md
msimberg: I think changing `CMAKE_PREFIX_PATH` is a perfectly fine way to help CMake find packages. It has the benefit of users not having to know in which subdirectory the DLAF CMake config file is (you're also using `GNUInstallDirs` which means that the config file can be in `lib/cmake` or `lib64/cmake` depending on the system.

albestro: Good point. I've just made it the first choice and re-arranged a bit the information of this section.

README.md
teonnik: I think these are at least worth listing. Most of the are self-explanatory. Perhaps something like this: 

- `DLAF_ASSERT_ENABLE` : set to ON|OFF to enable/disable cheap assertions (default: ON)
- `DLAF_ASSERT_MODERATE_ENABLE` : set to ON|OFF to enable/disable moderate assertions (default: ON in Debug, OFF otherwise)
- `DLAF_ASSERT_HEAVY_ENABLE` : set to ON|OFF to enable/disable heavy assertions (default: ON in Debug, OFF otherwise)

- `DLAF_WITH_CUDA` : enable CUDA support 
- `HPX_DIR`: ???
- `blaspp_DIR`: ???
- `lapackpp_DIR`: ???

`blaspp`/`lapackpp` dependency:
- `DLAF_WITH_MKL` (if blaspp/lapackpp is built with MKL)
- `MKL_ROOT`: ???

- `DLAF_BUILD_MINIAPPS` : set to ON|OFF to enable/disable building miniapps
- `DLAF_WITH_TEST` : set to ON|OFF to enable/disable building tests
- `DLAF_INSTALL_TESTS` : set to ON|OFF to enable/disable installing tests

Sely85: What about adding also `BUILD_DOC` (that might probably be renamed to `DLAF_BUILD_DOC` or similar)?

albestro: I've just added a table trying to give an overview of all the options (it looks a bit cluttered to me, if you have any suggestion on how to improve it, it is absolutely welcome!).

README.md
```diff
@@ -25,34 +37,72 @@ Example installation:
...
 
-# ...
+Then, it is just as simple as adding these directives in your `CMakeLists.txt`:
 
-target_link_libraries(<your_target> PRIVATE DLAF)
+```
+find_package(DLAF)
+# ... and then for your executable/library target
+target_link_libraries(<your_target> PRIVATE DLAF::DLAF)
 ```
```
Sely85, line 103: I was thinking that it may be useful to show an example of a "real code" linked to DLAF. What about using miniapp?

rasolca (approved):

teonnik (approved):

PR #263, opened 2020-07-16
Fixing documentation format and equation in a comment

Sely85:
Fixing documentation: added punctuation and removed some `@brief` keyword, forgotten in the previous PR about moving to the new error handling style (#178).

Fixed the equation to obtain the elements of matrix B in test_gemm.

### DIFF
--- a/include/dlaf/matrix/layout_info.h
+++ b/include/dlaf/matrix/layout_info.h
@@ -32,15 +32,15 @@ public:
   ///
   /// See misc/matrix_distribution.md for for more detail about the parameters.
   ///
-  /// If size.isEmpty():
+  /// @pre If size.isEmpty():
   /// @pre tile_ld >= 1,
   /// @pre tile_offset_row >= 1,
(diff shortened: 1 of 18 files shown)


include/dlaf/matrix/layout_info.h
```diff
@@ -36,7 +36,7 @@ class LayoutInfo {
   /// @pre tile_ld >= 1,
   /// @pre tile_offset_row >= 1,
   /// @pre tile_offset_col >= 1;
-  /// If !size.isEmpty():
+  /// if !size.isEmpty():
```
albestro, line 39: I think it is not going to be rendered as you expect...

image

Sely85: You're right. Now it should be (slightly) better. What do you think? 

image

Sely85:
> I would suggest something similar to what I did for `norm` that should result in something like
> 
> What do you think? @Sely85 @rasolca

Definitely better!

rasolca (approved):

PR #265, opened 2020-08-20
[spack] Enforce same build type

teonnik:
Related to #84 . In particular, this from HPX :

> We currently don’t guarantee ABI compatibility between Debug and Release builds. Please make sure that applications built against HPX use the same build type as you used to build HPX. For CMake builds, this means that the CMAKE_BUILD_TYPE variables have to match and for projects not using CMake, the HPX_DEBUG macro has to be set in debug mode.


Sely85 (approved):
Looks good!

PR #275, opened 2020-10-19
Utility to print matrix/tile elements

Sely85:
First implementation of a utility that allows elements of matrix or tile to be printed to standard output. 

To do:

- [ ] verify how to use `iterate_range2d` instead of nested `if` without losing formatting

- [x] use a format more user-friendly (_i.e._ numpy/ipthon, @albestro) 

Ref. comments by @albestro and @rasolca  in https://github.com/eth-cscs/DLA-Future/pull/261#discussion_r506356142


albestro:
About the output format, for both matrices and tiles, I would propose a numpy-friendly one. I have some code about this, since I use it when I debug things, and I find it useful because I can easily copy output from DLAF to my python script/jupyter notebook and manipulate it with numpy.

A matrix example (as it is now, it requires a string with a symbol for the matrix, e.g. in this example `"Z"`)

```python
Z = np.zeros((6, 6), dtype=np.complex)
Z[0,0] = complex(-0.679585,0)
Z[1,0] = complex(0.753421,-0.529475)
Z[0,1] = complex(0.753421,0.529475)
Z[1,1] = complex(-0.0791742,0)
Z[2,0] = complex(-1.10087,-0)
...
Z[4,4] = complex(-0.754793,0)
Z[5,4] = complex(0.0424268,0.0343959)
Z[4,5] = complex(0.355474,-0.0861055)
Z[5,5] = complex(0.134015,0)
```

A tile example
```python
np.array([complex(-0.754793,0), complex(0.0424268,0.0343959), complex(0.355474,-0.0861055
), complex(0.134015,0), ]).reshape(2, 2).T
```

This is what I use currently. There is one main difference between matrices and tiles: matrix can be distributed, so by specifying element indices, it is easier to copy all chunks from different ranks and "manually" collect all of them locally in a python script.

I think it would be nice to have it at least as an additional/optional output format (e.g. it may not be the most straightforward to look at in the console, for which the matrix representation would be for sure nicer).

albestro:
*Reminder to self*

In case we want to pursue the numpy output format, IMHO it would be nice to distinguish column/row vectors, with the aim of avoiding the wrong shape matching in python.

This would require not using 1D `np.array`, but always using 2D ones. (see https://stackoverflow.com/a/17428859/12494143)

Sely85:
> About the output format, for both matrices and tiles, I would propose a numpy-friendly one. I have some code about this, since I use it when I debug things, and I find it useful because I can easily copy output from DLAF to my python script/jupyter notebook and manipulate it with numpy. [...]

I agree: a numpy-friendly format allows further manipulation. It would be nice to have both output formats.

albestro:
I've just branched this PR to #284 adding the code for the numpy-ready format.

rasolca:
For the numpy matrix output can I suggest the following?
```
a=zeros([5, 5])
a[0:2,0:2] = [[1, 2], [3, 4]]
a[4:5,2:4] = [[-1, 2]]
a[0:2,4:5] = [[-3], [-2]]
```
In this way, for the right-hand-side, the print tile utility can be used...

albestro:
> For the numpy matrix output can I suggest the following?
> 
> ```
> a=zeros([5, 5])
> a[0:2,0:2] = [[1, 2], [3, 4]]
> a[4:5,2:4] = [[-1, 2]]
> a[0:2,4:5] = [[-3], [-2]]
> ```
> 
> In this way, for the right-hand-side, the print tile utility can be used...

It's a good suggestion. The only possible downside I see, it is that it worsen a bit the readability for giving a quick look in the console (e.g. it would be a bit more error-prone if I have to check a specific element value). But, I agree that it may be a good choice anyway.

Let me know what do you think about it, even after this consideration, and as soon as I have time I will adapt it 😉

rasolca:
I don't think you will ever check just an element.
Anyway already with a distributed 10x10 matrix the output would be 100 lines of mangled elements: very difficult to inspect in the console anyway.

teonnik:
Is this PR discontinued in favor of #284 ?

albestro:
Quoting @Sely85 
> I agree: a numpy-friendly format allows further manipulation. It would be nice to have both output formats.

Personally, I just see the need of being able to easily work with the matrices (especially for debugging purposes) with other tools (in my case python-based), so I may be biased but I would say that to me the print utility is valuable just if it is in a machine-ready format. It may be true that this can be quickly adapted to become a `csv` format.

My fork in #284 was simply to show what I was doing on the same topic. I will for sure work on that PR.

include/dlaf/matrix/print_human.h
```diff
@@ -0,0 +1,69 @@
...
+    for (SizeType icol = 0; icol < ncol; ++icol) {
+      SizeType tilecol = icol / blockcol;
+      SizeType elcol = icol % blockcol;
+
+      const LocalTileIndex idx = {tilerow, tilecol};
+      auto& tile = mat.read(idx).get();
+
+      std::cout << tile({elrow, elcol}) << " ";
+    }
+    std::cout << " " << std::endl;
```
rasolca, line 50: Why do you add a space?

include/dlaf/matrix/print_human.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace format {
+struct human {};
+}
+
+namespace matrix {
+
+namespace internal {
+
+/// Print a matrix in a human readable format to standard output
```
rasolca, line 28: I would use the csv format instead of elements separated by space.

include/dlaf/matrix/print_human.h
```diff
@@ -0,0 +1,69 @@
...
+// TODO: now is printed to standard output, would it be better to save it into a file?
```
rasolca, line 29: Simply use an ostream argument.

include/dlaf/matrix/print_human.h
```diff
@@ -0,0 +1,69 @@
...
+template <class T>
+void print(format::human, Matrix<T, Device::CPU>& mat) {
```
rasolca, line 31: Please use the same API as print(format::numpy, ...)

include/dlaf/matrix/print_human.h
```diff
@@ -0,0 +1,69 @@
...
+  SizeType nrow = mat.size().rows();
```
rasolca, line 32: As a distributed matrix will produce incorrect output, if the matrix is distributed just print the basic matrix info.

rasolca (changes requested):

include/dlaf/matrix/print_human.h
```diff
@@ -0,0 +1,77 @@
...
+/// Print a tile in csv format to standard output
+template <class T>
+void print(format::csv, const Tile<const T, Device::CPU>& tile, std::ostream& os = std::cout) {
+  for (SizeType ii = 0; ii < tile.size().rows(); ++ii) {
+    for (SizeType jj = 0; jj < tile.size().cols(); ++jj) {
+      os << std::setprecision(5) << tile({ii, jj}) << ",";
+    }
+    os << std::endl;
+  }
+  os << std::endl;
```
rasolca, line 38: This new line is not needed. The user should put it if an empty line is needed.

include/dlaf/matrix/print_human.h
```diff
@@ -0,0 +1,77 @@
...
+      SizeType elcol = icol % blockcol;
+
+      const LocalTileIndex idx = {tilerow, tilecol};
+      auto& tile = mat.read(idx).get();
+
+      os << tile({elrow, elcol}) << ",";
+    }
+    os << std::endl;
+  }
+  os << std::endl;
```
rasolca, line 72: Not needed.

include/dlaf/matrix/print_human.h
```diff
@@ -0,0 +1,77 @@
+//
```
rasolca, line 1: Please update the filename to `print_csv.h` as well.

test/unit/matrix/test_matrix_output.cpp
```diff
@@ -206,7 +272,11 @@ struct test_matrix_dist_output {
 
     const std::string output{"M = np.zeros((0, 0), dtype=np.single)\n"};
 
-    return std::make_pair(std::move(mat), output);
+    std::string output_csv{"size=(0, 0), block_size=(2, 3), tiles_grid=(0, 0), rank_index=(" +
+                           std::to_string(comm_grid_.rank().row()) + ", " +
+                           std::to_string(comm_grid_.rank().col()) + "), comm_grid=(3, 2)\n"};
+
+    return std::make_tuple(std::move(mat), output, output_csv);
```
rasolca, line 279:
```suggestion
    std::stringstream output_csv;
    output_csv << mat << std::endl;

    return std::make_tuple(std::move(mat), output, output_csv.str());
```

rasolca: Please fix similar cases as well.

Sely85: true, sorry. done!

include/dlaf/matrix/print_csv.h
```diff
@@ -0,0 +1,75 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2019, ETH Zurich
```
albestro, line 4: 2021

test/unit/matrix/test_matrix_output.cpp
```diff
@@ -206,7 +272,10 @@ struct test_matrix_dist_output {
 
     const std::string output{"M = np.zeros((0, 0), dtype=np.single)\n"};
 
-    return std::make_pair(std::move(mat), output);
+    std::stringstream output_csv;
+    output_csv << mat << std::endl;
```
albestro, line 276: Here I would specify what's the expected output, without relying on code for producing it.

test/unit/matrix/test_matrix_output.cpp
```diff
@@ -239,7 +308,10 @@ struct test_matrix_dist_output {
     else if (5 == linear_rank)
       output += "M[4:5,3:4] = np.array([-19,], dtype=np.single).reshape(1, 1).T\n";
 
-    return std::make_pair(std::move(mat), output);
+    std::stringstream output_csv;
+    output_csv << mat << std::endl;
```
albestro, line 312: see previous comment.

test/unit/matrix/test_matrix_output.cpp
```diff
@@ -260,7 +332,10 @@ struct test_matrix_dist_output<std::complex<T>> {
 
     const std::string output{"M = np.zeros((0, 0), dtype=np.csingle)\n"};
 
-    return std::make_pair(std::move(mat), output);
+    std::stringstream output_csv;
+    output_csv << mat << std::endl;
```
albestro, line 336: here too

test/unit/matrix/test_matrix_output.cpp
```diff
@@ -292,7 +367,10 @@ struct test_matrix_dist_output<std::complex<T>> {
     else if (5 == linear_rank)
       o += "M[4:5,3:4] = np.array([4-3j,], dtype=np.csingle).reshape(1, 1).T\n";
 
-    return std::make_pair(std::move(mat), o);
+    std::stringstream output_csv;
+    output_csv << mat << std::endl;
```
albestro, line 371: here too.

test/unit/matrix/test_matrix_output.cpp
rasolca: I don't agree.
As discussed before (unfortunately it is not documented in the code), if the matrix is distributed the basic info of the matrix are printed using the operator<<.
This is the expected behaviour, independently from the implementation of matrix operator<<.

include/dlaf/matrix/print_csv.h
```diff
@@ -0,0 +1,75 @@
...
+void print(format::csv, const Tile<const T, Device::CPU>& tile, std::ostream& os = std::cout) {
+  for (SizeType ii = 0; ii < tile.size().rows(); ++ii) {
+    for (SizeType jj = 0; jj < tile.size().cols(); ++jj) {
+      os << std::setprecision(5) << tile({ii, jj}) << ",";
+    }
+    os << std::endl;
+  }
+}
+
+/// Print a matrix in csv format to standard output
```
rasolca, line 40: Specify what happens for distributed matrices.

rasolca:
bors try

rasolca (approved):

PR #279, opened 2020-10-22
Setup tile

Sely85:
Implementation of utilities to quickly set-up Tile and const Tile. 

These functions work as follows (thanks @albestro):
1. create a Tile,
2. set values of elements of Tile,
3. if the tiles must be constant, move the tile setup so far into a const Tile.

 Ref. comments by @albestro and @rasolca in https://github.com/eth-cscs/DLA-Future/pull/261#discussion_r506340909


Sely85:
bors try

test/include/dlaf_test/matrix/util_tile_setup.h
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/// Create a read-only tile and fill with selected values
+///
+/// @pre val argument is an index of type const TileElementIndex&,
+/// @pre val return type should be T;
+/// @pre size is the dimension of the tile to be created (type: TileElementSize);
+/// @pre ld is the leading dimension of the tile to be created.
+template <class T, class CT = const T, class ElementGetter>
+Tile<CT, Device::CPU> setup_readonly_tile(ElementGetter val, const TileElementSize& size,
```
albestro, line 65: IMHO given what the function is doing, the name is a bit misleading.

By looking at the implementation, this function allocates a new tile of type T and then move this into another tile with the type `CT`, which by default is `const T`. If I call this function with `setup_readonly_tile<double, double>` it won't return a readonly tile.

A starting point for fixing this could be renaming `CT` to make more clear that it is the output type for the tile, and not the CONST TYPE. Then, it may be easier to think of a better name for the function.

albestro (changes requested):
First quick thoughts.

test/include/dlaf_test/matrix/util_tile_setup.h
rasolca: I agree the name should be setupTile.
@albestro, @Sely85 Regarding the type, is it not better to have just the output type (calling it simply `T`), and use `std::remove_const_t` to get the type to first construct the tile?

test/include/dlaf_test/matrix/util_tile.h
```diff
@@ -74,6 +74,17 @@ void print(const Tile<T, Device::CPU>& tile, int precision = 4, std::ostream& ou
   out.precision(out_precision);
 }
 
+/// Create a MemoryView and initialize a Tile on it.
+///
+/// @pre size is the dimension of the tile to be created (type: TileElementSize);
+/// @pre ld is the leading dimension of the tile to be created.
+template <class T>
+Tile<T, Device::CPU> create_tile(const TileElementSize& size, const SizeType& ld) {
```
rasolca, line 82: Please follow function naming conventions.

test/include/dlaf_test/matrix/util_tile_setup.h
```diff
@@ -0,0 +1,90 @@
...
+/// Create a read/write tile and fill with selected values
+///
+/// @pre val argument is an index of type const TileElementIndex&,
+/// @pre val return type should be T;
+/// @pre size is the dimension of the tile to be created (type: TileElementSize);
+/// @pre ld is the leading dimension of the tile to be created,
+/// @pre op is the blas::Op to be applied to the tile.
+template <class T, class ElementGetter>
+Tile<T, Device::CPU> setup_tile(ElementGetter val, const TileElementSize& size, const SizeType& ld,
+                                const blas::Op& op) {
```
rasolca, line 52: Note that the distinction between `set(tile, values)` and `set(tile, values, op)` is that the first one doesn't require any blas header.
I would simply add the method without blas::Op in `util_tile.h` and the other in `util_tile_blas.h`

rasolca (changes requested):

test/include/dlaf_test/matrix/util_tile_setup.h
albestro: Yep, I completely agree.

test/include/dlaf_test/matrix/util_tile.h
```diff
@@ -74,6 +74,32 @@ void print(const Tile<T, Device::CPU>& tile, int precision = 4, std::ostream& ou
...
+/// @pre val return type should be T;
+/// @pre size is the dimension of the tile to be created (type: TileElementSize);
+/// @pre ld is the leading dimension of the tile to be created.
+template <class T, class ElementGetter>
+Tile<T, Device::CPU> createTile(ElementGetter val, const TileElementSize& size, const SizeType& ld) {
+  using nonconstT = std::remove_const_t<T>;
+  Tile<nonconstT, Device::CPU> support = createTile<nonconstT>(size, ld);
+  set(support, val);
+  Tile<T, Device::CPU> tileout(std::move(support));
+  return tileout;
```
rasolca, line 100: I would try to avoid to give names to types or object used just once:
```suggestion
  auto tile = createTile<std::remove_const_t<T>>(size, ld);
  set(tile, val);
  return Tile<T, Device::CPU>(std::move(tile));
```

rasolca (changes requested):

test/include/dlaf_test/matrix/util_tile_blas.h
```diff
@@ -56,6 +56,24 @@ void set(Tile<T, Device::CPU>& tile, Func el, blas::Op op) {
...
+/// @pre ld is the leading dimension of the tile to be created,
+/// @pre op is the blas::Op to be applied to the tile.
+template <class T, class ElementGetter>
+Tile<T, Device::CPU> createTile(ElementGetter val, const TileElementSize& size, const SizeType& ld,
+                                const blas::Op& op) {
+  using nonconstT = std::remove_const_t<T>;
+  Tile<nonconstT, Device::CPU> support = createTile<nonconstT>(size, ld);
+  set(support, val, op);
+  Tile<T, Device::CPU> tileout(std::move(support));
+  return tileout;
```
rasolca, line 74: Please update also this part of the code.

test/unit/test_blas_tile/test_gemm.h
```diff
@@ -93,14 +84,9 @@ void testGemm(blas::Op op_a, blas::Op op_b, SizeType m, SizeType n, SizeType k,
...
 
-  // Set tile elements.
-  set(a0, el_op_a, op_a);
-  set(b0, el_op_b, op_b);
-  set(c, el_c);
-
-  // Read-only tiles become constant if CT is const T.
-  Tile<CT, Device::CPU> a(std::move(a0));
-  Tile<CT, Device::CPU> b(std::move(b0));
+  Tile<CT, Device::CPU> a = createTile<CT>(el_op_a, size_a, lda, op_a);
```
rasolca, line 87: I'm wondering if 
```suggestion
  auto a = createTile<CT>(el_op_a, size_a, lda, op_a);
```
would be better.
What do you think?

Sely85: It's more compact and helps to avoid mistakes while writing the whole type. I will substitute them all, if you agree.

test/include/dlaf_test/matrix/util_tile.h
```diff
@@ -74,6 +74,30 @@ void print(const Tile<T, Device::CPU>& tile, int precision = 4, std::ostream& ou
...
+/// Create a MemoryView and initialize a Tile on it.
+///
+/// @pre size is the dimension of the tile to be created (type: TileElementSize);
+/// @pre ld is the leading dimension of the tile to be created.
+template <class T>
+Tile<T, Device::CPU> createTile(const TileElementSize& size, const SizeType& ld) {
+  using dlaf::util::size_t::mul;
+  memory::MemoryView<T, Device::CPU> support_mem(mul(ld, size.cols()));
+  return Tile<T, Device::CPU>(size, std::move(support_mem), ld);
+}
```
teonnik, line 86: Nitpicks
 
1. Why `createTile()` and not `create_tile()`? I though that we use snake case for free functions (e.g. in `range2d.h`, `util_matrix.h` or in `tile.h`).

2. It's OK to pass the parameters of the function by value, these are small types, no need to have a const ref. Sometimes this helps the compiler with aliasing rules, although, in our case, that's not going to help us much. It's also easier to grok.

rasolca: 1. From the coding style:
```
Function Names

Function names should start with a lowercase character and have a capital letter for each new word. Underscores (_) are not allowed.

The only exception to this rule are getter and setter methods for class data members. They follow the naming convention set_variable and get_variable, respectively (see Class Format).
```
Some wrong names might have been missed in reviews.

2. I agree.

rasolca:
bors try

rasolca (approved):

PR #280, opened 2020-10-23
Introduce AllReduce wrapper

albestro:
This PR introduces the DLAF `all_reduce` wrapper for the `MPI_AllReduce` function.

Tests are clones of the `reduce` ones, just slightly changed for checking that results are available on all ranks involved.

Even for this implementation, I used the workaround from #177 for the problem #148, about the incompatibility between pre-defined `MPI_Op` and custom type (used for non-contiguous memory). The logic has been refactored.

I'm going to integrate test for new functions (i.e. `Buffer::operator bool`, `make_contiguous).

### DIFF
--- a/test/unit/common/test_data_descriptor.cpp
+++ b/test/unit/common/test_data_descriptor.cpp
@@ -545,19 +545,22 @@ TYPED_TEST(DataDescriptorTest, CopyCtorFromStridedArray) {
 }
 
 template <class Data>
-void DLAF_CHECK_CREATE_TEMPORARY_BUFFER(const Data& data) {
+::testing::AssertionResult checkCreateTemporaryBuffer(const Data& data) {
   auto data_temp = create_temporary_buffer(data);
 
-  EXPECT_NE(common::data_pointer(data), common::data_pointer(data_temp));
-  EXPECT_EQ(common::data_count(data), common::data_count(data_temp));
-  EXPECT_TRUE(common::data_iscontiguous(data_temp));
+  if (common::data_pointer(data) == common::data_pointer(data_temp))
+    return ::testing::AssertionFailure() << "temporary buffer points to the same memory location";
+  if (common::data_count(data) != common::data_count(data_temp))
+    return ::testing::AssertionFailure() << "temporary buffer does not have the same number of elements";
+  if (!common::data_iscontiguous(data_temp))
+    return ::testing::AssertionFailure() << "temporary buffer is not contiguous";
+
+  return ::testing::AssertionSuccess();
 }
 
-template <bool from_contiguous, class Data>
-void DLAF_CHECK_MAKE_CONTIGUOUS(const Data& data) {
+template <class Data>
+::testing::AssertionResult checkMakeContiguous(const Data& data) {
   using T = std::remove_const_t<typename common::data_traits<Data>::element_t>;
-
-  ASSERT_EQ(from_contiguous, common::data_iscontiguous(data));
 
   common::Buffer<T> temp_buffer;
   auto data_contiguous = common::make_contiguous(data, temp_buffer);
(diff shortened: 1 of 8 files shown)


test/unit/communication/test_all_reduce.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+using namespace dlaf;
+using namespace dlaf_test;
+using namespace dlaf::comm;
+
+using AllReduceTest = SplittedCommunicatorsTest;
+
+template <class T, class = std::enable_if_t<std::is_arithmetic<T>::value>>
+void DLAF_EXPECT_NEAR(const T a, const T b, const T abs_err) {
+  EXPECT_NEAR(a, b, abs_err);
```
rasolca, line 29: Using the macro in a separate function completely mess up the output message in case of error. How do you deal with this problem?

rasolca (changes requested):
LGTM, only a couple of small details to check

test/unit/common/test_data_descriptor.cpp
```diff
@@ -543,68 +544,102 @@ TYPED_TEST(DataDescriptorTest, CopyCtorFromStridedArray) {
   check_copy_ctor<const TypeParam>(data_const);
 }
 
-template <class TypeParam, class Buffer>
-void check_temporary(Buffer& data) {
+template <class Data>
+void DLAF_CHECK_CREATE_TEMPORARY_BUFFER(const Data& data) {
```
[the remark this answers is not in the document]
albestro: I used UPPERCASE name because I thought it would me more clear that they are "on the same level" as `EXPECT_EQ` and other gtest macros. That's the only reason, I can revert it easily if you don't like it.

test/unit/communication/test_all_reduce.cpp
albestro: I didn't think about this.

I'm going to check if `AssertionResult` can be useful in this context (or if something else exists for this purpose).

test/unit/common/test_data_descriptor.cpp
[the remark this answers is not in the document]
rasolca: The difference that IMO has to be highlighted, is that the macros can be "line aware" to write the output when an error occour, while functions cannot.
Looking forward to C++20 and `std::source_location` to avoid all this macro non-sense...

test/unit/communication/test_all_reduce.cpp
rasolca: What about using always `EXPECT_LE(std::abs(a-b), error)`?

albestro: FIY, in this specific instance, we can easily use your proposal (it does not need to be generic, because this test runs just for `TypeParam = std::complex<double>`. IIRC, this is because it is not possible to mix TestFixtures and TypedTests in GTEST (AFAIK, I have to check if they improved it).

However, I did a quick test with `AssertionResult` and it works nicely. It is possible to define a function and then check its result with `EXPECT_TRUE`.

```
template <class T, class U>
::testing::AssertionResult DLAF_EXPECT_NEAR(const T& a, const T& b, const U& abs_err) {
  if (std::abs(a-b) < abs_err)
    return ::testing::AssertionSuccess();
  return ::testing::AssertionFailure() << "They are not equal " << a << " and " << b;
}
```

The message construction is optional, but then testing it with `EXPECT_TRUE(DLAF_EXPECT_NEAR(a, b, err))` allows to report the error in the right source location.

```
dla-future/test/unit/communication/test_all_reduce.cpp:209: Non-fatal failure:
Value of: DLAF_EXPECT_NEAR(value * static_cast<TypeParam>(NUM_MPI_RANKS), data_strided[mem_pos], NUM_MPI_RANKS * TypeUtilities<TypeParam>::error)
  Actual: false (They are not equal (52,104) and (52,104))
Expected: true
```

It may be an overshoot for this particular instance of the problem, but I thought it was worth sharing it.

rasolca: Note that `EXPECT_LE(std::abs(a-b), error)` works both for real and complex cases:
https://en.cppreference.com/w/cpp/numeric/math/fabs
https://en.cppreference.com/w/cpp/numeric/complex/abs

rasolca:
bors try

rasolca (approved):
LGTM

Sely85 (approved):
LGTM!

teonnik (approved):

rasolca:
bors try

albestro:
bors try

Warning fixed

PR #284, opened 2020-10-30
Print Matrix/Tile in numpy ready format

albestro:
I opened this PR just for starting sharing my code.

I implemented a test for both functions based on output parsing with regex.

The code needs refinements, I can work on it if we decide to proceed on this path.


albestro:
This is the updated proposal for a tile

```
np.array([6+6j, 7+6j, 6+7j, 7+7j, ]).reshape(2, 2).T
```

and for a matrix
```
mat = np.zeros((9, 7), dtype=np.complex)
mat[0:3, 0:3] = np.array([0+0j, 1+0j, 2+0j, 0+1j, 1+1j, 2+1j, 0+2j, 1+2j, 2+2j, ]).reshape(3, 3).T
mat[3:6, 0:3] = np.array([3+0j, 4+0j, 5+0j, 3+1j, 4+1j, 5+1j, 3+2j, 4+2j, 5+2j, ]).reshape(3, 3).T
mat[6:9, 0:3] = np.array([6+0j, 7+0j, 8+0j, 6+1j, 7+1j, 8+1j, 6+2j, 7+2j, 8+2j, ]).reshape(3, 3).T
mat[0:3, 3:6] = np.array([0+3j, 1+3j, 2+3j, 0+4j, 1+4j, 2+4j, 0+5j, 1+5j, 2+5j, ]).reshape(3, 3).T
mat[3:6, 3:6] = np.array([3+3j, 4+3j, 5+3j, 3+4j, 4+4j, 5+4j, 3+5j, 4+5j, 5+5j, ]).reshape(3, 3).T
mat[6:9, 3:6] = np.array([6+3j, 7+3j, 8+3j, 6+4j, 7+4j, 8+4j, 6+5j, 7+5j, 8+5j, ]).reshape(3, 3).T
mat[0:3, 6:7] = np.array([0+6j, 1+6j, 2+6j, ]).reshape(1, 3).T
mat[3:6, 6:7] = np.array([3+6j, 4+6j, 5+6j, ]).reshape(1, 3).T
mat[6:9, 6:7] = np.array([6+6j, 7+6j, 8+6j, ]).reshape(1, 3).T
```

Wait for your feedback before adapting the tests to the new output.

test/unit/test_matrix_output.cpp
```diff
@@ -67,3 +76,174 @@ TYPED_TEST(MatrixOutputLocalTest, printElements) {
...
+struct numpy_type;
+
+template <>
+struct numpy_type<float> {
+  static constexpr auto name = "float";
+};
+
+template <>
+struct numpy_type<double> {
+  static constexpr auto name = "float";
```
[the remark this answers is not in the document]
albestro: I investigated a bit. In numpy, there are specific types for single and double precision with various synonyms. I quickly tried and there seems to be a problem with `np.complex128` (matching with std::complex<double>).

Anyway, I don't think that it is necessary to use the exact match for the type, and the single precision is way more than enough, since we are printing values with just few digits.

Do you agree @Sely85 @rasolca @teonnik ?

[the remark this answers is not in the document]
teonnik: Unless I am missing something, currently we are using the default precision which is 6. If we don't plan to increase the precision, then yes, it wouldn't make a difference if we have single and double precision.

Regardless, we should probably use `np.csingle` and `np.cdouble` instead of the fixed width aliases (`np.complex64` `np.complex128`) because `float` and `double` are technically platform-dependent.

teonnik:
Is the idea to be able to copy the output directly into a python notebook/script?

This will work well for relatively small matrices but if the matrix is large, shouldn't we print it in a CSV format? That way we can easily read it into a 2D array in numpy.

albestro:
> Is the idea to be able to copy the output directly into a python notebook/script?

Yes, it is something I use during debugging. I think it is quite handy.

> This will work well for relatively small matrices but if the matrix is large, shouldn't we print it in a CSV format? That way we can easily read it into a 2D array in numpy.

My use-case is during debugging, for which generally I tend to keep matrices small. I agree with you, but at the same I don't know how much CSV can help us with bigger matrices (the number of total characters is approximately the same for large matrices).

IMHO, this work does not address the serialization of a matrix, but it is just a quick way to play with them for debugging purposes. For this reason, my initial thought was to not expose it in the public API, and make it available just in `DLAF_test`.

teonnik:
> My use-case is during debugging, for which generally I tend to keep matrices small. I agree with you, but at the same I don't know how much CSV can help us with bigger matrices (the number of total characters is approximately the same for large matrices).

I see, that is fine! Regarding CSV, I think it will be significantly faster for big matrices because the proposed output format slices the matrix incrementally while a CSV format can be read without much internal indexing logic. However, this is not a priority atm.

> IMHO, this work does not address the serialization of a matrix, but it is just a quick way to play with them for debugging purposes. For this reason, my initial thought was to not expose it in the public API, and make it available just in DLAF_test.

It's still worth keeping in the public API. It doesn't hurt users to have more APIs to work with. In general, I'm open to moving almost all stuff from DLAF_test into DLAF proper, in particular those that do not involve some GTEST specific asserts.

test/unit/test_matrix_output.cpp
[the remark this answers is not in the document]
albestro: > Unless I am missing something, currently we are using the default precision which is 6. If we don't plan to increase the precision, then yes, it wouldn't make a difference if we have single and double precision.

Yes, it is what I meant in my previous comment.

> Regardless, we should probably use `np.csingle` and `np.cdouble` instead of the fixed width aliases (`np.complex64` `np.complex128`) because `float` and `double` are technically platform-dependent.
> 
> https://numpy.org/doc/stable/user/basics.types.html

I'm not completely getting your point. Let me know if you agree on this mapping (ctypes -> numpy):
- `float/double` -> **np.single** (_"Platform-defined single precision float"_)
- `complex<float/double>` -> **np.csingle** (_"Complex number, represented by two single-precision floats"_)

albestro:
Do you agree if I move this from being a fork of the other PR to be an independent branch? so as soon as it will be ready we can merge it in master.

test/unit/test_matrix_output.cpp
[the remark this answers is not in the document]
teonnik: Yep, I do.

albestro:
bors try

It is ready for the review. Let me just try if it builds or there is any macro problem.

albestro:
bors try

This should build correctly...

albestro:
bors try

this should really work (if master does not change anymore) 😁

albestro:
I had to rerun manually one of the steps of the pipeline, but it succeeded.

include/dlaf/matrix/print_numpy.h
```diff
@@ -0,0 +1,117 @@
...
+  const auto imag = std::imag(value);
+  const auto sign = (imag < 0) ? '-' : '+';
+  os << std::real(value) << sign << std::abs(imag) << "j";
+  return os.str();
+}
+}
+
+/// Print a tile as a numpy array
+template <class Stream, class T>
+Stream& print_numpy(Stream& os, const Tile<const T, Device::CPU>& tile) {
```
rasolca, line 58: Is there a reason to use a template instead of `std::ostream`?

rasolca, line 58: Wouldn't it be better to name the function `print` and have an argument for the format of the output?

albestro: Would you go for a template parameter or a tag parameter?

What quickly comes to my mind is:
```cpp
enum class FORMAT{
  NUMPY
};

// usage
print<FORMAT::NUMPY>(std::cout, tile);
```

or
```cpp
struct numpy {}; // optionally in a dedicate namespace `format`

//usage
print(std::cout, tile, numpy{});
```

The first one at first look appears more handy, since you can do something like `using print_numpy = print<numpy>`.

Do you have any suggestion/alternative?

albestro: No, it was a leaked experiment. 😁

rasolca: A third alternative can be
```
enum class Format{
  numpy
};

// usage
print(std::cout, tile, Format::numpy);
```
Not sure about which is the better structure.

albestro: That's true! I quickly thought about this and a proposal can be:
- reorder parameters like `print(tile, stream)` so that the stream can have a default parameter (e.g. `std::cout`)
- with the previous point in mind, I would use the format as template parameter. This would allow to specify or use the default value for stream and format, independently. In fact, if both of them are classic function parameters, depending on the order, if we want to specify the second one, we are forced to specify also the first one (assuming we want default values).

So, this proposal would be something like
```cpp
// enum/struct for describing available formats

print(tile);            // default format and default stream
print(tile, std::cout); // default format and output on cout
print<numpy>(tile);     // numpy format and default stream
print(tile, std::cout); // default format and output on cout
```

albestro: An addition/correction to my previous comment. Since it is already a templated function, adding a default parameter as a template would mean partial specialization for it, which is not allowed. On the other hand, we could use enable_if, but I think it is a big machinery not worth it.

I would go for not having a default format, and have it as a normal parameter before the stream (which could still be optional). And having the format as parameter, I would opt for struct (possibly in a namespace) instead of enum, so we can use overloading for selecting different format implementation.

albestro:
This is the current output format.

```python
mat = np.zeros((9, 7), dtype=np.csingle)
mat[0:3,0:3] = np.array([0+0j,1-1j,2-2j,1+1j,2+0j,3-1j,2+2j,3+1j,4+0j,], dtype=np.csingle).reshape(3, 3).T
mat[3:6,0:3] = np.array([3-3j,4-4j,5-5j,4-2j,5-3j,6-4j,5-1j,6-2j,7-3j,], dtype=np.csingle).reshape(3, 3).T
mat[6:9,0:3] = np.array([6-6j,7-7j,8-8j,7-5j,8-6j,9-7j,8-4j,9-5j,10-6j,], dtype=np.csingle).reshape(3, 3).T
mat[0:3,3:6] = np.array([3+3j,4+2j,5+1j,4+4j,5+3j,6+2j,5+5j,6+4j,7+3j,], dtype=np.csingle).reshape(3, 3).T
mat[3:6,3:6] = np.array([6+0j,7-1j,8-2j,7+1j,8+0j,9-1j,8+2j,9+1j,10+0j,], dtype=np.csingle).reshape(3, 3).T
mat[6:9,3:6] = np.array([9-3j,10-4j,11-5j,10-2j,11-3j,12-4j,11-1j,12-2j,13-3j,], dtype=np.csingle).reshape(3, 3).T
mat[0:3,6:7] = np.array([6+6j,7+5j,8+4j,], dtype=np.csingle).reshape(1, 3).T
mat[3:6,6:7] = np.array([9+3j,10+2j,11+1j,], dtype=np.csingle).reshape(1, 3).T
mat[6:9,6:7] = np.array([12+0j,13-1j,14-2j,], dtype=np.csingle).reshape(1, 3).T
```

I'm investigating a problem I have with `np.single`/`np.csingle` in python, I will update as soon as I have more information.

albestro:
Ok, found out the problem I had with my scripts when I switched to `np.single`.

In my script, I use something very similar to `CHECK_MATRIX_NEAR` which is `np.allclose`. It allows to specify absolute and relative error, but it also provide defaults for them. Specifically `rtol=1e-05` and `atol=1e-08`(from https://numpy.org/doc/stable/reference/generated/numpy.allclose.html).

```
In [52]: np.finfo(np.single)
Out[52]: finfo(resolution=1e-06, min=-3.4028235e+38, max=3.4028235e+38, dtype=float32)

In [53]: np.finfo(np.float)
Out[53]: finfo(resolution=1e-15, min=-1.7976931348623157e+308, max=1.7976931348623157e+308, dtype=float64)

In [54]: np.finfo(np.double)
Out[54]: finfo(resolution=1e-15, min=-1.7976931348623157e+308, max=1.7976931348623157e+308, dtype=float64)
```

The point is that switching to a lower precision type the `allclose` routine with default values for absolute/relative error fails, because the lower precision value is closer to the default threshold, and the various operations leads to bigger errors.

So, it's a problem of my script and we can use the `np.single`/`np.csingle`.

test/unit/matrix/test_matrix_output.cpp
```diff
@@ -0,0 +1,267 @@
...
+    return {real, imag};
+  }
+};
+
+template <class T>
+::testing::AssertionResult parseAndCheckTile(const std::string numpy_output,
+                                             const Tile<const T, Device::CPU>& tile) {
+  const std::regex regex_syntax{
+      "np\\.array\\(\\[(.*)\\],\\s*dtype=np\\.(\\w+)\\)"  // np.array([<values>], dtype=np.<type>)
+      "\\.reshape\\((\\d+),\\s*(\\d+)\\)\\.T\\s*"};       // .reshape(<n>, <n>).T
```
teonnik, line 101: The unit tests are quite comprehensive but I find that their complexity exceeds that of what they are testing. I think it's worth aiming for simplicity. I had to google to understand all the parts of the regex here, it's very likely that I will have to do it again in a few weeks if I come back to this part of the code.

Instead of using `std::regex`, why don't you instead write down the complete expected string explicitly and compare it to what you get? I know it's tempting to automate it but I find it makes more sense to test several "interesting" cases, instead of a large number of similar cases. Sometimes it does make sense to automate and test in bulk (like for Cholesky or the triangular solver), but I think we can achieve a reasonable confidence in correctness by being more explicit about the expected results and by picking a few "interesting" cases here.

However, I can understand if you would like to keep the code as it is, but I do think there is certain maintainability and comprehensibility costs that have to be taken into account.

albestro: Yes, I agree, it's way more "complex" the testing code than the code to be tested.

For me it's ok, we can pick up some specific local/distributed cases and just test them.

If others agree I can simplify them, it's not a problem. @rasolca @Sely85

albestro:
In the last commit d85df57 there is a proposal for test simplification avoiding regex and using fixed expected output.

Waiting for your suggestions/comments before proceeding with the rest of the tests.

test/unit/matrix/test_matrix_output.cpp
Sely85: I agree that the `regex `expressions are really "complex" to understand. A "simplified" version of unit tests (when it is possible) like that used in `test_matrix_local.cpp` looks more clear to me.

test/unit/matrix/test_matrix_local.cpp
```diff
@@ -99,6 +100,67 @@ TYPED_TEST(MatrixLocalTest, Copy) {
...
+
+    return std::make_pair(std::move(mat), expected_output);
+  }
+};
+
+TYPED_TEST(MatrixLocalTest, OutputNumpyForamt) {
+  const auto test_config = test1<TypeParam>{}();
+
+  std::ostringstream stream_matrix_output;
+  print(format::numpy{}, "mat", test_config.first, stream_matrix_output);
```
Sely85, line 159: I have tried to use this code in a python script: the result looks really good! I think this will be a useful utility for users as well. Great job!

teonnik:
@albestro looks good to me! Very clear. :+1:

albestro:
bors try

albestro:
bors-

albestro:
bors try

albestro:
It seems there is a difference between `clang` and `gcc` (probably related to SFINAE, locally on my notebook it builds while on daint it doesn't). I will investigate further tomorrow.

albestro:
bors try

let's try this fix...still don't know why it fixes the problem

albestro:
It built correctly, I just had to rerun one step manually.

albestro:
bors try

let's test the real fix, then it should be ready for review/merge!

rasolca (approved):

teonnik (approved):

Issue #285, opened 2020-11-02
Adapt CHECK_TILE_NEAR to check only absolute error


Sely85:
Using relative error to check small numbers can be misleading, see https://github.com/eth-cscs/DLA-Future/pull/261#discussion_r515132060.  
While CHECK_MATRIX_NEAR can be used only to verify only absolute error, CHECK_TILE_NEAR macro does not work if only this error (not both) is inspected.  

_Originally posted by @albestro in https://github.com/eth-cscs/DLA-Future/pull/261#discussion_r514989781

rasolca:
Only the documentation (preconditions) has to be updated, the macro already works:
- if you pass 0 to relative error, only the absolute error is checked,
- if you pass 0 to absolute error, only the relative error is checked.

albestro:
Not really sure about that.

this is the tile utility

https://github.com/eth-cscs/DLA-Future/blob/52ea1a0f82e06193f74c28c1abc6e39f09a695ca/test/include/dlaf_test/matrix/util_tile.h#L164-L168

compared to matrix one

https://github.com/eth-cscs/DLA-Future/blob/52ea1a0f82e06193f74c28c1abc6e39f09a695ca/test/include/dlaf_test/matrix/util_matrix.h#L130-L135

Sely85:
I gave it a try and it works. 
I updated the preconditions in `util_tile.h` to look like those in `util_matrix.h`:
```
void checkNear(ElementGetter expected, const Tile<T, Device::CPU>& tile, BaseType<T> rel_err,
               BaseType<T> abs_err, const char* file, const int line) {
  ASSERT_GE(rel_err, 0);
  ASSERT_GE(abs_err, 0);
  ASSERT_TRUE(rel_err > 0 || abs_err > 0);
```
Sometimes you may need to increase the "coefficient" of the absolute error. For example this one:
https://github.com/eth-cscs/DLA-Future/blob/52ea1a0f82e06193f74c28c1abc6e39f09a695ca/test/unit/test_blas_tile/test_gemm.h#L108-L109
needs to be re-written as: 
```
 CHECK_TILE_NEAR(res_c, c, 0,
                  20 * (k + 1) * TypeUtilities<T>::error);
```
to pass the test (coefficient increased from 2 to 20).

rasolca:
I consider precondition asserts part of the doc.
checkNear can be fixed in a PR.
[...]

PR #286, opened 2020-11-05
Fixing preconditions in checkNear (Tile)

Sely85:
Fixing preconditions in `checkNear`, used for `Tile`, to allow checking only relative or absolute error (instead of forcing both checks), as implemented for `Matrix`. 

This solves issue #285.


rasolca:
Please fix the preconditions in the documentation as well.

rasolca (approved):

PR #288, opened 2020-11-09
Fix a couple of trailing_executor initializations

msimberg:
The high priority executor was not being used at all with the old code.


Sely85 (approved):
Nice catch! LGTM.

rasolca (approved):

rasolca:
bors try

PR #289, opened 2020-11-09
Add GPU matrix copy specializations

msimberg:
Organizational questions:
- Do you prefer to move out the GPU copy specializatoins to another file like the host `lacpy`? I find it rather pointless considering that the implementations are just a couple of assertions and a cuBLAS call. `lacpy` has some other wrappers to go with it so the separate file makes sense. I'll be adding similar wrappers for cuBLAS (non-copy) operations after this so maybe a `dlaf/cublas_tile.hpp` file makes sense considering that?
- ~There is no GPU to GPU copy in cuBLAS. I've currently stuck a `static_assert` that always fails in the implementation for a slightly better error message. Do you prefer this, or just no specialization at all? Or a runtime error?~ Using `cudaMemcpy2D` now.

Note that the copies are currently blocking copies. We can update these to be asynchronous copies once the CUDA/cuBLAS executors etc. are in place.


include/dlaf/matrix/copy.h
```diff
@@ -37,8 +36,8 @@ void copy(MatrixTypeSrc<const T, device>& source, MatrixTypeDst<T, device>& dest
 
   for (SizeType j = 0; j < local_tile_cols; ++j)
     for (SizeType i = 0; i < local_tile_rows; ++i)
-      hpx::dataflow(hpx::util::unwrapping(dlaf::copy<T>), source.read(LocalTileIndex(i, j)),
-                    dest(LocalTileIndex(i, j)));
+      hpx::dataflow([](auto&& source,
+                       auto&& destination) { dlaf::copy<>(source.get(), destination.get()); },
```
msimberg, line 40: This wrapping lambda is only here because of overload resolution not working through `unwrapping`. An alternative would be to rename the `Tile` overload (or the `Matrix` overload) to e.g. `copy_tile`.

rasolca:
Please note that you can use cudaMemcpy2D (and later cudaMemcpy2DAsync) for device to device copies.
I usually prefer this approach as, you can copy from any to any other type of devices with the same call using `cudaMemcpyDefault` (which determine the direction of the copies through the pointer value).

msimberg:
> Please note that you can use cudaMemcpy2D (and later cudaMemcpy2DAsync) for device to device copies.
> I usually prefer this approach as you can copy from any to any other type of devices with the same call using `cudaMemcpyDefault` (which determine the direction of the copies through the pointer value).

Very good point, thanks. I'll update this to use `cudaMemcpy2D` instead.

msimberg:
What's the status of CUDA CI (I see no mention in the config files)? Is GitHub actions (although no GPUs here), GitLab runners, or Jenkins the preferred environment?

albestro:
> What's the status of CUDA CI (I see no mention in the config files)? Is GitHub actions (although no GPUs here), GitLab runners, or Jenkins the preferred environment?

The expert is @haampie ! He surely can help you about it. I think he will help you with the GitLab runners option.

haampie:
Does a CUDA version need separate release/debug&codecov builds as well? I should also condense those dockerfiles, they're super verbose.

rasolca:
@haampie Yes, the 3 variants are needed. What need to be condensed? the dockerfile or the output itself?

haampie:
The dockerfiles, currently it's a lot of copy & paste.

msimberg:
Oh, and by the way: I need to open this PR with a branch from this repo, not my fork, to get the GitLab CI to run, right?

rasolca:
@msimberg No you can run the tests from a fork too. See e.g. #288.

haampie:
Doesn't matter if it's a fork or not, CI is triggered by commenting `bors try` on a new line, see https://bors.tech/

include/dlaf/matrix/copy_tile.h
```diff
@@ -12,14 +12,73 @@
...
 }
 
+template <typename T, Device Source, Device Destination>
+void copy(const matrix::Tile<const T, Source>& source, const matrix::Tile<T, Destination>& destination) {
+  DLAF_ASSERT_HEAVY(
+      source.size().rows() <= destination.size().rows(),
+      "number of rows in destination must be at least as large as number of rows in source for tile copy");
+  DLAF_ASSERT_HEAVY(
+      source.size().cols() <= destination.size().cols(),
+      "number of columns in destination must be at least as large as number of columns in source for tile copy");
```
rasolca, line 81: I'm not sure we should allow copy to larger tiles... What is the use case you have in mind?

msimberg: None. I left it like this only because it's possible. I'll happily make it stricter if you prefer that.

msimberg: I've updated the assertions to require equality.

rasolca (approved):
LGTM.
Sorry to have deleted a comment, it was reported as a pending review due to a refresh problem, but in fact it was the original review.

msimberg:
Converted to draft until CUDA CI is in place.

teonnik (approved):

msimberg:
(just to see that I didn't break anything without CUDA enabled...)

bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
Rebased on #300.

bors try

rasolca:
bors try

rasolca:
Note that the copies are currently blocking copies. We can update these to be asynchronous copies once the CUDA/cuBLAS executors etc. are in place.

Issue #236, opened 2020-06-02
Lighten the folder structure


albestro:
@teonnik pointed out during his reviews that the tree structure of algorithms (i.e. factorization, solver, auxiliary) can be lighten a bit.

Mainly the proposals are about:
- CMakeFiles organization for `test/unit` (https://github.com/eth-cscs/DLA-Future/pull/224#discussion_r427290606 and https://github.com/eth-cscs/DLA-Future/pull/224#discussion_r427291333)
- Evaluate moving implementation details to `src` (https://github.com/eth-cscs/DLA-Future/pull/224#discussion_r427273595)

As soon as we decide how to streamline, we can proceed on changing all existing instances at the same way in one shot.

teonnik:
Do we have consensus on the following suggestions or do we want some changes? :

- Move contents of `.tpp` files into corresponding `.cpp` files before the ETI. If there are multiple `.tpp` corresponding to a `.h`, move the contents of all `.tpp` to the `.cpp` (e.g. `matrix.h` with `matrix.tpp`, `matrix_const.tpp` to `matrix.cpp`). 
- Headers only used within `.cpp` files are private/internal and should be in the `src` directory.
- Headers that are private/internal but are used within our public headers should stay in the `include/dlaf` directory.

The current structure was discussed here : https://github.com/eth-cscs/DLA-Future/pull/146 

There are a few issues I have with it:

- deeply nested hierarchies with a lot of small, "glue" files like `factorization/internal.h`. 
- often several directory layers have to be navigated with similarly named files/folders (e.g. `mc`/`mc.h`/`mc.tpp`) to get to the actual files containing the implementation of the algorithms. 
- The structure is designed to have separate `mc` (multicore) and e.g. `gpu` (or other backends) hierarchies, however it is not clear yet whether the separation is beneficial as it is not clear whether the `mc`/`gpu` back ends will have divergent or similar implementations.
- overuse of namespaces : it's good to group things into namespaces but we should use them judiciously and for their main purpose : to avoid name clashes https://stackoverflow.com/questions/713698/c-namespaces-advice
- using struct with static methods when function overloading is sufficient (e.g. in `factorization/mc.h`)

For example, instead of this:

```
solver
├── internal.h
├── mc.h
└── triangular/
    ├── mc
    │   ├── triangular_LLN.h
    │   ├── triangular_LLT.h
    │   ├── triangular_LUN.h
    │   ├── triangular_LUT.h
[...]

PR #290, opened 2020-11-09
first proposal according #236

rasolca:
I refactored Cholesky according to Mikael suggestion.

new API (which dispatches to internal API):
```
dlaf::factorization::cholesky<backend, device, Type>(...)
```
new internal API:
```
template <Backend b, Device d, class T>
Cholesky{};

template <class T>
Cholesky<MC, CPU, T>{
// Note that call_L cannot be defined here, otherwise ETI doesn't work. 
static void call_L(..);
};
// non inline definition needed by ETI
template <class T>
void Cholesky<MC, CPU, T>::call_L(..){
  ..
}
```

The new directory structure is the following:
```
factorization/
    cholesky.h  <-  user API (template functions that dispatch to internal API)
    cholesky/
         cholesky.h  <-  internal template struct declaration
         cholesky_mc.h  <-  template partial specialization for MC backend
    # other algorithms e.g. QR
    qr.h
    qr/
         qr.h
         qr_mc.h
         t_factor.h
         t_factor_mc.h
```
 
@albestro @teonnik @msimberg What do you think?


teonnik:
I like where this is heading. A few points of discussion:

The `cholesky/` contains the implementation details of the API. Can we rename things a little bit, for example:

`cholesky/cholesky.h` -> `cholesky/api.h` (i.e. the internal API of cholesky)
`cholesky/cholesky_mc.h` -> `cholesky/mc.h` (to avoid name repetitions)

include/dlaf/factorization/cholesky/mc.h
```diff
@@ -22,18 +22,24 @@
...
-namespace mc {
 
-// Local implementation of Lower Cholesky factorization.
-template <class T>
-void cholesky_L(Matrix<T, Device::CPU>& mat_a) {
+template <Device device, class T>
+struct Cholesky<Backend::MC, device, T> {
+  static void call_L(Matrix<T, Device::CPU>& mat_a);
+  static void call_L(comm::CommunicatorGrid grid, Matrix<T, Device::CPU>& mat_a);
+};
```
teonnik, line 39: Do we need the `Device device` template parameter here? It doesn't seem to be used in `call_L()`. Perhaps we can just set it to `Device::CPU` and remove it from `template<...>`

teonnik (changes requested):

include/dlaf/factorization/factorization.h
```diff
@@ -9,11 +9,4 @@
...
 
-#include "dlaf/types.h"
-
-namespace dlaf {
-
-template <Backend execution>
-struct Solver;
-
-}
+#include "dlaf/factorization/cholesky.h"
```
teonnik, line 12: This header should probably replace the `factorization.h` header outside of the `factorization/` directory. That header has a wrong include statement in it (dlaf/factorization/mc.h), this one has the right include statement.

teonnik (approved):

rasolca:
bors try

include/dlaf/auxiliary/norm/api.h
```diff
@@ -12,8 +12,12 @@
 #include "dlaf/types.h"
 
 namespace dlaf {
+namespace auxiliary {
+namespace internal {
 
-template <Backend execution>
-struct Eigensolver;
+template <Backend backend, Device device, class T>
+struct Norm {};
```
albestro, line 19: like this.

albestro (approved):
Very good job!

Just a technical question: is there any benefit of using empty structs instead of just declaring them? See my comment just for reference in the code.

PR #291, opened 2020-11-09
Matrix helper for tests

albestro:
The aim of this PR is introducing a simple data-structure for a Matrix, useful during tests writing.

For this reason I would go for keeping it in the `dlaf::test` namespace, not making it part of our library.

The main idea is to have a `MatrixLocal` that has these main properties:

- it is not async, so it does not have to deal with futures;
- it is not thread-safe;
- it is fully locally stored;
- it should be possible to work with it in terms of tiles or directly elements: this double-abstraction over it allows to keep a direct connection with a distributed matrix (if it has been "gathered" from it) but at the same time it is easy to access (address) of elements for inter-operability with BLAS/LAPACK;

At the moment I wouldn't mind about having it device portable (e.g. stored on GPU or CPU). The philosophy can be, as soon as we need it, we can improve it.

I already have some helper functions that I'm going to add soon. Some examples:
- `set`, similar to what it is already available for `dlaf::matrix::Matrix`
- `all_gather` that allows to collect locally a distributed `dlaf::matrix::Matrix`

### DIFF
--- a/test/include/dlaf_test/matrix/matrix_local.h
+++ b/test/include/dlaf_test/matrix/matrix_local.h
@@ -28,13 +28,15 @@ struct MatrixLocal;
 
 /// MatrixLocal is a local matrix with column-layout, not thread-safe
 ///
-/// It is a useful helper object that allows you to access element directly given their
-/// GlobalElementIndex, or by accessing the tile via its GlobalTileIndex and then the
+/// It is a useful helper object that allows you to access elements directly given their
+/// GlobalElementIndex, or by accessing the tile containing it via its GlobalTileIndex and then the
 /// related LocalTileIndex inside it.
 ///
 /// It uses Index/Size with the Global tag instead of the Local one, because its main task
 /// is to create a local copy of a distributed matrix. So, it is generally easier to think
 /// of it as the global matrix.
+///
+/// About its semantic, it is similar to a std::unique_ptr.
 template <class T>
 struct MatrixLocal<const T> {
   static constexpr auto CPU = Device::CPU;
@@ -52,6 +54,9 @@ struct MatrixLocal<const T> {
                                    layout_.minTileMemSize(tile_index)},
                           layout_.ldTile());
   }
+
+  MatrixLocal(const MatrixLocal&) = delete;
+  MatrixLocal& operator=(const MatrixLocal&) = delete;
 
   MatrixLocal(MatrixLocal&&) = default;
 
@@ -116,6 +121,9 @@ struct MatrixLocal : public MatrixLocal<const T> {
 
   MatrixLocal(GlobalElementSize size, TileElementSize blocksize) noexcept
       : MatrixLocal<const T>(size, blocksize) {}
+
+  MatrixLocal(const MatrixLocal&) = delete;
+  MatrixLocal& operator=(const MatrixLocal&) = delete;
 
   MatrixLocal(MatrixLocal&&) = default;
 
--- a/test/include/dlaf_test/matrix/matrix_local.h
+++ b/test/include/dlaf_test/matrix/matrix_local.h
@@ -88,19 +88,6 @@ struct MatrixLocal<const T> {
     return GlobalTileSize{layout_.nrTiles().rows(), layout_.nrTiles().cols()};
   }
 
-  friend std::ostream& operator<<(std::ostream& os, const MatrixLocal& matrix) {
-    using dlaf::util::size_t::mul;
-    using dlaf::util::size_t::sum;
-
-    os << matrix.size() << matrix.ld() << std::endl;
-    for (const auto& index : iterate_range2d(matrix.size())) {
-      auto linear_index = dlaf::to_sizet(sum(mul(index.col(), matrix.ld()), index.row()));
-      os << "[" << index.row() << ", " << index.col() << "]" << '=' << *matrix.memory_(linear_index)
-         << "\n";
-    }
-    return os;
-  }
-
 protected:
   SizeType elementLinearIndex(const GlobalElementIndex& index) const noexcept {
     using dlaf::util::size_t::mul;
--- a/test/include/dlaf_test/matrix/matrix_local.h
+++ b/test/include/dlaf_test/matrix/matrix_local.h
@@ -57,16 +57,17 @@ struct MatrixLocal<const T> {
                       tiles_.size());
   }
 
-  /// Access directly elements
+  /// Access elements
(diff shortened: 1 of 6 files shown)


albestro:
In this work, I found myself in need of generalizing a bit also the `CHECK_TILE`/`CHECK_MATRIX` macros and machinery.

This topic covers the entire MatrixLike structures (`Matrix` and `MatrixView`), and so the new `MatrixLocal` is affected too.

The proposal in few points:
1. For both matrix and tile utils, `ElementGetter` can be anything that has a call operator that given a `GlobalElementIndex` returns a value; a lambda/function, but also a `MatrixLocal`.
2. For matrix utils we have two levels in the architecture: the specific check function (e.g. `checkNear`) that passes to `internal::check` all the tools to do the check. The idea is to make the access to check functions like `checkNear` more generic, so that it can accepts any "MatrixLike" type, being it a `Matrix`, a `MatrixView` or a `MatrixLocal`, and then implement the specific version of `internal::check` that may differ (e.g. `MatrixLocal` can access directly elements, while `Matrix` has to go get the future of the tile and then check the value).

The second point for the moment is easy to implement because the `MatrixLocal` differs in number of templates parameter (i.e. it implicitly uses `Device::CPU`), but as soon as we need to extend it, we could extend `matrix_traits` so that it returns and use it for the enable if (e.g. by checking if operator(GlobalElementIndex) exists).

About the first, IMHO it makes it more versatile and useful, even more with the new `MatrixLocal`. A problem that I see is that there is no check about the size: a lambda is generally implemented for represeting an "infinte-size" matrix, so it returns values for any coordinate, while a `MatrixLocal` is limited to its size. IMHO, it is a secondary problem and the benefit overcomes the problems.

test/include/dlaf_test/matrix/matrix_local.h
```diff
@@ -0,0 +1,159 @@
...
+///
+/// It is a useful helper object that allows you to access elements directly given their
+/// GlobalElementIndex, or by accessing the tile containing it via its GlobalTileIndex and then the
+/// related LocalTileIndex inside it.
+///
+/// It uses Index/Size with the Global tag instead of the Local one, because its main task
+/// is to create a local copy of a distributed matrix. So, it is generally easier to think
+/// of it as the global matrix.
+///
+/// About its semantic, it is similar to a std::unique_ptr.
```
rasolca, line 39: I would add that that it uses column major layout and the full matrix can be used in blas and lapack.

test/include/dlaf_test/matrix/matrix_local.h
```diff
@@ -0,0 +1,159 @@
...
+  const TileT& tile(const GlobalTileIndex& index) const noexcept {
+    return tiles_[static_cast<size_t>(tileLinearIndex(index))];
+  }
+
+protected:
+  using MatrixLocal<const T>::elementLinearIndex;
+  using MatrixLocal<const T>::tileLinearIndex;
+  using MatrixLocal<const T>::layout_;
+  using MatrixLocal<const T>::memory_;
+  using MatrixLocal<const T>::tiles_;
```
teonnik, line 154: Using `BaseT = MatrixLocal<const T>` to make this a little shorter.

include/dlaf/matrix_const.tpp
```diff
@@ -77,7 +77,7 @@ hpx::shared_future<Tile<const T, device>> Matrix<const T, device>::read(
 }
 
 template <class T, Device device>
-Matrix<const T, device>::Matrix(Distribution&& distribution,
+Matrix<const T, device>::Matrix(Distribution distribution,
                                 std::vector<hpx::future<TileType>>&& tile_futures,
                                 std::vector<hpx::shared_future<ConstTileType>>&& tile_shared_futures)
```
teonnik, line 82: Not sure if this is a fly-by change but you can take the remaining arguments by value (i.e. std::vector<...>) and move them the same way you did for `Distribution`. If the class is going to take ownership of the objects (which it does), declaring them by value may have some optimization benefits and covers the const& case (i.e when you anyway have to make a copy). There was a talk about that by Chandler Carruth sometime ago: https://www.youtube.com/watch?v=eR34r7HOU14&feature=youtu.be

teonnik (approved):

test/include/dlaf_test/matrix/matrix_local.h
albestro: I agree about BLAS/LAPACK, I'm going to add it.

About the column major layout I'm interested in knowing if **"MatrixLocal is a local matrix with column-layout, not thread-safe"** is ok like this or you have interpreted it differently.

include/dlaf/matrix_const.tpp
albestro: Yes, I agree with you. It is something I learnt months ago, but unfortunately after I implemented part of the structures in DLAF. I can change them here too if you agree.

test/include/dlaf_test/matrix/matrix_local.h
rasolca: Please note that the correct wording is "column-major layout".
With "local matrix with column-layout" I understand a matrix with m >> n (like a col panel).
As brief doc I would use `MatrixLocal is a column-major local matrix, not thread-safe.`
In the doc body I would connect the column-major layout to the possibility to use it in blas and lapack, and expand also the meaning of thread-safe (i.e. no futures used => cannot be used to schedule tasks)

include/dlaf/matrix_const.tpp
albestro: While I was looking at it, I found out that it was not even needed anymore. So, I took the chance and moved the majority of these Matrix changes to a dedicated PR.

test/include/dlaf_test/matrix/matrix_local.h
albestro: Do you think is it more clear now? Do you think it would be better to change/add anything else?

Sely85 (approved):
Looks good to me!

test/include/dlaf_test/matrix/matrix_local.h
rasolca: LGTM

rasolca (approved):

rasolca:
bors try

albestro:
bors try

GCC (at least 8.3) doesn't like the auto keyword as return argument, saying that it was not able to deduce the template type. I don't know if it is somehow related to the problem that @msimberg and @teonnik had with `copy`.

PR #276, opened 2020-10-19
Making arguments "const" in Lapack and Blas wrappers

Sely85:
Added `const` declaration to Lapack and Blas functions arguments (`blas::Uplo`, `blas::Side`, ...) to avoid erroneous modifications.

Ref. comment by @albestro in https://github.com/eth-cscs/DLA-Future/pull/261#discussion_r506315003

PR #292, opened 2020-11-10
Add cuBLAS wrappers for tile operations

msimberg:
This adds the basic wrappers that do argument conversion (blaspp enums to cuBLAS enums, `std::complex` to `cuComplex`/`cuDoubleComplex`), and tests.

Questions:
- Namespaces and names? The blaspp wrappers are in `dlaf::tile`. The signatures won't be the same since the cuBLAS wrappers will take a cuBLAS handle. Is this functionality meant to be public? I've currently prefixed the wrappers with `cublas_`.
- Do you rely only on `-Wswitch-enum` to cover all cases in a switch statement? I currently added a false assertion on the default case, but I see that you haven't done that elsewhere.

Includes #289.

### DIFF
--- a/include/dlaf/cublas_tile.h
+++ b/include/dlaf/cublas_tile.h
@@ -1,0 +1,200 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2020, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+
+/// @file
+
+#ifdef DLAF_WITH_CUDA
+
+#include <cublas_v2.h>
+#include <blas.hh>
+
+#include "dlaf/cublas/error.h"
+#include "dlaf/matrix/tile.h"
+#include "dlaf/types.h"
+#include "dlaf/util_cublas.h"
+
+namespace dlaf {
+namespace tile {
+namespace internal {
+template <typename T>
+struct CublasTrsm;
+
+template <>
+struct CublasTrsm<float> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasStrsm(std::forward<Args>(args)...));
+  }
+};
+
+template <>
+struct CublasTrsm<double> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasDtrsm(std::forward<Args>(args)...));
+  }
+};
+
+template <>
+struct CublasTrsm<std::complex<float>> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasCtrsm(std::forward<Args>(args)...));
+  }
+};
+
+template <>
+struct CublasTrsm<std::complex<double>> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasZtrsm(std::forward<Args>(args)...));
+  }
+};
+
+template <typename T>
+struct CublasGemm;
+
+template <>
+struct CublasGemm<float> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasSgemm(std::forward<Args>(args)...));
+  }
+};
+
+template <>
+struct CublasGemm<double> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasDgemm(std::forward<Args>(args)...));
+  }
+};
+
+template <>
+struct CublasGemm<std::complex<float>> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasCgemm(std::forward<Args>(args)...));
+  }
+};
+
+template <>
+struct CublasGemm<std::complex<double>> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasZgemm(std::forward<Args>(args)...));
+  }
+};
+
+template <typename T>
+struct CublasHerk;
+
+template <>
+struct CublasHerk<std::complex<float>> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasCherk(std::forward<Args>(args)...));
+  }
+};
+
+template <>
+struct CublasHerk<std::complex<double>> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasZherk(std::forward<Args>(args)...));
+  }
+};
+
+}
+
+template <class T>
+void cublasTrsm(cublasHandle_t handle, const blas::Side side, const blas::Uplo uplo, const blas::Op op,
+                const blas::Diag diag, const T alpha, const matrix::Tile<const T, Device::GPU>& a,
+                const matrix::Tile<T, Device::GPU>& b) {
+  SizeType m = b.size().rows();
+  SizeType n = b.size().cols();
+
+  DLAF_ASSERT(a.size().rows() == a.size().cols(), "`a` is not square!", a);
+
+  auto left_side = (side == blas::Side::Left ? m : n);
+  DLAF_ASSERT(a.size().rows() == left_side, "`a` has an invalid size!", a, left_side);
+
+  internal::CublasTrsm<T>::call(handle, util::blasToCublas(side), util::blasToCublas(uplo),
+                                util::blasToCublas(op), util::blasToCublas(diag), m, n,
+                                util::blasToCublasCast(&alpha), util::blasToCublasCast(a.ptr()), a.ld(),
+                                util::blasToCublasCast(b.ptr()), b.ld());
+}
+
+template <class T>
+void cublasGemm(cublasHandle_t handle, const blas::Op op_a, const blas::Op op_b, const T alpha,
+                const matrix::Tile<const T, Device::GPU>& a, const matrix::Tile<const T, Device::GPU>& b,
+                const T beta, const matrix::Tile<T, Device::GPU>& c) {
+  SizeType m;
+  SizeType k;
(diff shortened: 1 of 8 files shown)


include/dlaf/util_cublas.h
```diff
@@ -0,0 +1,110 @@
...
+#include <cublas_v2.h>
+#include <blas.hh>
+
+// TODO: Which namespace?
+namespace dlaf {
+namespace internal {
+namespace gpu {
+
+// TODO: Are these necessary?
+inline constexpr cublasSideMode_t blas_to_cublas_side(blas::Side side) {
```
teonnik, line 24: Perhaps you can overload `blas_to_cublas()` and remove the suffixes `_side`, `_uplo`, etc. 

Also an alternative name for the function could be `blas2cublas()`, in case someone prefers it. I am OK with the current name too.

teonnik (approved):
LGTM! Just a minor comment.

EDIT:

> Namespaces and names? The blaspp wrappers are in dlaf::tile. The signatures won't be the same since the cuBLAS wrappers will take a cuBLAS handle. Is this functionality meant to be public? I've currently prefixed the wrappers with cublas_.

Yes, I think the handle should be in the API in case a user has multiple handles (e.g. when using multiple GPUs from a single process).

I am not sure about the namespace, I use `dlaf::cuda::` and `dlaf::cublas::` in #234 .

> Do you rely only on -Wswitch-enum to cover all cases in a switch statement? I currently added a false assertion on the default case, but I see that you haven't done that elsewhere.

It seems we do: `cmake/DLAF_AddTargetWarnings.cmake` defines `target_add_warnings()` which includes the switch. The function is used in `miniapp/CMakeLists.txt` to include the flags to the `DLAF` target.

include/dlaf/util_cublas.h
msimberg: I'm indifferent about the particular name. Is there precedent for `x2y`? The `_type` suffix is indeed unnecessary, thanks! I'll remove that.

msimberg:
> LGTM! Just a minor comment.
> 
> EDIT:
> 
> > Namespaces and names? The blaspp wrappers are in dlaf::tile. The signatures won't be the same since the cuBLAS wrappers will take a cuBLAS handle. Is this functionality meant to be public? I've currently prefixed the wrappers with cublas_.
> 
> Yes, I think the handle should be in the API in case a user has multiple handles (e.g. when using multiple GPUs from a single process).

My question was confusingly worded. It was not about having the handle in the signature (that I take as a given), but rather about the `internal` or non-`internal` namespace. The blaspp wrappers are now in `dlaf::tile`, i.e. public. `dlaf::cuda::internal` is obviously not meant for public use, and the scheme doesn't fit where the blaspp wrappers are. I'd think either the blaspp wrappers should move to `dlaf::something::internal` or the cuBLAS wrappers should move to `dlaf::tile`. Preferences? I can also leave things as they are and leave it for another PR.

> I am not sure about the namespace naming, I use `dlaf::cuda::` and `dlaf::cublas::` in #234 .

Indeed, that's why I followed the same namespace.

include/dlaf/cublas_tile.h
```diff
@@ -0,0 +1,202 @@
...
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasZtrsm(std::forward<Args>(args)...));
+  }
+};
+
+template <class T>
+void cublas_trsm(cublasHandle_t handle, blas::Side side, blas::Uplo uplo, blas::Op op, blas::Diag diag,
+                 T alpha, const matrix::Tile<const T, Device::GPU>& a,
+                 const matrix::Tile<T, Device::GPU>& b) {
```
Sely85, line 68: A couple of weeks ago we added `const` to all arguments of LAPACK and BLAS wrappers (see #276). I actually don't know how cuBLAS works, sorry. But do you think it is worth marking them as `const` here too?

msimberg:
> > I am not sure about the namespace naming, I use `dlaf::cuda::` and `dlaf::cublas::` in #234 .
> 
> Indeed, that's why I followed the same namespace.

Uhh, I thought I did :D but I did not... I'll change this once there's a concensus on where the functionality should live.

include/dlaf/cublas_tile.h
msimberg: Yeah, absolutely, thanks!

include/dlaf/util_cublas.h
msimberg: I've removed the suffixes for the enumeration conversion functions. In principle `blas_to_cublas_cast` could also do without the `_cast` suffix. I slightly prefer to have it there to explicitly say that we're casting and returning the same underlying value in a different form, rather than a new value, but I can just as well go without the suffix as well.

include/dlaf/cublas_tile.h
msimberg: Updated.

include/dlaf/util_cublas.h
teonnik: > I've removed the suffixes for the enumeration conversion functions. In principle blas_to_cublas_cast could also do without the _cast suffix. I slightly prefer to have it there to explicitly say that we're casting and returning the same underlying value in a different form, rather than a new value, but I can just as well go without the suffix as well.

I also think it's worth keeping the suffix for that one.

> I'm indifferent about the particular name. Is there precedent for x2y

I think that's some UNIX-sm I may have subconsciously picked up (utilities like pdf2svg, etc). The current name is pretty clear, let's just keep it as it is.

rasolca:
> > LGTM! Just a minor comment.
> > EDIT:
> > > Namespaces and names? The blaspp wrappers are in dlaf::tile. The signatures won't be the same since the cuBLAS wrappers will take a cuBLAS handle. Is this functionality meant to be public? I've currently prefixed the wrappers with cublas_.
> > 
> > 
> > Yes, I think the handle should be in the API in case a user has multiple handles (e.g. when using multiple GPUs from a single process).
> 
> My question was confusingly worded. It was not about having the handle in the signature (that I take as a given), but rather about the `internal` or non-`internal` namespace. The blaspp wrappers are now in `dlaf::tile`, i.e. public. `dlaf::cuda::internal` is obviously not meant for public use, and the scheme doesn't fit where the blaspp wrappers are. I'd think either the blaspp wrappers should move to `dlaf::something::internal` or the cuBLAS wrappers should move to `dlaf::tile`. Preferences? I can also leave things as they are and leave it for another PR.
>
Tile operations should be available to users (with a public API), in case they have to develop custom algorithms.
I would use `dlaf::tile`.

rasolca:
Please note that function names are supposed to be lowerCamelCase. (see misc/coding.md)

msimberg:
I moved the cuBLAS wrappers to `dlaf::tile`. Looking around a bit more I thought `dlaf::util` might even fit better for the conversion functions. I moved them there now, but again I have no real preference (except no nesting...). `dlaf::cublas` also works.

test/unit/test_cublas_tile/test_herk.h
```diff
@@ -0,0 +1,103 @@
...
+  auto a = createTile<CT>(el_op_a, size_a, lda, op_a);
+  auto c = createTile<T>(el_c, size_c, ldc);
+
+  Tile<T, Device::GPU> a0d(size_a, memory::MemoryView<T, Device::GPU>(mul(lda, size_a.cols())), lda);
+  Tile<T, Device::GPU> cd(size_c, memory::MemoryView<T, Device::GPU>(mul(ldc, size_c.cols())), ldc);
+
+  copy(a, a0d);
+  copy(c, cd);
+
+  Tile<CT, Device::GPU> ad(std::move(a0d));
```
Sely85, line 92: Do you think it would be possible, in case it is useful, to extend the usage of `createTile` to deal with GPU matrices/tiles? 
I a thinking about these functions:
https://github.com/eth-cscs/DLA-Future/blob/c8bca01662d7aad4eccc03a76272494694e72897/test/include/dlaf_test/matrix/util_tile.h#L77-L98
and 
https://github.com/eth-cscs/DLA-Future/blob/c8bca01662d7aad4eccc03a76272494694e72897/test/include/dlaf_test/matrix/util_tile_blas.h#L59-L72

msimberg: I think that could be useful, yes! If you don't mind I'd like to look at that in a separate PR though. Would that be ok?

Sely85: Yes, sure!

msimberg:
bors try

msimberg:
bors try

albestro (approved):
LGTM

include/dlaf/cublas_tile.h
```diff
@@ -0,0 +1,200 @@
...
+  SizeType k2;
+  SizeType n;
+  if (op_b == blas::Op::NoTrans) {
+    k2 = b.size().rows();
+    n = b.size().cols();
+  }
+  else {
+    k2 = b.size().cols();
+    n = b.size().rows();
+  }
```
albestro, line 160: Just a minor thought.

I'm wondering if in the future wouldn't be worth to factor out this kind of checks, that in general are not device-dependent.

msimberg: Good idea. Let me also follow up on this in a separate PR?

albestro: Absolutely! 😉

test/unit/test_cublas_tile/test_trsm.h
```diff
@@ -0,0 +1,100 @@
...
+  auto s = t.size();
+  auto n = s.cols();
+  auto m = s.rows();
+  for (int i = 0; i < m; ++i) {
+    for (int j = 0; j < n; ++j) {
+      std::cout << t(TileElementIndex{i, j}) << " ";
+    }
+    std::cout << std::endl;
+  }
+}
```
rasolca, line 50: Is this debug code supposed to me added to master?

rasolca (changes requested):

test/unit/test_cublas_tile/test_trsm.h
msimberg: Nope, good catch!

msimberg: Removed.

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

rasolca (approved):

PR #293, opened 2020-11-11
Refactor Cholesky

teonnik:
- Use global indices instead of local to improve readibility
- Refactor send/recv logic to separate functions to avoid repetitions
- Refactor most kernels (e.g. trsm, gemm, etc) to improve readibility of
  the important parameters
- Refactor Pipeline to and improve name and docs

### DIFF
--- a/include/dlaf/factorization/cholesky/mc/cholesky_L.h
+++ b/include/dlaf/factorization/cholesky/mc/cholesky_L.h
@@ -11,36 +11,107 @@
 
 #include <hpx/include/parallel_executors.hpp>
 #include <hpx/include/threads.hpp>
-#include <hpx/include/util.hpp>
-#include <hpx/local/future.hpp>
+#include <hpx/util/annotated_function.hpp>
+
+#include <unordered_map>
 
 #include "dlaf/blas_tile.h"
 #include "dlaf/common/index2d.h"
 #include "dlaf/common/pipeline.h"
 #include "dlaf/common/range2d.h"
 #include "dlaf/common/vector.h"
+#include "dlaf/communication/communicator.h"
 #include "dlaf/communication/communicator_grid.h"
 #include "dlaf/communication/executor.h"
 #include "dlaf/communication/functions_sync.h"
 #include "dlaf/lapack_tile.h"
 #include "dlaf/matrix.h"
 #include "dlaf/matrix/distribution.h"
+#include "dlaf/memory/memory_view.h"
 #include "dlaf/util_matrix.h"
 
 namespace dlaf {
 namespace internal {
 namespace mc {
 
+// If the diagonal tile is on this node factorize it
+// Cholesky decomposition on mat_a(k,k) r/w potrf (lapack operation)
+template <class T>
+void potrf_diag_tile(hpx::threads::executors::pool_executor executor_hp,
+                     hpx::future<matrix::Tile<T, Device::CPU>> matrix_tile) {
+  hpx::dataflow(executor_hp, hpx::util::unwrapping(tile::potrf<T, Device::CPU>), blas::Uplo::Lower,
+                std::move(matrix_tile));
+}
+
+template <class T>
+void trsm_panel_tile(hpx::threads::executors::pool_executor executor_hp,
+                     hpx::shared_future<matrix::Tile<const T, Device::CPU>> kk_tile,
+                     hpx::future<matrix::Tile<T, Device::CPU>> matrix_tile) {
+  hpx::dataflow(executor_hp, hpx::util::unwrapping(tile::trsm<T, Device::CPU>), blas::Side::Right,
+                blas::Uplo::Lower, blas::Op::ConjTrans, blas::Diag::NonUnit, 1.0, std::move(kk_tile),
+                std::move(matrix_tile));
+}
+
+template <class T>
+void herk_trailing_diag_tile(hpx::threads::executors::pool_executor trailing_matrix_executor,
+                             hpx::shared_future<matrix::Tile<const T, Device::CPU>> panel_tile,
+                             hpx::future<matrix::Tile<T, Device::CPU>> matrix_tile) {
+  hpx::dataflow(trailing_matrix_executor, hpx::util::unwrapping(tile::herk<T, Device::CPU>),
+                blas::Uplo::Lower, blas::Op::NoTrans, -1.0, panel_tile, 1.0, std::move(matrix_tile));
+}
+
+template <class T>
+void gemm_trailing_matrix_tile(hpx::threads::executors::pool_executor trailing_matrix_executor,
+                               hpx::shared_future<matrix::Tile<const T, Device::CPU>> panel_tile,
+                               hpx::shared_future<matrix::Tile<const T, Device::CPU>> col_panel,
+                               hpx::future<matrix::Tile<T, Device::CPU>> matrix_tile) {
+  hpx::dataflow(trailing_matrix_executor, hpx::util::unwrapping(tile::gemm<T, Device::CPU>),
+                blas::Op::NoTrans, blas::Op::ConjTrans, -1.0, std::move(panel_tile),
+                std::move(col_panel), 1.0, std::move(matrix_tile));
+}
+
+template <class T>
+void send_tile(hpx::threads::executors::pool_executor ex,
+               common::Pipeline<comm::CommunicatorGrid>& task_chain, Coord rc_comm,
+               hpx::shared_future<matrix::Tile<const T, Device::CPU>> tile) {
+  using ConstTile_t = matrix::Tile<const T, Device::CPU>;
+  using PromiseComm_t = common::PromiseGuard<comm::CommunicatorGrid>;
+
+  // Broadcast the (trailing) panel column-wise
+  auto send_bcast_f = hpx::util::annotated_function(
+      [rc_comm](hpx::shared_future<ConstTile_t> ftile, hpx::future<PromiseComm_t> fpcomm) {
+        PromiseComm_t pcomm = fpcomm.get();
+        comm::sync::broadcast::send(pcomm.ref().subCommunicator(rc_comm), ftile.get());
+      },
+      "send_tile");
+  hpx::dataflow(ex, std::move(send_bcast_f), tile, task_chain());
+}
+
+template <class T>
+hpx::future<matrix::Tile<const T, Device::CPU>> recv_tile(
+    hpx::threads::executors::pool_executor ex, common::Pipeline<comm::CommunicatorGrid>& mpi_task_chain,
+    Coord rc_comm, TileElementSize tile_size, int rank) {
+  using ConstTile_t = matrix::Tile<const T, Device::CPU>;
+  using PromiseComm_t = common::PromiseGuard<comm::CommunicatorGrid>;
+  using MemView_t = memory::MemoryView<T, Device::CPU>;
+  using Tile_t = matrix::Tile<T, Device::CPU>;
+
+  // Update the (trailing) panel column-wise
+  auto recv_bcast_f = hpx::util::annotated_function(
+      [rank, tile_size, rc_comm](hpx::future<PromiseComm_t> fpcomm) -> ConstTile_t {
+        PromiseComm_t pcomm = fpcomm.get();
+        MemView_t mem_view(tile_size.linear_size());
+        Tile_t tile(tile_size, std::move(mem_view), tile_size.rows());
+        comm::sync::broadcast::receive_from(rank, pcomm.ref().subCommunicator(rc_comm), tile);
+        return std::move(tile);
+      },
+      "recv_tile");
+  return hpx::dataflow(ex, std::move(recv_bcast_f), mpi_task_chain());
+}
+
 // Local implementation of Lower Cholesky factorization.
 template <class T>
 void cholesky_L(Matrix<T, Device::CPU>& mat_a) {
-  constexpr auto NonUnit = blas::Diag::NonUnit;
-  constexpr auto ConjTrans = blas::Op::ConjTrans;
-  constexpr auto NoTrans = blas::Op::NoTrans;
-  constexpr auto Right = blas::Side::Right;
-  constexpr auto Lower = blas::Uplo::Lower;
-
-  using hpx::util::unwrapping;
   using hpx::threads::executors::pool_executor;
   using hpx::threads::thread_priority_high;
   using hpx::threads::thread_priority_default;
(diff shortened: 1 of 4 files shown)


teonnik:
bors try

include/dlaf/factorization/cholesky/mc/cholesky_L.h
```diff
@@ -11,36 +11,103 @@
 
 #include <hpx/include/parallel_executors.hpp>
 #include <hpx/include/threads.hpp>
-#include <hpx/include/util.hpp>
-#include <hpx/local/future.hpp>
+#include <hpx/util/annotated_function.hpp>
```
msimberg, line 14: I would recommend you keep `hpx/include/util.hpp` even though it may include more than necessary. `hpx/util/annotated_function.hpp` is deprecated (or will be in the next release). We haven't decided yet into what smaller public header that functionality would go.

include/dlaf/solver/triangular/mc/triangular_LLN.h
```diff
@@ -129,26 +129,27 @@ void triangular_LLN(comm::CommunicatorGrid grid, blas::Diag diag, T alpha,
...
+                                                     auto&& comm_wrapper) -> ConstTileType {
+                              memory::MemoryView<T, Device::CPU> mem_view(
+                                  util::size_t::mul(tile_size.rows(), tile_size.cols()));
+                              TileType tile(tile_size, std::move(mem_view), tile_size.rows());
+                              comm::sync::broadcast::receive_from(index,
+                                                                  comm_wrapper.ref().rowCommunicator(),
+                                                                  tile);
+                              return std::move(tile);
+                            }),
+                            k_rank_col, mat_a.tileSize(GlobalTileIndex(k, k)), serial_comm());
```
msimberg, line 152: Would it not make sense to name this function as well, and the ones further down in the file (apologies if you already planned to do this)?

teonnik: Yes, definitely, I am preparing a separate PR for `triangular_LLN()`. I had to touch it because I made changes to `Pipeline`.

rasolca (commented):
I'm not sure that using global indices makes the code more readable, I personally prefer to use local indices.

Moreover I think we shouldn't neglect the overhead introduced, especially for small tile sizes (for which we are already dealing with a large scheduler overhead) and large mpi grids.

Assuming we have to schedule several sgemm with tile size 128 on ranks with 64 cores, we should be able to schedule at least 64 tasks in the time one of this task execute.
Assuming 2 FMA AVX units a sgemm requires `2*128^3/32` cycles + function call and check overhead + scheduling overhead.
Assuming that the calculation of the rank and the comparison `if (my_rank == rank_of_tile(i, j))` requires 10 cycles and that (for simplicity) the task creation overhead in 1/64 of the sgemm overhead we obtain that we can run less than `2*128^3/32/64/10=204` ranks, otherwise scheduling will become the bottleneck.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
```diff
@@ -87,189 +152,88 @@ void cholesky_L(Matrix<T, Device::CPU>& mat_a) {
...
-        // Avoid useless communications if one-row communicator grid
-        if (row_comm_size > 1) {
-          // Broadcast the panel row-wise
-          auto send_bcast_f = unwrapping([](auto&& tile, auto&& comm_wrapper) {
-            comm::sync::broadcast::send(comm_wrapper().rowCommunicator(), tile);
-          });
-          hpx::dataflow(executor_mpi, std::move(send_bcast_f), mat_a.read(ik_local_idx), serial_comm());
-        }
+    // Iterate over the k-th column
+    std::vector<SizeType> bindices;
```
rasolca, line 195: I think this is not used anywhere.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
```diff
@@ -87,189 +152,88 @@ void cholesky_L(Matrix<T, Device::CPU>& mat_a) {
...
-        hpx::dataflow(trailing_matrix_executor, unwrapping(tile::gemm<T, Device::CPU>), NoTrans,
-                      ConjTrans, -1.0, panel[i_local], col_panel, 1.0,
-                      std::move(mat_a(LocalTileIndex{i_local, j_local})));
+      for (SizeType i = j + 1; i < nrtile; ++i) {
+        // Update the ij-tile using the ik-tile and jk-tile
+        GlobalTileIndex ij_idx(i, j);
+        comm::Index2D ij_rank = distr.rankGlobalTile(ij_idx);
+        if (this_rank == ij_rank) {
+          gemm_trailing_matrix_tile(executor_normal, panel[i], panel[j], mat_a(ij_idx));
+        }
```
rasolca, line 236: Can I suggest to further improve the readability a method called e.g. `herkTrailingMatrix` which performs the two loops (line 212 and line 230) and schedule (full) herk tasks (herks and gemms)?
(This helps reducing code duplication as well, as functions that are introduced can be easily reused in other algorithms).

Similarly we can introduce methods for the panel bcasts (lines 200-228 (BLAS ops excluded)) that can be used in other algorithms (e.g. generalized to standard transformation and the reduction to band need the same communication pattern for trailing matrix her2k)

teonnik: Yes, certainly. I just implemented `herkTrailingMatrix()`, you can have a look.

teonnik:
> I'm not sure that using global indices makes the code more readable, I personally prefer to use local indices.

One reason I prefer global indices is because it avoids loops like this:

```
      for (SizeType i_local = distr.nextLocalTileFromGlobalTile<Coord::Row>(k + 1);
           i_local < localnrtile_rows; ++i_local) {
        auto i = distr.globalTileFromLocalTile<Coord::Row>(i_local);
```

and frequent conversions between local and global indices.

However, the main reason to like global indices is that it makes it easier for me to understand the algorithm. Mathematically, the indices used to describe the algorithm are global. Moreover, if you notice the loops in the local and distributed versions, they are essentially the same, except the distributed has 1)  guards (if statements) that control tile ownership of processes and 2) tile communication logic between processes. I find the similarity between the local and distributed versions of the algorithms quite nice, but I agree this may be subjective. I'd happy to hear what others think.

> Moreover I think we shouldn't neglect the overhead introduced, especially for small tile sizes (for which we are already dealing with a large scheduler overhead) and large mpi grids.

> Assuming we have to schedule several sgemm with tile size 128 on ranks with 64 cores, we should be able to schedule at least 64 tasks in the time one of this task execute.
Assuming 2 FMA AVX units a sgemm requires 2*128^3/32 cycles + function call and check overhead + scheduling overhead.
Assuming that the calculation of the rank and the comparison if (my_rank == rank_of_tile(i, j)) requires 10 cycles and that (for simplicity) the task creation overhead in 1/64 of the sgemm overhead we obtain that we can run less than 2*128^3/32/64/10=204 ranks, otherwise scheduling will become the bottleneck.

I can do some benchmarks on some edge-case scenarios if you are concerned about the extra costs coming from the loops. However, I'd be very surprised if it's noticeable in practice.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
rasolca: I meant to separate the communication from the computation part, as this communication pattern is found elsewhere (e.g. trailing updates with her2k).
But for now it is ok. Reuse of this pattern is not part of this PR.

teonnik: Did you mean just this part (this is from the local but it's the same for the global)?:

```
179     for (SizeType j = k + 1; j < nrtile; ++j) {
180       // Choose queue priority
181       auto trailing_matrix_executor = (j == k + 1) ? executor_hp : executor_normal;
182
183       // Update trailing matrix: diagonal element mat_a(j,j), reading mat_a.read(j,k), using herk (blas operation)
184       herk_trailing_diag_tile(trailing_matrix_executor, mat_a.read(LocalTileIndex{j, k}),
185                               mat_a(LocalTileIndex{j, j}));
186
187       for (SizeType i = j + 1; i < nrtile; ++i) {
188         // Update remaining trailing matrix mat_a(i,j), reading mat_a.read(i,k) and mat_a.read(j,k),
189         // using gemm (blas operation)
190         gemm_trailing_matrix_tile(trailing_matrix_executor, mat_a.read(LocalTileIndex{i, k}),
191                                   mat_a.read(LocalTileIndex{j, k}), mat_a(LocalTileIndex{i, j}));
192       }
193     }
```

include/dlaf/factorization/cholesky/mc/cholesky_L.h
```diff
@@ -12,35 +12,147 @@
...
+      if (j != nrtile - 1)
+        send_tile(ex_mpi, task_chain, Coord::Col, panel[j]);
+    }
+    else if (this_rank.col() == jj_rank.col()) {
+      GlobalTileIndex jk_idx(j, k);
+      if (j != nrtile - 1)
+        panel[j] = recv_tile<T>(ex_mpi, task_chain, Coord::Col, mat_a.tileSize(jk_idx), jj_rank.row());
+    }
+
+    for (SizeType i = j + 1; i < nrtile; ++i) {
```
rasolca, line 142: what about executing this loop only when I own the column?
```
if (this_rank.col() == jj_rank.col()) {
  for (....) {
     ...
  }
}
```

rasolca (commented):
> > I'm not sure that using global indices makes the code more readable, I personally prefer to use local indices.
> 
> One reason I prefer global indices is because it avoids loops like this:
> 
> ```
>       for (SizeType i_local = distr.nextLocalTileFromGlobalTile<Coord::Row>(k + 1);
>            i_local < localnrtile_rows; ++i_local) {
>         auto i = distr.globalTileFromLocalTile<Coord::Row>(i_local);
> ```
> 
> and frequent conversions between local and global indices.
> 
> However, the main reason to like global indices is that it makes it easier to understand the algorithm. The mathematical description of the algorithm uses global indices. Moreover, if you notice the loops in the local and distributed versions, they are essentially the same, except the distributed has 1) guards (if statements) that control tile ownership of processes and 2) tile communication logic between processes. I find the similarity between the local and distributed versions of the algorithms quite nice, but I agree this may be subjective. I'd happy to hear what others think.
> 
I understand your point, and as I stated in my comment is my personal preference and I was hoping for other opinions as well.

> > Moreover I think we shouldn't neglect the overhead introduced, especially for small tile sizes (for which we are already dealing with a large scheduler overhead) and large mpi grids.
> 
> > Assuming we have to schedule several sgemm with tile size 128 on ranks with 64 cores, we should be able to schedule at least 64 tasks in the time one of this task execute.
> > Assuming 2 FMA AVX units a sgemm requires 2_128^3/32 cycles + function call and check overhead + scheduling overhead.
> > Assuming that the calculation of the rank and the comparison if (my_rank == rank_of_tile(i, j)) requires 10 cycles and that (for simplicity) the task creation overhead in 1/64 of the sgemm overhead we obtain that we can run less than 2_128^3/32/64/10=204 ranks, otherwise scheduling will become the bottleneck.
> 
> I can do some benchmarks on some edge-case scenarios if you are concerned about the extra costs coming from the loops. However, I'd be very surprised if it's noticeable in practice.

I would not be concerned of performance if you agree to add the if statement I suggested.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
rasolca: Don't worry. This PR generated some ideas in my mind, I will elaborate them to see if they are feasible and discuss them separately.

teonnik: OK, then, should I revert the commit and leave `herkTrailingMatrix()` the function to you?

Sely85:
> I'm not sure that using global indices makes the code more readable, I personally prefer to use local indices.
I understand both your points of view. It is true that the mathematical description of the algorithm uses global indices. However, I find it easier to understand how it works applied to tiles with the use of local indices.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
```diff
@@ -87,191 +197,61 @@ void cholesky_L(Matrix<T, Device::CPU>& mat_a) {
...
-
-    auto kk_idx = GlobalTileIndex(k, k);
-    auto kk_tile_rank = distr.rankGlobalTile(kk_idx);
-
-    if (this_rank.col() == kk_tile_rank.col()) {
-      hpx::shared_future<ConstTile_t> kk_tile;
-
-      if (this_rank.row() == kk_tile_rank.row()) {
-        LocalTileIndex kk = distr.localTileIndex(kk_idx);
+    std::unordered_map<SizeType, hpx::shared_future<ConstTile_t>> panel;
```
albestro, line 219: What is the benefit of using `std::unordered_map` over `std::vector`?

teonnik: I have to use `std::unordered_map` because I am using global indices which may have large gaps in-between as keys.

albestro: ah yeah, good point! I didn't thought about it. because you changed everything to global indices, I forgot that.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
teonnik: @rasolca Yes, I will do something to the equivalent effect, which will avoid the nesting of the for loop:

```
if (this_rank.col() != jj_rank.col()) 
      continue;
``` 

However, I would like to know whether to drop the last commit (the refactoring of `herkTrailingMatrix()`) before that?

include/dlaf/factorization/cholesky/mc/cholesky_L.h
rasolca: OK.

include/dlaf/factorization/cholesky/mc/cholesky_L.h
teonnik: Is that OK? : 1ce31d4

rasolca: LGTM.

teonnik:
FYI: since the global vs local indices decision proved somewhat controversial, in the branch (future PR) refactoring the LLN triangular solver (https://github.com/eth-cscs/DLA-Future/tree/teo/triang_LLN), I left the indices as they were.

teonnik:
bors try

rasolca (approved):

Sely85 (approved):
LGTM! 
The refactor of functions is really useful to avoid repetitions and improve readability.

PR #294, opened 2020-11-12
Add a general tile `copy()`

teonnik:
- Generalize `copy()`: copy a 2D sub-region from an input tile to an output
  tile. The operation is needed when batching multiple times or
  splitting a tile (separate PRs for these are coming).

### DIFF
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -401,11 +401,5 @@ Size2D<IndexT, Tag> operator+(const Size2D<IndexT, Tag>& lhs, const Size2D<Index
   return Size2D<IndexT, Tag>(lhs.rows() + rhs.rows(), lhs.cols() + rhs.cols());
 }
 
-/// Returns true if @param small is within @param big starting at @param idx
-template <class IndexT, class Tag>
-bool isWithin(Size2D<IndexT, Tag> small, Index2D<IndexT, Tag> idx, Size2D<IndexT, Tag> big) {
-  return idx.row() + small.rows() <= big.rows() && idx.col() + small.cols() <= big.cols();
-}
-
 }
 }
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -401,5 +401,11 @@ Size2D<IndexT, Tag> operator+(const Size2D<IndexT, Tag>& lhs, const Size2D<Index
   return Size2D<IndexT, Tag>(lhs.rows() + rhs.rows(), lhs.cols() + rhs.cols());
 }
 
+/// Returns true if @param small is within @param big starting at @param idx
+template <class IndexT, class Tag>
+bool isWithin(Size2D<IndexT, Tag> small, Index2D<IndexT, Tag> idx, Size2D<IndexT, Tag> big) {
+  return idx.row() + small.rows() <= big.rows() && idx.col() + small.cols() <= big.cols();
+}
+
 }
 }
--- a/include/dlaf/lapack_tile.h
+++ b/include/dlaf/lapack_tile.h
@@ -19,6 +19,7 @@
 #undef I
 #endif
 
+#include "dlaf/matrix/index.h"
 #include "dlaf/matrix/tile.h"
 #include "dlaf/types.h"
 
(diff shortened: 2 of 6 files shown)


teonnik:
bors try

teonnik:
bors try

rasolca (approved):

include/dlaf/common/index2d.h
```diff
@@ -401,5 +401,11 @@ Size2D<IndexT, Tag> operator+(const Size2D<IndexT, Tag>& lhs, const Size2D<Index
   return Size2D<IndexT, Tag>(lhs.rows() + rhs.rows(), lhs.cols() + rhs.cols());
 }
 
+/// Returns true if @param small is within @param big starting at @param idx
+template <class IndexT, class Tag>
+bool isWithin(Size2D<IndexT, Tag> small, Index2D<IndexT, Tag> idx, Size2D<IndexT, Tag> big) {
```
albestro, line 406: Just a minor suggestion. I like free functions more than member functions, and I'm just wondering if moving also `Index2D::isIn()` function wouldn't be a good idea, just to keep them "close" since they are in some way semantically similar.

include/dlaf/common/index2d.h
```diff
@@ -401,5 +401,11 @@ Size2D<IndexT, Tag> operator+(const Size2D<IndexT, Tag>& lhs, const Size2D<Index
...
+  return idx.row() + small.rows() <= big.rows() && idx.col() + small.cols() <= big.cols();
```
albestro, line 407: Just supporting my previous comment, I think it is possible to implement this one using coordinates arithmetic and the previously cited `Index2D::isIn`.

albestro (approved):
I gave a very quick view because I'm interested for another work I'm doing on a similar topic. In fact, I may discuss with you about making this thing even more generic for dealing with a "concept" of column-major memory with a size and ld.

Anyway, LGTM, just a minor comment, I'd like to know what you think about.

Sely85 (approved):
LGTM!

include/dlaf/common/index2d.h
teonnik: I can certainly move it out, the issue is that I will have to touch quite a lot of files as the function is used in many places. Perhaps we can leave that to a separate PR.

teonnik:
bors try

teonnik:
I am little confused by the CI. Before the last commit, I was getting the following error:

```
In file included from /home/teonnik/code/dlaf/test/unit/matrix/test_matrix.cpp:12:
/home/teonnik/code/dlaf/include/dlaf/matrix/copy.h:40:21: error: no matching function for call to 'unwrapping'
      hpx::dataflow(hpx::util::unwrapping(dlaf::copy<T>), source.read(LocalTileIndex(i, j)),
```

The error was due to type deduction for `hpx::util::unwrapping()` which wasn't able to pick-up the right overload for tile `copy()`.

I am using clang@10.0.1. I am interested if someone else could reproduce it? (you have to checkout the commit before the last one. - 0bfc4e3)

bors try

include/dlaf/common/index2d.h
teonnik: I removed `isWithin()`, as you pointed out, I can actually just use `isIn()` with some workarounds. fd08acb

msimberg:
> I am little confused by the CI. Before the last commit, I was getting the following error:
> 
> ```
> In file included from /home/teonnik/code/dlaf/test/unit/matrix/test_matrix.cpp:12:
> /home/teonnik/code/dlaf/include/dlaf/matrix/copy.h:40:21: error: no matching function for call to 'unwrapping'
>       hpx::dataflow(hpx::util::unwrapping(dlaf::copy<T>), source.read(LocalTileIndex(i, j)),
> ```
> 
> The error was due to type deduction for `hpx::util::unwrapping()` which wasn't able to pick-up the right overload for tile `copy()`.
> 
> I am using clang@10.0.1. I am interested if someone else could reproduce it? (you have to checkout the commit before the last one. - [0bfc4e3](https://github.com/eth-cscs/DLA-Future/commit/0bfc4e351eeea7cb6d3abc00f18f46c845d50966) , the last commit fixes it)

This is expected, and I have the same problem over here: https://github.com/eth-cscs/DLA-Future/pull/289/files#r519862729. I wrap it in a lambda instead of using unwrapping, but both approaches work. There's no way to make this work with unwrapping and plain functions. The only other solution I know of is making `copy` a callable object instead, which defers overload resolution to inside unwrapping.

teonnik:
@msimberg I used a function reference, which seems to work as well:

```
void (&cpy)(const matrix::Tile<const T, Device::CPU>&, const matrix::Tile<T, Device::CPU>&) = copy<T>;
```

teonnik:
@albestro 

> I gave a very quick view because I'm interested for another work I'm doing on a similar topic. In fact, I may discuss with you about making this thing even more generic for dealing with a "concept" of column-major memory with a size and ld.

Where `Tile<>` are concerned, I think this is as generic as we can get. If you want to generalize it in a sense of the Data concept, we can do that. Overall, I like the Tile<> interface as tiles are what I find most intuitive to think about and work with. I try to avoid doing much with Buffer and Message except when absolutely needed. If I have to copy some buffers around, I will try to represent them as tiles before doing so.

albestro:
Is it because when unwrapping is defined the copy function, which is a template, is not defined yet? (see https://clang.llvm.org/compatibility.html#dep_lookup)

IMHO the solution by @msimberg of wrapping it into a lambda looks more readable to me.

rasolca (approved):

teonnik:
> IMHO the solution by @msimberg of wrapping it into a lambda looks more readable to me.

@albestro  No love for function references? ;)

The lambda approach is more LOCs but if that's what other prefer as well, I will change it.

rasolca:
> The lambda approach is more LOCs but if that's what others prefer as well, I will change it.
I'm fine with both approaches.

albestro:
I find more readable and clean the lambda, even if it may produce LOCs.

However, just a matter of preferences, absolutely not a request of change. 😁

rasolca:
bors try

rasolca:
bors try

PR #297, opened 2020-11-13
Refactoring tile future management and avoid continuation duplication.

rasolca:
Was part of #154.

Reduce code duplication of tile continuation management.


include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,85 @@
...
+#include <hpx/hpx.hpp>
+#include "dlaf/matrix/tile.h"
+
+namespace dlaf {
+namespace matrix {
+namespace internal {
+
+template <class ReturnTileType, class TileType>
+hpx::future<ReturnTileType> setPromiseTileFuture(hpx::future<TileType>& old_future,
+                                                 hpx::lcos::local::promise<TileType>& p) noexcept {
```
teonnik, line 20: I would accept these by value since they are consumed in the function.

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,85 @@
...
+  return old_future.then(hpx::launch::sync, [p = std::move(p)](hpx::future<TileType>&& fut) mutable {
```
teonnik, line 21: `std::move(old_future).then(...)`

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,85 @@
...
+    // The exception is set outside the catch block since set_exception may
+    // yield. Ending the catch block on a different worker thread than where it
+    // was started may lead to segfaults.
+    p.set_exception(current_exception_ptr);
+    std::rethrow_exception(current_exception_ptr);
+  });
+}
+
+template <class ReturnTileType, class TileType>
+hpx::future<ReturnTileType> getTileFuture(hpx::future<TileType>& tile_future) noexcept {
```
teonnik, line 40: Here also by value.

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,85 @@
...
+  hpx::future<TileType> old_future = std::move(tile_future);
+  hpx::lcos::local::promise<TileType> p;
+  tile_future = p.get_future();
+  return setPromiseTileFuture<ReturnTileType>(old_future, p);
```
teonnik, line 44: `std::move(old_future)` and `std::move(p)` in `setPromiseTileFuture<>(...)`

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,85 @@
...
+  using TileType = Tile<T, device>;
+  using ConstTileType = Tile<const T, device>;
+
+  TileFutureManager() {}
+
+  TileFutureManager(TileType&& tile) : tile_future_(hpx::make_ready_future(std::move(tile))) {}
+
+  hpx::shared_future<ConstTileType> getReadTileSharedFuture() noexcept {
+    if (!tile_shared_future_.valid()) {
+      tile_shared_future_ = std::move(getTileFuture<ConstTileType>(tile_future_));
```
teonnik, line 59: The returned `hpx::future` is an rvalue already, no need for `std::move()`, but the argument can be moved in `std::move(tile_future_)` (and has to be if the function is modified to take the parameter by value)

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,85 @@
...
+
+template <class T, Device device>
+class TileFutureManager {
+public:
+  using TileType = Tile<T, device>;
+  using ConstTileType = Tile<const T, device>;
+
+  TileFutureManager() {}
+
+  TileFutureManager(TileType&& tile) : tile_future_(hpx::make_ready_future(std::move(tile))) {}
```
teonnik, line 55: Here, can also be passed by value.

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,85 @@
...
+  hpx::shared_future<ConstTileType> getReadTileSharedFuture() noexcept {
+    if (!tile_shared_future_.valid()) {
+      tile_shared_future_ = std::move(getTileFuture<ConstTileType>(tile_future_));
+    }
+    return tile_shared_future_;
+  }
+
+  hpx::future<TileType> getRWTileFuture() noexcept {
+    tile_shared_future_ = {};
+    return getTileFuture<TileType>(tile_future_);
```
teonnik, line 66: Here, `tile_future_` can be moved in.

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,85 @@
...
+    tile_future_.get();
+  }
+
+protected:
+  // The future of the tile with no promise set.
+  hpx::future<TileType> tile_future_;
+
+  // If valid, a copy of the shared future of the tile,
+  // which has the promise set to the promise for tile_future_.
+  hpx::shared_future<ConstTileType> tile_shared_future_;
```
msimberg, line 80: A variant would in principle be appropriate here. We don't however have a variant in HPX and I doubt you want to go up to C++17?

include/dlaf/matrix/internal/tile_future_manager.h
msimberg, line 20: Just wanted to note that `future::then` is not `noexcept`, but if this is correctly set up the `noexcept` is of course warranted.

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -0,0 +1,85 @@
...
+
+  hpx::future<TileType> getRWTileFuture() noexcept {
+    tile_shared_future_ = {};
+    return getTileFuture<TileType>(tile_future_);
+  }
+
+  void clearSync() {
+    tile_shared_future_ = {};
+    tile_future_.get();
+  }
```
msimberg, line 72: The name of this function is 100% clear to me. Is it guaranteed that the non-shared future is always the active one here? Is an equivalent for the shared future required? Would an assert be appropriate if there are indeed preconditions?

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: Moving to C++17 is still an open question, but currently we are keeping the safest path.

Anyway, can you please explain what's wrong with is_valid?

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: I suppose that you mean "not 100% clear to me"...

The function clear the content of the object in a synchronous way.
This is needed to guarantee that the matrix object is not destroyed (and therefore deallocates memory) before the work on the tiles is completed.

By design, the only reason why tile_future_ wouldn't be active here is that clearSync has been called before. However this applies to all methods of this object.

I will add some doc about this.

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: I don't understand. Why should I move old_future to call its member?

On the other hand if you mean `std::move(old_future.then(...))` `.then` already returns an rvalue.

include/dlaf/matrix/internal/tile_future_manager.h
msimberg: Nothing wrong with it. I just think variant encodes the use here quite well. You can have exactly one or the other active at a time, never both. There is a small memory benefit as well. All in all I doubt that it's going to make much of a difference though. My comment was mainly about expressing intent better with variant.

include/dlaf/matrix/internal/tile_future_manager.h
teonnik: No, I meant what I wrote but it's OK either way. If you accept the parameter by value, by calling std::move, the compiler may produce better codegen as that would indicate that the data stored in the variable is not used later in the function. The function is small, so the compiler may already see that.

include/dlaf/matrix/internal/tile_future_manager.h
msimberg: Yes, that's indeed what I meant, sorry... I'll rephrase: is it not possible that `tile_shared_future_` is the active future when the `Matrix` destructor is called, not `tile_future_`? I.e. if the last access to a tile is read-only, `tile_future_` is moved into `tile_shared_future_`, and `get` on the now invalid `tile_future_` would fail. What am I missing? Note: I'm only asking because I may simply not have the full picture, and this is in fact impossible due to other constraints.

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: Sorry I interpreted your comment in the wrong way. The future is always valid while the shared future only for read operations. (See comment below for more info).

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: Please note that `tile_future_` is always active and valid, even when a shared_future is available. 

Please see [misc/synchronization.md](https://github.com/eth-cscs/DLA-Future/blob/master/misc/synchronization.md#matrix-mechanism) especially figures 1 and 2 to understand how the mechanism works.

msimberg: Ah, I see now. Thanks! In that case the comment about the variant indeed doesn't make sense.

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: No, it will not work as tile_future is an in/out argument.

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: No, as tile_future_ is an internal state that has to be updated.

teonnik (approved):
LGTM, with some minor unresolved comments ^^.

Sely85 (approved):
Looks good to me.

teonnik (commented):
I get a build error:

```
/home/teonnik/code/dlaf/test/unit/matrix/test_matrix_local.cpp:11:10: fatal error: 'dlaf/matrix.h' file not found
#include "dlaf/matrix.h"
         ^~~~~~~~~~~~~~~
```

rasolca:
Sorry, Some files remained uncommitted after the merge with master.

bors try

rasolca:
bors try

albestro (commented):
Just one minor comment. I see that you used double slash for the documentation. I'm not really sure it will be parsed as documentation by doxygen.

rasolca:
@albestro I didn't use doxygen by purpose, to avoid to clutter the doc with internal details.

teonnik (approved):

rasolca:
bors try

rasolca:
- Reduce code duplication of tile continuation management.
- Fixed header position
- Improved some doc

PR #298, opened 2020-11-13
algorithm: compute T factor (LARFT)

albestro:
This PR aims at implementing the equivalent of the lapack routine LARFT.

For the moment is completely work in progress.


albestro (commented):
@msimberg @rasolca @teonnik 

I tried to fit `computeTFactor` function inside this brand new structure and I'd like to share with you my concerns to understand how we want to address them. It's not a big deal, but I think it is worth thinking about it. I'm writing here and not in a new issue for the moment, because it is easier to use this as example.

We have:
- in `include/dlaf/factorization/qr/api.h` the `struct QR`
- that we specialize in `include/dlaf/factorization/qr/mc.h` by implementing the QR factorization algorithm
- the implementation requires a specialization of the `struct QR<MC, CPU>` and listing all its static members.

My first try for the internal function `computeTFactor` is:
- create a `include/dlaf/factorization/qr/t_factor_mc.h` file with the implementation
- but it also needs a specification of the same class `QR<MC,CPU>`, that is defined already in `include/dlaf/factorization/qr/mc.h`.

And it can't be done "incrementally", so the solutions I see are:
- we use a different struct for each algorithm (e.g. `QR`, `QRTFactor`);
- we centralize the definition of all the functions related to QR for the same specialization `<backend, device>`

What do you think?

teonnik:
I am not familiar with the T factor algorithm. I was looking for some resources: links, citations in the documentation of the code to learn more about it. Would you mind adding some? 

>  AFAIK you cannot add members "incrementally" to the same specialization

Yes

> we use a different struct for each algorithm (e.g. QR, QRTFactor);

I don't see a problem with using a separate struct. However, I don't know how this T-factor algorithm relates to QR, so I don't know if it makes sense to bundle them together.

msimberg:
> And the problem is: AFAIK you cannot add members "incrementally" to the same specialization, so the solutions I see are:
> 
>     * we use a different struct for each algorithm (e.g. `QR`, `QRTFactor`);
> 
>     * we centralize the definition of all the functions related to QR for the same specialization `<backend, device>`
> 
> 
> What do you think?

I'm absolutely in favor of separate structs. Lumping them together brings little benefit and mainly brings problems.

rasolca:
@albestro I would prefer a separate struct too.

@teonnik the T factor is needed to build the WY representation of the orthogonal matrix represented by householder vectors.
It has been introduced for QR, but it can be applied to reduction to band and multiplication of Q (householder form) with a general matrix. In lapack it is implemented as `xLARFT`

@albestro You can use this paper to cite it:
Schreiber, Robert & VanLoan, Charles. (1989). A Storage-Efficient WY Representation for Products of Householder Transformations. SIAM Journal on Scientific and Statistical Computing. 10. 10.1137/0910005.

albestro:
The algorithm structure is there. Even if it needs some additional refinements, it would be useful a quick round of review.

In particular, I'll start pointing out a few things:
- [ ] T is reset to zero at the beginning of the algorithm, it would be nice to embed it inside a bigger task
- [ ] Currently, all ranks in the same column allocate the non-distributed matrix T, which is needed as workspace during computations, but the result is reduced just on a single rank. The T factor is used together with V, which is distributed over the column, so I'm thinking about `all-reduce`-ing the result on all ranks.
- [ ] Feedbacks about the test and test cases are welcome
- [ ] I need to check if the algorithm assume/needs additional requirements (DLAF_ASSERT)

Looking forward for your comments!

@msimberg @rasolca @teonnik @Sely85

rasolca:
>     * [ ]  T is reset to zero at the beginning of the algorithm, it would be nice to embed it inside a bigger task
 
As the matrix has a single tile and everything is set to 0, I would use `laset`.

>     * [ ]  Currently, all ranks in the same column allocate the non-distributed matrix T, which is needed as workspace during computations, but the result is reduced just on a single rank. The T factor is used together with V, which is distributed over the column, so I'm thinking about `all-reduce`-ing the result on all ranks.

I agree. I don't see any reason to reduce here and broadcast later.

>     * [ ]  Feedbacks about the test and test cases are welcome

Tests looks ok .

include/dlaf/factorization/qr/t_factor_mc.h
```diff
@@ -0,0 +1,179 @@
...
+    return;
+
+  const LocalTileIndex v_start_loc{
+      dist.template nextLocalTileFromGlobalTile<Coord::Row>(v_start.row()),
+      dist.template nextLocalTileFromGlobalTile<Coord::Col>(v_start.col()),
+  };
+  const LocalTileIndex v_end_loc{dist.localNrTiles().rows(), v_start_loc.col() + 1};
+
+  // TODO it would be better to embed this reset inside a bigger task
+  matrix::util::set(t, [](auto&&) { return 0; });
```
albestro, line 75: I can absolutely use `laset` and I'll do, but the main point of my comment was: is it ok to have a task that just resets or do we want to try embed this reset in the first computing task in some way? @rasolca

Sely85:
Looks good! 

> * [ ]  Currently, all ranks in the same column allocate the non-distributed matrix T, which is needed as workspace during computations, but the result is reduced just on a single rank. The T factor is used together with V, which is distributed over the column, so I'm thinking about `all-reduce`-ing the result on all ranks.

I agree as well. "Reducing" the result on all ranks here looks clearer to me. 


> * [ ]  Feedbacks about the test and test cases are welcome

Tests cover both cases: all reflectors and only some reflectors. I think that's ok.

include/dlaf/factorization/qr/t_factor_mc.h
albestro: as discussed with @rasolca offline, `laset` of a tile is enough work for a task. we leave it as it is now.

albestro:
bors try

albestro:
There are a couple of TODO for secondary things that I left there.

The main obstacle I found is that I was not able to think about a quick and clean way to access "properties" of an object wrapped in a Pipeline/future. This lead to:
- I cannot ask a `Pipeline<CommunicatorGrid>` about the size of the grid (e.g. for assertions and conditionals)
- a Matrix has to be passed for `t` even if just a `future<Tile>` would be enough, because I cannot get tile sizes from the future (I would have to extract them the first time I wait on it, e.g. on `laset`) 

I don't like the workaround, so I preferred to go straight using the Matrix and just leaving there a couple of TODO. Let me know your feedback about this, if you prefer we can do it differently.

albestro:
The test is failing due to a wrongly set error margin for the checks.

In the test, there are two places where the comparison between results is performed:
1. test orthogonality, where it is performed a `M . M* == I`
2. test for equality between the two H transformations, respectively computed sequentially or in block with the T factor

For the former, being a `gemm`  "m by n by m", it may be enough allowing an error up to `2 * n *  error` (`n` muls and `n` adds), perhaps using just the absolute margin and not the relative one.

For the latter, it may be harder to find a "precise" boundary for the error, and it may be reasonable to go for a "close-enough" approximation.

Looking forward for your feedback.

include/dlaf/factorization/qr/t_factor_mc.h
```diff
@@ -0,0 +1,191 @@
...
+              va_size.rows(), va_size.cols(),
+              -tau,
+              tile_v.ptr(va_start), tile_v.ld(),
+              tile_v.ptr(vb_start), 1,
+              1, tile_t.ptr(t_start), 1);
+          // clang-format on
+        }
+      });
+
+      hpx::dataflow(gemv_func, v.read(v_i_loc), taus[j], t(t_idx));
```
rasolca, line 156: I see a problem here: All the gemv are serialized as all need to write to the same t.

1st solution (easy): keep the gemv serialized, but exchange the j and v_i_loc the loop and put the j loop in the task (such that the V tile is read only once in cache).
2nd solution: 1st solution + have multiple T and reduce them before all_reduce.

include/dlaf/factorization/qr/t_factor_mc.h
```diff
@@ -0,0 +1,191 @@
...
+
+  // T(0:j, j) = T(0:j, 0:j) . -tau(j) . V(j:, 0:j)* . V(j:, j)
+  for (SizeType j = 0; j < k; ++j) {
+    // this is the x0 element of the reflector j
+    const TileElementIndex x0{j, j};
+
+    const TileElementIndex t_start{0, x0.col()};
+    const TileElementSize t_size{x0.row(), 1};
+
+    // 2A First step GEMV
```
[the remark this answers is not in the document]
albestro: I improved a bit the documentation in general for the algorithm, by adding some notes.

Please, let me know if you think it is enough.

include/dlaf/factorization/qr/t_factor_mc.h
albestro: Good point. I followed your first suggestion for the moment, for the second I left there a comment so we can improve it in the future.

test/unit/factorization/mc/test_compute_t_factor.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+      MatrixLocal<TypeParam> h_expected({m, m}, block_size);
+      set(h_expected, preset_eye<TypeParam>);
+
+      for (auto j = 0; j < k; ++j) {
+        const SizeType reflector_size = m - j;
+
+        const TypeParam* data_ptr = v.ptr({j, j});
+        const auto norm = blas::nrm2(reflector_size, data_ptr, 1);
+        const TypeParam tau = 2 / std::pow(norm, 2);
```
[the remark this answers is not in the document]
albestro: After an offline discussion with @rasolca, it has been decided that it may be better to leave it as it is with `nrm2+pow` instead of `dot`.

The main point could be the accuracy of the results, since `norm2` internally ensures that an overflow does not occur (if I get it correctly).

albestro:
There is still an open point about error thresholds.

Please, let me know if you have any suggestions.

rasolca:
> The test is failing due to a wrongly set error margin for the checks.
> 
> In the test, there are two places where the comparison between results is performed:
> 
>     1. test orthogonality, where it is performed a `M . M* == I`
> 
>     2. test for equality between the two H transformations, respectively computed sequentially or in block with the T factor
> 
> 
> For the former, being a `gemm` "m by n by m", it may be enough allowing an error up to `2 * n * error` (`n` muls and `n` adds), perhaps using just the absolute margin and not the relative one.
> 
> For the latter, it may be harder to find a "precise" boundary for the error, and it may be reasonable to go for a "close-enough" approximation.
> 
> Looking forward for your feedback.

You are just considering the error needed to check the unitary matrix and not the error constructing it.
To construct the orthogonal/unitary matrix each element needs 4 * n * nb arithmetic operations (I-VTV*).

For the first test you are testing an identity. So you can't check the zeros with relative error.
Anyway as `abs` of each element of the identity or unitary matrices is smaller than 1 you can set the relative error to 0 without any problem.

include/dlaf/factorization/qr/t_factor_mc.h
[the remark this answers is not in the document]
rasolca: Looks good.

Sely85 (commented):
Looks good. Documentation is really clear as well.

albestro:
bors try

albestro:
bors try

rasolca (approved):

PR #299, opened 2020-11-16
Only build triang miniapp if tests are enabled

teonnik:
This prevents surprises when building the package with spack. For example, this will fail:

```
spack install dla-future@develop %clang build_type=Debug ^mpich -fortran ^blaspp~cuda
```

unless we add `--test=all` which enables tests.

With the fix in this PR, we can build with spack without the `--test=all` flag by skipping the triangular solver miniapp.


msimberg:
Would it make sense to have `DLAF_WITH_MINIAPPS` guard the whole minapps directory?

teonnik:
> Would it make sense to have DLAF_WITH_MINIAPPS guard the whole minapps directory?

There is a `DLAF_BUILD_MINIAPPS` variable which does that. There was talk about renaming the variable to `DLAF_WITH_MINIAPPS` for consistency in the past, I can do it in this PR too.

 > Edit: I see the change is not 100% related to what I initially thought, but the question still stands. The DLAF_test target could also be added even if tests are disabled (possibly moving it to a more independent location, not to tie it too much with the tests).

Yes, let's move the parts of DLAF_test that we need within the miniapps into the DLAF library and remove the DLAF_test dependency.

The proposed fix in this PR is a temporary work-around. I can move this PR in a direction of a more permanent solution by moving parts of DLAF_test into DLAF if that's what we want.

rasolca (approved):
Im general miniapps shouldn't rely on DLAF_test, since its functions are not designed to scale. As the TODO mentions, the dependency should be removed.

As for now, I agree with the temporary fix.

PR #301, opened 2020-11-17
Introduce more variables to find CUDA with

teonnik:
- Do not rely on `enable_language(CUDA)` which was removed
- Introduce/Document alternative commonly used environment/CMake variables for finding CUDA libraries

### DIFF
--- a/cmake/FindCUDALIBS.cmake
+++ b/cmake/FindCUDALIBS.cmake
@@ -10,9 +10,11 @@
 
 
 #
-# To find CUDA:
-#   1. Use `enable_language(CUDA)`
-#   2. Set `CMAKE_CUDA_COMPILER` if cuda in custom directory
+# To find CUDA, do one of the following:
+#   - Set `CUDALIBS_ROOT` as an environment variable or a CMake variable
+#   - Set the environment variable `CUDA_HOME`
+#   - Use `enable_language(CUDA)` and set `CMAKE_CUDA_COMPILER` if cuda in
+#      custom directory
 #
 # Imported Targets:
 #   dlaf::cudart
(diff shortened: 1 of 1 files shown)


haampie (approved):
LGTM, I'm just using `-DCUDALIBS_ROOT=/usr/local/cuda/targets/x86_64-linux` for the ubuntu:18.04 docker image now.

cmake/FindCUDALIBS.cmake
```diff
@@ -21,13 +23,21 @@
...
 find_path(CUDA_CUBLAS_INCLUDE cublas_v2.h
-          PATHS ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
+          PATHS ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}
+                $ENV{CUDA_HOME}
+          PATH_SUFFIXES include)
 find_library(CUDA_CUDART_LIB cudart
-             PATHS ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
+             PATHS ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES}
+                   $ENV{CUDA_HOME}
+             PATH_SUFFIXES lib64)
```
albestro, line 36: Is it always `lib64`?

albestro (approved):
LGTM

cmake/FindCUDALIBS.cmake
msimberg: That's in addition to `lib`, I think. @teonnik please correct me if I'm wrong. There's also this https://cmake.org/cmake/help/v3.12/prop_gbl/FIND_LIBRARY_USE_LIB64_PATHS.html#prop_gbl:FIND_LIBRARY_USE_LIB64_PATHS which I think would work as an alternative.

teonnik: >  That's in addition to lib

Yes, I think so too. The docs, point 1. seems to agree : https://cmake.org/cmake/help/v3.19/command/find_library.html#command:find_library

> here's also this https://cmake.org/cmake/help/v3.12/prop_gbl/FIND_LIBRARY_USE_LIB64_PATHS.html#prop_gbl:FIND_LIBRARY_USE_LIB64_PATHS which I think would work as an alternative.

I wasn't aware of that one. :)

PR #302, opened 2020-11-17
Refactor traingular LLN

teonnik:
- Share send/recv bcast kernels with Cholesky


rasolca:
This PR contains parts of #293. Please rebase or merge.

teonnik:
bors try

teonnik:
bors try

include/dlaf/communication/sync/broadcast.h
```diff
@@ -48,5 +52,42 @@ void receive_from(const int broadcaster_rank, Communicator& communicator, DataOu
...
+  using MemView_t = memory::MemoryView<T, Device::CPU>;
+  using Tile_t = matrix::Tile<T, Device::CPU>;
+
+  auto recv_bcast_f = hpx::util::annotated_function(
+      [rank, tile_size, rc_comm](hpx::future<PromiseComm_t> fpcomm) -> ConstTile_t {
+        PromiseComm_t pcomm = fpcomm.get();
+        MemView_t mem_view(tile_size.linear_size());
+        Tile_t tile(tile_size, std::move(mem_view), tile_size.rows());
+        comm::sync::broadcast::receive_from(rank, pcomm.ref().subCommunicator(rc_comm), tile);
+        return std::move(tile);
```
msimberg, line 87: I think the move here is unnecessary and potentially harmful, at least on the compilers you are likely to use. It's not compulsory to do RVO here (specifically NRVO), meaning it might be required in some cases, but both GCC and clang do it here. (See NRVO here: https://en.cppreference.com/w/cpp/language/copy_elision).

teonnik: We need it there otherwise we get a compiler error:

```
/home/teonnik/code/dlaf/include/dlaf/communication/sync/broadcast.h:88:16: error: call to deleted constructor of 'ConstTile_t' (aka 'Tile<const std::complex<double>, Device::CPU>')
        return tile;
```

The reason is that there is actually a conversion from Tile_t to ConstTile_t. The latter expects an rvalue of the former.

msimberg: Thanks for the explanation! I missed the different types.

rasolca (approved):

albestro (approved):

PR #303, opened 2020-11-19
Make extensive use of DLAF_MPI_CALL

albestro:
Wrap all the MPI calls with the custom DLAF_MPI_CALL.

Other unrelated changes:
- DLAF_test is now an INTERFACE library
- add some missing assert messages (I'm incrementally adding the one I found during debugging);

Another proposal for which I'd like to ask you is about using the custom error handler for all tests. In this way if something bad happens during our tests, we have some more information (and a way to attach the debugger) compared to the default fatal from MPI.


teonnik (approved):

albestro:
@msimberg @rasolca @teonnik 

What do you think about setting our DLAF handler in our custom `gests_main`, where MPI is involved, with 
`MPI_Comm_set_errhandler(MPI_COMM_WORLD, MPI_ERRORS_RETURN)`?

IMHO, this might help with test errors and debugging them, offering an hook for a breakpoint otherwise impossible because of the default management with a FATAL.

rasolca (approved):
LGTM.

Regarding setting the return error handler, I don't see any issue.

albestro:
bors try

teonnik:
> What do you think about setting our DLAF handler in our custom gests_main, where MPI is involved, with
MPI_Comm_set_errhandler(MPI_COMM_WORLD, MPI_ERRORS_RETURN)?

Sounds good!

PR #304, opened 2020-11-23
Allow empty range2d

albestro:
I don't know if it was intended or not, but I think that empty ranges are quite useful.

Currently, my use case is a distributed matrix where a rank may or may not have local elements, so it would be nice that, once local range boundaries are computed, the `for (auto local : iterate_range2d)` would just skip without any additional check.

For the moment, I would propose just to remove strictness in the condition, leaving any major implementation changes to the iterator design for future improvements (just if needed).


rasolca (approved):
lgtm

teonnik (approved):

albestro:
bors try

PR #307, opened 2020-11-25
Add an option to disable thread pinning for tests

teonnik:


albestro (approved):

rasolca (commented):
LGTM. Just a small remark that @msimberg can address:
Is it safe to add both `--hpx:use-process-mask` and `--hpx:bind=none` or is better to use only the second?

test/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ set(DLAF_MPI_PRESET "plain-mpi" CACHE STRING "Select a preset to use")
 set_property(CACHE DLAF_MPI_PRESET PROPERTY STRINGS ${DLAF_PRESET_OPTIONS})
 option(DLAF_CI_RUNNER_USES_MPIRUN "Remove mpiexec command for tests executed by ctest. This option is to be used if the CI runner executes the tests with <mpiexec + options> ctest -L RANK_<spawned MPI ranks>" OFF)
 
+# On some machines, tests using multiple ranks + oversubscribing run
+# significantly faster when threads are not pinned.
+option(DLAF_TEST_DISABLE_THREAD_BINDING "If ON disables HPX thread binding." OFF)
```
msimberg, line 18: `DLAF_TEST_ENABLE_THREAD_BINDING` and default to `ON`? Not a big deal, I just prefer not having to do double negation in my head :P

msimberg:
> LGTM. Just a small remark that @msimberg can address:
> Is it safe to add both `--hpx:use-process-mask` and `--hpx:bind=none` or is better to use only the second?

It works in principle, and it should be safe to keep both on. `--hpx:use-process-mask` also limits the default number of threads, so there are situations where having one or both actually behaves differently.

test/CMakeLists.txt
teonnik: Sounds good! I just changed it to `DLAF_TEST_THREAD_BINDING_ENABLED`.

teonnik:
I did tests with both flags: `--hpx:use-process-mask` and `--hpx:bind=none` included and with just `--hpx:bind=none`. There was not much of a performance difference on my laptop.

msimberg:
> I did tests with both flags: `--hpx:use-process-mask` and `--hpx:bind=none` included and with just `--hpx:bind=none`. There was not much of a performance difference on my laptop.

Ok, good, that's expected (I assume you don't have a mask on your machine, which is why you want to disable pinning). The case I was thinking about where it would make a difference is if there actually is a mask which is not the full machine. In that case the number of threads would be limited, but the threads would still be unpinned.

msimberg (approved):

rasolca (approved):

PR #308, opened 2020-11-27
Add a miniapps option to the spack recipe

teonnik:


spack/packages/dla-future/package.py
```diff
@@ -53,5 +56,8 @@ def cmake_args(self):
 
         # TESTs
         args.append(self.define('DLAF_WITH_TEST', self.run_tests))
+        
+        # MINIAPPS 
+        args.append(self.define('DLAF_BUILD_MINIAPPS', 'miniapps'))
```
albestro, line 61: Shouldn't this be [self.define_from_variant](https://spack.readthedocs.io/en/latest/spack.build_systems.html#spack.build_systems.cmake.CMakePackage.define_from_variant)?

albestro (changes requested):

spack/packages/dla-future/package.py
teonnik: Yes, good catch!

albestro (approved):
LGTM

teonnik:
I've recently started using a code formatter for python called "Black". The formatter has very few options, so it pretty much does the same for everyone (unlike clang-format which has a lot of options). Do you want me to run it on the file?

albestro:
In general it is a good really idea. For this specific case I think you should instead run [*flake8*](https://spack.readthedocs.io/en/latest/contribution_guide.html#flake8-tests), the one that spack uses (and they require this formatting in case you want to add the package to the official repo).

teonnik:
I think it should be fine because Black is a PEP 8 compliant formatter.

albestro:
Good point. I don't know about their "compatibility"; if you know that PEP8 is strictly required by spack and that Black strictly enforces it, IMHO it is a good idea doing it. 😉

teonnik:
My impression is that most python projects try to be PEP8 compliant. PEP8 doesn't specify everything about how the code should be formatted, that is why different formatters, such as Black and autopep8, have differences in the final formatting while still being PEP8 compliant. The flake8 tool just enforces that PEP8 is followed, it doesn't do formatting.

The reason I bring this up is that if I start formatting with Black and then someone else decides that they like autopep8 better, and starts using it, we have a problem the formatting styles will clash. So, either we all use it or we don't use it. :D

albestro:
It seems that [a spack maintainer is already using Black](https://github.com/spack/spack/pull/19610#issuecomment-719576262), so, IMHO, it may be a good and solid starting point. 😉

Sely85 (approved):
LGTM!

PR #309, opened 2020-11-30
Panel and common communication patterns

albestro:
This PR tries to address a common need in writing algorithms about workspaces, and this is the result of a discussion we had with @rasolca a couple of months ago (I developed this feasibility prototype during for reduction to band algorithm).

In writing algorithms, it is common using additional panel workspaces for storing partial results or operands that are shared among multiple ranks for usage in local computations. The workspace represents a set of tiles locally stored, so it may be thought as a Matrix, non-distributed, with few additional features.

The main feature is that each tile in the workspace can be either internally allocated or mapped to an external one (e.g. from another Matrix). This is useful both for using local available tiles without copying their content and for coordinate mapping, (e.g. making a tile reachable at another coordinate, such as in panel transposition).

This feature deserves more detailed explanations in order to fully understand its utility. I will comment with a couple of examples, which are the main use-cases I addressed writing algorithms and that may help understanding.

The idea and the implementation is not fixed in stone and can be surely improved.
Waiting for your feedback!


albestro:
### Working with a column panel
Let's consider a simple situation, where a matrix is distributed over a (1x2) grid of ranks, and just the first rank has all the stuff for computing the panel. The next computation is the multiplication `M . P`. The situation looks more or less like this:

```
note: `--` means that the tile is not locally stored

rank(0, 0)
|P0|	|M0 --|
|P1|	|M1 --|

rank(0, 1)
|--|	|-- M0|
|--|	|-- M1|
```

The rank(0,0) that owns the panel has everything it needs to compute the partial result, while the rank(1,1) before proceeding with the computations must receive the panel.

Each matrix row deals with the corresponding row in the panel, but there is just one rank that owns it. The owner will then communicate row-wise the tile, while receiving ranks will store it inside the panel at the correct position, using the memory allocated by the panel itself.


In addition to this, let's consider a slightly different situation. Rank(0,0) computes the panel in-place in M.

```
note: `&` means that the tile points to the externally allocated tile specified

rank(0, 0)
|P0=&M0|	|M0 --|
|P1=&M1|	|M1 --|

rank(0, 1)
|P0|		|-- M0|
|P1|		|-- M1|
```

In this way, the workspace helps unifying the code over ranks, because it hides all the machinery to access the needed tile from the different sources.

albestro:
### Working with transposed panel

In this example we have a matrix distributed over a (2,2) grid of ranks, where just the first column of ranks compute the panel values. We want to do a computation like `transpose(P) . M`.

```
rank(0,0)
	|-- --|

|P0|	|M0 --|
|--|	|-- --|

rank(1,0)
	|-- --|

|--|	|-- --|
|P1|	|M1 --|

rank(0,1)
	|-- --|

|--|	|-- M2|
|--|	|-- --|

rank(1,1)
	|-- --|

|--|	|-- --|
|--|	|-- M3|
```

Given the previous initial setup, the next communication are going to happen.

1. Following the communication pattern described in the previous example, each rank communicate row-wise the panel (the tile each one owns respectively).
```
rank(0,0)
	|-- --|

|P0|	|M0 --|
|--|	|-- --|

rank(1,0)
	|-- --|

|--|	|-- --|
|P1|	|M1 --|

rank(0,1)
	|-- --|

|P0|	|-- M2|
|--|	|-- --|

rank(1,1)
	|-- --|

|--|	|-- --|
|P1|	|-- M3|
```

2. Now it is the turn of the transposed row panel: because of the tranposition, we need to do coordinates mapping with transposed ones (the column has to be mapped with the corresponding row). Thanks to the previous step, each column of ranks have access to all rows of the panel, so by communicating column-wise, there will always be a rank that owns the tile that maps to the transposed coordinate of each column. Let's this in two steps:

  a. Ranks already owning the wanted transposed tile, just points to the related one in the column panel. In this specific setup, all but rank(0,1) already own what they need.
```
rank(0,0)
	|T0=&P0 --|

|P0|	|M0 --|
|--|	|-- --|

rank(1,0)
	|-- T1=&P1|

|--|	|-- --|
|P1|	|M1 --|

rank(0,1)
	|-- --|

|P0|	|-- M2|
|--|	|-- --|

rank(1,1)
	|-- T1=&P1|

|--|	|-- --|
|P1|	|-- M3|
```

In the last step rank(0,1) receives from rank(1,1) column-wise the required tile
```
rank(0,0)
	|T0=&P0 --|

|P0|	|M0 --|
|--|	|-- --|

rank(1,0)
	|-- T1=&P1|

|--|	|-- --|
|P1|	|M1 --|

rank(0,1)
	|-- T1|

|P0|	|-- M2|
|--|	|-- --|

rank(1,1)
	|-- T1=&P1|

|--|	|-- --|
|P1|	|-- M3|
```

At this point everyone has access to the tile it needs locally to do the wanted calculation.

albestro:
The design of the proposal is changed a lot (hopefully in better) compared to the one it was at the beginning of the PR.
The implementation is still a bit cluttered and far from being perfect, but the main idea is there.

The `Workspace` (any name suggestion is welcome) is a non-distributed panel of tiles, i.e. a row or a column of tiles, in a close relationship with a distributed `Matrix`. In particular, it allocates a panel of tiles which can be used as support for the part of the distributed matrix stored locally.
From the column workspace on the left, it can be seen that:
- just locally stored tiles of the matrix, have a support tile
- for each tile in the row, there is a single support tile

Let's see an example of a Matrix (4,2) distributed over a (1x2) grid (red and blue ranks). On the left of the matrix we can see the size of column workspaces, respectively for the red and blue ranks, while at the top we can see row workspaces.
image

The current design decision is to have Workspace access via `LocalTileIndex` of the matrix (i.e. the projection of the matrix local index on the first col/row). This allows to easily manipulate coordinates in algorithms easing the access to the related support tile for a specific tile in the matrix.

The previous design decision is quite helpful when the workspace is allocated for a submatrix, e.g. when we deal with trailing matrix operations. In fact, it is possible to specify an offset from the top-left corner, and the workspace will manage internally the coordinate mapping.
In the example below the same matrix in which workspaces are allocated for working just with the sub-matrix starting at offset (1,1).

image

As said in the previous comment illustrating the general concept, memory for all the tiles in the workspace is allocated but it is possible to map a single tile to an external tile (e.g. from another matrix).

In the next comments I will provide additional information, restrictions and design aspects, so we can discuss them.

albestro:
The main aspects I'd like to discuss are:
1. How to specify if it is a column or row workspace. Should it be part of the type or not?
2. Design of the `share_panel` function (at the moment it's not so nice, it just works)
3. Workspace expose just part of the Matrix API and it does not expose its real distribution because of the mapping coordinate feature, so I made it iterable.
4. Workspace uses the same Matrix mechanism to provide const-correctness. However, it is not allowed to instantiate a `Workspace<const>` (I don't find a reason to have a support workspace const) and it is just used for allowing const-correctness passing it as a parameter (by-reference).

These are just the main points, I will try to give more details and behind the scene aspects in the next comments.

albestro:
### Should be Col/Row part of the type or not?

As a starting point I used the `enum Coord` already available in DLAF, but it is not fixed in stones, it may be easier to use an empty class exploiting function overloading mechanism to select the correct implementation (instead of using conditionals or `enable_if`).

However, IMHO, the main decision about this point is _**`Workspace<col/row, T, device>` vs `Workspace<T, device>`**_

Having the panel geometry (row/col) in the type (first solution), allows to easily specify if we want to deal with a column or row panel, which I find useful and informative when passing workspaces as arguments. On the other side, it makes col and row panels two distinct types.

Personally, I'm for the first solution which embeds the geometry in the type, because IMHO it makes the code more readable and for the moment I don't see the utility of making row/col panel the same type.

albestro:
### 3. Workspace does not expose the Distribution

Workspace privately inherits from Matrix, but it does not expose the distribution because of the coordinate mapping feature. In fact, the distribution would allow to manipulate the Workspace as we manipulate a Matrix, but the access operators `read` and `operator()` have been overridden in order to deal with the coordinate mapping. This is extremely useful when the Workspace origin is offset wrt the Matrix origin.

For this reason, I opted for not exposing the full Matrix API (e.g. it does not return its size, it does not allow to access tiles through Global indices), but at the same time I made it iterable allowing to iterate over all its tiles, being them internal or external.

albestro:
### 4. Const-correctness + R/W tiles

Workspace has reference semantic as Matrix, and it uses the same implementation mechanism enabling 
```cpp
Workspace<const T>& const_ws = Workspace<T>
```

However, the `const` version of the Workspace cannot be instantiated by the user, since I don't see any reason at the moment.

Another important aspect is about read-write permission of tiles in both const and non-const Workspaces.

|  | read() internal or external | operator() internal | operator() external |
| --------------- | :---------------: | :---------------: | :---------------: |
| `Workspace<const T>` | 👓 `shared_future<ConstTile>` | ⛔️ | ⛔️ |
| `Workspace<T>` | 👓 `shared_future<ConstTile>` |  ✍️ `future<Tile>` | ⚠️ |

⚠️ In the current implementation external tiles are just read-only, so it is not possible to get a read-write tile from an external source. I didn't think too long about making them read-writeable (any problem with future dependencies like with MatrixView?!) since for the moment I don't see the use-case.

One last thing somehow related to this topic. External tiles and internal tiles are mutually exclusive.
- Once an internal tiles is retrieved with access operators, it is not possible anymore to "link" that specific index with an external tile;
- Once a specific index is "linked" to an external tile the internal one cannot be retrieved anymore.

The mask status can be reset in order to re-use the same workspace for other purposes.

albestro:
### 3. `share_panel` API

A nice and very useful feature of the Workspace is the ability to broadcast it in two different way:
1. broadcast the panel: every rank in the same row/column will have the same panel
2. transpose + broadcast the panel: make a row panel a column panel (or viceversa)

#### 1. Broadcast the panel
This is a quite simple and straightforward operation. In the next example a column workspace is used, but it is a valid operation between row panels too.

In this example we have (4,2) matrix distributed over (2,2) ranks. On the left the column workspaces for the various ranks (see the top 4x1 "header" row). Some of them are gray, it means that they have been not initialized, while the others are colored as the rank, meaning that on that rank the tile of the workspace has a valid value (e.g. it has been computed locally).
By broadcasting the column panel row-wise, we will get the situation on the right, where one rank per row broadcasts the value to its row mates.

image

This operation has two points that I'd like to highlight:
- It may be useful to specify a different master rank per row, i.e. by passing a function that given a LocalTileIndex it returns who is the root;
- This operation has been described with the broadcast operation, but I think it can be quite easily generalized to other collective operations (e.g. I may a use-case for the `reduce` in reduction to band algorithm)

#### 2. Transpose + Broadcast
This operation is a bit more "complicated", mainly because it deals with transposed local indices, but I will try to give an overview.

Let's start from one of the simplest use cases: a column of ranks has just computed a column workspace `X`, but the next operation sees X used both non-transposed and transposed, like in `X V' * V X'` (where the apex is the symbol for the transposition).

Since these operations will happen locally on each rank, it is required before proceeding that each rank has access to both version of the workspace.

The **non-transposed** version is not a big deal, it can be collected, if not computed locally, by the column workspace; the broadcast operations described before is enough.

For the **transposed** version it is needed a coordinate mapping. In fact, being transposed, the column maps to the row, and it may happen that a rank is not on the required row, so an additional step is required.
- share the panel row-wise, so that each column of ranks has the same panel;
- at this point, there is by-construction a rank that owns, for each column, the tranposed row of a column

In this last step it is clear how the first point highlighted in the previous paragraph, about being able to compute the root rank given a specific index, comes into play.

albestro:
It is clear that the lifetime of this object, in case of externally mapped tiles, has to be managed carefully by the developer for avoiding deadlocks.

Sely85 (commented):
Great job! Workspaces will be handy for algorithms.

About the aspects you'd like to discuss, here my first, very general opinions. 
1. I personally prefer adding the panel geometry in the type: `Workspace<Row/Col, T, device>` seems the clearest option to me. 
2. The broadcasts look good to me. I'm still figuring the `transpose` out.  Quick question: is it (or will it be) possible to transpose also from row to col? This https://github.com/eth-cscs/DLA-Future/blob/1dbf3c0aa9f244cb6be98b66df02a29fabdf3cf1/include/dlaf/matrix/workspace.h#L225 only do the transpose from col to row, right?
3. I agree not to expose the real distribution of the Matrix.
4. I can not think of a use for `Workspace<const T>` neither. The mechanism of "read/read-write" looks good for both internal and external tiles. 


ps: I find the name `Workspace` appropriate.

albestro:
>     2. The broadcasts look good to me. I'm still figuring the `transpose` out.  Quick question: is it (or will it be) possible to transpose also from row to col? This https://github.com/eth-cscs/DLA-Future/blob/1dbf3c0aa9f244cb6be98b66df02a29fabdf3cf1/include/dlaf/matrix/workspace.h#L225
>         only do the transpose from col to row, right?

The current implementation just works in one direction, i.e. row->col, but it can be quite easily adapted the other way, by:
- changing (inverting) the logic of `whos_root`
- inside the if condition, use the col instead of the row component of the rank
- invert the index computation and the next use for accessing the tile

> ps: I find the name `Workspace` appropriate.

The main thing I don't like about it, it is that it is a very generic term, while in this case we can be more specific since we are dealing with (row/col) panel workspaces

PanelWorkspace? WorkspacePanel? if we opt for embedding the row/col in the type, we get something like`WorkspacePanel<Col, double, CPU>`.

Sely85:
> > ```
> > 2. The broadcasts look good to me. I'm still figuring the `transpose` out.  Quick question: is it (or will it be) possible to transpose also from row to col? This https://github.com/eth-cscs/DLA-Future/blob/1dbf3c0aa9f244cb6be98b66df02a29fabdf3cf1/include/dlaf/matrix/workspace.h#L225
> >     only do the transpose from col to row, right?
> > ```
> 
> The current implementation just works in one direction, i.e. row->col, but it can be quite easily adapted the other way, by:
> 
> * changing (inverting) the logic of `whos_root`
> * inside the if condition, use the col instead of the row component of the rank
> * invert the index computation and the next use for accessing the tile
> 
Good! 

> > ps: I find the name `Workspace` appropriate.
> 
> The main thing I don't like about it, it is that it is a very generic term, while in this case we can be more specific since we are dealing with (row/col) panel workspaces
> 
> PanelWorkspace? WorkspacePanel? if we opt for embedding the row/col in the type, we get something like`WorkspacePanel<Col, double, CPU>`.
You're right (can it be extended for example to Trailing Matrix?). 
I vote for `PanelWorkspace<Row/Col, double, CPU>`. I think it gives you the idea of what you're dealing with more quickly.

rasolca:
1) template over `<Row/Col, double, CPU>` helps readability for sure. What about calling it `Panel`?

2) bcast/reduce are ok, but only in the correct direction. It doesn't make any sense to bcast a column panel columnwise as the number of tiles and their size (thinking about the last tile) might be different. (Anyway is it not better to call them bcast instead of share_panel?)
Maybe the transpose name is not the best choice, as no transposition happens. Just a broadcast among the owner of the transposed panel is performed.
What about `broadcast(Panel<Row,..>, root)`, `broadcast(Panel<Col,..>, root)`,`broadcast(Panel<Col,..>, Panel<Row, ...>)`  and `broadcast(Panel<Row,..>, Panel<Col, ...>)`?

3) If you want to expose a distribution, it should be the correct one. I.e. for a col panel it should be a nb*m panel distributed on a 1*Q grid.

4) If writing external tiles is not needed, I would not care about it. Anyway if this will be needed in future the same mechanism used in MatrixView can be used.

include/dlaf/matrix/panel.h
```diff
@@ -0,0 +1,362 @@
...
+void broadcast(hpx::execution::parallel_executor ex, comm::IndexT_MPI rank_root,
+               Panel<axis, T, device>& ws, common::Pipeline<comm::CommunicatorGrid>& serial_comm) {
+  using namespace comm::sync::broadcast;
+  using hpx::dataflow;
+
+  constexpr auto comm_dir = orthogonal(axis);
+
+  // TODO
+  // if (grid_size.get(component(comm_dir)) < 1)
+  //  return;
```
albestro, line 292: This is an open topic (it applies also to the other `broadcast` function): there is no way of knowing the size of the communicator before scheduling the tasks, unless we explicitly decide to pass it as argument.

If not checked, we may schedule almost "empty" tasks because the check will be carried out in nested calls once the task is in execution (e.g. in the MPI send/recv function).

include/dlaf/matrix/panel.h
```diff
@@ -0,0 +1,362 @@
...
+  using hpx::dataflow;
+
+  DLAF_ASSERT(ws_from.parent_distribution() == ws_to.parent_distribution(),
+              "they must refer to the same matrix");
+
+  // TODO add check about sizes?!
+  // DLAF_ASSERT(ws_from.localNrTiles() >= ws_to.localNrTiles(),
+  //            "sizes", ws_from.localNrTiles(), ws_to.localNrTiles());
+
+  // TODO do I have to check for offset?!
```
albestro, line 330: note-to-self: a check about the compatibility of the two panel should be done.

albestro:
Suggestions also about file structure are welcome (e.g. moving `broadcast` functions somewhere else in the `communication` module.

albestro:
note-to-self (open to suggestions as always).

Currently, the API for `broadcast` hides the executor and accepts a `Pipeline<CommunicatorGrid>`.

Looking at the work done in #344, it seems it will be required to get as arguments two `Pipeline<Communicator>`, respectively for row and col communications, instead of a single `Pipeline<CommunicatorGrid`.

albestro:
While I was looking at #339 , I was wondering how the `Panel` can be useful in that particular context.

In general, it looks helpful there, but something I had in the back of my mind has been exposed, in particular when there is the need to deal with "incomplete" tiles.

Currently, a Panel is built starting from a Matrix and an offset. The offset indicates where the computation needs support for the Panel, so it allows calculation of how tall/wide the Panel should be (i.e. from the offset till the last row/col of the matrix). Once allocated, the memory chunk won't be changed in size, but it will still be possible to change the offset (new_offset >= construction_offset), which is useful for re-using the same Panel, and so the same memory chunk, for example when the algorithm works on trailing submatrices.

This mechanism takes into account that the last tile may be of different size.
For instance, let's consider a Matrix with `m=n=8` and `mb=nb=3`, which will have the last tile of the first column (i.e. [2,0]) with size (2,3) instead of the full blocksize (3, 3).

```
00 01 02 | 03 04 05 | 06 07
10 11 12 | 13 14 15 | 16 17
20 21 22 | 23 24 25 | 26 27
---------|----------|------
30 31 32 | 33 34 35 | 36 37
40 41 42 | 43 44 45 | 46 47
50 51 52 | 53 54 55 | 56 57
---------|----------|------
60 61 62 | 63 64 65 | 66 67
70 71 72 | 73 74 75 | 76 77
```

Let's see how the Panel will be created with two different offsets (same applies to `Panel<Row>`):
- A) `Panel<Col>(offset=[0, 0])` -> size=(8, 3)
- B) `Panel<Col>(offset=[0, 2])` -> size=(8, 2)

IMHO it is reasonable to think that if a panel is created explicitly for the last column (B), so with a potentially "different" size, it will be used as support just for that column and it is not supposed to be used for a "full" column. It can still be problematic, but at least this reduces a bit the domain of problems (i.e. size(panel_tile) >= size(mat_tile)).

If, after the creation, panel offset is changed, e.g. by setting the offset to the immediately next row, the panel will provide access to the tiles skipping the first not needed tiles, so the last tile will always be used (*note1).

Just a couple of example usages, to show how the problem of incomplete tiles may show up.

## Panel as support for results
If we use the panel to store the computation of a related matrix tile, it can work without problems or not. Think about a gemm. If we are working on the second column [:, 1] of the matrix above, doing something like `panel = mat_tile * x`, the result will have the same rows of the mat_tile (so it is ok), but the number of columns will fit exactly just if `x` is a full tile, otherwise we may end up using just part of the tile in the panel (considering to work with tiles of maximum size (mb, mb)).

Then, if you start working with the same panel on the last column, problems are more or less the same, considering also that the very last tile at the bottom right has size (2,2).

*TL;DR* It depends on each result size, we may end up (reasonably) to have more memory than requested and use just a portion of the available tile (so the size of the tile in the panel does not really represent the size of the result).

## Panel as support for temporary copy of a column

Until we use a "full tiles" panel to store a "full tiles" column, there is a perfect match. But, when we reach the last column, the panel has more room for each tile than needed (given my previous note above about size(panel_tile) >= size(mat_tile)).

## Takeout
It is not the first time we realize there will be the need of a special management for tiles with "reduced size" (aka "incomplete tiles") #152.

I would propose for the moment to still ignore the problem, and let the algorithms assume (with a `DLAF_ASSERT`) that the size of the matrix is a multiple of the blocksize, assuring that there won't be the need to deal with incomplete tiles.

What do you think? @rasolca @teonnik @msimberg @Sely85


*note1: this is interesting also from a dependency point of view, as I already experienced in Reduction2Band algorithm, because it may become a "synchronization point", because tiles near the head are used less than the ones in near the tail of the panel).*

rasolca:
The non-complete tiles is a more general problem, therefore I created an issue to discuss about the possible solutions #354.

include/dlaf/communication/rdma.h
```diff
@@ -54,7 +54,7 @@ auto prepareSendTile(hpx::shared_future<matrix::Tile<const T, D>> tile) {
 /// the CPU. This helper duplicates to the GPU if the first template parameter
 /// is a GPU device. The first template parameter must be given.
 template <Device D, typename T>
-auto handleRecvTile(hpx::future<matrix::Tile<const T, CommunicationDevice<D>::value>> tile) {
+auto handleRecvTile(hpx::future<matrix::Tile<T, CommunicationDevice<D>::value>> tile) {
```
albestro, line 57: @msimberg is it ok if we drop this `const`? can you think of any specific case where this is needed?

I'm asking because with the `recvBcast`, the tile has to be an RW one, because it has to be passed from the user then it gets filled with data from comm, differently from `recvBcastAlloc` where the tile is created inside, filled and then returned wrapped in a RO tile.

msimberg: Definitely ok with me. It'll match both const and non-const elements after this anyway. The const was just the tightest requirement for the cases where I put `handleRecvTile` into use, but it's not really a requirement.

albestro: I think I'm going to revert it for the moment, together with changing `recvBcast` to return a `Tile<const T>` instead of a `Tile<T>`.

I was thinking about a case ahead of time, so it would be better to leave it as it is now for the moment, and just in case a specific requirement will come up, we may evaluate the change. Thanks for the feedback!

include/dlaf/matrix/panel.h
albestro: @rasolca instead of passing as argument the communicator size, we could opt for getting the information from the `parentDistribution` as alternative.

What do you think it is better?

teonnik (changes requested):

rasolca (commented):
I see some TODO in the tests.
Do you plan to resolve them or to postpone them in a later PR?

include/dlaf/common/index2d.h
```diff
@@ -26,6 +26,20 @@ namespace dlaf {
...
 
+// Given a Coord direction, returns its orthogonal
+constexpr Coord orthogonal(const Coord rc) {
+  return rc == Coord::Row ? Coord::Col : Coord::Row;
+}
+
+// Given a Coord direction, return the component that varies
+constexpr Coord component(const Coord rc) {
+  return rc == Coord::Row ? Coord::Col : Coord::Row;
+}
```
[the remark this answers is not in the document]
albestro: Yep, I know but to me it looked more explanatory to use two different function names, for the two different context where we can use the `Coord`:
- as an axis, e.g. in `Panel<Col>`
- as an index component, e.g. in `size.get(Coord::Col)`

Let me know if you prefer to just keep a single function name, taking into account also how the code will change/look like.

A good reference for usage is the function `broadcast`, look for `orthogonal` and `component` in it.

https://github.com/eth-cscs/DLA-Future/blob/c8767dd2e39e62801b2a30ee3ab95a88f81c5ec2/include/dlaf/matrix/panel.h#L332

include/dlaf/matrix/panel.h
```diff
@@ -0,0 +1,450 @@
...
+
+/// Panel (1D array of tiles)
+///
+/// 1D array of tiles, i.e. a Row or Column panel strictly related to a given dlaf::Matrix (from the
+/// coords point of view)
+template <Coord axis, class T, Device device>
+struct Panel;
+
+template <Coord axis, class T, Device device>
+struct Panel<axis, const T, device> : protected Matrix<T, device> {
```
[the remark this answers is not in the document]
albestro: IMHO having the Panel axis as a template parameter makes more clear and easier for the user to understand.

See also https://github.com/eth-cscs/DLA-Future/pull/309#pullrequestreview-548351017 and https://github.com/eth-cscs/DLA-Future/pull/309#issuecomment-743178896

include/dlaf/matrix/panel.h
```diff
@@ -0,0 +1,450 @@
...
+    }
+    else {
+      internal_.insert(internal_linear_idx);
+      return BaseT::read(fullIndex(index));
+    }
+  }
+
+  /// Set the panel to a new offset (with respect to the "parent" matrix)
+  ///
+  /// @pre offset cannot be less than the offset has been specifed on construction
```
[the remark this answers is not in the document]
albestro: It's a good point. I think we can easily deal with this change:
- using two different member functions
- using an additional parameter (being it a boolean, a tag, ...)

A remark: the coordinate system in the panel is, as it has been implemented, always equal to the one of the local parent matrix. This means that no matter the hook point for the shrinking, the tiles will always be accessed using LocalTileIndex of the parent matrix.

See the next example, where a panel `P` with its parent matrix `M` (just local part) is depicted togheter with row indices to be used for accessing the tiles.
```
# full panel			# shrink from top		# shrink from bottom
0 P  M M M M M M		  P  M M M M M M		0 P  M M M M M M
1 P  M M M M M M		  P  M M M M M M		1 P  M M M M M M
2 P  M M M M M M		2 P  M M M M M M		2 P  M M M M M M
3 P  M M M M M M		3 P  M M M M M M		3 P  M M M M M M
4 P  M M M M M M		4 P  M M M M M M		  P  M M M M M M
5 P  M M M M M M		5 P  M M M M M M		  P  M M M M M M
```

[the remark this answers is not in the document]
rasolca: I understand that indexing is coherent, however I think that you need to know the correct size for the bcast in the shrink from bottom case, otherwise all tiles will be communicated.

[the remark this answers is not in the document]
rasolca: >     * using two different member functions
>     * using an additional parameter (being it a boolean, a tag, ...)

Maybe a single member function that takes both offset and size?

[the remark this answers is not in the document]
albestro: It looks reasonable, I'm working on it.

Considering the use-case you have in mind, do you think it is enough to allow the shrinking from both edges just once initialized, or do you want also to be able to build a panel that allocates just a (start, end) range of tiles?

Currently it is possible to build a panel that does not begin with the beginning of the matrix, but the end is fixed to the last tile of the parent matrix.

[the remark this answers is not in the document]
rasolca: I don't see the use case for this, therefore I would not.

rasolca (approved):

include/dlaf/matrix/panel.h
```diff
@@ -0,0 +1,479 @@
...
+  ///> It represents the last
+  SizeType end_;
+
+  ///> Container for references to external tiles
+  common::internal::vector<hpx::shared_future<ConstTileType>> external_;
+  ///> Keep track of usage status of internal tiles (accessed or not)
+  std::set<SizeType> internal_;
+};
+
+template <Coord axis, class T, Device device>
```
teonnik, line 261: Do we need `axis` to be a compile-time parameter? It seems to only be used in runtime context.

albestro: Is your comment related to this?
https://github.com/eth-cscs/DLA-Future/pull/309#discussion_r625544384

teonnik: Argh, yes, I thought my previous comment was lost. Thanks for the reply, I will look into it!

teonnik (approved):

albestro:
bors try

albestro:
Before proceeding with merging this, I'm going to rebase the Cholesky factorization refactoring branch #293, in order to check that at least that one works after the last changes to the Panel.

albestro:
> Before proceeding with merging this, I'm going to rebase the Cholesky factorization refactoring branch #293, in order to check that at least that one works after the last changes to the Panel.

I did the rebase and it worked, so this is ready for the last round of review and eventually get merged.

rasolca:
bors try

PR #310, opened 2020-12-02
Add utilities for getting and checking tile operation parameters based on input parameters

msimberg:
Reduces duplication between the blas and cublas tile operations.

### DIFF
--- a/include/dlaf/util_blas.h
+++ b/include/dlaf/util_blas.h
@@ -1,0 +1,151 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2020, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+
+#include <blas.hh>
+
+#include <dlaf/common/assert.h>
+#include <dlaf/matrix/tile.h>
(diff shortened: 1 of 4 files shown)


msimberg:
borst try

include/dlaf/util_blas.h
```diff
@@ -0,0 +1,146 @@
...
+//
+
+#pragma once
+
+#include <blas.hh>
+
+#include <dlaf/types.h>
+
+namespace dlaf {
+namespace util {
```
rasolca, line 18: I would put these methods in dlaf::tile::internal as I think that they are not part of the public API.

rasolca (approved):
LGTM, just a minor comment.

include/dlaf/util_blas.h
msimberg: Yes, agreed! I'll change that.

msimberg:
> borst try

That didn't quite go as planned.

bors try

include/dlaf/util_blas.h
msimberg: Done.

albestro (approved):
I quickly went through and it LGTM.

Just one minor comment: in DLAF there are `Size2D` and `Index2D`, which I think they can be useful even for these things. They can be transposed and they accept a tag (for strong typing). In addition, they have allow few other manipulations (e.g. comparison and arithmetic operations).

PR #312, opened 2020-12-09
Library initialization/finalization

msimberg:
Note that this builds on #234, and thus the diff looks a bit larger than it really is. I've squashed the current changes into the last commit.

This addresses #100/#306.

This is a draft for a couple things:
1. `dlaf::initialize/finalize` for (at this point) initializing static CUDA stream/cuBLAS handle pools, but this could be useful for other things as well.
2. A way to generically use functionality for different backends (with different executors) (not really new, but just applied to a new thing: copy).

As a bonus, it also demonstrates the way I've used so far for extending tile lifetimes generically. I can eventually split this out into a separate PR, but it's related as it makes use of the possibility to get executors for different backends.

To expand on what this includes right now:
1. `dlaf::initialize` currently takes `argc/argv` but does nothing with them. It initializes the static pools for the GPU backend. I could see DLAF eventually getting some command line options for controlling e.g. the number of streams per worker thread. Do we want to do that now? If yes, with `hpx::program_options`? With something else?
2. Currently the only thing I've added is `getCopyExecutor` which returns different types of executors for CPU and GPU (note that the only reason why it's specifically *copy*, not e.g. normal priority, is because I don't have the cuBLAS executor forward to the superclass `async_execute` at the moment. I'm looking into making this work.) For this to actually be generic I've added some helper functionality.

The first helper is a macro for creating callable objects for a particular function name. This simplifies (i.e. avoids) having to do things like this: https://github.com/eth-cscs/DLA-Future/blob/4d66c894389ff99c974ed9f69383cc48cc2088fe/include/dlaf/matrix/copy.h#L39-L41. The callable object simply passes through any and all arguments it gets and dispatches to the actual function. This doesn't have to be done with a macro, but I think in many cases it simplifies having to create the wrapper manually.

The second helper is `unwrapExtendTiles` (name needs some bikeshedding). This is essentially a custom version of `hpx::util::unwrapping` which does the unwrapping, but also returns a tuple of the arguments to extend the lifetimes of the tiles. Currently it simply returns a tuple of *all* the arguments, but this could perhaps be made to return only the tile arguments. I'm not sure yet how to do that.

Note that I've made no attempt to put things in appropriate places yet. I'd appreciate feedback on the functionality. I realize some of this significantly adds complexity, but I think at least some of it is justified. Here I've only done the copy operation (the CUDA versions are now asynchronous!), but I think this would allow having a single implementation of *any* algorithm.

### DIFF
--- a/include/dlaf/common/callable_object.h
+++ b/include/dlaf/common/callable_object.h
@@ -1,7 +1,7 @@
 //
 // Distributed Linear Algebra with Future (DLAF)
 //
-// Copyright (c) 2020, ETH Zurich
+// Copyright (c) 2020-2021, ETH Zurich
 // All rights reserved.
 //
 // Please, refer to the LICENSE file in the root directory.
--- a/include/dlaf/common/callable_object.h
+++ b/include/dlaf/common/callable_object.h
@@ -17,11 +17,12 @@
 /// require explicit selection of the overload when passing the function to
 /// higher-order functions. With this wrapper the overload is selected inside
 /// the wrapper's operator().
-#define DLAF_MAKE_CALLABLE_OBJECT(fname)     \
-  constexpr struct fname##_t {               \
-    template <typename... Ts>                \
-    decltype(auto) operator()(Ts&&... ts) {  \
-      return fname(std::forward<Ts>(ts)...); \
-    }                                        \
-  } fname##_o {                              \
+#define DLAF_MAKE_CALLABLE_OBJECT(fname)                                                 \
+  constexpr struct fname##_t {                                                           \
+    template <typename... Ts>                                                            \
+    auto operator()(Ts&&... ts) const noexcept(noexcept(fname(std::forward<Ts>(ts)...))) \
+        -> decltype(fname(std::forward<Ts>(ts)...)) {                                    \
+      return fname(std::forward<Ts>(ts)...);                                             \
+    }                                                                                    \
+  } fname##_o {                                                                          \
   }
--- a/include/dlaf/common/callable_object.h
+++ b/include/dlaf/common/callable_object.h
@@ -1,0 +1,27 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2020, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+
+/// Given a function name @fname generates a constexpr object with name fname_o
+/// with type fname_t. The generated type has one static operator() which
+/// transparently forwards all arguments to a call to fname. This macro is
+/// useful for creating wrappers of overloaded functions that would otherwise
+/// require explicit selection of the overload when passing the function to
+/// higher-order functions. With this wrapper the overload is selected inside
+/// the wrapper's operator().
+#define DLAF_MAKE_CALLABLE_OBJECT(fname)     \
+  constexpr struct fname##_t {               \
+    template <typename... Ts>                \
+    decltype(auto) operator()(Ts&&... ts) {  \
+      return fname(std::forward<Ts>(ts)...); \
+    }                                        \
+  } fname##_o {                              \
+  }
--- a/include/dlaf/executors.h
+++ b/include/dlaf/executors.h
@@ -1,7 +1,7 @@
 //
 // Distributed Linear Algebra with Future (DLAF)
 //
-// Copyright (c) 2020, ETH Zurich
+// Copyright (c) 2020-2021, ETH Zurich
 // All rights reserved.
 //
 // Please, refer to the LICENSE file in the root directory.
--- a/include/dlaf/executors.h
+++ b/include/dlaf/executors.h
@@ -1,0 +1,120 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2020, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+#pragma once
+
+#include <hpx/execution.hpp>
+#include <hpx/thread.hpp>
+
+#ifdef DLAF_WITH_CUDA
+#include <hpx/modules/async_cuda.hpp>
+#endif
+
+#include <dlaf/communication/executor.h>
+#include <dlaf/init.h>
+#include <dlaf/types.h>
+
+#ifdef DLAF_WITH_CUDA
+#include <dlaf/cublas/executor.h>
+#endif
+
+namespace dlaf {
+namespace internal {
+template <Backend B>
+struct GetMPIExecutor {
+  static auto call() {
+    return hpx::execution::parallel_executor(&hpx::resource::get_thread_pool(
+                                                 comm::internal::mpi_pool_exists() ? "mpi" : "default"),
+                                             hpx::threads::thread_priority_high);
+  }
+};
+
+template <Backend B>
+struct GetNpExecutor {
+  static auto call() {
+    return hpx::execution::parallel_executor{hpx::threads::thread_priority::normal};
+  }
+};
+
+#ifdef DLAF_WITH_CUDA
+template <>
+struct GetNpExecutor<Backend::GPU> {
+  static auto call() {
+    return dlaf::cublas::Executor{getNpCudaStreamPool(), getCublasHandlePool()};
+  }
+};
+#endif
+
+template <Backend B>
+struct GetHpExecutor {
+  static auto call() {
+    return hpx::execution::parallel_executor{hpx::threads::thread_priority::high};
+  }
+};
+
+#ifdef DLAF_WITH_CUDA
+template <>
+struct GetHpExecutor<Backend::GPU> {
+  static auto call() {
+    return dlaf::cublas::Executor{getHpCudaStreamPool(), getCublasHandlePool()};
+  }
+};
+#endif
+
+template <Device S, Device D>
+struct GetCopyExecutor {
+  static auto call() {
+    return hpx::execution::parallel_executor{};
+  }
+};
+
+#ifdef DLAF_WITH_CUDA
+template <>
+struct GetCopyExecutor<Device::GPU, Device::CPU> {
+  static auto call() {
+    return dlaf::cuda::Executor{getNpCudaStreamPool()};
+  }
+};
+
+template <>
+struct GetCopyExecutor<Device::CPU, Device::GPU> {
+  static auto call() {
+    return dlaf::cuda::Executor{getNpCudaStreamPool()};
+  }
+};
+
+template <>
+struct GetCopyExecutor<Device::GPU, Device::GPU> {
+  static auto call() {
+    return dlaf::cuda::Executor{getNpCudaStreamPool()};
+  }
+};
+#endif
+}
+
+template <Backend B>
+decltype(auto) getMPIExecutor() {
+  return internal::GetMPIExecutor<B>::call();
+}
+
+template <Backend B>
+decltype(auto) getNpExecutor() {
+  return internal::GetNpExecutor<B>::call();
+}
+
+template <Backend B>
+decltype(auto) getHpExecutor() {
+  return internal::GetHpExecutor<B>::call();
+}
+
+template <Device S, Device D>
+decltype(auto) getCopyExecutor() {
+  return internal::GetCopyExecutor<S, D>::call();
+}
+}
--- a/include/dlaf/executors.h
+++ b/include/dlaf/executors.h
@@ -28,7 +28,8 @@ namespace internal {
 template <Device S, Device D>
 struct GetCopyExecutor {
   static auto call() {
-    return hpx::execution::parallel_executor{&hpx::resource::get_thread_pool("default")};
+    return hpx::execution::parallel_executor{&hpx::resource::get_thread_pool("default"),
+                                             hpx::threads::thread_priority::normal};
   }
 };
 
@@ -72,7 +73,8 @@ auto getMPIExecutor() {
 /// @tparam B backend with which the executor should be used.
 template <Backend B>
 auto getNpExecutor() {
-  return hpx::execution::parallel_executor{hpx::threads::thread_priority::normal};
+  return hpx::execution::parallel_executor{&hpx::resource::get_thread_pool("default"),
+                                           hpx::threads::thread_priority::normal};
 }
 
 #ifdef DLAF_WITH_CUDA
@@ -88,7 +90,8 @@ inline auto getNpExecutor<Backend::GPU>() {
 /// @tparam B backend with which the executor should be used.
 template <Backend B>
 auto getHpExecutor() {
-  return hpx::execution::parallel_executor{hpx::threads::thread_priority::high};
+  return hpx::execution::parallel_executor{&hpx::resource::get_thread_pool("default"),
+                                           hpx::threads::thread_priority::high};
 }
 
 #ifdef DLAF_WITH_CUDA
--- a/include/dlaf/executors.h
+++ b/include/dlaf/executors.h
@@ -16,7 +16,6 @@
 #include <hpx/modules/async_cuda.hpp>
 #endif
 
-#include <dlaf/communication/executor.h>
 #include <dlaf/init.h>
 #include <dlaf/types.h>
 
@@ -26,51 +25,10 @@
 
 namespace dlaf {
 namespace internal {
-template <Backend B>
-struct GetMPIExecutor {
-  static auto call() {
-    return hpx::execution::parallel_executor(&hpx::resource::get_thread_pool(
-                                                 comm::internal::mpi_pool_exists() ? "mpi" : "default"),
-                                             hpx::threads::thread_priority_high);
-  }
-};
-
-template <Backend B>
-struct GetNpExecutor {
-  static auto call() {
-    return hpx::execution::parallel_executor{hpx::threads::thread_priority::normal};
-  }
-};
-
-#ifdef DLAF_WITH_CUDA
-template <>
-struct GetNpExecutor<Backend::GPU> {
-  static auto call() {
-    return dlaf::cublas::Executor{getNpCudaStreamPool(), getCublasHandlePool()};
-  }
-};
-#endif
-
-template <Backend B>
-struct GetHpExecutor {
-  static auto call() {
-    return hpx::execution::parallel_executor{hpx::threads::thread_priority::high};
-  }
-};
-
-#ifdef DLAF_WITH_CUDA
-template <>
-struct GetHpExecutor<Backend::GPU> {
-  static auto call() {
-    return dlaf::cublas::Executor{getHpCudaStreamPool(), getCublasHandlePool()};
-  }
-};
-#endif
-
 template <Device S, Device D>
 struct GetCopyExecutor {
   static auto call() {
-    return hpx::execution::parallel_executor{};
+    return hpx::execution::parallel_executor{&hpx::resource::get_thread_pool("default")};
   }
 };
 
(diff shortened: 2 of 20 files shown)


include/dlaf/init.h
```diff
@@ -0,0 +1,114 @@
...
+namespace dlaf {
+namespace internal {
+bool& initialized();
+
+template <Backend D>
+struct Init {
+  // Initialization and finalization does nothing by default. Behaviour can be
+  // overridden for backends.
+  static void initialize(int, char**) {}
+  static void finalize() {}
```
msimberg, line 30: We can also explicitly make the `CPU` backend do nothing. However, I thought this was appropriate default behaviour.

include/dlaf/matrix/copy_tile.h
```diff
@@ -46,6 +49,16 @@ struct CopyTile<T, Device::CPU, Device::GPU> {
...
+
+  static void call(const matrix::Tile<const T, Device::CPU>& source,
+                   const matrix::Tile<T, Device::GPU>& destination, cudaStream_t stream) {
+    SizeType m = source.size().rows();
+    SizeType n = source.size().cols();
+
+    DLAF_CUDA_CALL(cudaMemcpy2DAsync(destination.ptr(), destination.ld() * sizeof(T), source.ptr(),
+                                     source.ld() * sizeof(T), m * sizeof(T), n, cudaMemcpyDefault,
+                                     stream));
+  }
```
msimberg, line 61: I've now added stream versions of the CUDA copy operations. We don't strictly need both right now, but I don't know if there's a use case for some user who wants to directly use the blocking version?

include/dlaf/matrix/copy_tile.h
```diff
@@ -89,5 +122,63 @@ void copy(TileElementSize region, TileElementIndex in_idx, const matrix::Tile<co
...
+
+    // Call f with all futures (not just future<tile>) unwrapped.
+    hpx::invoke_fused(hpx::util::unwrapping(f), t);
+
+    // Finally, we extend the lifetime of read-write tiles directly and
+    // read-only tiles wrapped in shared_futures by returning them here in a
+    // tuple
+    // TODO: *Only* return tiles, no other parameters?
+    return t;
+  }
```
msimberg, line 174: This is lots of unwrapping... I still need to see if this impacts compile-times in any way.

test/unit/matrix/test_matrix.cpp
```diff
@@ -1077,6 +1078,8 @@ TYPED_TEST(MatrixTest, GPUCopy) {
   using GPUMemoryViewT = dlaf::memory::MemoryView<TypeParam, Device::GPU>;
   using GPUMatrixT = dlaf::Matrix<TypeParam, Device::GPU>;
 
+  hpx::cuda::experimental::enable_user_polling p;
+
```
msimberg, line 1082: This is due for removal. I've been bitten by forgetting to have this too many times... I'm looking into what we can do in HPX to simplify using it.

include/dlaf/matrix/copy_tile.h
msimberg: @albestro FYI this is the "magic" (combined with the CUDA/cuBLAS executors) that extends the tile lifetimes.

msimberg:
> Here I've only done the copy operation (the CUDA versions are now asynchronous!)

I just remembered that's a lie! We're not yet using pinned host memory, which is required for the transfers to be asynchronous.

msimberg:
> > > Here I've only done the copy operation (the CUDA versions are now asynchronous!)
> > 
> > 
> > I just remembered that's a lie! We're not yet using pinned host memory, which is required for the transfers to be asynchronous.
> 
> It's not a lie, as it should be already using pinned memory unless you are allocating it by yourself... https://github.com/eth-cscs/DLA-Future/blob/master/include/dlaf/memory/memory_chunk.h

Ah, indeed... thanks for the correction! I remembered some changes related to memory management on the cuBLAS executor branch, and I remembered the change as enabling the use of pinned host memory, but it seems I got it mixed up. Oh well, this is even better!

include/dlaf/matrix/copy_tile.h
```diff
@@ -89,5 +122,63 @@ void copy(TileElementSize region, TileElementIndex in_idx, const matrix::Tile<co
...
+struct UnwrapFuture<hpx::future<Tile<T, D>>> {
+  template <typename U>
+  static auto call(U&& u) {
+    auto t = u.get();
+    return t;
+  }
+};
+
+template <typename T>
+struct dummy;
```
albestro, line 155: What is this for?

spack/packages/dla-future/package.py
```diff
@@ -27,7 +27,7 @@ class DlaFuture(CMakePackage, CudaPackage):
     depends_on("lapackpp")
     depends_on("hpx cxxstd=14 networking=none")
     depends_on("hpx@1.5.0", when="~cuda")
-    depends_on("hpx@master +cuda", when="+cuda")
+    depends_on("hpx@unwrapping-noncopyable +cuda", when="+cuda")
```
msimberg, line 30: Note to self: remove (the change was already merged to master).

include/dlaf/matrix/copy_tile.h
msimberg: Debugging... I'll remove it. It prints the type `T` with an error since there's no definition.

albestro: good to know for template debugging!

albestro:
LGTM in general.

test/unit/matrix/test_matrix.cpp
msimberg: The need for enabling polling is not gone, but I've moved this to `dlaf::initialize`.

spack/packages/dla-future/package.py
msimberg: Done.

test/src/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ target_link_libraries(DLAF_gtest_hpx_main
   PUBLIC
     gtest
   PRIVATE
+    DLAF
```
rasolca, line 16: Shouldn't this be `dlaf.core`?

rasolca (commented):
In general it LGTM.
I like the fact of having the possibility to use command line options to control the library. (hpx::program_option is ok as we already depend on hpx...)
However I'm thinking if (command line -> DLAF default) is enough. In some cases an application might want to overwrite some default values (command line -> hard coded user defaults -> DLAF default).
What do you think?

test/src/CMakeLists.txt
msimberg: Most likely yes!

msimberg:
> In general it LGTM.
> I like the fact of having the possibility to use command line options to control the library. (hpx::program_option is ok as we already depend on hpx...)
> However I'm thinking if (command line -> DLAF default) is enough. In some cases an application might want to overwrite some default values (command line -> hard coded user defaults -> DLAF default).
> What do you think?

Yes, I think it only makes sense to reuse what's in HPX. Also agreed on giving the possibility to provide defaults within the application.

I think the simple way of doing that is to have some sort of config struct (nested or not) that gets passed to `dlaf::initialize` along with `argc`/`argv`. A bit like the `hpx::init_params` struct we have in HPX (except that that one has a member which is the actual configuration as a vector of strings). The more complicated option is to do something like what HPX does with configuration files that can be read from a file or passed in as an argument. However, I think it's overly complicated and overkill. The potential missing level of configuration is environment variables: command line options -> environment variables -> user defaults -> DLAF defaults. That's also straightforward to add afterwards.

```
struct config {
    struct nested_options {
        int option1 = 42;
        double option2 = 3.14;
    };

    nested_options no{};
    std::string another_option = "hello";
    std::size_t cuda_streams_per_worker_thread = 3;
};

...

config cfg;
cfg.no.option1 = -1;

initialize(argc, argv, opts);
```
This is simple, but then again it provides direct access to the members. Whether this is sufficient or not depends a bit on if we want/need to do validation immediately on setting values, or if we validate everything at the end. For some things in HPX validating at the end is a must since the options depend on each other, but this may not be necessary in DLAF.

msimberg:
I've pushed some more changes.

1. I've updated the algorithms to use `getXExecutor`. I've added `getNpExecutor`, `getHpExecutor`, and `getMPIExecutor`, in addition to `getCopyExecutor`. Note that I decided to keep `getCopyExecutor` because it's templated on the source and destination devices, not the backend. I have separate changes that would allow the cuBLAS executor to be used in place of the CUDA executor, if needed. I won't be adding that here though.
2. I've added command line handling. It uses a struct for the configuration (I slightly prefer this over a `map<std::string, std::string>` since it's typed, but the latter is also an option). It has DLAF defaults, which can be overridden by user defaults, environment variables, and command line options in increasing order of precedence. I only have the CUDA-related options in there at the moment.

- I'm not 100% sure this is the best way to deal with different `options_description`s. However, it works relatively well. The main quirk is that if the application uses program options, the DLAF `options_description` has to be explicitly added by the user. If not, `argc/argv` can be passed to `dlaf::initialize`, but HPX has to be told to ignore unknown options.
- Do we want a `dlaf::initialize_guard` (`dlaf::init_guard`), i.e. an RAII `initialize/finalize` wrapper?
- I've moved things around into different files and namespaces. I think they're in reasonable places now, but feedback is welcome.

Does this still look like a good direction? I'm starting to be quite happy with this. Definitely still to do is:

- [ ] docstrings
- [ ] checking that all tests are still ok (mainly that DLAF is initialized)
- [ ] some basic tests for the configuration handling

msimberg:
By the way, the miniapps use a special `mpi_initter` RAII object. Is this something that can or should be integrated with `dlaf::initialize`? Are there any special requirements there (except that MPI needs to be threaded)?

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
I should mention the custom unwrapping functionality now requires HPX master (there was a bug in `hpx::util::unwrapping` that prevented it from compiling). Do you want to still stay with 1.5.X? The unwrapping functionality is very handy. I can work around it for now if needed, but long term I'd definitely want to use the custom unwrapper. Is there someone already relying on DLA-Future?

rasolca:
I would prefer to stick with released dependencies... What are the plans for next HPX release?

rasolca:
> By the way, the miniapps use a special `mpi_initter` RAII object. Is this something that can or should be integrated with `dlaf::initialize`? Are there any special requirements there (except that MPI needs to be threaded)?

No other requirements beside the threading level.
You can integrate it, but it should be an option, as MPI can be initialized elsewhere.

msimberg:
> I would prefer to stick with released dependencies... What are the plans for next HPX release?

We've started preparations for 1.6.0. January is likely, latest Febrary. I can definitely wait for the release with this (and try to speed up the release ;)).

msimberg:
> > By the way, the miniapps use a special `mpi_initter` RAII object. Is this something that can or should be integrated with `dlaf::initialize`? Are there any special requirements there (except that MPI needs to be threaded)?
> 
> No other requirements beside the threading level.
> You can integrate it, but it should be an option, as MPI can be initialized elsewhere.

Looking at this a bit more, it may be best left independent of `dlaf::initialize`. We want to initialize MPI in the main thread whereas `dlaf::initialize` should be called after HPX has been initialized (1. HPX can do the command line parsing, 2. the stream/handle pools use the number of worker threads which is only known after starting HPX). Not impossible to overcome, but may be awkward.

include/dlaf/common/callable_object.h
```diff
@@ -0,0 +1,27 @@
...
+/// with type fname_t. The generated type has one static operator() which
+/// transparently forwards all arguments to a call to fname. This macro is
+/// useful for creating wrappers of overloaded functions that would otherwise
+/// require explicit selection of the overload when passing the function to
+/// higher-order functions. With this wrapper the overload is selected inside
+/// the wrapper's operator().
+#define DLAF_MAKE_CALLABLE_OBJECT(fname)     \
+  constexpr struct fname##_t {               \
+    template <typename... Ts>                \
+    decltype(auto) operator()(Ts&&... ts) {  \
```
albestro, line 23: I was playing with something similar...

wouldn't it be betteer to have it `const noexcept`?

msimberg: Good point! I think it should be `const` and conditionally `noexcept`. I'll leave this open to remind me to change this.

msimberg: Also, as discussed earlier with @albestro an alternative to this would be something like this `LIFT` macro: https://github.com/rollbear/lift/blob/42293d66f0bd5a92b2a0ab42284e56950d56d853/include/lift.hpp#L20-L32, which uses a lambda instead of creating a callable object. It's less boilerplate when defining functions, but requires a `LIFT(fun)` when used.

msimberg:
bors try

msimberg:
This is rebased on #315 since it requires that anyway.

include/dlaf/common/callable_object.h
msimberg: This is now updated to be const and conditionally noexcept.

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

rasolca (approved):

include/dlaf/executors.h
```diff
@@ -0,0 +1,112 @@
...
+template <>
+struct GetCopyExecutor<Device::GPU, Device::GPU> {
+  static auto call() {
+    return dlaf::cuda::Executor{getNpCudaStreamPool()};
+  }
+};
+#endif
+}
+
+/// Returns an MPI executor approprate for use with the given backend.
```
albestro, line 60: minor comment: there are a couple of typos "approprate"

include/dlaf/matrix/copy_tile.h
```diff
@@ -70,17 +91,27 @@ struct CopyTile<T, Device::GPU, Device::GPU> {
...
+                                     source.ld() * sizeof(T), m * sizeof(T), n, cudaMemcpyDefault,
+                                     stream));
+  }
 };
 #endif
 }
 
-template <typename T, Device Source, Device Destination>
-void copy(const Tile<const T, Source>& source, const Tile<T, Destination>& destination) {
+template <typename T, Device Source, Device Destination, typename... Ts>
```
albestro, line 108: Just asking, what is this variadic template used/needed for?

include/dlaf/matrix/copy_tile.h
```diff
@@ -46,6 +47,16 @@ struct CopyTile<T, Device::CPU, Device::GPU> {
     DLAF_CUDA_CALL(cudaMemcpy2D(destination.ptr(), destination.ld() * sizeof(T), source.ptr(),
                                 source.ld() * sizeof(T), m * sizeof(T), n, cudaMemcpyDefault));
   }
+
+  static void call(const matrix::Tile<const T, Device::CPU>& source,
+                   const matrix::Tile<T, Device::GPU>& destination, cudaStream_t stream) {
+    SizeType m = source.size().rows();
```
albestro, line 53: What about a check for compatibility between source and destination (in terms of sizes)?

I always struggle finding a general guideline about this, because who should check? the caller (like we are doing in `copy` now), the one which is responsible for the action (e.g. this one I would say), or both?

It's more a philosophic question than a comment...look forward to your thoughts!

test/src/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ target_link_libraries(DLAF_gtest_hpx_main
   PUBLIC
     gtest
   PRIVATE
+    dlaf.core
```
albestro, line 16: `gtest_mpihpx_main` has been changed with the new `dlaf::initialize`, but it's not the case for `gtest_hpx_main`. 
Since I see you are adding the dependency here, I think that something is missing there.

miniapp/miniapp_triangular_solver.cpp
```diff
@@ -65,6 +65,8 @@ linear_system_t sampleLeftTr(blas::Uplo uplo, blas::Op op, blas::Diag diag, T al
 }
 
 int hpx_main(hpx::program_options::variables_map& vm) {
+  dlaf::initialize(vm);
```
albestro, line 68: is it something we have to do also in other miniapp, e.g. `miniapp_cholesky`, right?

albestro (changes requested):

src/init.cpp
```diff
@@ -0,0 +1,192 @@
...
+                     "Number of normal priority CUDA streams per worker thread");
+  desc.add_options()("dlaf:num-hp-cuda-streams-per-thread", hpx::program_options::value<std::size_t>(),
+                     "Number of high priority CUDA streams per worker thread");
+
+  return desc;
+}
+
+void initialize(hpx::program_options::variables_map const& vm, configuration const& user_cfg) {
+  if (vm.count("dlaf:help") > 0) {
+    std::cout << getOptionsDescription() << std::endl;
```
albestro, line 154: Sorry, I didn't test it, but it recalled me something from HPX: does it exit after "dlaf:help" or not?

albestro (changes requested):
...and here it is my second and last part of this round of review.

In general it looks absolutely good to me, just a couple of "minor" comments.

teonnik (approved):
LGTM! I didn't find issues that weren't mentioned already.

test/src/CMakeLists.txt
msimberg: Indeed, it should be in both. Good catch!

test/unit/init/test_init.cpp
```diff
@@ -0,0 +1,217 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#include <cstdlib>
+
+#include <gtest/gtest.h>
+
+#include <hpx/init.hpp>
+
+#include "dlaf/init.h"
```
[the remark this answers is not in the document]
msimberg: Fixed.

msimberg:
bors try

msimberg (commented):
Thanks @albestro for the comments! Let me know if I there's something unclear about my replies.

include/dlaf/matrix/copy_tile.h
albestro: no problem, it was just for realizing what was the rationale. thanks for the clarification.

msimberg:
bors try

test/src/CMakeLists.txt
albestro: What about the `gtest_mpi_main`?

As soon as we enable CUDA support, `dlaf::initialize` starts using HPX for the queue, so it implies the presence of MPI.

Are we going to drop this since DLAF has HPX as strong requirement, so there is no need to have just MPI based tests?

@rasolca ?

src/init.cpp
albestro: I think that the general expectation is that if any help is printed, then it stops without "running" the application.
But I don't know what's the best way of doing it together with HPX.

albestro (approved):
LGTM in general.

Just a couple of comments left open, in particular `gtest_mpi_main` and "exit on help".

test/src/CMakeLists.txt
rasolca: MPI only test should be kept to test basic MPI functionalities (e.g. communicator management, ...).
As they do not require dlaf::init, I don't see any problem to keep them as they are now.

src/init.cpp
msimberg: Yes, agreed. I'll see what I can do here given the limitations of already running inside a task. As an aside, `--hpx:help` will already print the DLAF-specific options as well. I added this option to print only the DLAF-specific options. The former may be enough? I'll check what I can do in any case.

albestro: Yes, exactly. I think it can be enough `--hpx:help`, even if it is not straightforward, it is what I do currently.

rasolca:
bors try

rasolca:
@msimberg I just realized that in the miniapp_cholesky dlaf options are not added and cannot be used. Is this an expected behaviour?

msimberg:
@rasolca I've fixed `miniapp_cholesky` now, but would you mind giving it a try as well? I added the `exit` when `--dlaf:help` is passed now, and I also changed the order things a bit so that if `--dlaf:help` and `--dlaf:print-config` are given the config will be printed before the exit happens, but after the help text. @albestro please also have a look to see if this looks good to you.

albestro (approved):
LGTM

Issue #311, opened 2020-12-09
Scripts for benchmark analysis


albestro:
In the past year we have created a set of python scripts for analysis of benchmark results.

It has been recognized to be helpful for the entire team, especially for evaluating new features. So the time has come to share our efforts on improving and fixing these scripts too.

Do we want to provide them as part of our library (e.g. in tools, in miniapp, ...)? Or do we want to keep them in a private team git repo?

Looking forward to your proposals.

msimberg:
I'm in support of simply having them in the scripts subdirectory.

Sely85:
I think that such scripts can be useful for users too. 
I agree that the scripts subdirectory would be a good location.

albestro:
> I think that such scripts can be useful for users too.
> I agree that the scripts subdirectory would be a good location.

What do you have in mind about "being useful for users"?

Sely85:
If we provide them

> > I think that such scripts can be useful for users too.
> > I agree that the scripts subdirectory would be a good location.
> 
> What do you have in mind about "being useful for users"?

Not sure I've understood correctly.
But besides the fact that these scripts are a great tool for us when developing new features, I think that users approaching a new library would probably make some benchmarks on their own clusters as well. And if we are able to provide "tools" (aka scripts working on our cluster) for analyzing benchmarks results quickly it might be useful.

teonnik:
My concern is that we all have different workflows and may need/want to change the scripts considerably. However, I do agree that having a starting point is very useful, that is what I would potentially use the scripts for. 

I prefer having them in the scripts directory. It's less of a hassle than managing multiple repositories. I would also prefer if they are python / bash scripts and not jupyter notebooks. Notebooks can be converted to scripts and vice-versa:

```bash
jupyter nbconvert  foo.ipynb --to="python" --output="foo.py"
```  

How do we decide whose scripts we upload?

rasolca:
> My concern is that we all have different workflows and may need/want to change the scripts considerably. However, I do agree that having a starting point is very useful, that is what I would potentially use the scripts for.

I have the same concern as @teonnik. The scripts can vary a lot depending on the benchmark.
How we would deal with it? If we just upload a basic version, we risk to increase the effort of maintaining the personal variations.

> I prefer having them in the scripts directory. It's less of a hassle than managing multiple repositories.

Fine with me.

> I would also prefer if they are python / bash scripts and not jupyter notebook.

[...]

PR #313, opened 2020-12-10
Scripts

teonnik:
As requested in #311 these are the scripts I use.

- `miniapps.py` - a small library for writing a set of job files
- `gen.py` - the script that actually generates the job files using functions from `miniapps.py`
- `postprocess_cholesky.py` - a postprocessing script used to analyze the data produced for Cholesky in the latest runs.

The `gen.py` script generates the following jobs and files in it's current form:

```
<run_dir>
├── 16
│   ├── job.sh
├── 32
│   ├── job.sh
├── 64
│   ├── job.sh
└── 8
    ├── job.sh
```

Each `job.sh` produces a number of files, depending on the number of executables and combination of flags:

```
├── chol_master.txt    # output files for this particular combination of executable and flags
├── chol_nbmpi_default_polling_default.txt  # ^^
├── chol_nbmpi_default_polling_mpi.txt    # ^^
├── chol_nbmpi_default_yielding_default.txt  # ^^
├── chol_nbmpi_default_yielding_mpi.txt  # ^^
├── chol_nbmpi_shared_polling_default.txt   # ^^
├── chol_nbmpi_shared_polling_mpi.txt  # ^^
├── chol_nbmpi_shared_yielding_default.txt  # ^^
├── chol_nbmpi_shared_yielding_mpi.txt  # ^^
├── env.txt             # printout of the environment of the job at the time of execution (for debugging)
├── error.txt           # any errors
├── job.sh              # the job file
├── modules.txt     # a list of modules loaded at the time of execution of the job script
└── output.txt         # output file -- empty in most cases as output is redirected to separate files for easier post-processing
```

<ins>Note</ins>: 

In my workflow I have a file `<build_dir>/.envrc` which contains the environment that the executables were built in. This is sourced in the `job.sh` file (see `miniapps.py`). If you are using modules, you can just do:

```bash
printenv > <build_dir>/.envrc
```

after the modules are loaded.

If you are using spack and you are building against a spec, you can do:

```bash
spack build-env --dump <build_dir>/.envrc /<spec>
```

Often sourcing the file is not necessary as the executable can find it's dependencies with RPATH. This is just as a means of precaution, I've noticed issues with `intel-mkl` in the past unless some env vars are sourced.

<ins>Issue</ins>: 

Currently there is a single `.envrc` file sourced for all executables. That may need to be generalized such that each executable can use a different environment file.


albestro:
I think that in your refactoring you did quite some nice improvements with respect to first version of the scripts:
- directory structure needed an improvement, and your proposal is interesting
- script generation was done with a bash script, and even on this side it is now way easier to generate a set of configurations by combining various parameter domains

I think that in this version a missing part is the modularity that allowed to keep a backbone for various miniapp and just specifying custom functions for parsing. I think it is mainly due to the fact that you are comparing variants inside DLAF, while I was comparing results of miniapps from different libraries (DLAF vs SLATE vs DPLASMA ...) AND different variants (like you are doing).

I would try to write a more detailed comment about this showing what's available, so we can merge the two works together in case we find it useful.

teonnik:
> I think that in this version a missing part is the modularity that allowed to keep a backbone for various miniapp and just specifying custom functions for parsing. I think it is mainly due to the fact that you are comparing variants inside DLAF, while I was comparing results of miniapps from different libraries (DLAF vs SLATE vs DPLASMA ...) AND different variants (like you are doing).

Ahh, yes indeed, I forgot about SLATE and DPLASMA. I will try to write some functions for them in miniapps.py based on your bash scrip

Regarding the parsing part of the post-processing script, it currently depends on the how the `suffix` for each output file was generated. So there is a need for changes there whenever `gen.py` is changed to generate different suffixes.

I forgot to mention that when the jobs are done I copy the whole tree generated by the `gen.py` file and the job's outputs from the remote to a local directory and I rename `<run_name>` to `<data>`. This is reflected in the post-processing script, but I can potentially generalize it.

albestro:
> Regarding the parsing part of the post-processing script, it currently depends on the how the `suffix` for each output file was generated. So there is a need for changes there whenever `gen.py` is changed to generate different suffixes.

What about adding a "marker" to the filename? Like an additional extension which tells us that the file follows our "format" and we have some expectation about being able to parse those files?

> I forgot to mention that when the jobs are done I copy the whole tree generated by the `gen.py` file and the job's outputs from the remote to a local directory and I rename `<run_name>` to `<data>`. This is reflected in the post-processing script, but I can potentially generalize it.

It's ok for me. In general, I think that it is nice to have the chance to separate group of results in different folders, because I have some post-processing scripts that read from various data folders, which allows to easily compare results between different campaigns (even from the past).

teonnik:
This is ready for another round of reviews. There is an up-to-date [README.md](https://github.com/eth-cscs/DLA-Future/blob/8bc42f8c8cff5e5438517a2e50ff248eeb85abd0/scripts/README.md) with information on the job-generation and post processing workflows.

> What about adding a "marker" to the filename? Like an additional extension which tells us that the file follows our "format" and we have some expectation about being able to parse those files?

The name of the file provides information on the format of the file and how it should be parsed.

scripts/miniapps.py
```diff
@@ -0,0 +1,128 @@
...
+# Note: the second factor is larger or equal to the first factor
+def _sq_factor(n):
+    for i in range(1, int(n ** 0.5) + 1):
+        if n % i == 0:
+            f = (i, n // i)
+    return f
+
+
+def _check_ranks_per_node(rpn):
+    if not (rpn == 1 or rpn == 2):
```
rasolca, line 16: I don't understand why the choice of ranks per node is constrained.
Depending on the architecture (e.g. multiple numa nodes or multiple gpu per node might require more).

scripts/miniapps.py
```diff
@@ -0,0 +1,128 @@
...
+# where `dlaf_nb_[0|1][0|1][0|1]` is `dlaf_nb_<hpx queue><non-blocking mechanism><mpi pool>`, for example
+#       `dlaf_nb_111` : dlaf with non-blocking MPI(nb), shared priority queue(1), polling (1) and a dedicated core for an MPI pool (1).
+#
+# rpn: ranks per node
+#
+def chol(lib, build_dir, nodes, rpn, m_sz, mb_sz, nruns):
+    _check_ranks_per_node(rpn)
+
+    total_ranks = nodes * rpn
+    cpus_per_rank = 72 // rpn
```
rasolca, line 88: I would not hardcode system dependent parameter in a file that (citing the README) usually should not be modified. (Applies to other part of this file.)

scripts/miniapps.py
```diff
@@ -0,0 +1,128 @@
...
+    grid_cols, grid_rows = _sq_factor(total_ranks)
+
+    if lib.startswith("dlaf"):
+        extra_flags = _dlaf_flags(lib)
+        cmd = f"{build_dir}/miniapp/miniapp_cholesky --matrix-size {m_sz} --block-size {mb_sz} --grid-rows {grid_rows} --grid-cols {grid_cols} --nruns {nruns} --hpx:use-process-mask {extra_flags}"
+    elif lib == "slate":
+        cmd = f"{build_dir}/test/slate_test potrf --dim {m_sz}x${m_sz}x0 --nb {mb_sz} --p {grid_rows} --q {grid_cols} --repeat {nruns} --check n --ref n --type d"
+    elif lib == "dplasma":
```
rasolca, line 96: dplasma should be run only with 1 rank per node. I would add a check here.

scripts/miniapps.py
```diff
@@ -0,0 +1,128 @@
...
+def chol(lib, build_dir, nodes, rpn, m_sz, mb_sz, nruns):
+    _check_ranks_per_node(rpn)
+
+    total_ranks = nodes * rpn
+    cpus_per_rank = 72 // rpn
+    grid_cols, grid_rows = _sq_factor(total_ranks)
+
+    if lib.startswith("dlaf"):
+        extra_flags = _dlaf_flags(lib)
+        cmd = f"{build_dir}/miniapp/miniapp_cholesky --matrix-size {m_sz} --block-size {mb_sz} --grid-rows {grid_rows} --grid-cols {grid_cols} --nruns {nruns} --hpx:use-process-mask {extra_flags}"
```
rasolca, line 93: The possibility to investigate multiple options shouldn't be limited to this specific case.
I would suggest to add a parameter to pass the function that generates the extra_flags (which allow None when no extra flags are needed).

scripts/miniapps.py
```diff
@@ -0,0 +1,128 @@
...
+# Debug
+module list &> modules.txt
+printenv > env.txt
+"""
+    return job_text[1:-1]
+
+
+# Create the job directory tree and submit jobs.
+#
+def submit_jobs(run_dir, nodes, job_text):
```
rasolca, line 66: I would add the possibility to add a suffix to the job file. In this way you can split huge jobs without loosing the batch scripts used.

rasolca (changes requested):
Currently I reviewed only the submit part (using them in a real benchmark campaign).
I will review the analysis part later.

scripts/miniapps.py
```diff
@@ -0,0 +1,128 @@
...
+#SBATCH --time={time_min}
+#SBATCH --nodes={nodes}
+#SBATCH --partition=normal
+#SBATCH --account=csstaff
+#SBATCH --constraint=mc
+#SBATCH --output=output.txt
+#SBATCH --error=error.txt
+
+# Env
+export CRAYPE_LINK_TYPE=dynamic
```
rasolca, line 54: No need, as nothing is linked in this script.

scripts/miniapps.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+
+def _err_msg(lib):
+    return f"No such `lib`: {lib}! Allowed values are : `dlaf|dlaf_nb_[0|1][0|1][0|1]`, `slate` and `dplasma`."
+
+
+# Job preamble
+#
+def init_job_text(run_name, nodes, time_min):
```
rasolca, line 41: The batch template is system dependent too.

scripts/miniapps.py
teonnik: > The possibility to investigate multiple options shouldn't be limited to this specific case.

Yes, that is the case. That is a good point.

> I would suggest to add a parameter to pass the function that generates the extra_flags (which allow None when no extra flags are needed).

Wouldn't it be simpler to pass a string of the extra flags than the function/mechanism generating it? The mechanism/function can be moved to `gen.py`. 

I would also like to add a parameter for an output suffix instead of deducing it from `lib`. That way, `lib` can be simplified to just `dlaf` and there will be no need for suffixes on this parameter (e.g `_nb_111` etc).

scripts/postprocess.py
```diff
@@ -0,0 +1,206 @@
...
+            alpha=0.2,
+        )
+
+    ax.set_ylabel(ylabel)
+    ax.set_xlabel("nodes")
+    ax.set_xticks(df["nodes"].sort_values().unique())
+    ax.legend(loc="upper right", prop={"size": 6})
+    ax.set_title(title)
+    fig.savefig(f"{file_name}.png", dpi=300)
+
```
rasolca, line 38: `plt.close(fig)` is missing.

rasolca (commented):
The plot part LGTM. It is a good starting point to visualize data.

scripts/miniapps.py
rasolca: Maybe we should add config files (json?) for different systems
and let the user specify the system in the gen.py file.

teonnik:
@rasolca The remaining comments are all related to the fact that I had assumed Piz Daint's MC partition as the system and haven't generalized the generation scripts to other architectures just yet. 

Depending on the system that may not be very easy. For example, do we assume SLURM as a scheduler? All clusters I currently work on use SLURM, but I've encountered other schedulers in the past (LSF). Which systems/architectures do we want supported (e.g. Piz Daint, Cineca, ...)

I can add a `system` (with values `daint_mc`, `daint_gpu`) parameter to `init_job_text()`, `chol()`, `trsm()` which changes the job preamble, the checks on `rpn` and the `cpus_per_rank` calculations depending on the desired system.

rasolca:
@teonnik I added a commit that generalize the miniapp.py script to deal with multiple systems.
This approach allows to support non-slurm systems, and any specific srun/mpirun/... flag that the system might need.
Feel free to drop it if you have a better solution or to modify it. (Note: currently I just tried a few cases and it run correctly. I have to run more tests if we decide to keep it)

I also added a second commit to use better run config for slate.

teonnik:
@rasolca LGTM!

scripts/miniapps.py
```diff
@@ -0,0 +1,107 @@
...
+        cmd = f"{build_dir}/test/tester potrf --dim {m_sz}x{m_sz}x0 --nb {mb_sz} --p {grid_rows} --q {grid_cols} --repeat {nruns} --check n --ref n --type d {extra_flags}"
+    elif lib == "dplasma":
+        env = "OMP_NUM_THREADS=1"
+        cmd = f"{build_dir}/tests/testing_dpotrf -N {m_sz} --MB {mb_sz} --NB {mb_sz} --grid-rows {grid_rows} --grid-cols {grid_cols} -c {cpus_per_rank} -v {extra_flags}"
+    else:
+        raise ValueError(_err_msg(lib))
+
+    run_cmd = run_command(system, total_ranks, cpus_per_rank)
+    return (
+        "\n" + f"{env} {run_cmd} {cmd} >> chol_{lib}_{suffix}.out 2>&1".strip()
```
msimberg, line 76: Note: I added `2>&1` to also redirect stderr to the output file. DPLASMA outputs the header lines to stderr. From my testing this doesn't seem to break anything, but could of course lead to unwanted output in the file.

scripts/gen.py
```diff
@@ -0,0 +1,72 @@
...
+        job_text += mp.chol(
+            system,
+            "dplasma",
+            dplasma_build_dir,
+            nodes,
+            ranks_per_node,
+            m_sz,
+            mb_sz,
+            nruns,
+        )
```
rasolca, line 66: trsm was changed to cholesky, but iteration over n_sz (not used) remained.

rasolca (commented):
I added some enhancements:
- possibility to use a log scale for x axis,
- possibility to read results from multiple directories
- possibility to submit the same batch script multiple times (reduce node config dependency)

I also added full gen examples for cholesky and for trsm.

rasolca:
> possibility to submit the same batch script multiple times (reduce node config dependency)

Just re-thinking about it... Maybe it is not the best idea as it can lead to mangled output...

albestro:
I'm giving it a try right now, I'll let give you my feedback as soon as possible!

albestro:
What about adding a `requirements.txt` with

```
seaborne
parse
```

so that it is enough `pip install -r requirements.txt` to install what scripts need?

albestro (approved):
I had the chance to do some basic performance analysis, and I decided to switch from my old ones to these new shared ones.

I think it is a very nice and valuable effort! I found them very handy and useful.

I didn't went through all the details of the scripts, but I tested them both on daint and on eiger and they worked like a charm!

See just my previous comment about `requirements.txt`, moreover you may consider to add `+x` permissions to other `gen_*` scripts that have been added later, too.

For the rest...thank you again everyone @teonnik @rasolca @msimberg , very nice job!

rasolca (approved):
I fixed the last problems.

The only thing to adress before merging is the `requirements.txt`.
@albestro mentioned
```
seaborne
parse
```
but what about
```
numpy
matplotlib
pandas
```?
Shouldn't they be considered requirements too?

albestro:
> I fixed the last problems.
> 
> The only thing to adress before merging is the `requirements.txt`.
> @albestro mentioned
> 
> ```
> seaborne
> parse
> ```
> 
> but what about
> 
> ```
> numpy
> matplotlib
> pandas
> ```
> 
> ?
> Shouldn't they be considered requirements too?

It's a good point, since we use all of them directly, it is better to specify them.

rasolca:
Everything is addressed. I'll wait @teonnik's green light before merging.

teonnik:
LGTM!

PR #317, opened 2021-01-15
A couple of minor fixes

albestro:
This PR addresses few minor fixes:
- solve build problem for miniapp triangular solver if tests are not enabled (see also #299)
- add compiler warnings also for cublas (@msimberg do you see any problem?)
- test_range2d was not reserving space for the vectors


teonnik (approved):

albestro:
bors try

msimberg (approved):

msimberg:
>     * add compiler warnings also for cublas (@msimberg do you see any problem?)

Nope, thanks for fixing that. It's an oversight on my part.

albestro:
Sorry, I just realized there is another subtle problem with CMake.

Given that `miniapp/CMakeLists.txt` depends on the `DLAF_WITH_TEST` and this option is parsed just after the miniapp section, it results in the wrong behavior: even if tests are enabled the `miniapp_triangular_solver` won't be built.

Since it seems enough sneaky, instead of reordering miniapp and test sections in the main `CMakeLists.txt` in order to reflect this dependency (that we are willing to remove), my proposal would be to move all options together in a single place somewhere near the begin of the main `CMakeLists.txt`.

@msimberg @teonnik

albestro:
bors try

CMakeLists.txt
```diff
@@ -79,26 +82,20 @@ add_subdirectory(src)
...
 
 # ---------------------------------------------------------------------------
 # Test
 # ---------------------------------------------------------------------------
-set(DLAF_WITH_TEST ON CACHE BOOL "Enables tests")
-if (DLAF_WITH_TEST)
-  include(CTest)
-  enable_testing()
-
+if (BUILD_TESTING)
```
teonnik, line 92: Why remove the prefix here (`DLAF_`)?

CMakeLists.txt
```diff
@@ -17,13 +17,17 @@ project(DLAF VERSION 0.1.0)
...
 include(DLAF)
 
 if (NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Build Type" FORCE)
 endif()
 
 option(BUILD_SHARED_LIBS "Build shared libraries." OFF)
+option(DLAF_WITH_CUDA "Enable CUDA support" OFF)
+option(DLAF_BUILD_MINIAPPS "Build miniapps" ON)
+option(BUILD_DOC "Build documentation" OFF)
```
teonnik, line 30: We should probably add the prefix `DLAF_` to `BUILD_DOC`.

teonnik (changes requested):
The spack package has to be updated too.

CMakeLists.txt
albestro: In the last months I realized that it is a CMake standard thing. Same for `BUILD_DOC` and `BUILD_SHARED_LIBS`.

I thought it was a good thing to do, in case I can revert it.

albestro:
> The spack package has to be updated too.

Sorry, I replied to your comments and I missed this one. I'm not getting your point, the options have been matched to the new ones. If you meant to remember reverting also those ones, I'll do it in case 😉.

Looking forward to your feedback @teonnik to my previous reply (also suggestions from @msimberg who have experience from HPX and all the CMake options).

CMakeLists.txt
teonnik: Sorry, something weird happened with GitHub here, I replied yesterday but my reply was marked as pending and you seem not to have seen it....

I like having a prefix `DLAF_` instead of `BUILD_TESTING` / `BUILD_DOC` because if `add_subdirectory()` is used to add DLA-F as a subproject, we need to take special precautions to avoid activating our tests when the top project makes use of the same variable to activate their tests as discussed here : https://cliutils.gitlab.io/modern-cmake/chapters/testing.html .

teonnik:
> Sorry, I replied to your comments and I missed this one. I'm not getting your point, the options have been matched to the new ones. If you meant to remember reverting also those ones, I'll do it in case wink.

I may have missed the spack changes when I made the comment, looks good to me now!

albestro:
bors try

teonnik (approved):

PR #318, opened 2021-01-15
Communication helpers

albestro:
This PR for discussing about communication helpers and extend the set of available ones.

IMHO their main purpose is hiding all the needed boilerplate for communicating using a specific (sub-)`Communicator` wrapped+embedded in `PromiseGuard<CommunicatorGrid>`.

About design decisions, following what it has been already implemented:
1. Manage variants with or without the executor as first parameter
2. Helpers are task schedulers and not task functions (i.e. they are not the thing to be passed to `hpx::dataflow` but the thing that internally calls `hpx::dataflow`)
3. The available `recv_tile` helper also manage the allocation of the destination tile and returns it

About 3, I think we should improve separation of concerns: IMHO currently it seems more a "need" of the code where it is used, but for them I can see a use case for the #309 or something very similar.

Last thing, I played a bit with templates for a possible alternative/complementary option ((see [alby/proposal/comm_helpers_template](https://github.com/eth-cscs/DLA-Future/pull/new/alby/proposal/comm_helpers_template)). It is mainly about `unwrap_guards` and `selector` that acts as "extensions" of `hpx::unwrapping`: former one unwraps the `PromiseGuards<T>` and the latter one selects the specified sub-`Communicator` in a `CommunicatorGrid`, allowing something like `hpx::dataflow(unwrap_guards(selector<Coord::Row>(broadcast::send))), ...)`

### DIFF
--- a/include/dlaf/communication/sync/broadcast.h
+++ b/include/dlaf/communication/sync/broadcast.h
@@ -55,59 +55,44 @@ void receive_from(const int broadcaster_rank, Communicator& communicator, DataOu
 }
 }
 
+/// Task for broadcasting (send endpoint) a Tile in a direction over a CommunicatorGrid
 template <class T>
-void send_tile(hpx::future<common::PromiseGuard<comm::CommunicatorGrid>> task_chain, Coord rc_comm,
+void send_tile(hpx::future<common::PromiseGuard<comm::CommunicatorGrid>> mpi_task_chain, Coord rc_comm,
                hpx::shared_future<matrix::Tile<const T, Device::CPU>> tile) {
-  using ConstTile_t = matrix::Tile<const T, Device::CPU>;
   using PromiseComm_t = common::PromiseGuard<comm::CommunicatorGrid>;
 
-  auto send_bcast_f = [rc_comm](hpx::shared_future<ConstTile_t> ftile,
-                                hpx::future<PromiseComm_t> fpcomm) {
-    PromiseComm_t pcomm = fpcomm.get();
-    comm::sync::broadcast::send(pcomm.ref().subCommunicator(rc_comm), ftile.get());
-  };
-
-  send_bcast_f(tile, std::move(task_chain));
+  PromiseComm_t pcomm = mpi_task_chain.get();
+  comm::sync::broadcast::send(pcomm.ref().subCommunicator(rc_comm), tile.get());
 }
 
 DLAF_MAKE_CALLABLE_OBJECT(send_tile);
 
+/// Task for broadcasting (receiving endpoint) a Tile in a direction over a CommunicatorGrid
 template <class T>
 void recv_tile(hpx::future<common::PromiseGuard<comm::CommunicatorGrid>> mpi_task_chain, Coord rc_comm,
                hpx::future<matrix::Tile<T, Device::CPU>> tile, comm::IndexT_MPI rank) {
   using PromiseComm_t = common::PromiseGuard<comm::CommunicatorGrid>;
-  using Tile_t = matrix::Tile<T, Device::CPU>;
 
-  auto recv_bcast_f = hpx::util::annotated_function(
-      [rank, rc_comm](hpx::future<PromiseComm_t> fpcomm, hpx::future<Tile_t> ftile) {
-        PromiseComm_t pcomm = fpcomm.get();
-        Tile_t tile = ftile.get();
-        comm::sync::broadcast::receive_from(rank, pcomm.ref().subCommunicator(rc_comm), tile);
-      },
-      "recv_tile");
-  recv_bcast_f(std::move(mpi_task_chain), std::move(tile));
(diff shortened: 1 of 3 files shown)


msimberg:
I'll throw in another "separate concern": when RDMA isn't used for CUDA tiles need to be duplicated before and after sending them and it's handy to hide this at least form the main algorithm bodies.

Regarding 3. I think it's fair that helpers would combine would combine otherwise independent functionality. Whether the current `send/recv_tile` helpers should be the helpers that do that, or if there should be another level of helpers is of course another question, so I'm not necessarily arguing against this point.

albestro:
> Regarding 3. I think it's fair that helpers would combine would combine otherwise independent functionality. Whether the current `send/recv_tile` helpers should be the helpers that do that, or if there should be another level of helpers is of course another question, so I'm not necessarily arguing against this point.

I'm absolutely with you on this. In the case you propose, I agree that we should "hide" in the helper the intermediate CPU tile...even if we should pay attention to how this tile is created, because it may be important to fetch it from a pool of available temporary resources pre-allocated...

However, keeping the focus on this PR, my point about point 3 is more on the CPU variant, that in my opinion should not internally allocate the tile. In fact, my concern started because I'm working on the `Panel` proposal (#309) that would change the management of "support" tiles in algorithms, so providing also the space for received tiles and removing the need of allocating support tiles inside the helper `recv_tile`.

albestro:
> Regarding 3. I think it's fair that helpers would combine would combine otherwise independent functionality. Whether the current `send/recv_tile` helpers should be the helpers that do that, or if there should be another level of helpers is of course another question, so I'm not necessarily arguing against this point.

I'm absolutely with you on this. In the case you propose, I agree that we should "hide" in the helper the intermediate CPU tile...even if we should pay attention to how this tile is created, because it may be important to fetch it from a pool of available temporary resources pre-allocated...

However, keeping the focus on this PR, my point about point 3 is more on the CPU variant, that in my opinion should not internally allocate the tile. In fact, my concern started because I'm working on the `Panel` proposal (#309) that would change the management of "support" tiles in algorithms, so providing also the space for received tiles and removing the need of allocating support tiles inside the helper `recv_tile`.

albestro:
> Regarding 3. I think it's fair that helpers would combine would combine otherwise independent functionality. Whether the current `send/recv_tile` helpers should be the helpers that do that, or if there should be another level of helpers is of course another question, so I'm not necessarily arguing against this point.

I'm absolutely with you on this. In the case you propose, I agree that we should "hide" in the helper the intermediate CPU tile...even if we should pay attention to how this tile is created, because it may be important to fetch it from a pool of available temporary resources pre-allocated...

However, keeping the focus on this PR, my point about point 3 is more on the CPU variant, that in my opinion should not internally allocate the tile. In fact, my concern started because I'm working on the `Panel` proposal (#309) that would change the management of "support" tiles in algorithms, so providing also the space for received tiles and removing the need of allocating support tiles inside the helper `recv_tile`.

albestro:
About 2, "task scheduler vs task function", by working with these I started having some interesting use cases that may support the idea of having them as task function instead of task schedulers.

By having them as task functions:
- in case of tracing/profiling, it is possible to wrap differently the same task function easily (it may require creation of callable objects, there is a macro in another PR from @msimberg which is useful for this purpose). A couple of examples: it would be possible to specify indices of the task for debugging purposes, or additional info about the communication phase (diagonal tile broadcasting, column panel row-wise broadcasting, ...)
- we can still use them to hide any boilerplate/machinery, and we can avoid having variants with/without executor to be passed to dataflow. The price to pay is that we will have the call `dataflow(ex, helper, arg1, arg2)` inside the algorithm (IMHO not much).

rasolca (commented):
In general it looks good.

Please check function naming style.

albestro:
> In general it looks good.
> 
> Please check function naming style.

Are you thinking about renaming `send_tile` to `sendTile`? or just `recv_tile_with_alloc` to `recvTileWithAlloc` which has been just introduced?

note-to-self: add back `hpx::annotate_function`

rasolca:
Yes it would be great. What about a shorter name for the last case? `recvAllocTile`?

albestro:
About annotations, I had a quick call with @msimberg, and he reported that there might be some overhead with RAII `hpx::annotate_function` compared to functional wrapper `hpx::annotated_function`.

I think it may be worth sharing a couple of thoughts about this topic, so I would go for opening an issue and delay the annotation thing to a dedicated PR.

albestro:
bors try

albestro:
It seems all tests passed correctly.

albestro:
@teonnik do you see any problem with this change? I'm also talking about what you are currently working on.

teonnik (approved):
I will have to change most of this but it won't be hard. I don't see issues with this PR, I think It can go through, I will change things in my branch later.

rasolca (approved):

PR #322, opened 2021-01-26
Update the FindMKL module to squash warnings

teonnik:
The module is based on https://github.com/eth-cscs/cmake-recipes/blob/master/FindMKL.cmake .


cmake/FindMKL.cmake
```diff
@@ -175,135 +127,183 @@ endif()
...
+#
 __mkl_find_library(MKL_SEQ_LIB mkl_sequential)
-if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT APPLE)
+if(NOT APPLE AND (CMAKE_C_COMPILER_ID STREQUAL "GNU" OR
+                  CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR
+                  CMAKE_Fortran_COMPILER_ID STREQUAL "GNU"))
     __mkl_find_library(MKL_OMP_LIB mkl_gnu_thread)
 else()
     __mkl_find_library(MKL_OMP_LIB mkl_intel_thread)
 endif()
```
rasolca, line 200: I'm wondering about Clang openMP... Does anyone know if it is compatible with any of the MKL library?

teonnik:
bors try

cmake/FindMKL.cmake
teonnik: I am not sure... unless someone chimes in, I suggest we address it when/if we encounter issues.

rasolca (approved):
LGTM

PR #323, opened 2021-01-27
Add missing DLAF_BUILD_TESTING option

msimberg:
From what I can tell this option went missing in #317. Probably due to the discussion about using `BUILD_TESTING` vs `DLAF_BUILD_TESTING`.


albestro:
Good catch! Thanks!

albestro (approved):

PR #315, opened 2020-12-23
Porting to HPX 1.6.0

albestro:
This PR aims at collecting all required changes for porting to HPX 1.6.0 which, AFAIK, will be released early next year.

I don't think there is any doubt about updating, since we are already looking for it for new features and improvements.

PR #326, opened 2021-02-08
HPX version compatibility in spack package

albestro:
I realized this just now that I started using our spack package for DLAF.

It may be a bit late for this, since we are close to HPX 1.6.0, iirc. I'm going to add a similar thing to #315.

### DIFF
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -26,7 +26,7 @@ class DlaFuture(CMakePackage, CudaPackage):
     depends_on("blaspp")
     depends_on("lapackpp")
     depends_on("hpx cxxstd=14 networking=none")
-    depends_on("hpx@1.5.0", when="~cuda")
+    depends_on("hpx@1.5.0:1.5.99", when="~cuda")
     depends_on("hpx@master +cuda", when="+cuda")
 
     depends_on("hpx build_type=Debug", when="build_type=Debug")
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -26,7 +26,10 @@ class DlaFuture(CMakePackage, CudaPackage):
     depends_on("blaspp")
     depends_on("lapackpp")
     depends_on("hpx cxxstd=14 networking=none")
-    depends_on("hpx@1.5", when="~cuda")
+    # it would have been "more correct" specifying @1.5 as constraint but, if clingo
+    # solver works as expected, the default solver has problems with that.
+    # Eventually, we opted for specifying the range more explicitly, so it works with both
+    depends_on("hpx@1.5.0:1.5.1", when="~cuda")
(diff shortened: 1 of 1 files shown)


spack/packages/dla-future/package.py
```diff
@@ -26,7 +26,7 @@ class DlaFuture(CMakePackage, CudaPackage):
     depends_on("blaspp")
     depends_on("lapackpp")
     depends_on("hpx cxxstd=14 networking=none")
-    depends_on("hpx@1.5.0", when="~cuda")
+    depends_on("hpx@1.5.0:1.5.99", when="~cuda")
```
msimberg, line 29: Wasn't `hpx@1.5` equivalent to `hpx@1.5.<any patch version>`? Not that we're going to have more than 100 patch versions for 1.5, or more than 1 for that matter :D

teonnik (approved):
1.5.1 instead of 1.5.99 seems better but I don't see an issue either way.

spack/packages/dla-future/package.py
albestro: yeah...you're right! I always struggle with this...thanks for pointing it out!

albestro:
@haampie with the change proposed by @msimberg, which looks absolutely the right thing to do, I'm addressing some problems.

The problem happens during the concretization, I'm going to give it a try to clingo.

However, what @teonnik suggested may be a nice solution, i.e. `@1.5.0:1.5.1`.

albestro:
> @haampie with the change proposed by @msimberg, which looks absolutely the right thing to do, I'm addressing some problems.
> 
> The problem happens during the concretization, I'm going to give it a try to clingo.

Concretization with clingo works as expected with the versioning mechanism!
+1 for clingo! @haampie

haampie:
I bumped this PR https://github.com/spack/spack/pull/20258, it's weird to not have closed-open version ranges

spack/packages/dla-future/package.py
```diff
@@ -26,7 +26,10 @@ class DlaFuture(CMakePackage, CudaPackage):
     depends_on("blaspp")
     depends_on("lapackpp")
     depends_on("hpx cxxstd=14 networking=none")
-    depends_on("hpx@1.5.0", when="~cuda")
+    # it would have been "more correct" specifying @1.5 as constraint but, if clingo
+    # solver works as expected, the default solver has problems with that.
+    # Eventually, we opted for specifying the range more explicitly, so it works with both
+    depends_on("hpx@1.5.0:1.5.1", when="~cuda")
```
haampie, line 32:
```suggestion
    depends_on("hpx@1.5.0:1.5.999", when="~cuda")
```
Just my 2 cents: it's horrible, but assuming hpx respects semver any future patch version should be ok, and whenever the open ranges PR lands in spack, it'll be easier to grep for `depends_on.*99` and fix it

PR #327, opened 2021-02-09
Fix deploy of CMake configuration files

albestro:
According to CMake documentation, `find_package` automatically looks for library configuration files in [specific places](https://cmake.org/cmake/help/latest/command/find_package.html?highlight=find_package#search-procedure).

Currently used one is not one of them, so it requires to manually specify with `DLAF_DIR` to the folder containing the `DLAFConfig.cmake`. With this change, it would be enough setting the root where the library is installed, which is quite helpful in case the library is installed in a common root (e.g. spack environment view, `/usr/local`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 
 option(BUILD_SHARED_LIBS "Build shared libraries." OFF)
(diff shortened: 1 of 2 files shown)


albestro:
bors try

albestro:
bors try-

albestro:
bors try

teonnik (approved):

PR #330, opened 2021-03-05
Extend API for in-place MPI reduction

albestro:
This PR extends a bit the MPI reduce API.

In particular, before this, there was just the "single access" signature, valid for both endpoints, resembling `MPI_Reduce` (apart from the order, arguments are the same).
```cpp
reduce(
  const IndexT_MPI rank_root,
  Communicator& communicator,
  MPI_Op reduce_op,
  const DataIn input,
  const DataOut output)
```

This has a couple of downsides:
- On the sender side, you have to specify a "fake" `DataOut`, even if it is not needed at all;
- On the receiver side, you have to specify the `rank_root`

With this PR, three functions are going to be added:
- sender endpoint: it won't be needed anymore to specify the `output` argument
`reduce(rank_root, comm, op, input)`
- receiver endpoint:
  - the classic version, but implying that if you call this function, the calling rank is the one that collects results
  `reduce(comm, op, in, out)`
  - an alternative version, similar to previous one, but it uses internally `MPI_IN_PLACE`, allowing to use the same memory for input and output
  `reduce(comm, op, inout)`

The main point of this PR is being able to reduce in-place.

In addition, a minor thing: move `IndexT_MPI` to communicator.

### DIFF
--- a/include/dlaf/common/data_descriptor.h
+++ b/include/dlaf/common/data_descriptor.h
@@ -123,8 +123,8 @@ struct DataDescriptor {
   }
 
   bool operator==(const DataDescriptor& rhs) const noexcept {
-    return (this == &rhs) && this->data_ == rhs.data_ && this->nblocks_ == rhs.nblocks_ &&
-           this->blocksize_ == rhs.blocksize_ && this->stride_ == rhs.stride_;
+    return this == &rhs || (this->data_ == rhs.data_ && this->nblocks_ == rhs.nblocks_ &&
+                            this->blocksize_ == rhs.blocksize_ && this->stride_ == rhs.stride_);
   }
 
   bool operator!=(const DataDescriptor& rhs) const noexcept {
--- a/include/dlaf/common/data_descriptor.h
+++ b/include/dlaf/common/data_descriptor.h
@@ -122,6 +122,15 @@ struct DataDescriptor {
     return nblocks_ == 1 || stride_ == blocksize_;
   }
 
+  bool operator==(const DataDescriptor& rhs) const noexcept {
+    return (this == &rhs) && this->data_ == rhs.data_ && this->nblocks_ == rhs.nblocks_ &&
+           this->blocksize_ == rhs.blocksize_ && this->stride_ == rhs.stride_;
+  }
+
+  bool operator!=(const DataDescriptor& rhs) const noexcept {
+    return !(*this == rhs);
+  }
+
 protected:
   T* data_;
   SizeType nblocks_;
--- a/include/dlaf/communication/sync/reduce.h
+++ b/include/dlaf/communication/sync/reduce.h
@@ -25,13 +25,12 @@ namespace sync {
 namespace internal {
 namespace reduce {
 
-/// MPI_Reduce wrapper (sender side).
+/// MPI_Reduce wrapper (collector side, i.e. send and receive).
 ///
 /// MPI Reduce(see MPI documentation for additional info).
-/// @param reduce_operation MPI_Op to perform on @p input data coming from ranks in @p communicator.
+/// @param reduce_op MPI_Op to perform on @p input data coming from ranks in @p communicator.
 template <class DataIn, class DataOut>
-void collector(Communicator& communicator, MPI_Op reduce_operation, const DataIn input,
-               const DataOut output) {
+void collector(Communicator& communicator, MPI_Op reduce_op, const DataIn input, const DataOut output) {
   using T = std::remove_const_t<typename common::data_traits<DataIn>::element_t>;
 
   // Wayout for single rank communicator, just copy data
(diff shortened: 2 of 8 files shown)


albestro:
Tests are still missing...I'll add them soon. Looking forward to a feedback about this PR.

albestro:
note to self: the same extension applies to `all_reduce`

include/dlaf/communication/sync/reduce.h
```diff
@@ -51,22 +49,60 @@ void collector(Communicator& communicator, MPI_Op reduce_operation, const DataIn
...
+/// MPI Reduce(see MPI documentation for additional info).
+/// It uses the MPI_IN_PLACE option, so the result of the reduce is stored in the same buffer
+/// used for the input, i.e. @p inout
+///
+/// It must be highlighted that if the buffer is not contiguous, it will get copied to a support
+/// buffer, and then copied back. From the user perspective it is still in-place, but in that specific
+/// case there will be internal memory copies.
+///
+/// @param reduce_op MPI_Op to perform on @p inout data coming from ranks in @p communicator,
+/// @pre @p rank_root < @p communicator.size(),
```
rasolca, line 70: As ranks are signed integers I would use `0 <= rank_root < communicator.size()`

rasolca (commented):
You are proposing:
```
1) reduce(rank, in) -> send
2) reduce(inout) -> recv
3) reduce(in, out) -> recv
4) reduce(rank, in, out) -> send/recv
```
I think that this API can be misleading ((1) is only send, while (4) can be send or recv).
Is 4) needed? If yes, why a corresponding in-place reduce has not been considered?

include/dlaf/communication/sync/reduce.h
albestro: Good point, updated!

include/dlaf/common/data_descriptor.h
```diff
@@ -122,6 +122,15 @@ struct DataDescriptor {
     return nblocks_ == 1 || stride_ == blocksize_;
   }
 
+  bool operator==(const DataDescriptor& rhs) const noexcept {
+    return (this == &rhs) && this->data_ == rhs.data_ && this->nblocks_ == rhs.nblocks_ &&
+           this->blocksize_ == rhs.blocksize_ && this->stride_ == rhs.stride_;
```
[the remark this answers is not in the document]
albestro: Good catch!
I meant a different thing and I implemented it wrongly (mistyped the second operator, it should be an OR).

It is just a way to check that they are not representing the same memory chunk, i.e. same memory with exactly the same layout. It does not check about aliasing, it looks just for a perfect match.

Please check the condition I've just changed.

albestro:
> You are proposing:
> 
> ```
> 1) reduce(rank, in) -> send
> 2) reduce(inout) -> recv
> 3) reduce(in, out) -> recv
> 4) reduce(rank, in, out) -> send/recv
> ```
> 
> I think that this API can be misleading ((1) is only send, while (4) can be send or recv).
> Is 4) needed? If yes, why a corresponding in-place reduce has not been considered?

It's a good point.

As it is now, adding a (5) "send+recv in-place" would complicate a bit things, because the function signature will look exactly the same as (1) (apart from constness, but AFAIC now, it would require a bit of template magic).

I can think of:
- removing (4), but it may require the user-code to be changed to add the simple if based on the rank (not my first option)
- renaming functions: we can rename (4) and (5) or all the others (e.g. `reduce_send` + `reduce_recv` vs `reduce_together`). With different names for "single" vs "both" categories, we can easily add also (5) (because the name avoids the overload resolution)

teonnik (approved):

albestro:
After talking offline, we opted for the following naming:

1) `reduceSend`
2) `reduceRecvInPlace`
3) `reduceRecv`
4) `reduce`
5) `reduceInPlace`

 And, for what concerns the `AllReduce`

1) `allReduce` (previously known as `all_reduce`, changed to match the coding style guidelines)
2) `allReduceInPlace`

I still have a couple of things to do before opening the PR for the review:
1) add tests for the new variants, and specifically for the in-place ones
2) think about the check of "overlapping" buffers (not really an overlap, just a more loose check of perfect overlapping)

For 2) I would go for adding the check on the buffers really used by the MPI function and not on buffers passed by the user. Indeed, since the wrappers may create a contiguous copy of the data, it looks superfluous to check the user buffers  when intermediate buffers are used internally.

What do you think about this? @teonnik and @rasolca

albestro:
bors try

albestro:
bors try

albestro:
Re-ran manually the problematic jobs, got all green

rasolca (approved):

albestro:
bors try

Sely85 (approved):
Looks good to me. New names depict functions very well.

PR #331, opened 2021-03-12
computeTFactor: API change

albestro:
After a discussion with @rasolca, this PR proposes two major API changes:
- `taus` are passed as `shared_future<vector<T>>` and not as `vector<shared_future<T>>`;
- pass a `future<Tile>` instead of a `Matrix` as storage for the output

The former suggests better the idea that all the `taus` for this panel are considered together instead as a results of single columns. This reduces the atomicity of the operation, but the granularity at column level is not interesting because it leads to numerous "microtasks" (not advisable due to overhead and management by the scheduling).

The latter allows to use any Tile (of the right size) from whatever Matrix and in the preferred position (not strictly the top-left one). Since there is no more the Matrix that manages the "pipelining" of the accesses to the tile, it required to do it inside the algorithm (keeping/returning the future<Tile> from tasks).

### DIFF
--- a/include/dlaf/factorization/qr.h
+++ b/include/dlaf/factorization/qr.h
@@ -25,9 +25,10 @@ namespace internal {
 // SIAM Journal on Scientific and Statistical Computing. 10. 10.1137/0910005.
 template <Backend backend, Device device, class T>
 void computeTFactor(const SizeType k, Matrix<const T, device>& v, const GlobalTileIndex v_start,
-                    hpx::shared_future<common::internal::vector<T>> taus, Matrix<T, device>& t,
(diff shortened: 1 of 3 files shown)


Sely85 (approved):
Nice work! Look good to me! 

> taus are passed as shared_future<vector<T>> and not as vector<shared_future<T>>;

This way is actually more clear.

> pass a future<Tile> instead of a Matrix as storage for the output.

This is very useful!

albestro:
note-to-self: I have to change the documentation of the function

rasolca:
bors try

rasolca (approved):

PR #333, opened 2021-03-15
Refactor Cholesky factorization with Panel

albestro:
Cholesky factorization is a nice use-case for the `Panel` in #309.

The purpose of this PR is testing `Panel` and showing its usage in an algorithm, with the aim of improving its API.

In addition to the refactoring of the cholesky algorithm making use of the `Panel`, at the moment of writing there is also a very basic additional proposal whose name is `RoundRobin`, that we may want to consider separately.

Eventually, there is also a minor bug fix in `miniapp_cholesky`.


include/dlaf/factorization/cholesky/mc.h
```diff
@@ -126,74 +141,98 @@ void Cholesky<Backend::MC, Device::CPU, T>::call_L(comm::CommunicatorGrid grid,
...
-    // Iterate over the trailing matrix
-    for (SizeType j = k + 1; j < nrtile; ++j) {
-      GlobalTileIndex jj_idx(j, j);
-      comm::Index2D jj_rank = mat_a.rankGlobalTile(jj_idx);
+    // row panel has been used for temporary storage of diagonal panel for column update
+    panel_col_t.reset();
 
-      if (this_rank.col() != jj_rank.col())
+    // TODO skip last step tile
+    broadcast(executor_mpi, kk_rank.col(), panel_col, panel_col_t, mpi_task_chain);
```
albestro, line 204: This is a point we may want to discuss.

test/unit/factorization/mc/test_cholesky.cpp
```diff
@@ -145,10 +145,6 @@ void testDistCholesky(CholeskyDistributedTest<TypeParam>& test) {
   }
 }
 
-TYPED_TEST(CholeskyDistributedTest, DistCholesky_Yielding) {
```
albestro, line 148: @teonnik am I wrong saying it is duplicated? did am I missing something here?

teonnik: Yes, you are right, one of them should be removed!

include/dlaf/factorization/cholesky/mc.h
albestro: Regarding this, we may want to open an issue about the limitation, so that we can address this in another PR and keep it as it is for the moment (at each iteration there will be a superfluous communication).

albestro:
Discussing with @rasolca a couple of improvements have been identified. I'm going to work on them soon.

teonnik (changes requested):
LGTM overall, some small remarks!

The broadcast methods in `panel.h` should probably be moved somewhere under the `communication` folder.

include/dlaf/common/round_robin.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <vector>
+
+namespace dlaf {
+namespace common {
+
+template <class T>
+struct RoundRobin {
+  template <class... Args>
+  RoundRobin(std::size_t n, Args... args) : curr_index_(0) {
```
albestro, line 20: @teonnik I also switched to passing `args` by value instead of universal reference since we are going to use it multiple times (for constructing the n resources) and it may be problematic if we pass in a movable argument.

teonnik: I would also add a `std::move` to potentially help the compiler a little bit : `pool_.emplace_back(std::move(args)...);`

albestro:
@rasolca IIRC in our discussion the points were:
- use the panel broadcasting facilities also for the communication of the diagonal (by shrinking the panelT to a single tile)
- fix the size of panels to work with the trailing matrix (skip the `kk` tile that it was communicated without any reason)

I think I addressed these two in the last commit, please give it a look.

About other points still open/to address (possible in another PR, or partially in this one):
- panel `setRange` vs `setTile` / `reset` (e.g. what should happen if we set an external tile and then we mask it out from the active panel range). I think this may deserve a dedicated discussion before deciding/fixing/improving it
- improve `broadcast` to operate just on a subset of indices of the panel (allowing to solve the TODO "skip last tile communication")
- move Panel communication function to the `communication` folder (as also @teonnik suggested)

rasolca (approved):

rasolca:
> About other points still open/to address (possible in another PR, or partially in this one):
> 
>     * panel `setRange` vs `setTile` / `reset` (e.g. what should happen if we set an external tile and then we mask it out from the active panel range). I think this may deserve a dedicated discussion before deciding/fixing/improving it

I would use the safest approach for now: setRange can be called only after construction/reset before setting or accessing any tile.
 
>     * improve `broadcast` to operate just on a subset of indices of the panel (allowing to solve the TODO "skip last tile communication")

another PR.

>     * move Panel communication function to the `communication` folder (as also @teonnik suggested)

I would do it for this PR.

include/dlaf/common/round_robin.h
msimberg: Note that you can still take them by universal reference, as long as you don't forward them in the `emplace_back` call.

albestro:
> > About other points still open/to address (possible in another PR, or partially in this one):
> > ```
> > * panel `setRange` vs `setTile` / `reset` (e.g. what should happen if we set an external tile and then we mask it out from the active panel range). I think this may deserve a dedicated discussion before deciding/fixing/improving it
> > ```
> 
> I would use the safest approach for now: setRange can be called only after construction/reset before setting or accessing any tile.

Please check the newly updated documentation for the three member functions.

> > ```
> > * improve `broadcast` to operate just on a subset of indices of the panel (allowing to solve the TODO "skip last tile communication")
> > ```
> 
> another PR.

Fully agree, I will open an issue about it before merging this PR.

> > ```
> > * move Panel communication function to the `communication` folder (as also @teonnik suggested)
> > ```
> 
> I would do it for this PR.

For the moment I moved out the broadcast functions to a separate file under `communication`, and I also isolated the specific unit tests. I'm going to also change the namespace to `dlaf::comm` instead of `dlaf::matrix` for `broadcast` functions. Do you agree?

@rasolca @teonnik

rasolca (commented):
Please use a consistent name. Currently:
header: panel_bcast.h
test: test_broadcast_panel.cpp

rasolca (approved):

rasolca:
bors try

teonnik (approved):

Issue #332, opened 2021-03-12
Implicit synchronization barrier in Matrix::~Matrix d'tor


albestro:
Discussing with @rasolca we started evaluating to change the behavior of `dlaf::Matrix` d'tor.

Currently, `Matrix::~Matrix()` creates a synchronization barrier that waits for all operations scheduled on all its tiles, and this is a clear limit.

Let's think about a simple example:

```cpp
void ONE_SEC_op(hpx::future<Tile>) {
    std::this_thread::sleep_for(std::chrono::seconds(1));
}

void TEN_SEC_op(hpx::future<Tile>) {
    std::this_thread::sleep_for(std::chrono::seconds(10));
}

void algorithm(Matrix& a) {
  Matrix b;
  for (auto i = 0 ; i < n; ++i) {
    Matrix c;

    a({i, i}).then(ONE_SEC_op);
    b({i, i}).then(ONE_SEC_op);
    c({i, i}).then(TEN_SEC_op);
  }
}
```

In this case, the algorithm is scheduling independent operations on `a`, `b` and `c` (diagonal) tiles.
Something like this:

```
a0 -> a1 -> ... -> an
b0 -> b1 -> ... -> bn
c0 -> c1 -> ... -> cn
```

But, since `c` is defined inside the for-loop scope, it synchronizes at the end of each iteration. This means that each iteration will have a synchronization barrier, because the scheduled operation on `c` tile has to finish before "unlocking" the Matrix d'tor. It will look like this:

```
a0 |               | -> a1 |               | -> an |                             | an is ready!
b0 |               | -> b1 |               | -> bn |                             | bn is ready!
c0 | run(c0)=10sec | -> c1 | run(c1)=10sec | -> c2 | ... | -> cn | run(cn)=10sec |
```

It should be clear how it limits the efficiency of execution, in fact it blocks the scheduling of next iterations ops limiting parallellization.

So, what if we remove the barrier on exit?

```
a0 -> a1 -> ... -> an // ideally an can be ready after n seconds, instead of 10*n seconds
b0 -> b1 -> ... -> bn // ideally bn can be ready after n seconds, instead of 10*n seconds
c0 -> c1 -> ... -> cn
```

## Workaround the barrier

[...]

PR #270, opened 2020-09-30
algorithm: reduction to band

albestro:
This PR aims at introducing the algorithm for reduction to band matrix, which is part of the Eigensolver we target to implement.

After a period working off-the-radar, now there is a quite stable solution from which it is possible to work on and refine.

TODO:
- [x] Documentation
- [x] Miniapp
  - [x] Compute FLOPs
  - [x] What about the check option? Do we want to delay it to a next PR?
- [x] Test `Communicator::clone()`

Dependencies:
- [x] #334
- [x] #388
- [x] #339 (for `trmm` wrappers)
- [x] #402

PR #335, opened 2021-03-17
Select a range of tiles (for creating a barrier)

albestro:
This is a proposal for adding a way to get a set of tiles from a Matrix, in readwrite or readonly mode.

In particular, two functions are proposed, `select` and `select_read`, which return a vector of futures or shared_futures of Tile respectively.

With the return vector, it would be possible to do something like

```cpp
Matrix m;

// create a barrier
hpx::when_all(select_read(m, iterate_range2d(m.nrTiles())).get();

// select a range of tiles, e.g. a panel, and then operate on them
hpx::dataflow(do_something_on_panel, select(m, iterate_range2d({1, 1}, {2, 3}));
```

These extension to the API could represent an interesting building block for #332, but it can play an important role in algorithms in order to avoid problems like micro-tasking.

Just for giving a bit of context about latter case with an example, in `redutionToBand` algorithm it is necessary to work with a single matrix column spread across multiple tiles (and the next column has to wait for the first column result). The micro-tasking problem arises due to code like this

```cpp
for (auto index : panel_range) {
 ...
 hpx::dataflow(work_on_single_column, matrix(index));
}
```

while we can "reverse-out" that part of the algorithm with

```cpp
hpx::dataflow([](std::vector<hpx::future<Tile>> tiles){
  for (auto& tile : tile)
    work_on_single_column(tile);
}, select(matrix, panel_range));
```

Getting a single HPX task that internally has a for-loop that groups a set of small operations, instead of having a moltitude of micro-tasks which may overload the scheduler and on which the overhead may easily become unsustainable.

Clearly, this is not possible everywhere because it creates a barrier and it "locks/unlocks" dependencies all at once, but if there is anyway an implicit dependency, it helps improving the performances.

Just one last comment, I still have some concerns about the API, in particular about keeping the "index" of the `Tile` inside the for-loop, because with the current proposal we loose the information of the tile position in the matrix, and we have it just like an ordered linear sequence of tiles.

### DIFF
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -38,6 +38,7 @@ auto selectGeneric(Func&& f, common::IterableRange2D<SizeType, LocalTile_TAG> ra
   using RetT = decltype(f(LocalTileIndex{}));
 
   std::vector<RetT> tiles;
+  tiles.reserve(to_sizet(std::distance(range.begin(), range.end())));
   std::transform(range.begin(), range.end(), std::back_inserter(tiles),
                  [&](auto idx) { return f(idx); });
   return tiles;
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -28,6 +28,21 @@
 namespace dlaf {
 namespace matrix {
 
+namespace internal {
+
+/// Helper function returning a vector with the results of calling a function over a IterableRange2D
+///
+/// @param f non-void function accepting LocalTileIndex as parameter
+template <class Func>
+auto selectGeneric(Func&& f, common::IterableRange2D<SizeType, LocalTile_TAG> range) {
+  using RetT = decltype(f(LocalTileIndex{}));
+
+  std::vector<RetT> tiles;
+  std::transform(range.begin(), range.end(), std::back_inserter(tiles), [&](auto idx) { return f(idx); });
+  return tiles;
+}
+}
+
 /// A @c Matrix object represents a collection of tiles which contain all the elements of a matrix.
 ///
 /// The tiles are distributed according to a distribution (see @c Matrix::distribution()),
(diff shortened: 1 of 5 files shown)


albestro:
note-to-self: in case we make `IterableRange2D` more "standard", i.e. compatible with std iterators traits (at least a subset of them), we could use any kind of iterable thing as range (e.g. we could create a vector with a custom set of indexes and pass that).

rasolca (approved):
LGTM.
I don't think keeping the index is needed. As a vector is returned, you need more details about what it represent in the function that uses the vector, and can be done in two ways (or a mix of the two):
- precondition on what vector represent
- input parameters which describe whats in the vector.

albestro:
I will soon test this
- in #270 for improving performances by grouping together micro tasks
- and I will open a PR also about #332

include/dlaf/matrix/matrix.h
```diff
@@ -333,6 +335,38 @@ Matrix<T, device> createMatrixFromTile(const GlobalElementSize& size, const Tile
   return createMatrixFromTile<device>(size, block_size, ld_tile, tiles_per_col, comm, {0, 0}, ptr);
 }
 
+namespace internal {
+template <class RetT, class T, dlaf::Device d, class RangeTile>
+auto select_generic(Matrix<T, d>& matrix, RangeTile range,
+                    RetT (Matrix<T, d>::*f)(const typename RangeTile::index2d_t&)) {
+  std::vector<RetT> tiles;
```
msimberg, line 342: With the assumption that the range has random access iterators:
```suggestion
  std::vector<RetT> tiles;
  tiles.reserve(std::distance(std::begin(range), std::end(range));
```
Unlikely to make a big difference, but doesn't hurt.

msimberg (approved):
Looks good to me! Just one minor suggestion.

include/dlaf/matrix/matrix.h
albestro: In order to do it, I will have to check what are the requirements of `std::distance` for the iterator (they should be quite basic requirements), and it is somehow linked to what I wanted to do when i wrote this: standardize a bit our IterableRange2D iterator.

I'll check it soon 😉 

> note-to-self: in case we make `IterableRange2D` more "standard", i.e. compatible with std iterators traits (at least a subset of them), we could use any kind of iterable thing as range (e.g. we could create a vector with a custom set of indexes and pass that).

albestro:
@rasolca I gave a quick look at the chance to extend the mechanism also to ranges with `GlobalTileIndex`es.


The idea is to add an helper that converts a "global" range to a "local" range, by using the helpers from `Distribution` of the matrix passed as parameter. This conversion hides a couple of obstacles:
- `IterableRange2D` does not have the same `end` semantic for constructor and `end()` member function
e.g. a 2-by-3 range starting a `(1, 0)` is `[(1, 0), (2, 0), (1, 1), (2, 1), (1, 2), (2, 2)]`
which can be created passing `end = (3, 3)`, i.e. the bottom right limit outside the range, but then calling `end()` will return the index (1, 3), i.e. the next "potential" index in the range
- Mapping global to local via `nextLocalElement` requires that the given global index is valid. But, converting the iterator may require converting an index outside of valid ones (e.g. a range covering the full matrix has the end iterator outside the valid domain of indices).

include/dlaf/matrix/matrix.h
albestro: As said in a meeting, it would not be so straightforward to adapt `IterableRange2D` to be an [`InputIterator`](https://en.cppreference.com/w/cpp/named_req/InputIterator) (as requested by `std::distance`) and at the same time it would not be so clean to compute the distance (e.g. we should convert the result of the range boundaries to a linear index).

For reference, see also https://github.com/eth-cscs/DLA-Future/pull/335#issuecomment-822397375

albestro:
note-to-self: in `miniapp_triangular_solver` there is a function `waitall_tiles` that can be replaced with the new function. Probably it similarly applies to other miniapps.

include/dlaf/matrix/matrix.h
albestro: Following what you said, the IteratorRange2D has been "standardized" a bit and now it is possible to do what you suggested here.

albestro:
UPDATE: (FIY @rasolca)
- add `select` and `selectRead` functions that allow selecting a range of tiles from a matrix (with basic test)
- IteratorRange2D has been standardized a bit (the problem with conversion global <-> local is still there, see previous comment)

albestro:
bors try

rasolca (approved):

albestro:
bors try

rasolca (approved):

rasolca:
bors try

albestro:
bors try

PR #338, opened 2021-03-18
Add default Device and Backend enumerations

msimberg:
Simply adds `Device::Default` and `Backend::Default` to the `Device` and `Backend` enums, and they are equal to `MC/CPU` when CUDA is disabled and `GPU/GPU` when CUDA is enabled. This helps avoiding ifdefs in user code (i.e. miniapps) to choose the most "capable" backend/device that is available. I wanted to use this in #321 but I think it deserves its own PR.

It's a simple change so I may be overthinking the implications of it, but given that those enums are quite fundamental I'd be curious to hear if you think there might be another way to do this.


rasolca:
I don't see any drawback about adding the defaults.

albestro:
LGTM

msimberg:
bors try

rasolca (approved):

PR #336, opened 2021-03-17
CI: Small fix

rasolca:
- Removed Jenkins unused files from `ci` directory.
- Fixed the calculation of the value for the SLURM `-c` flag.

PR #342, opened 2021-03-24
CI: revert CI workaround

rasolca:
This PR is needed to debug the correct CI setup.

### DIFF
--- a/ci/ctest_to_gitlab.sh
+++ b/ci/ctest_to_gitlab.sh
@@ -14,13 +14,14 @@ image: $IMAGE
 
 stages:
   - allocate
-{{TEST_STAGES}}
+  - test
   - upload
   - cleanup
 
 variables:
   ALLOCATION_NAME: dlaf-ci-job-\$CI_PIPELINE_ID
   SLURM_EXCLUSIVE: ''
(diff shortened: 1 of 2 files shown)


rasolca:
bors try

haampie:
when pulling without srun:

bors try

haampie:
well... i'm calling it a slurm bug

rasolca:
bors try
Remove 5 rank tests, maybe it's causing problems as not all the cores are used.

rasolca:
bors try
This time really without 5 ranks...

rasolca:
bors try
Let's ensure that the previous success was not a lucky alignment of all the stars, planets and asteroids....

haampie:
So the idea right now is that there's a bug in slurm where it realizes not all CPUs are used, then it tried to schedule a job, but it's marked exclusive, there are not enough CPUs for it, and it fails?

rasolca:
I think that now we can keep the broadcast test with 4 ranks and go back to the original CI pipeline.

albestro (approved):
It embeds also #346, which can be closed.

Compared differences with #336, it seems all back to normal. So, LGTM.

PR #344, opened 2021-03-25
New MPI API

teonnik:
This is the new MPI API based on non-blocking MPI communication.

Three different approaches were considered as follows:
1. yielding
2. polling
3. blocking

`1.` and `2.` work with non-blocking MPI communication. The difference between the two approaches is in the internal mechanism for testing the `MPI_Request` for completion. `1.` tests each request individually (with `MPI_Test()`), `2.` places the request in a global array which is then tested in bulk (with `MPI_Testall()`) whenever a task yields or completes.

Version `3.` requires that each communicator has it's own OS thread. This is necessary to avoid interleaving MPI tasks from different communicators which may lead to deadlocks as each task blocks the OS thread on which it executes. To make this work, oversubscription (within a pool, i.e. having >1 OS thread per processing unit(PU)) was enabled and task stealing within a pool was disabled (i.e. by using a static pool). Tasks were executed on a specific thread using hints passed to `parallel_executor`. Unfortunately, this proved slow within HPX. Having >1 OS thread for the MPI pool which had 1 PU assigned resulted in slowdowns for all approaches. That is why I removed it from the current version of the PR, it can be found in past commits (in `hints.h`).

The MPI executor implemented in this PR supports both the `async()` and `dataflow()` HPX APIs. The APIs is such that the last parameter of any function scheduled using the executor is required to be `MPI_Request* req_ptr` (see `kernels.h`).

`common::PromiseGuard<Communicator> pcomm` is necessary to supply the communicator AND to impose order between tasks launched on that communicator. The purpose of this object hasn't changed. 

Attempts to manage MPI task ordering within the MPI executor proved to be challenging and fragile due to internal details of the dataflow API. The `dataflow_finalize()` function implemented in the executor is often executed in a continuation which requires that the executor is copied and complicates the lifetime-management of the ordering mechanism within the executor considerably.

The switch between approaches `1.` and `2.` was initially implemented at compile-time. A few days ago I changed the switch to a runtime parameter. Measurements showed no performance difference either way. One reason for the switch was to reduce the number of explicit template instantiations that were needed for each algorithm. Another is the avoid having to introduce additional struct within `api.h` to manage an extra template parameter.

Request for help!!

- The polling approach `2.` executes fine but hangs at the very end somewhere in `hpx::init()` but after `hpx::finalize()`! This is why I have temporarily commented out `polling` approach in `test_cholesky.cpp`.  If someone can help identify what the issue is, that would be awesome!

- !!! Please, try to run some benchmarks with this PR (for the approach that works - `1.`)! My own experiments show that performance in certain configuration is slightly better and in certain it is slightly worse. Try to see if you can find an issue! We should try to get this merged ASAP!


include/dlaf/communication/executor.h
```diff
@@ -0,0 +1,162 @@
...
+    auto fn = [mech = mech_, frame_p = std::move(frame_p), f = std::forward<F>(f),
+               args = std::forward<TupleArgs>(args)]() mutable {
+      MPI_Request req;
+      auto all_args = detail::make_mpi_tuple(std::move(args), &req);
+      using result_t = decltype(hpx::util::invoke_fused(f, all_args));
+      detail::invoke_fused_wrapper<result_t> wrapper(std::move(f), std::move(all_args));
+      detail::handle_request(mech, req);
+      frame_p->set_data(wrapper.dataflow_return());
+    };
+    hpx::async(ex_, std::move(fn));
```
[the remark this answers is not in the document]
msimberg: Have I understood it correctly that you currently do the polling on the default pool? Also, does the pipeline wrapper have to ensure that there is only ever a single MPI request in flight, or that they are only submitted in order? Is the wrapper in args, and in that case the continuation is triggered only once `fn` has finished?

For the polling case, it would be slightly preferable to submit the MPI operation when entering `dataflow_finalize` and attach a continuation to the future from `hpx::mpi::experimental::get_future` which calls `set_data` which would avoid a probable yield when calling `get` on the future. However, this is just a potential optimization.

I don't see anything obvious right now that would cause it to deadlock but I'll have a closer look still. When you say it hangs after `hpx::finalize`, does that mean you see `test_cholesky` actually successfully run all the tests, but it hangs after that?

include/dlaf/communication/kernels.h
```diff
@@ -0,0 +1,58 @@
...
+  using ConstTile_t = matrix::Tile<const T, Device::CPU>;
+  using MemView_t = memory::MemoryView<T, Device::CPU>;
+
+  MemView_t mem_view(util::size_t::mul(tile_size.rows(), tile_size.cols()));
+  Tile_t tile(tile_size, std::move(mem_view), tile_size.rows());
+
+  auto msg = comm::make_message(common::make_data(tile));
+  MPI_Ibcast(msg.data(), msg.count(), msg.mpi_type(), root_rank, pcomm.ref(), req);
+  return ConstTile_t(std::move(tile));
+}
```
msimberg, line 55: Should we make these callable objects as well? What about the naming, in the sync versions there was `recvAllocTile` and `recvTile`, do we want the same here?

msimberg (changes requested):
Thanks @teonnik for putting this up, looks good! Mostly minor style comments.

include/dlaf/factorization/cholesky/mc.h
```diff
@@ -39,43 +43,75 @@ namespace internal {
...
                      hpx::future<matrix::Tile<T, Device::CPU>> matrix_tile) {
-  hpx::dataflow(executor_hp, hpx::util::unwrapping(tile::potrf<T, Device::CPU>), blas::Uplo::Lower,
-                std::move(matrix_tile));
+  auto kern_f = hpx::util::
+      annotated_function([](hpx::future<matrix::Tile<T, Device::CPU>>
+                                mtile) { tile::potrf<T, Device::CPU>(blas::Uplo::Lower, mtile.get()); },
+                         "potrf");
+  hpx::dataflow(executor_hp, std::move(kern_f), std::move(matrix_tile));
+  // hpx::dataflow(executor_hp, hpx::util::unwrapping(tile::potrf<T, Device::CPU>), blas::Uplo::Lower,
+  //                    std::move(matrix_tile));
```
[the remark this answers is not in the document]
albestro: I agree, I have in my mind backlog to work on this. I'm looking for information before proceeding.

Just for reference, see this https://github.com/STEllAR-GROUP/hpx/discussions/5263

spack/packages/dla-future/package.py
```diff
@@ -31,9 +31,7 @@ class DlaFuture(CMakePackage, CudaPackage):
 
     depends_on("hpx build_type=Debug", when="build_type=Debug")
     depends_on("hpx build_type=Release", when="build_type=Release")
-    depends_on(
-        "hpx build_type=RelWithDebInfo", when="build_type=RelWithDebInfo"
-    )
+    depends_on("hpx build_type=RelWithDebInfo", when="build_type=RelWithDebInfo")
```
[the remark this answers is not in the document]
albestro: With `spack`, we should consider using the official one: `flake8`. See their contribution guide.

miniapp/miniapp_triangular_solver.cpp
```diff
@@ -279,4 +288,18 @@ linear_system_t sampleLeftTr(blas::Uplo uplo, blas::Op op, blas::Diag diag, T al
...
+
+MPIMech parse_mech(const std::string& mech) {
+  if (mech == "yielding") {
+    return MPIMech::Yielding;
+  }
+  else if (mech == "polling") {
+    return MPIMech::Polling;
+  }
+  else {
+    DLAF_ASSERT(false, mech);
```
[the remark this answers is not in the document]
albestro: Yes, it's customizable via CMake options, but by default the `DLAF_ASSERT`s are always on

https://github.com/eth-cscs/DLA-Future/blob/a911db696d82b51e7b1d9b4aa498724b2e16a715/src/CMakeLists.txt#L42-L50

include/dlaf/communication/kernels.h
```diff
@@ -0,0 +1,58 @@
...
+  auto msg = comm::make_message(common::make_data(tile));
+  MPI_Ibcast(const_cast<T*>(msg.data()), msg.count(), msg.mpi_type(), pcomm.ref().rank(), pcomm.ref(),
+             req);
+}
+
+// Non-blocking receiver broadcast
+template <class T>
+matrix::Tile<const T, Device::CPU> bcast_recv(TileElementSize tile_size, int root_rank,
+                                              common::PromiseGuard<Communicator> pcomm,
+                                              MPI_Request* req) {
```
albestro, line 44: During your work you deleted the `bcast_recv` using an externally allocated tile.
Do you see any problem adding it back?

I'm asking because I think that this solution with the embedded allocation is going to be dropped soon (e.g. with the `Panel` where the matrix will be re-used and not allocated at each communication).

include/dlaf/factorization/cholesky/mc.h
```diff
@@ -114,19 +151,26 @@ void Cholesky<Backend::MC, Device::CPU, T>::call_L(Matrix<T, Device::CPU>& mat_a
...
   using ConstTileType = typename Matrix<T, Device::CPU>::ConstTileType;
+  using hpx::util::unwrapping;
 
   auto executor_hp = dlaf::getHpExecutor<Backend::MC>();
   auto executor_np = dlaf::getNpExecutor<Backend::MC>();
-  auto executor_mpi = dlaf::getMPIExecutor<Backend::MC>();
 
-  common::Pipeline<comm::CommunicatorGrid> mpi_task_chain(std::move(grid));
+  // Set up MPI executor pipelines
+  std::string mpi_pool = (hpx::resource::pool_exists("mpi")) ? "mpi" : "default";
```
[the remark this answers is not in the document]
albestro: I comment here because it is somehow related.

I was wondering if passing the `MPIMech` to all algorithms is needed. In particular:
- is it possible to mix different `MPIMech`? If not, passing it to different algorithms would not make sense to me.
- I think one reason you have to pass it as algorithm argument is because you have to initialize the `Executor`, and here there might be the link with @msimberg comment: would it be possible to "centralize" `Executor`s definition and then re-use them as other (and as @msimberg was suggesting)?

include/dlaf/matrix/distribution.h
```diff
@@ -292,6 +292,12 @@ class Distribution {
     return util::matrix::tileElementFromElement(global_element, block_size_.get<rc>());
   }
 
+  /// Returns the size of the Tile with global index @p index.
+  TileElementSize tileSize(const GlobalTileIndex& index) const noexcept {
```
albestro, line 296: I strongly agree with this move from `MatrixBase`! 😁

include/dlaf/communication/executor.h
```diff
@@ -0,0 +1,162 @@
...
+      int flag;
+      mpi_invoke(MPI_Test, &req, &flag, MPI_STATUS_IGNORE);
+      return flag == 0;
+    });
+  }
+  else if (mech == MPIMech::Polling) {
+    hpx::mpi::experimental::get_future(req).get();
+  }
+  else {
+    std::cout << "UNIMPLEMENTED!" << std::endl;
```
albestro, line 56: very minor: you can use `DLAF_UNIMPLEMENTED`

albestro (changes requested):
First general comments, I still have to dive a bit more in order to fully understand the `Executor` and the new HPX experimental features.

include/dlaf/communication/executor.h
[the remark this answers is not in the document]
teonnik: > Have I understood it correctly that you currently do the polling on the default pool? 

Polling can be done both on the default and the mpi pools. Unit tests, do it on the default pool. Miniapps do it on the mpi pool currently, but I will implement a command line option allowing to switch between the default and mpi pools.

> Also, does the pipeline wrapper have to ensure that there is only ever a single MPI request in flight, or that they are only submitted in order?

Only that they are submitted in order on the same communicator.

> Is the wrapper in args, and in that case the continuation is triggered only once fn has finished?

Yes, the pipeline wrapper is in `args`. The underlying promise is set when `fn` finishes using RAII (if that is what you are asking).

> For the polling case, it would be slightly preferable to submit the MPI operation when entering dataflow_finalize and attach a continuation to the future from hpx::mpi::experimental::get_future which calls set_data which would avoid a probable yield when calling get on the future. However, this is just a potential optimization.

I agree that this is the case for the non-blocking MPI approach (which is what we are left with in this PR). However, for blocking MPI, the call can't happen in `dataflow_finalize()` and has to be using the executor `ex_`, because it will block the underlying OS thread on which `dataflow_finalize()` is scheduled.

> I don't see anything obvious right now that would cause it to deadlock but I'll have a closer look still. When you say it hangs after hpx::finalize, does that mean you see test_cholesky actually successfully run all the tests, but it hangs after that?

Thanks for helping me on this one. The issue was that polling was disabled before all outstanding non-blocking communications were done.

spack/packages/dla-future/package.py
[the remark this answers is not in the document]
teonnik: @msimberg the formatter I am using is [black](https://github.com/psf/black), it is flake8 compatible.

miniapp/miniapp_triangular_solver.cpp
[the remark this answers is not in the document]
teonnik: This is indeed somewhat hacky. I changed to print a more descriptive message and terminate instead.

include/dlaf/communication/executor.h
msimberg: Very minor tangent: would it be useful to have a `DLAF_UNREACHABLE` as well?

include/dlaf/factorization/cholesky/mc.h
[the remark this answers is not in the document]
teonnik: @albestro Yes, good point!

> is it possible to mix different MPIMech? If not, passing it to different algorithms would not make sense to me.

This is technically possible but I don't see why someone might one to do it.

I can make `MPIMech` initialized once and remove it from all algorithms. I will check how this can be integrated with Mikaels initialization work.

include/dlaf/communication/kernels.h
albestro: @teonnik in the last meeting we had, I was talking about this. Do you have any insight to share about it?

include/dlaf/communication/kernels.h
teonnik: Yes, I will change the names, but why do we need to make these callable objects? Is it because of the GPU kernels?

msimberg: Not a must (I think) to make them callable objects, but it saves having to specify the (some) template parameters. It's a small simplification for the caller that I personally prefer, but I can see how it's nice not to have to make them callable objects as well. I'm indifferent on which way we go. The only reason I ask is because they are now callable objects on master.

src/init.cpp
```diff
@@ -108,22 +115,45 @@ void Init<Backend::GPU>::finalize() {
...
 #endif
 
-template <typename T>
-void updateConfigurationValue(hpx::program_options::variables_map const& vm, T& var,
-                              std::string const& env_var, std::string const& cmdline_option);
+template <>
+void Init<Backend::MC>::initialize(configuration const& cfg) {
+  // TODO: Consider disabling polling in finalize()
+  if (cfg.mpi_mech == comm::MPIMech::Polling) {
+    hpx::mpi::experimental::init(false, cfg.mpi_pool);
```
msimberg, line 122: I suppose this should be enabled independently of the backend.

src/init.cpp
```diff
@@ -132,6 +162,21 @@ void updateConfiguration(hpx::program_options::variables_map const& vm, configur
...
+  if (mpi_mech_str == "yielding") {
+    cfg.mpi_mech = comm::MPIMech::Yielding;
+  }
+  else if (mpi_mech_str == "polling") {
+    cfg.mpi_mech = comm::MPIMech::Polling;
+  }
+  else if (mpi_mech_str != "") {
+    std::cout << "Unknown value for --mech=" << mpi_mech_str << "!" << std::endl;
+    std::terminate();
+  }
```
msimberg, line 177: Could this be moved into `convertFromString` instead?

src/init.cpp
```diff
@@ -149,6 +194,10 @@ hpx::program_options::options_description getOptionsDescription() {
                      "Number of normal priority CUDA streams per worker thread");
   desc.add_options()("dlaf:num-hp-cuda-streams-per-thread", hpx::program_options::value<std::size_t>(),
                      "Number of high priority CUDA streams per worker thread");
+  desc.add_options()("dlaf:mpi-mech",
+                     hpx::program_options::value<std::string>()->default_value("yielding"),
```
msimberg, line 198: Do you have benchmarks comparing yielding and polling now? I'd expect polling to never be worse than yielding.

src/init.cpp
```diff
@@ -17,13 +20,20 @@
...
-  os << "  num_hp_cuda_streams_per_thread = " << cfg.num_hp_cuda_streams_per_thread;
-
+  os << "  num_hp_cuda_streams_per_thread = " << cfg.num_hp_cuda_streams_per_thread << std::endl;
+  os << "  mpi_pool = " << cfg.mpi_pool << std::endl;
+  if (cfg.mpi_mech == comm::MPIMech::Polling) {
+    os << "  mpi_mech = polling" << std::endl;
+  }
+  else if (cfg.mpi_mech == comm::MPIMech::Yielding) {
+    os << "  mpi_mech = yielding" << std::endl;
+  }
```
msimberg, line 36: This could probably be put into a `operator<<` for `MPIMech`.

src/init.cpp
teonnik: We don't need it for `MPIMech::Yielding`. Why would you prefer having it enabled unconditionally?

msimberg: Ugh, sorry. Ignore my comment. My mind was telling me that it won't be enabled for the CUDA case, but of course it will since the `MC` backend is always enabled.

src/init.cpp
teonnik: I had some old benchmarks that showed that they are approximately equivalent in the case of Cholesky (with perhaps a slight advantage for the polling approach). My intention was to have both for now until we can test more algorithms and then have a more clear picture if one approach is superior to the other.

src/init.cpp
teonnik: Yes, I refactored this part of the code. (I had to introduce another function but it's cleaner)

teonnik:
bors try

miniapp/miniapp_triangular_solver.cpp
```diff
@@ -109,7 +110,7 @@ int hpx_main(hpx::program_options::variables_map& vm) {
     sync_barrier();
 
     dlaf::common::Timer<> timeit;
-    dlaf::solver::triangular<Backend::MC>(comm_grid, side, uplo, op, diag, alpha, a, b);
+    dlaf::solver::triangular<Backend::MC, Device::CPU, T>(comm_grid, side, uplo, op, diag, alpha, a, b);
```
msimberg, line 113: Just curious: are the added template parameters required or just for explicitness?

msimberg (approved):
Looks good to me!

miniapp/miniapp_triangular_solver.cpp
teonnik: For explicitness, but I can revert if you prefer.

msimberg: No preference, I'm happy to keep your change.

rasolca:
bors try

rasolca:
bors try

include/dlaf/factorization/cholesky.h
```diff
@@ -11,6 +11,8 @@
 
 #include <blas.hh>
 #include "dlaf/communication/communicator_grid.h"
+#include "dlaf/communication/mech.h"
```
albestro, line 14: minor: you don't need it anymore, right? it applies also in other parts.

include/dlaf/factorization/cholesky/mc.h
```diff
@@ -115,18 +121,24 @@ void Cholesky<Backend::MC, Device::CPU, T>::call_L(Matrix<T, Device::CPU>& mat_a
...
+  comm::Executor executor_mpi_row{};
+  common::Pipeline<comm::Communicator> mpi_col_task_chain(grid.colCommunicator());
+  common::Pipeline<comm::Communicator> mpi_row_task_chain(grid.rowCommunicator());
 
-  const comm::Index2D this_rank = grid.rank();
+  const matrix::Distribution& distr = mat_a.distribution();
 
-  matrix::Distribution const& distr = mat_a.distribution();
   SizeType nrtile = mat_a.nrTiles().cols();
+  comm::Index2D this_rank = grid.rank();
```
albestro, line 141: `const` has been lost during changes

miniapp/miniapp_triangular_solver.cpp
```diff
@@ -40,6 +41,7 @@ using dlaf::TileElementSize;
 using dlaf::comm::Communicator;
 using dlaf::comm::CommunicatorGrid;
 using dlaf::common::Ordering;
+using dlaf::comm::MPIMech;
```
albestro, line 44: This is not needed as well.

albestro (commented):
Great work!

Just minor comments left.

I still have one question: what about the lifetime problem with kernels?

rasolca:
bors try

teonnik:
@albestro 

> I still have one question: what about the lifetime problem with kernels?

Let's resolve it in a different PR. It's not urgent (I think).

albestro (approved):

PR #346, opened 2021-03-31
Minor change: remove test with 5 RANKS

albestro:
FIY @rasolca


albestro:
It has been embedded in #342.

It may be the underlying reason that triggers a bug in slurm.

PR #348, opened 2021-04-06
Fix Compiler ID in CMake (relevant for warnings)

albestro:
I just realized that I used the wrong compiler ID as a trigger for a couple of warnings...

PR #349, opened 2021-04-07
Fix warnings on GCC (after CMake bug fix)

albestro:
This PR is just a side-PR of #348.

It is going to be merged into that one and then they will be merged on master all together.

### DIFF
--- a/include/dlaf/matrix/copy_tile.h
+++ b/include/dlaf/matrix/copy_tile.h
@@ -41,20 +41,24 @@ template <typename T>
 struct CopyTile<T, Device::CPU, Device::GPU> {
   static void call(const matrix::Tile<const T, Device::CPU>& source,
                    const matrix::Tile<T, Device::GPU>& destination) {
-    SizeType m = source.size().rows();
-    SizeType n = source.size().cols();
+    const std::size_t m = to_sizet(source.size().rows());
+    const std::size_t n = to_sizet(source.size().cols());
+    const std::size_t ld_source = to_sizet(source.ld());
+    const std::size_t ld_destination = to_sizet(destination.ld());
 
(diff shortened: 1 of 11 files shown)


albestro:
There is still a problem to be solved in lapack tests.

I'm going to push a commit with a fix but it requires a better definition of the maximum error allowed, otherwise the test fails.

albestro:
bors try

albestro:
bors try

albestro:
bors try

albestro:
The problem there was in the previous failed build from bors, it seems was due to GCC 7.

The problem was
```
/DLA-Future/include/dlaf/types.h:123:54: error: conversion to 'long unsigned int' from 'const int' may change the sign of the result [-Werror=sign-conversion]
   DLAF_ASSERT_MODERATE(std::numeric_limits<U>::max() >= static_cast<std::size_t>(signed_value),
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

After rebasing it on the new master which uses GCC 9, it went away (and I don't see any related change).

albestro:
bors try
(after switching from functional cast to static_cast)

albestro:
bors try

(let's try again)

rasolca (approved):

albestro:
I will merge it by myself (it will merged in another PR, not on master), just waiting if @msimberg would like to give a quick feedback. 😉

PR #355, opened 2021-04-19
Remove implicit synchronization in Matrix d'tor

### LINKED ISSUE #332: Implicit synchronization barrier in Matrix::~Matrix d'tor

albestro:
Close #332 

This PR is a breaking change for codes dealing with `Matrix`es built from user provided memory, since a manual synchronization barrier would be needed from now on.

On the opposite, it represents an interesting improvement for `Matrix`es that internally allocates memory, since it removes implicit synchronization barriers that could limit performance and flexibility.

See issue #332 for additional details.

### DIFF
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -175,6 +175,22 @@ public:
     return read(distribution().localTileIndex(index));
   }
 
+  /// Synchronization barrier for all tiles in the matrix
+  void syncAll() noexcept {
+    // Note:
+    // Using a readwrite access to the tile ensures that the access is exclusive and not shared
+    // among multiple tasks.
+
+    auto readwrite_f = [this](const LocalTileIndex& index) {
+      std::size_t i = tileLinearIndex(index);
+      return this->tile_managers_[i].getRWTileFuture();
+    };
+
+    const auto range_local = common::iterate_range2d(distribution().localNrTiles());
+    auto all_local_tiles_rw = internal::selectGeneric(readwrite_f, range_local);
+    hpx::when_all(std::move(all_local_tiles_rw)).get();
+  }
+
 protected:
   Matrix(Distribution distribution) : internal::MatrixBase{std::move(distribution)} {}
 
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -176,7 +176,7 @@ public:
   }
 
   /// Synchronization barrier for all local tiles in the matrix
-  void syncAll() noexcept;
+  void waitLocalTiles() noexcept;
 
 protected:
   Matrix(Distribution distribution) : internal::MatrixBase{std::move(distribution)} {}
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -175,21 +175,8 @@ public:
     return read(distribution().localTileIndex(index));
   }
 
-  /// Synchronization barrier for all tiles in the matrix
-  void syncAll() noexcept {
-    // Note:
-    // Using a readwrite access to the tile ensures that the access is exclusive and not shared
-    // among multiple tasks.
-
-    auto readwrite_f = [this](const LocalTileIndex& index) {
-      const auto i = tileLinearIndex(index);
-      return this->tile_managers_[i].getRWTileFuture();
-    };
-
-    const auto range_local = common::iterate_range2d(distribution().localNrTiles());
-    auto all_local_tiles_rw = internal::selectGeneric(readwrite_f, range_local);
-    hpx::when_all(std::move(all_local_tiles_rw)).get();
-  }
+  /// Synchronization barrier for all local tiles in the matrix
+  void syncAll() noexcept;
 
 protected:
   Matrix(Distribution distribution) : internal::MatrixBase{std::move(distribution)} {}
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -157,8 +157,6 @@ public:
   Matrix(const Matrix& rhs) = delete;
   Matrix(Matrix&& rhs) = default;
 
-  virtual ~Matrix();
-
   Matrix& operator=(const Matrix& rhs) = delete;
   Matrix& operator=(Matrix&& rhs) = default;
 
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -182,7 +182,7 @@ public:
     // among multiple tasks.
 
     auto readwrite_f = [this](const LocalTileIndex& index) {
-      std::size_t i = tileLinearIndex(index);
+      const auto i = tileLinearIndex(index);
       return this->tile_managers_[i].getRWTileFuture();
     };
 
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -176,6 +176,9 @@ public:
   }
 
   /// Synchronization barrier for all local tiles in the matrix
+  ///
+  /// This blocking call does not return until all operations, i.e. both RO and RW,
+  /// involving any of the locally available tiles are completed.
   void waitLocalTiles() noexcept;
 
 protected:
--- a/include/dlaf/matrix/matrix_const.tpp
+++ b/include/dlaf/matrix/matrix_const.tpp
@@ -38,6 +38,21 @@ hpx::shared_future<Tile<const T, device>> Matrix<const T, device>::read(
   return tile_managers_[i].getReadTileSharedFuture();
 }
 
+void Matrix<const T, device>::syncAll() noexcept {
+  // Note:
+  // Using a readwrite access to the tile ensures that the access is exclusive and not shared
+  // among multiple tasks.
+
+  auto readwrite_f = [this](const LocalTileIndex& index) {
+    const auto i = tileLinearIndex(index);
+    return this->tile_managers_[i].getRWTileFuture();
+  };
+
+  const auto range_local = common::iterate_range2d(distribution().localNrTiles());
+  auto all_local_tiles_rw = internal::selectGeneric(readwrite_f, range_local);
+  hpx::when_all(std::move(all_local_tiles_rw)).get();
+}
+
 template <class T, Device device>
 void Matrix<const T, device>::setUpTiles(const memory::MemoryView<ElementType, device>& mem,
                                          const LayoutInfo& layout) noexcept {
--- a/include/dlaf/matrix/matrix_const.tpp
+++ b/include/dlaf/matrix/matrix_const.tpp
@@ -39,7 +39,7 @@ hpx::shared_future<Tile<const T, device>> Matrix<const T, device>::read(
 }
 
 template <class T, Device device>
-void Matrix<const T, device>::syncAll() noexcept {
+void Matrix<const T, device>::waitLocalTiles() noexcept {
   // Note:
   // Using a readwrite access to the tile ensures that the access is exclusive and not shared
   // among multiple tasks.
--- a/include/dlaf/matrix/matrix_const.tpp
+++ b/include/dlaf/matrix/matrix_const.tpp
@@ -38,6 +38,7 @@ hpx::shared_future<Tile<const T, device>> Matrix<const T, device>::read(
   return tile_managers_[i].getReadTileSharedFuture();
 }
 
+template <class T, Device device>
 void Matrix<const T, device>::syncAll() noexcept {
   // Note:
   // Using a readwrite access to the tile ensures that the access is exclusive and not shared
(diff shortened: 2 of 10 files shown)


albestro:
Minor thing, do we want to keep this?

https://github.com/eth-cscs/DLA-Future/blob/93d739e8ba1e63bc2ad434db843a459d185b6cdc/include/dlaf/matrix/internal/tile_future_manager.h#L75-L83

It was used just in the Matrix d'tor, and the new `syncAll` cannot use it because it would invalidate the object status.

It's not a big deal keeping it, I'm reporting it just to make it transparent.

include/dlaf/matrix/matrix_const.tpp
```diff
@@ -31,25 +31,28 @@ Matrix<const T, device>::Matrix(Distribution distribution, const matrix::LayoutI
...
-
 template <class T, Device device>
 hpx::shared_future<Tile<const T, device>> Matrix<const T, device>::read(
     const LocalTileIndex& index) noexcept {
-  std::size_t i = to_sizet(tileLinearIndex(index));
+  const auto i = tileLinearIndex(index);
   return tile_managers_[i].getReadTileSharedFuture();
 }
 
+void Matrix<const T, device>::syncAll() noexcept {
```
albestro, line 41: I don't know if the name of this barrier function is the best.

My first idea is it should just synchronize local tiles and just care about them, leaving flexibility to the user to decide to create a more "global" barrier involving other ranks (e.g. via an `MPI_Barrier`).

Do you like this kind of API? Should it be something more `syncLocal()`?

@msimberg @teonnik @rasolca

teonnik: How about `waitAllLocalTiles()`?

albestro:
This PR can be reviewed, but it depends on #335 (part of its changes are in here), so merge that one before this.

include/dlaf/matrix/matrix.h
```diff
@@ -23,9 +23,27 @@
...
+namespace internal {
+
+/// Helper function returning a vector with the results of calling a function over a IterableRange2D
+///
+/// @param f non-void function accepting LocalTileIndex as parameter
+template <class Func>
+auto selectGeneric(Func&& f, common::IterableRange2D<SizeType, LocalTile_TAG> range) {
+  using RetT = decltype(f(LocalTileIndex{}));
+
+  std::vector<RetT> tiles;
```
teonnik, line 40: To avoid potentially avoid some `malloc` calls, I suggest to preallocate the vector :`tiles.reserve()`.

teonnik (changes requested):
One of the test cases fails occasionally:

```
25: [ RANK   3 ]
25: /home/teonnik/code/dlaf/test/unit/matrix/test_matrix.cpp:1310: Non-fatal failure:
25: Value of: is_exited_from_scope
25:   Actual: false
25: Expected: true
25: [  FAILED  ] MatrixDestructorFutures.NonConstAfterRead_UserMemory (21 ms)
```

include/dlaf/matrix/matrix.h
albestro: yes, I force pushed the updated branch where I had already fixed this (sorry for that, it is because this change belongs to another PR which this one depends onto)

albestro:
> Some test cases of `test_matrix` fail occasionally:
> 
> ```
> 25: [ RANK   3 ]
> 25: /home/teonnik/code/dlaf/test/unit/matrix/test_matrix.cpp:1310: Non-fatal failure:
> 25: Value of: is_exited_from_scope
> 25:   Actual: false
> 25: Expected: true
> 25: [  FAILED  ] MatrixDestructorFutures.NonConstAfterRead_UserMemory (21 ms)
> ```
> 
> and
> 
> ```
> 25: [ RANK   2 ]
> 25: /home/teonnik/code/dlaf/test/unit/matrix/test_matrix.cpp:1329: Non-fatal failure:
> 25: Value of: is_exited_from_scope
> 25:   Actual: false
> 25: Expected: true
> 25: [  FAILED  ] MatrixDestructorFutures.NonConstAfterReadWrite_UserMemory (22 ms)
> 25: [  INFO    ] 1 of 6 ranks failed
> 25: [ RUN      ] MatrixDestructorFutures.ConstAfterRead_UserMemory
> ```

I asked @teonnik to extend the `wait_guard` time and the problem is not showing up anymore on his machine. We went from 10ms to 100ms, I think it is still a reasonable duration.

include/dlaf/matrix/matrix_const.tpp
```diff
@@ -31,25 +31,29 @@ Matrix<const T, device>::Matrix(Distribution distribution, const matrix::LayoutI
...
+  // among multiple tasks.
+
+  auto readwrite_f = [this](const LocalTileIndex& index) {
+    const auto i = tileLinearIndex(index);
+    return this->tile_managers_[i].getRWTileFuture();
+  };
+
+  const auto range_local = common::iterate_range2d(distribution().localNrTiles());
+  auto all_local_tiles_rw = internal::selectGeneric(readwrite_f, range_local);
+  hpx::wait_all(std::move(all_local_tiles_rw));
```
msimberg, line 54: Purely a subjective style comment. I prefer this, but I understand if you prefer to name the intermediate variable as well.
```suggestion
  hpx::wait_all(internal::selectGeneric(readwrite_f, range_local));
```

msimberg (approved):

test/unit/matrix/test_matrix.cpp
```diff
@@ -1135,90 +1228,189 @@ TYPED_TEST(MatrixTest, GPUCopy) {
...
   {
     auto matrix = createMatrix<TypeParam>();
 
     auto shared_future = matrix.read(LocalTileIndex(0, 0));
-    last_task = shared_future.then(hpx::launch::async, [&guard](auto&&) {
-      hpx::this_thread::sleep_for(WAIT_GUARD);
-      EXPECT_FALSE(guard);
+    last_task = shared_future.then(hpx::launch::async, [&is_exited_from_scope](auto&&) {
+      hpx::this_thread::sleep_for(wait_guard);
+      EXPECT_TRUE(is_exited_from_scope);
```
rasolca, line 1273: I'm wondering if giving more time (e.g. 2 seconds) wouldn't be even better:
(I suggest to use `wait_guard = std::chrono::milliseconds(20);`)
```suggestion
      for (int i=0; i < 10 && !is_exited_from_scope; ++i) {
        hpx::this_thread::sleep_for(wait_guard);
      }
      EXPECT_TRUE(is_exited_from_scope);
```

rasolca:
> I asked @teonnik to extend the `wait_guard` time and the problem is not showing up anymore on his machine. We went from 10ms to 100ms, I think it is still a reasonable duration.

Maybe even longer is better. See my suggestion in the review.

test/unit/matrix/test_matrix.cpp
albestro: Ok, so you're proposal is to wait the least time possible with a granularity defined by `wait_guard`, up until a maximum duration which is eventually way longer than the currently extended one.

I'll do that.

albestro: See if the change here 136ea48ec12729304eee62d86169680491002b76 is ok for you.

Moreover, pay attention to the last change of the commit, where I extended the time for the only guard that works differently from the others.

include/dlaf/matrix/matrix_const.tpp
```diff
@@ -31,25 +31,28 @@ Matrix<const T, device>::Matrix(Distribution distribution, const matrix::LayoutI
...
-  std::size_t i = to_sizet(tileLinearIndex(index));
+  const auto i = tileLinearIndex(index);
   return tile_managers_[i].getReadTileSharedFuture();
 }
 
+template <class T, Device device>
+void Matrix<const T, device>::waitLocalTiles() noexcept {
+  // Note:
+  // Using a readwrite access to the tile ensures that the access is exclusive and not shared
+  // among multiple tasks.
```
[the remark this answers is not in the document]
albestro: Good point! Just changed it, please give it a look if it is clear enough. 😉

teonnik (approved):

PR #359, opened 2021-04-22
Range2D, a standard solution

albestro:
Thanks @msimberg for pointing this out!

If we agree on this, I can try extend this also to `std::random_access_iterator_tag`.

This is a replacement of the naive proposal #358

### DIFF
--- a/include/dlaf/common/range2d.h
+++ b/include/dlaf/common/range2d.h
@@ -69,7 +69,7 @@ struct IteratorRange2D {
   // Postfix
   IteratorRange2D& operator++(int) noexcept {
     auto tmp = *this;
-    current_ = (++i_, computeIndex2D());
+    operator++();
     return tmp;
   }
 
--- a/include/dlaf/common/range2d.h
+++ b/include/dlaf/common/range2d.h
@@ -36,6 +36,7 @@
 /// ```
 
 #include <cstddef>
+#include <iterator>
 
 #include "dlaf/common/assert.h"
 #include "dlaf/common/index2d.h"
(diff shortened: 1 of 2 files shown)


msimberg (approved):
Nice! As I mentioned earlier to @albestro I can't judge if there's some corner case which makes this a non-compliant iterator, but to the best of my knowledge this looks good!

teonnik (approved):

include/dlaf/common/range2d.h
```diff
@@ -45,29 +46,113 @@ namespace common {
...
 /// An Iterator returning indices in column-major order.
 template <typename IndexT, class Tag>
-class IteratorRange2D {
-  using index2d_t = Index2D<IndexT, Tag>;
+struct IteratorRange2D {
+  using iterator_category = std::random_access_iterator_tag;
+  using difference_type = SizeType;
+  using value_type = Index2D<IndexT, Tag>;
+  using reference = const Index2D<IndexT, Tag>&;
+  using pointer = const Index2D<IndexT, Tag>*;
```
albestro, line 54: @msimberg

IMHO `IteratorRange2D` should be immutable, since it does not refer to any container and so it cannot change the pointed/referenced value.

AFAIK, this should be legit https://en.cppreference.com/w/cpp/named_req/ForwardIterator, in particular in reference to these lines

>     The type std::iterator_traits<It>::reference must be exactly 
>        - T& if It satisfies LegacyOutputIterator (It is mutable)
>        - const T& otherwise (It is constant)

teonnik (approved):

PR #361, opened 2021-04-23
Move codecov init and fetch of upload codecov script to deploy image

rasolca:
We are often hitting test timelimit in codecov tests.

The simplest solution is to increase the limit, but as codecov initialization is done in all tests by each rank in the same way I moved it directly in the deploy image.

Moreover I also moved the fetch of the codecov upload script (and checksum check) in the deploy image as well.
(I don't think it is a good idea to rebuild the full build image, just because a test script has changed.)

### DIFF
--- a/ci/docker/codecov/build.Dockerfile
+++ b/ci/docker/codecov/build.Dockerfile
@@ -14,11 +14,6 @@ RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
     git tar wget curl gpg-agent jq tzdata && \
     rm -rf /var/lib/apt/lists/*
 
-RUN cd /usr/local/bin && \
-  curl -Ls https://codecov.io/bash > codecov.sh && \
-  echo "d6aa3207c4908d123bd8af62ec0538e3f2b9f257c3de62fad4e29cd3b59b41d9 codecov.sh" | sha256sum --check --quiet && \
-  chmod +x codecov.sh
-
 # Install cmake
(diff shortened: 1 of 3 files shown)


rasolca:
bors try

rasolca:
bors try

haampie:
I added it to the build image because codecov does not have a versioned url for their script :( the checksum might fail more often than you want :p

rasolca:
bors try

rasolca:
> I added it to the build image because codecov does not have a versioned url for their script :( the checksum might fail more often than you want :p

But this mean that any time you have to update the checksum you have to fully rebuild the build image... 😱

rasolca:
bors try

haampie:
> > I added it to the build image because codecov does not have a versioned url for their script :( the checksum might fail more often than you want :p
> 
> But this mean that any time you have to update the checksum you have to fully rebuild the build image... 😱 

You can move it towards the end, but then it the checksum fails after rebuilding hpx you wont be happy :D

rasolca:
bors try

rasolca:
bors try

rasolca:
bors try

rasolca:
> > > I added it to the build image because codecov does not have a versioned url for their script :( the checksum might fail more often than you want :p
> > 
> > 
> > But this mean that any time you have to update the checksum you have to fully rebuild the build image... 😱
> 
> You can move it towards the end, but then it the checksum fails after rebuilding hpx you wont be happy :D

Yep, exactly, and is the reason why I put it in the deploy image... 😉

albestro (approved):
LGTM

PR #362, opened 2021-04-23
mpi race condition

rasolca:
Tile lifetime for future is not guaranteed, while for shared_future it is.

### DIFF
--- a/test/unit/communication/test_comm_matrix.cpp
+++ b/test/unit/communication/test_comm_matrix.cpp
@@ -20,7 +20,7 @@
 
 using namespace dlaf;
 
-TEST(BcastMatrixTest, Dataflow) {
+TEST(BcastMatrixTest, DataflowFuture) {
   using namespace std::literals;
   comm::Communicator comm(MPI_COMM_WORLD);
   comm::CommunicatorGrid grid(comm, 1, 2, common::Ordering::ColumnMajor);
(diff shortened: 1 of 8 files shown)


rasolca:
bors try

rasolca:
bors try

msimberg:
Is this exactly about the issue we've been discussing in the last few meetings, i.e. the MPI executor does not keep the tiles alive for the duration of the communication? If yes, could you give unwrapExtendTiles a try? You may need the updated one from the #319.

rasolca:
Yes exactly. What puzzles me is the fact that shared futures work correctly.

msimberg:
I'll admit the reason for the difference between `future` and `shared_future` was a bit surprising to me. The reason is not 100% due to the futures themselves. The reason the test fails with `future` is that `get`ting the tile from the future invalidates the future, and so the tile goes out of scope at the end of the `sendBcast` function. With the `shared_future` `get` only gets a `const&` to the tile. However, what I was expecting to happen is that the tuple that gets passed to `invoke_fused` in https://github.com/eth-cscs/DLA-Future/blob/48d39ac023fc2e5abf22e3b8da8e574b932b21ed/include/dlaf/communication/executor.h#L141 would go out of scope at the latest at the end of the `invoke_fused_wrapper` constructor. However, it doesn't, and the tuple instead goes out of scope at the end of the lambda scope at https://github.com/eth-cscs/DLA-Future/blob/48d39ac023fc2e5abf22e3b8da8e574b932b21ed/include/dlaf/communication/executor.h#L144, i.e. after the request has been handled and so the tile lives long enough. A little godbolt example to show what happens: https://godbolt.org/z/rv89nsczq.

TL;DR: `invoke_fused` takes the tuple by rvalue reference but never moves it internally. `unwrapping` only invalidates `future` because it has no other choice when calling `get`. This doesn't happen with `shared_future`. As a result the tuple passed to `invoke_fused` lives until the end of the scope where it's created.

test/unit/communication/test_comm_matrix.cpp
```diff
@@ -0,0 +1,73 @@
...
+  comm::Executor ex{};
+
+  int root = 0;
+  int sz = 10000;
+
+  if (comm.rank() == root) {
+    LocalTileIndex index(0, 0);
+    dlaf::Matrix<double, Device::CPU> mat({sz, 1}, {sz, 1});
+    mat(index).get()({sz - 1, 0}) = 1.;
+    hpx::dataflow(ex, hpx::util::unwrapping(comm::sendBcast<double>), mat(index), ccomm());
```
msimberg, line 37:
```suggestion
    hpx::dataflow(ex, dlaf::matrix::unwrapExtendTiles(comm::sendBcast<double>), mat(index), ccomm());
```

test/unit/communication/test_comm_matrix.cpp
```diff
@@ -0,0 +1,73 @@
...
+  comm::Executor ex{};
+
+  int root = 0;
+  int sz = 10000;
+
+  if (comm.rank() == root) {
+    LocalTileIndex index(0, 0);
+    dlaf::Matrix<double, Device::CPU> mat({sz, 1}, {sz, 1});
+    mat(index).get()({sz - 1, 0}) = 1.;
+    hpx::dataflow(ex, hpx::util::unwrapping(comm::sendBcast<double>), mat.read(index), ccomm());
```
msimberg, line 63:
```suggestion
    hpx::dataflow(ex, dlaf::matrix::unwrapExtendTiles(comm::sendBcast<double>), mat.read(index), ccomm());
```

msimberg (changes requested):
https://github.com/eth-cscs/DLA-Future/blob/48d39ac023fc2e5abf22e3b8da8e574b932b21ed/include/dlaf/communication/executor.h#L122 and https://github.com/eth-cscs/DLA-Future/blob/48d39ac023fc2e5abf22e3b8da8e574b932b21ed/include/dlaf/communication/executor.h#L140 also need to change to at least move the `all_args` tuple.

test/unit/communication/test_comm_matrix.cpp
```diff
@@ -0,0 +1,73 @@
...
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2019, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#include <hpx/futures/future.hpp>
```
msimberg, line 11:
```suggestion
#include <hpx/local/future.hpp>
```

rasolca:
bors try

rasolca:
bors try
Hoping that the strange failure of the CI doesn't happen again...

rasolca:
bors try

msimberg (approved):

Issue #118, opened 2019-12-12
Add NonPositiveDefiniteException test for Cholesky factorization


Sely85:
Write a test to check if the matrix is non-positive definite to be add to the Cholesky factorization method test.

rasolca:
As we are not using exceptions anymore, a return code should be added to Cholesky factorization.

PR #363, opened 2021-04-28
CUDA Cholesky factorization

msimberg:
Mostly cleaned up and rebased on master (and #321).

Definitely still to do is handling of the error code from potrf. I think this is mostly ready for review if someone feels like taking a look, but I'll still do a pass through it to see what I might've missed/messed up.


rasolca:
Handling the potrf return code is something in the TODO list. (See #118 (note it was created when we were still using exceptions extensively.))
It is not so simple as it requires a collective operation before scheduling each iteration of the algorithm.
Currently we are fine with abort if not positive definite.

msimberg:
> Handling the potrf return code is something in the TODO list. (See #118 (note it was created when we were still using exceptions extensively.))
> It is not so simple as it requires a collective operation before scheduling each iteration of the algorithm.
> Currently we are fine with abort if not positive definite.

Sounds good. Right now I'm adding an assertion similar to the CPU case. The assertion is terribly custom-made for the cuBLAS potrf, and will likely need rewriting. This is another one of those that could be more easily handled/customized with senders.

test/unit/test_cusolver_tile/test_potrf.h
```diff
@@ -0,0 +1,138 @@
...
+using namespace dlaf::matrix;
+using namespace dlaf::matrix::test;
+using namespace dlaf::tile;
+using namespace testing;
+
+template <class T, bool return_info>
+void testPotrf(const blas::Uplo uplo, const SizeType n, const SizeType extra_lda) {
+  const TileElementSize size_a = TileElementSize(n, n);
+  const SizeType lda = std::max<SizeType>(1, size_a.rows()) + extra_lda;
+
```
msimberg, line 38: Note that the CPU version, and the `NonPosDef` version below, of this test has a `SCOPED_TRACE` here. However, I had to remove it here because it turns out `SCOPED_TRACE` is not thread-safe in the sense that it uses a thread local vector internally to push and pop the trace messages. This is problematic if the scoped trace is constructed on one worker thread and destroyed on another, which can happen when yielding the HPX thread (there's at least a `then` in `potrf` that can occasionally yield). In general HPX attempts to reschedule yielded and suspended threads on the same worker thread where they were yielded or suspended, but does not guarantee it with the work-stealing schedulers.

It looks like all other uses of `SCOPED_TRACE` just don't end up yielding while it's in scope.

msimberg:
bors try

msimberg:
bors try

miniapp/miniapp_cholesky.cpp
```diff
@@ -141,15 +154,15 @@ int hpx_main(hpx::program_options::variables_map& vm) {
...
+                << " " << hpx::get_os_thread_count() << std::endl;
 
     // (optional) run test
     if ((opts.do_check == CholCheckIterFreq::Last && run_index == (opts.nruns - 1)) ||
         opts.do_check == CholCheckIterFreq::All) {
-      MatrixType original(matrix_size, block_size, comm_grid);
+      HostMatrixType original(matrix_size, block_size, comm_grid);
       copy(matrix_ref, original);
-      check_cholesky(original, matrix, comm_grid);
+      check_cholesky(original, matrix_host, comm_grid);
```
teonnik, line 165: Don't we need a `matrix.syncTargetToSource()` before checking if the result is correct?

teonnik (changes requested):
Great work! Some minor comments.

include/dlaf/matrix/matrix_mirror.h
```diff
@@ -0,0 +1,182 @@
...
+    syncTargetToSource();
+  }
+
+  /// Return a reference to the mirror matrix on the target device.
+  Matrix<T, Target>& get() {
+    return mat_target;
+  }
+
+  /// Copies the source to the target matrix.
+  void syncSourceToTarget() {
```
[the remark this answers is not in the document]
msimberg: I have nothing against that! @rasolca, @albestro, @Sely85 ok with you?

miniapp/miniapp_cholesky.cpp
msimberg: It's synchronized in the destructor, when possible (depends on constness; in this case it's possible). Do you think having it explicit outside of the destructor would be clearer?

teonnik: I didn't notice that a copy occurs upon destruction. Thanks for clarifying, I find it clear enough as it is.

teonnik (approved):

include/dlaf/matrix/matrix_mirror.h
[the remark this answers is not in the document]
msimberg: Updated to use `copy` in place of `sync`.

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
Let's see if this will work after the update... I've made preliminary changes to make this work (possibly) after the panel changes. I did see some worrying non-zero status codes from the cusolver potrf calls which need investigating. Besides the panel changes I tried to leave it as unchanged as possible so not quite sure what could be causing that.

bors try

msimberg:
bors try-

msimberg:
@haampie, @rasolca it looks like that build is not planning to start. I also can't seem to cancel the job in gitlab ("An error occurred..."). Could one of you cancel it please?

msimberg:
Converted to draft until I understand what's going on in the cholesky miniapp...

haampie:
thanks @rasolca :)

msimberg:
bors try

msimberg:
How complicated would it be to set up the triangular solver miniapps to run as tests? Alternatively, would it make sense to change the existing cholesky (and others) test to run with larger blocksizes (probably with the method used to generate the matrix from the miniapp), or just to add a new test like that? As mentioned to @rasolca the existing unit tests at least for cholesky only run with small blocksizes and this did not catch the tile lifetime bug that the last commit fixes. Long term I suppose the miniapps should ideally be performance tests.

I can open a separate issue for this as there's enough to discuss only on that topic.

msimberg:
bors try

rasolca:
bors try

rasolca (approved):

include/dlaf/communication/kernels.h
```diff
@@ -68,16 +68,53 @@ matrix::Tile<const T, D> recvBcastAlloc(TileElementSize tile_size, comm::IndexT_
...
+// This implements scheduleRecvBcast for when the device of the given tile and
+// the communication device are different. In that case we make use of
+// recvBcastAlloc, which will allocate a tile on the correct device for
+// communication, receive the data, and return the tile. When communication is
+// ready, we copy the data from the tile on the communication device to the tile
+// passed into scheduleRecvBcast.
+template <Device D, Device CommunicationD>
+struct scheduleRecvBcastImpl {
+  template <class T>
+  static void call(const comm::Executor& ex, hpx::future<matrix::Tile<T, D>> tile,
```
albestro, line 84: @rasolca @msimberg

This is something that I'd like to know, since I'm working on a PR with similar functions.

Do we want to return a `Future` to the tile with the result of this function for enabling the user to (possibly) attach a continuation?

include/dlaf/communication/kernels.h
```diff
@@ -68,16 +68,53 @@ matrix::Tile<const T, D> recvBcastAlloc(TileElementSize tile_size, comm::IndexT_
...
 void scheduleSendBcast(const comm::Executor& ex, Future<matrix::Tile<const T, D>> tile,
                        hpx::future<common::PromiseGuard<comm::Communicator>> pcomm) {
-  hpx::dataflow(ex, hpx::util::unwrapping(sendBcast_o), internal::prepareSendTile(std::move(tile)),
+  hpx::dataflow(ex, matrix::unwrapExtendTiles(sendBcast_o), internal::prepareSendTile(std::move(tile)),
                 std::move(pcomm));
 }
 
+// This implements scheduleRecvBcast for when the device of the given tile and
+// the communication device are different. In that case we make use of
+// recvBcastAlloc, which will allocate a tile on the correct device for
```
albestro, line 77: @msimberg @rasolca 

should we keep in mind that maybe this will in the next future accept a "workspace" tile from the user?

albestro (commented):
Nice job! just a few comments here and there.

include/dlaf/communication/kernels.h
msimberg: I don't really have an answer to this, except that I'd leave it void until it's needed. This is again about the conflict between the Matrix handling lifetimes and access to tiles vs tiles being passed as arguments and being returned from tasks. When one is using only one of the methods things are mostly clean but when they get mixed things get messy.

include/dlaf/communication/kernels.h
msimberg: From the user using dlaf algorithms or the user writing dlaf algorithms? This already takes a tile from the panel (which I assume you're referring to as a workspace). Are you thinking about something else?

test/unit/matrix/test_matrix_mirror.cpp
```diff
@@ -177,14 +177,14 @@ void syncConstTest(CommunicatorGrid const& comm_grid, TestSizes const& test) {
...
     for (const auto& test : sizes_tests) {
-      syncTest<TypeParam, Device::CPU, Device::CPU>(comm_grid, test);
+      copyTest<TypeParam, Device::CPU, Device::CPU>(comm_grid, test);
 #ifdef DLAF_WITH_CUDA
-      syncTest<TypeParam, Device::CPU, Device::GPU>(comm_grid, test);
-      syncTest<TypeParam, Device::GPU, Device::CPU>(comm_grid, test);
-      syncTest<TypeParam, Device::GPU, Device::GPU>(comm_grid, test);
+      copyTest<TypeParam, Device::CPU, Device::GPU>(comm_grid, test);
+      copyTest<TypeParam, Device::GPU, Device::CPU>(comm_grid, test);
+      copyTest<TypeParam, Device::GPU, Device::GPU>(comm_grid, test);
```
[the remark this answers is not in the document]
msimberg: Yep, very good catch!

include/dlaf/factorization/cholesky/api.h
```diff
@@ -9,16 +9,35 @@
 //
 #pragma once
 
-#include "dlaf/communication/mech.h"
+#include "blas.hh"
```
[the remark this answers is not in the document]
msimberg: Nope, looks like it's not needed anymore. I've removed it in both `api.h` headers.

include/dlaf/lapack/tile.h
```diff
@@ -0,0 +1,278 @@
...
+///
+/// If @p itype = 2 or 3, the problem is A*B*x = lambda*x or
+/// B*A*x = lambda*x, and A is overwritten by U*A*(U**H) or (L**H)*A*L.
+/// B must have been previously factorized as (U**H)*U or L*(L**H) by potrf().
+///
+/// @pre a must be a complex Hermitian matrix or a symmetric real matrix (A),
+/// @pre b must be the triangular factor from the Cholesky factorization of B,
+/// @throw std::runtime_error if the tile was not positive definite.
+template <class T, Device device>
+void hegst(const int itype, const blas::Uplo uplo, const Tile<T, device>& a, const Tile<T, device>& b) {
```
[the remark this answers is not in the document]
msimberg: Good point. Since it'll be changed eventually anyway it sounds like a good idea to change it right away.

include/dlaf/communication/kernels.h
albestro: Sorry, it wasn't very clear what I meant. I should had pointed out these lines

https://github.com/eth-cscs/DLA-Future/blob/505e99981434189a06441f232ea3aa8d761920d6/include/dlaf/communication/kernels.h#L91-L93

where a temporary tile is allocated and used.

It is true that you are using it for the communication and then it is eventually copied back to the user passed one. I was just wondering if we can rely on the allocator (e.g. umpire) for these kind of temporary tiles or not.

include/dlaf/communication/kernels.h
albestro: Sorry, it wasn't very clear what I meant. I should had pointed out these lines

https://github.com/eth-cscs/DLA-Future/blob/505e99981434189a06441f232ea3aa8d761920d6/include/dlaf/communication/kernels.h#L91-L93

where a temporary tile is allocated and used.

It is true that you are using it for the communication and then it is eventually copied back to the user passed one. I was just wondering if we can rely on the allocator (e.g. umpire) for these kind of temporary tiles or not.

albestro (approved):

include/dlaf/communication/kernels.h
msimberg: I see now, thanks for the clarification. Yes, that's exactly what's happening, and indeed, one of the main reasons to use umpire is that it has a pool of memory that does not need to synchronize the device for every allocation. It's not guaranteed to never synchronize, but the pool is initialized with a large block to make it less likely, the initial size is configurable, and if it has to grow the newly allocated memory will be reused and again make device synchronization less likely.

With CUDA 11.something we could use `cudaMallocAsync` to guarantee no device synchronization, but until then we need some pool, and for that I found umpire the best choice to avoid having to implement a memory pool in dlaf.

PR #364, opened 2021-04-28
Don't disable CUDA polling when finalizing DLAF

msimberg:
Analogously to the MPI polling, we should not/don't need to disable CUDA polling at DLAF finalization (see: https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/jobs/1219016879).

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -135,7 +135,6 @@ struct Init<Backend::GPU> {
     finalizeNpCudaStreamPool();
     finalizeHpCudaStreamPool();
     finalizeCublasHandlePool();
-    hpx::cuda::experimental::detail::unregister_polling(hpx::resource::get_thread_pool("default"));
   }
 };
 #endif


msimberg:
bors try

albestro:
So, this is a doubt we don't have anymore?
https://github.com/eth-cscs/DLA-Future/blob/94942f785afe95d977f6d30face2532c1647bd3f/src/init.cpp#L57

I don't remember exactly the reason, but I remember that you talked about it during a meeting...it was something like waiting on exit for uncompleted tasks. Is it correct?

msimberg:
> So, this is a doubt we don't have anymore?
> 
> https://github.com/eth-cscs/DLA-Future/blob/94942f785afe95d977f6d30face2532c1647bd3f/src/init.cpp#L57
> 
> I don't remember exactly the reason, but I remember that you talked about it during a meeting...it was something like waiting on exit for uncompleted tasks. Is it correct?

Not sure if there's a doubt. We can only do that if we guarantee that all MPI and CUDA work is done before we finalize DLA-Future. Can or should we ensure that?

albestro:
> > So, this is a doubt we don't have anymore?
> > https://github.com/eth-cscs/DLA-Future/blob/94942f785afe95d977f6d30face2532c1647bd3f/src/init.cpp#L57
> > 
> > I don't remember exactly the reason, but I remember that you talked about it during a meeting...it was something like waiting on exit for uncompleted tasks. Is it correct?
> 
> Not sure if there's a doubt. We can only do that if we guarantee that all MPI and CUDA work is done before we finalize DLA-Future. Can or should we ensure that?

Not sure about that. I was just asking for confirmation I was on the same page 😉

rasolca:
Mhm, there are a couple of things I'm not sure about:

- Handels and streams are finalized. What happens if kernels are still running on them?
- What happens if a task is still running after calling finalize and try to queue another GPU task?

I think it would be better to check that everything terminated before executing the finalize.
I'm thinking about:
1) check that no HPX task is scheduled or running (Is there a way to do it?)
2) synchronize handels and streams
3) cleanup and deactivate cuda polling.

Note: 1) ensures that no other task will be scheduled, 2) ensures that the GPU terminated all the work.

msimberg:
> Mhm, there are a couple of things I'm not sure about:
> 
>     * Handels and streams are finalized. What happens if kernels are still running on them?
> 
>     * What happens if a task is still running after calling finalize and try to queue another GPU task?
> 
> 
> I think it would be better to check that everything terminated before executing the finalize.
> I'm thinking about:
> 
>     1. check that no HPX task is scheduled or running (Is there a way to do it?)
> 
>     2. synchronize handels and streams
> 
>     3. cleanup and deactivate cuda polling.
> 
> 
> Note: 1) ensures that no other task will be scheduled, 2) ensures that the GPU terminated all the work.

Those are very good points, didn't think about that. I think we should try to avoid as much as possible doing termination detection that could be done by HPX.

For the first point: The handle and stream pools are reference counted (behind a `shared_ptr`). We could make sure that they live long enough by capturing a copy in the continuation https://github.com/eth-cscs/DLA-Future/blob/ea92081c6593bd1d1836e3fba979f5710c9cbdd8/include/dlaf/cublas/executor.h#L162-L163 (and the same for the CUDA executor). This would imply more modifications to the atomic count in the shared pointer, but I think it would be negligible compared to the overheads of dealing with kernel launches anyway. However, I'd like to verify this.

For the second point: If a task is running after finalize and it tries to queue a GPU task it implies that there is an executor that can queue that task, and if there's an executor there's a `shared_ptr` to the handles/streams keeping it alive, with the caveat from the first point that they are currently only kept alive during the launch, not until a kernel is done.

Waiting that no more tasks are running (except the one calling finalize!) is possible, but this is exactly what HPX tries to do at shutdown. Doing that may or may not play well when DLAF is only part of the HPX functionality used by an application (why should it wait for non-DLAF-related tasks?). It is possible to check how many tasks are running, but if it's not necessary I'd like to avoid it (https://github.com/STEllAR-GROUP/hpx/pull/5291 is relevant btw; we need to make some improvements to HPX's termination detection). Synchronizing handles and streams, depending on how much waiting is done in the application before finalize is called, may end up blocking the worker thread calling finalize for a long time. If one were to check that no more tasks/events/requests are in flight it would likely only block for a short time though.

The one downside I can see of only making sure that handles/streams live long enough with `shared_ptr` is that calling finalize implies that user code will no longer spawn new dlaf algorithms after that (more specifically, not create new executors).

What do you think? How badly should `dlaf::finalize` be *the* finalize for an application?

msimberg:
bors try-

msimberg:
bors try

msimberg:
bors try

rasolca (approved):

PR #369, opened 2021-05-06
Panel: allow shrinking from both ends

albestro:
This PR addresses https://github.com/eth-cscs/DLA-Future/pull/309#discussion_r626486224 by @rasolca 

It also has some other changes (e.g. composition over inheritance and other minor ones)

### DIFF
--- a/include/dlaf/matrix/panel.h
+++ b/include/dlaf/matrix/panel.h
@@ -54,9 +54,8 @@ struct Panel<axis, const T, device> : protected Matrix<T, device> {
 
   /// Return an IterableRange2D with a range over all tiles of the panel (considering the offset)
   auto iterator() const noexcept {
-    return common::iterate_range2d(
-        LocalTileIndex(component(axis), start_),
-        LocalTileIndex(component(axis), end_, 1));
+    return common::iterate_range2d(LocalTileIndex(component(axis), start_),
+                                   LocalTileIndex(component(axis), end_, 1));
   }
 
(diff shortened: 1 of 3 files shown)


include/dlaf/matrix/panel.h
```diff
@@ -97,28 +98,53 @@ struct Panel<axis, const T, device> : protected Matrix<T, device> {
...
+    DLAF_ASSERT(start_loc < end_loc, start_loc, end_loc);
+
+    DLAF_ASSERT(start_loc >= bias_, start, bias_);
+    DLAF_ASSERT(end_loc <= dist_matrix_.localNrTiles().get(component(axis)), end,
+                dist_matrix_.localNrTiles().get(component(axis)));
+
+    start_ = start_loc;
+    end_ = end_loc;
+  }
+
```
rasolca, line 118: Maybe just call setRangeStart and setRangeEnd...

rasolca (approved):

include/dlaf/matrix/panel.h
albestro: I started with the solution you are proposing, but the assertions are a bit different.

By changing separately `start` or `end`, they have to be coherent with the other which is already set, while changing all together in a single step, they have to respect the "global" constraint, in addition to be coherent as a pair (but not considering existing setup).

Hope it is clear enough why I opted for the "duplicated" code, but I'm open to any idea that can improve this.

PR #370, opened 2021-05-11
Discussion (minor): handleRecvTile non-const

albestro:
This PR is just a change over the #309 to show a possible change in the API.

With reference to this comment
https://github.com/eth-cscs/DLA-Future/pull/309/commits/cb3933114a8cfdddfd2e95b60b94ec24934cc824/#r621351681

I forgot, in case there is, the reason why we prefer to return `ConstTile_t` from `recv` functions. Maybe I'm missing some use-cases.

In particular the main point about this I think it is if we prefer something like
> return `shared_future<ConstTile>` from schedule functions, so that there is just a read-only access to the result and in case a read-write access to the result is needed, the user has to go through the matrix/panel to get back a rw tile

I didn't change the `recvBcastAlloc` in the same way, because of limitations imposed by the current usage, i.e. cholesky stores returned tiles in a `vector<shared_future<Tile>>`.

### DIFF
--- a/include/dlaf/communication/kernels.h
+++ b/include/dlaf/communication/kernels.h
@@ -40,13 +40,11 @@ DLAF_MAKE_CALLABLE_OBJECT(sendBcast);
 
 // Non-blocking receiver broadcast
 template <class T, Device D>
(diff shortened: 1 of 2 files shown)


teonnik (approved):
LGTM!

msimberg (approved):

PR #374, opened 2021-05-19
Scheduler functions for MPI_Ireduce and MPI_Iallreduce

### LINKED ISSUE #368: MPI async collectives and tile lifetime
As they are implemented now, `MPI_Ireduce` and `MPI_Iallreduce` do an additional internal copy in case the memory being used for the operation is not contiguous.

This is a workaround we opted for because of the impossibility to use predefined `MPI_Op` with `MPI_Vector` (used by our tiles when the memory is not contiguous). AFAIK, the preset of `MPI_Op` (e.g. MPI_SUM) can work just with contiguous memory, while for `MPI_Vector` it is possible to define custom `MPI_Op` using the [`MPI_Op_create`](https://www.mpich.org/static/docs/latest/www3/MPI_Op_create.html).

In async collectives with copy workaround, we have to deal with different phases:
- make the original tile contiguous (if not yet)
- use the contiguous data for the asynchronous operation (being it the original tile or the temporary buffer)
- in case a tile has been used for output of the collective operation, and the temporary buffer was used, copy back the data to the original tile

Across all these phases, the original tile must be kept alive until the asynchronous operations is completed (due to MPI requirement).

For the moment I started investigating a possible solution with the workaround in branch [alby/test/reduce_async_mech](https://github.com/eth-cscs/DLA-Future/tree/alby/test/reduce_async_mech) (see [test/unit/communication/test_comm_matrix.cpp](https://github.com/eth-cscs/DLA-Future/blob/dc735c8b04dcb94d2b46c2745479db346c9c8e69/test/unit/communication/test_comm_matrix.cpp))

It does not look like very nice, and I'm wondering if it wouldn't be better to fix this workaround by implementing a custom `MPI_Op` for `MPI_Vector`. Probably it will not fix all our problems (I'm thinking about the CPU/GPU interaction), but it may improve the context a bit.

Looking forward to your feedbacks!

@msimberg @rasolca @teonnik

albestro:
Close #368

Extend MPI asynchronous kernels with `MPI_Ireduce` and `MPI_Iallreduce` operations.

What's going on:
- Created a `kernels/` folder together with new include-all header and moved existing async kernels to `kernels/broadcast.h`
- Add new reduce/all_reduce in respective files under the new `kernels/` folder
- minor change: add an assertion in matrix

EDIT:
- Fix annotations for custom MPI async executor

TODO/Open points:
- [ ] when using a communicator with size=1 (local) the copy mechanism gets executed when not not needed
- [x] where do we want to put the `Bag` facility? it sounds like `common` but it also has to deal with `matrix::Tile`
  - currently it is in `dlaf::common::internal`
- [x] add missing wrappers (maybe not all, but all the needed one at the moment are implemented)
- [x] executors to use with `makeContiguous` and `copyBack`
  - at the moment I'm using HP executors: do we want to open an issue to track that we'd like to investigate the creation of a dedicated executor on the MPI pool for the communication-related copies?)
- [ ] sort/cleanup tests for new functions
- [ ] refine helper function `getUnwrapRetValueAndArgs`
- [ ] ~order of parameters for schedule~
- [ ] ~simplify code (ideas)~
  - [ ] ~shared_future<T>~
  - [ ] ~tile lifetime control: is there any other way to keep the `shared_future<const T>`?~


include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -0,0 +1,109 @@
...
+  {
+    auto wrapped = getUnwrapRetValAndArgs(
+        hpx::dataflow(matrix::unwrapExtendTiles(common::internal::makeItContiguous_o),
+                      std::move(tile_out)));
+    bag_out = std::move(hpx::get<0>(wrapped));
+    tile_out = std::move(hpx::get<0>(std::move(hpx::get<1>(wrapped))));
+  }
+
+  bag_out = hpx::dataflow(ex, hpx::util::unwrapping(internal::allReduce_o), std::move(pcomm), reduce_op,
+                          std::move(bag_in), std::move(bag_out), tile_in);
```
msimberg, line 80:
```suggestion
  bag_out = hpx::dataflow(ex, unwrapExtendTiles(internal::allReduce_o), std::move(pcomm), reduce_op,
                          std::move(bag_in), std::move(bag_out), tile_in);
```
?
It's not quite obvious to me if plain `unwrapping` is enough in this case. What keeps `tile_in` otherwise alive? The bag doesn't, right?

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -0,0 +1,109 @@
...
+
+  tile_out = hpx::dataflow(hpx::util::unwrapping(common::internal::copyBack_o), std::move(tile_out),
+                           std::move(bag_out));
+
+  return tile_out;
+}
```
msimberg, line 86: I had a go at trying to formulate this using `let_value` from https://wg21.link/p1897 (but keeping the bag and dataflow etc.). It's definitely at least as verbose as the above, but is in my mind more explicit about keeping various things alive long enough. That might just be because I'm more familiar with the functionality though, so please let me know what you think. Keep in mind that I haven't compiled this, so it may contain bugs:

```
template <class T>
auto scheduleAllReduceSenders(const comm::Executor& ex,
                              hpx::future<common::PromiseGuard<comm::Communicator>> pcomm,
                              MPI_Op reduce_op,
                              hpx::shared_future<matrix::Tile<const T, Device::CPU>> tile_in,
                              hpx::future<matrix::Tile<T, Device::CPU>> tile_out) {
  namespace ex = hpx::execution::experimental;

  hpx::future<common::internal::Bag<const T>> bag_in =
      hpx::dataflow(hpx::util::unwrapping(common::internal::makeItContiguous_o), tile_in);

  // This async scope keeps tile_out alive for the duration of the operation                                                                                                                                                                   
  // returned by the lambda                                                                                                                                                                                                                    
  ex::let_value(std::move(tile_out), [ex, pcomm = std::move(pcomm), reduce_op,
                                      bag_in = std::move(bag_in), tile_in = std::move(tile_in)](
                                         matrix::Tile<T, Device::CPU>& tile_out) {
    auto bag_out = comm::internal::makeItContiguous(tile_out);

    // This async scope keeps tile_in, bag_in, and bag_out alive                                                                                                                                                                               
    // for the duration of the operation returned by the lambda                                                                                                                                                                                
    auto keep_alive = ex::when_all(ex::keep_future(std::move(tile_in)), std::move(bag_in),
                                   ex::just(std::move(bag_out)));
    return ex::let_value(keep_alive, [](hpx::shared_future<matrix::Tile<const T, Device::CPU>>& tile_in,
                                        common::internal::Bag<const T>& bag_in,
                                        common::internal::Bag<T>& bag_out) {
      // Note: no tile_in just for lifetime management, and void future returned                                                                                                                                                               
      hpx::future<void> =
          hpx::dataflow(ex, hpx::util::unwrapping(internal::allReduce_o), std::move(pcomm), reduce_op,
                        std::move(bag_in), std::move(bag_out));
      return hpx::dataflow(hpx::util::unwrapping(common::internal::copyBack_o), std::move(tile_out),
                           std::move(bag_out));
    });
  });
}                                                                                                                                                                                                                                              
```
I've now written out the types fully, but they could be replaced with auto in many places.

msimberg: I should add that I think this could be done without Bag as well, by using `let_value` more extensively.

albestro:
For the records: as per decision in other PRs (https://github.com/eth-cscs/DLA-Future/pull/363#discussion_r643026486), I'm not returning anything from the schedule functions, while I'm making an exception for `scheduleAllReduceInPlace`. There the return value is handy and eases operations chaining due to how the algorithm is structured and its API (it accepts a tile so we don't have a way to get back the tile from the original matrix).

include/dlaf/communication/kernels/all_reduce.h
albestro: Thanks for pointing out this, it looks more explicit to me too. I quickly read the standard proposal you linked and, if I got it right, `let_value` also returns the same sender we passed in. Is it right?

If yes, this returned future, will be available as soon as the async scope will complete, so it represents a future<Tile> after the operation. What will I get if I pass in a shared_future by copy? I suppose that it will keep alive the Tile and gets passed through the async_scope, so I will get a shared_future back, with the exact same Tile, but that will be ready once the async_scope will finish. It is a trivial question, but just to check if I get it correctly.

albestro:
bors try

test/unit/communication/test_collective_async.cpp
```diff
@@ -0,0 +1,159 @@
...
+  CHECK_TILE_EQ(exp_tile, mat_out.read(idx).get());
+}
+
+TEST_F(CollectiveTest, AllReduce) {
+  testAllReduce(world, newBlockMatrixContiguous<T, device>(), newBlockMatrixContiguous<T, device>());
+  testAllReduce(world, newBlockMatrixStrided<T, device>(), newBlockMatrixStrided<T, device>());
+  testAllReduce(world, newBlockMatrixContiguous<T, device>(), newBlockMatrixStrided<T, device>());
+}
+
+// TODO TEST AllReduce -> AllReduceInPlace Mixed
```
albestro, line 159: @rasolca do we really want to test also mixed or is it enough like this?

Do you have any suggestion on how to improve tests? would you suggest to split them over different TEST_F so that we get more information in case of failure?

albestro:
bors try

include/dlaf/communication/kernels/all_reduce.h
msimberg: I'll close this as resolved as this doesn't (and can't) block merging. This was just to start a discussion on how we might do this differently with sender adaptors. Let's continue this when we've made some more progress on using senders/receivers.

include/dlaf/common/bag.h
```diff
@@ -0,0 +1,70 @@
...
+
+/// A Bag:
+/// - owns the temporary buffer (optionally allocated)
+/// - contains a data descriptor to the contiguous data to be used for communication
+///   (being it the aforementioned temporary buffer if allocated, otherwise an externally managed memory,
+///   e.g. a tile)
+///
+/// It is up to the user to guarantee the integrity of the Bag and its properties
+template <class T>
+using Bag = hpx::tuple<common::Buffer<std::remove_const_t<T>>, common::DataDescriptor<T>>;
```
msimberg, line 34: Might this also benefit from being a struct with named fields rather than a tuple? Or did I miss a case where you're using this with `split_future`?

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -0,0 +1,151 @@
...
+                       hpx::future<matrix::Tile<T, Device::CPU>> tile_out) {
+  // Note:
+  //
+  //         +---------------------------------+
+  //         |                                 |
+  // TILE_I -+-> makeContiguous -----> BAG_I --+--> mpi_call --> BAG_O --+
+  //                                           |                         |
+  // TILE_O ---> makeContiguous --+--> BAG_O --+                         |
+  //                              |                                      |
+  //                              +----------------> TILE_O -------------+-> copyBack
```
msimberg, line 76: :+1:

More of this, I like it!

include/dlaf/common/bag.h
msimberg: I really don't want to block this on the name so let's keep `Bag` if we don't have a better alternative. But maybe `ContiguousBufferHolder`? Not a great name either but `Bag` is *very* generic ;)

msimberg (changes requested):
I just have minor naming comments. Looks good to me otherwise!

include/dlaf/communication/kernels/all_reduce.h
albestro: 😄 Just for reference, I used this https://asciiflow.com

include/dlaf/common/bag.h
albestro: Something like this? do you have any better suggestion for the field names?

```cpp
template <class T>
struct ContiguousBufferHolder {
  common::Buffer<std::remove_const_t<T>> buffer;
  common::DataDescriptor<T> descriptor;
};
```

msimberg: No better suggestion, this already looks much better! Thanks.

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -0,0 +1,152 @@
...
+template <class T>
+void scheduleAllReduce(const comm::Executor& ex,
+                       hpx::future<common::PromiseGuard<comm::Communicator>> pcomm, MPI_Op reduce_op,
+                       hpx::shared_future<matrix::Tile<const T, Device::CPU>> tile_in,
+                       hpx::future<matrix::Tile<T, Device::CPU>> tile_out) {
+  // Note:
+  //
+  //         +---------------------------------+
+  //         |                                 |
+  // TILE_I -+-> makeContiguous -----> BAG_I --+--> mpi_call --> BAG_O --+
```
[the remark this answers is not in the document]
albestro: do you think `contiguous_buffer` would be better?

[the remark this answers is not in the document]
albestro: acronym `cbh`/`CBH`? it will make also the life easier with the doc 😜

the name is not convincing me much, waiting for your suggestion @rasolca @msimberg 😄

test/unit/communication/test_collective_async.cpp
albestro: About mixed tests I meant having ranks in the same communication in which part uses `allReduce` and the others `allReduceInPlace`. To me it looks superfluous, but you may have some edge cases it is worth checking.

include/dlaf/communication/kernels/all_reduce.h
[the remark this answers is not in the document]
rasolca: buf/BUF?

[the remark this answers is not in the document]
msimberg: `cont_buf`? Is it only for the diagram or also for the code?

test/unit/communication/test_collective_async.cpp
rasolca: You are already mixing as you invert the matrices for odd ranks.

include/dlaf/communication/kernels/all_reduce.h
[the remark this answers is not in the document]
albestro: it's for both, but not strictly the same.

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -0,0 +1,152 @@
...
+
+  hpx::future<common::internal::ContiguousBufferHolder<T>> bag_out;
+  {
+    // clang-format off
+    auto wrapped = getUnwrapRetValAndArgs(
+        hpx::dataflow(
+          dlaf::getHpExecutor<Backend::MC>(),
+          matrix::unwrapExtendTiles(common::internal::makeItContiguous_o),
+          std::move(tile_out)));
+    // clang-format on
```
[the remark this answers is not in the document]
rasolca: Changes from https://github.com/eth-cscs/DLA-Future/pull/374/commits/bfb581a3ae7991d5aa7f97685c7ef50eae578f1f are OK

include/dlaf/communication/kernels/all_reduce.h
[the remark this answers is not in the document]
msimberg: Ok, for the diagram an abbreviation and a legend would work. For the code `buf` is also fine by me. It's a fairly local variable, so it's meaning is mostly clear from context.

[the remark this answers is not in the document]
albestro: in the meanwhile, I went for `cont_buf` 😄 not a big deal, but I went for the slightly more rich variant 😉

include/dlaf/matrix/matrix.tpp
```diff
@@ -60,6 +60,7 @@ Matrix<T, device>::Matrix(const LayoutInfo& layout, ElementType* ptr)
 template <class T, Device device>
 hpx::future<Tile<T, device>> Matrix<T, device>::operator()(const LocalTileIndex& index) noexcept {
   const auto i = tileLinearIndex(index);
+  DLAF_ASSERT_MODERATE(i < tile_managers_.size(), i, tile_managers_.size());
```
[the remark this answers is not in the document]
albestro: as discussed I added the pre-condition check in `MatrixBase::tileLinearIndex` and removed all the others in `Matrix` I added in this PR.

test/unit/communication/test_collective_async.cpp
albestro: In some sense, I agree, but in both cases the `scheduleAllReduceInPlace` is called, while I was thinking about mixing `scheduleAllReduceInPlace` and `scheduleAllReduce`.

Anyway, it is enough also for you, I would leave it as it is and remove the TODO.

rasolca (approved):

rasolca:
bors try

msimberg (approved):

include/dlaf/communication/kernels/all_reduce.h
msimberg: I seem to have completely forgotten to respond to this.

> `let_value` also returns the same sender we passed in. Is it right?

`let_value` returns a sender that sends the same values as the sender returned from the invocable. Note that I say "a sender that sends the same values" because it is not literally the same sender. The sender returned from the invocable will be internally connected and started, and the values themselves will then be sent to a connected receiver.

PR #378, opened 2021-05-26
Distributed GenToStd.

rasolca:
- Implemented distributed GenToStd.
- Updated the test naming and structure, such that GPU tests can be added easily.

I also refactored the local version in two steps:
1) Introducing functions which schedule tile operations (like e.g. in cholesky) https://github.com/eth-cscs/DLA-Future/commit/919c9874c9d3cffe87968a8321e636321b9f4992
2) Using loops on indices instead of ranges. https://github.com/eth-cscs/DLA-Future/commit/9ff372d84a72511950cb6d3cdcc5c75e094ff189

Regarding 2), I think that it makes the code more readable. What do you think?

Note:
- As the panel width is still fixed I needed a workaround during the last k iteration.
   When this problem will be fixed the code can become cleaner.

TODO:
- [x] GPU


msimberg:
>     1. Introducing functions which schedule tile operations (like e.g. in cholesky) [919c987](https://github.com/eth-cscs/DLA-Future/commit/919c9874c9d3cffe87968a8321e636321b9f4992)

I think this is nice. It also aligns quite nicely with the `scheduleX` functions (except for the naming). By the way, why are these helpers `snake_case` (I know that they are elsewhere as well, so it's not new for this PR)? :smile:

>     2. Using loops on indices instead of ranges. [9ff372d](https://github.com/eth-cscs/DLA-Future/commit/9ff372d84a72511950cb6d3cdcc5c75e094ff189)
>
> Regarding 2), I think that it makes the code more readable. What do you think?

I actually prefer the ranges, but that's purely subjective.
> TODO:
> 
>     * [ ]  GPU

I interpret this as you're planning to do this still in this PR? It looks like it would only need a few more cublas wrappers to work on the GPU, correct?

rasolca:
> > 1. Introducing functions which schedule tile operations (like e.g. in cholesky) [919c987](https://github.com/eth-cscs/DLA-Future/commit/919c9874c9d3cffe87968a8321e636321b9f4992)
> 
> I think this is nice. It also aligns quite nicely with the `scheduleX` functions (except for the naming). By the way, why are these helpers `snake_case` (I know that they are elsewhere as well, so it's not new for this PR)? 😄
> 
I'm just using the style used in Cholesky 😉 ... Maybe we should take one Algorithm, revise it and take it as a template...
> > 
> > 2. Using loops on indices instead of ranges. [9ff372d](https://github.com/eth-cscs/DLA-Future/commit/9ff372d84a72511950cb6d3cdcc5c75e094ff189)
> > Regarding 2), I think that it makes the code more readable. What do you think?
> 
> I actually prefer the ranges, but that's purely subjective.
> 
What I really don't like about range loops, is that you need to extract the indices anyway to create the 2D indices required. Usually it falls back to use:
```
range = ...;
for (const auto ij : range) {
  i = ij.row();
  j = ij.col();
  ...
}
instead of
for (SizeType j; ...) {
  for (SizeType i; ...) {
  ...
}
```
> > TODO:
> > GPU
> 
> I interpret this as you're planning to do this still in this PR? It looks like it would only need a few more cublas wrappers to work on the GPU, correct?

Yes, exactly, but I wanted to wait the merge of Cholesky...

rasolca:
> > > 1. Introducing functions which schedule tile operations (like e.g. in cholesky) [919c987](https://github.com/eth-cscs/DLA-Future/commit/919c9874c9d3cffe87968a8321e636321b9f4992)
> > 
> > I think this is nice. It also aligns quite nicely with the `scheduleX` functions (except for the naming). By the way, why are these helpers `snake_case` (I know that they are elsewhere as well, so it's not new for this PR)? 😄
> 
> I'm just using the style used in Cholesky 😉 ... Maybe we should take one Algorithm, revise it and take it as a template...
> 
Forgot to mention that `scheduleX` functions are part of the public API, while in this case they are just internal helper functions.

msimberg:
> What I really don't like about range loops, is that you need to extract the indices anyway to create the 2D indices required. Usually it falls back to use:

That's a good point. Maybe something for [structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding) once we require C++17... Then it could be 
```
range = ...;
for (const auto [i, j] : range) {
  ...
}
```

rasolca:
bors try

test/include/dlaf_test/blas/invoke.h
```diff
@@ -0,0 +1,61 @@
...
+#ifdef DLAF_WITH_CUDA
+template <>
+struct InvokeBlas<Device::GPU> {
+  template <class F, class... Args>
+  static void call(F&& f, Args&&... args) {
+    cublasHandle_t handle;
+    DLAF_CUBLAS_CALL(cublasCreate(&handle));
+  DLAF_CUBLAS_CALL(f(handle, std::forward<Args>(args)...);
+  DLAF_CUDA_CALL(cudaDeviceSynchronize());
+  DLAF_CUBLAS_CALL(cublasDestroy(handle));
```
albestro, line 46: it seems that this has not been detected by the format check.

test/include/dlaf_test/matrix/util_tile_blas.h
```diff
@@ -22,38 +24,46 @@ namespace dlaf {
...
-template <class T, class Func>
-void set(Tile<T, Device::CPU>& tile, Func el, blas::Op op) {
+namespace internal {
+template <class ElementGetter>
+auto opValFunc(ElementGetter val, const blas::Op op) {
+  std::function<decltype(val(std::declval<TileElementIndex>()))(TileElementIndex)> op_val;
   switch (op) {
     case blas::Op::NoTrans:
-      set(tile, el);
+      op_val = [&val](TileElementIndex i) { return val(i); };
```
albestro, line 33: I'm wondering if this was/is an error, since we are capturing by reference a variable with function scope inside a lambda, than the lambda outlives the value because it is returned outside of the function, so when it is called the referenced value has been destroyed.

test/include/dlaf_test/matrix/util_tile_blas.h
```diff
@@ -22,38 +24,46 @@ namespace dlaf {
...
       break;
     }
   }
+  return op_val;
+}
+}
+
+/// Sets the elements of the tile.
+///
+/// The (i, j)-element of the tile is set to el({i, j}) if op == NoTrans,
```
albestro, line 58: With (i,j) in the upper triangular part of the matrix?

I would take the chance to clarify in the documentation which part of the matrix acts as "source" for values.

include/dlaf/blas/tile.h
```diff
@@ -113,9 +113,95 @@ struct CublasGemm<std::complex<double>> {
...
+template <typename T>
+struct CublasHemm;
+
+template <>
+struct CublasHemm<float> {
+  template <typename... Args>
+  static void call(Args&&... args) {
+    DLAF_CUBLAS_CALL(cublasSsymm(std::forward<Args>(args)...));
+  }
+};
```
albestro, line 125: In your fight against repetitions, I would suggest you to think about using macros to cleanup this too.

e.g.
```
DLAF_CREATE_CUBLAS_OP(Hemm);
DLAF_CREATE_CUBLAS_OP(Hemm, float,                cublasSsymm);
DLAF_CREATE_CUBLAS_OP(Hemm, double,               cublasDsymm);
DLAF_CREATE_CUBLAS_OP(Hemm, std::complex<float>,  cublasChemm);
DLAF_CREATE_CUBLAS_OP(Hemm, std::complex<double>, cublasZhemm);
```

albestro (commented):
Super cleanup work!

note: I didn't went through details of gen2std algorithm yet. I'll do later.

rasolca:
bors try

rasolca:
This PR is ready, but depends on #379.

include/dlaf/lapack/tile.h
```diff
@@ -232,18 +209,40 @@ class CusolverPotrfInfo {
...
+                                             util::blasToCublasCast(a.ptr()), a.ld(),
+                                             util::blasToCublasCast(b.ptr()), b.ld(), &workspace_size);
+  internal::CusolverInfo<T> info{std::max(1, workspace_size)};
+  internal::CusolverHegst<T>::call(handle, itype, util::blasToCublas(uplo), n,
+                                   util::blasToCublasCast(a.ptr()), a.ld(),
+                                   util::blasToCublasCast(b.ptr()), b.ld(),
+                                   util::blasToCublasCast(info.workspace()), info.info());
+
+  return info;
+}
```
msimberg, line 231: Is this `hegstInfo`? Or is hegst different from potrf in some way?

msimberg: This comment probably belongs to #379...

rasolca: Please check new commits on #379.

rasolca:
bors try

rasolca:
bors try

rasolca:
bors try

rasolca:
bors try

msimberg (approved):
No more comments from me, looks good!

rasolca:
bors try

rasolca:
bors try-

albestro (approved):

rasolca:
bors try

teonnik (approved):
LGTM, Just some minor comment.

rasolca:
bors try

Sely85 (approved):
LGTM

rasolca:
bors try

PR #380, opened 2021-06-02
Python scripts: plot customization

albestro:
This PR slightly changes how internally plot generation works, introducing also some new features.

- `_gen_nodes_plot` now returns `(fig, ax)`. Its role is just generating plots, it does not save on file anymore;
- introduced `NodePlotWriter`, a generator that using the previous function, allow to customize the plots via returned `(fig, ax)` and on exit it writes to specified file
- existing specialized functions (e.g. `gen_chol_plots`), now accepts additional parameters `customize_ppn` and `customize_time` which are functions accepting `(fig, ax)` that are internally used (if set, they are optional), to customize the respective plots

Apart from being able to customize plots using `(fig, ax)`, other new features are that `_gen_nodes_plot` accepts:
- `subplots_args`: a dict with keywords to pass to `plt.subplots()` call (e.g. `gridspec_kw`)
- `styles`: a list with a structure similar to `filts` (i.e. `(regex, value)`), but accepting as values a dict with matplotlib styles for `plot` call (e.g. color, linewidth, ...)
- `fill_area`: enable/disable the min-max area plot

note: I gave a try to `black` for formatting the `postprocess.py` file (separate commit). let me know if you like it or not.

### DIFF
--- a/scripts/postprocess.py
+++ b/scripts/postprocess.py
@@ -29,9 +29,9 @@ def _gen_nodes_plot(
         title:          title of the plot
         df:             the pandas.DataFrame with the data for the plot
         combine_mb:     bool indicates if different mb has to be included in the same plot
-        filts:          [regex] for matching benchmark names to filter from the plot
-        replaces:       [(regex_replace_rule, newtext)] to apply to benchmark names for the legend
-        styles:         [(regex, dict())] where dict() contains kwargs valid for the plot
+        filts:          list of regex for selecting benchmark names to plot
+        replaces:       list of (regex_replace_rule, newtext) to apply to benchmark names for the legend
+        styles:         list of (regex, dict()) where dict() contains kwargs valid for the plot
         subplot_args:   kwargs to pass to pyplot.subplots
         fill_area:      switch on/off the min-max area on plots
     """
--- a/scripts/postprocess.py
+++ b/scripts/postprocess.py
@@ -11,7 +11,7 @@ from parse import parse
 
 sns.set_theme()
 
-# plt_type : ppn | time
+
 def _gen_nodes_plot(
     plt_type,
     title,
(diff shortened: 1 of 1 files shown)


scripts/postprocess.py
```diff
@@ -12,8 +12,21 @@
 sns.set_theme()
 
 # plt_type : ppn | time
-def _gen_nodes_plot(plt_type, title, ylabel, file_name, df, logx, logy=False, combine_mb=False, filts=None, replaces=None):
-    fig, ax = plt.subplots()
+def _gen_nodes_plot(
```
rasolca, line 15: As input parameters increased (and their value has to follow some constrains) I would improve the doc.

scripts/postprocess.py
```diff
@@ -220,7 +276,16 @@ def calc_trsm_metrics(df):
     )
 
 
-def gen_chol_plots(df, logx=False, combine_mb=False, filts=None, replaces=None, filename_suffix=None):
+# customize_* functions should accept fig and ax as parameters
```
rasolca, line 279: The same here... a little more doc doesn't hurt...

rasolca (commented):
I would improve a bit the doc and/or add an example that uses the new interface.

rasolca:
Regarding black: can you increase the line length to 105 chars as for C++ files?

albestro:
> Regarding black: can you increase the line length to 105 chars as for C++ files?

In the last commit I applied the line-length=105 modifier, just on the file changed by this PR. In local I have also a simple config file for the entire project (if we want to pursue this path).

scripts/postprocess.py
albestro: Can you give it a quick check, please?

scripts/postprocess.py
```diff
@@ -11,9 +11,33 @@
...
+    subplot_args=None,
+    fill_area=True,
+):
+    """
+    Args:
+        plt_type:       ppn | time
+        title:          title of the plot
+        df:             the pandas.DataFrame with the data for the plot
+        combine_mb:     group also by mb sizes
+        filts:          [regex] for matching benchmark names to filter from the plot
```
rasolca, line 32: - is `[regex]` a standard way to document an array containing one ore more regex? I'm used to fact that square brackets indicate optional arguments.
- I would interpret `filter from` as `matching are excluded from`. Am I wrong?
   Maybe rewriting as following is more clear:
   `benckmarks matching one of these regular expressions will appear in the plot (None: all benchmarks will appear).`

scripts/postprocess.py
```diff
@@ -11,9 +11,33 @@
...
+    styles=None,
+    subplot_args=None,
+    fill_area=True,
+):
+    """
+    Args:
+        plt_type:       ppn | time
+        title:          title of the plot
+        df:             the pandas.DataFrame with the data for the plot
+        combine_mb:     group also by mb sizes
```
rasolca, line 31:
```suggestion
        combine_mb:     bool indicates if different mb has to be included in the same plot
```

scripts/postprocess.py
albestro: Good points. I tried to address your comments. Please, give it a look 😉

rasolca (approved):

PR #382, opened 2021-06-07
Panel: skip last tile communication and minor improvements

albestro:
As a result of a discussion with @rasolca, these are small improvements/check that Panel requires:
- [x] skip last tile communication (and document it)
- [x] add a check in `setTile` for verifying that the external tile size matches the size of internal ones used by Panel
- [x] check for squareness of the broadcast transpose

About the latter one, I think it was already addressed by https://github.com/eth-cscs/DLA-Future/blob/e9e585d07cf5e479aea9ddc96d0f006c86d9e28f/include/dlaf/communication/broadcast_panel.h#L160-L161

For what concerns the `setTile` check, I look forward to any suggestion on how to do it.


rasolca (approved):
LGTM, I would remove TODOs in cholesky about last tile comm...

albestro:
bors try

albestro:
bors try

albestro:
I identified the problem introduced with this PR.

We don't communicate anymore the last tile, and it's perfectly fine, but the code is expecting that a tile stored locally, is accessible also via the transposed panel as an external tile.

Since `panelT` external tiles are set together with the communication, this "service" of being able to access a virtually transposed coordinate which is stored locally via the `panelT` does not work anymore.

I think that algorithmically it could be interesting to keep this "service", it should not be difficult to separate the communication logic from the set external tile one, but it does not looks perfectly clean, since we state that we transpose everything excepts the last tile. I would evaluate to document this "special" behavior with something like "if the last tile is locally available, it is set anyway as external tile of the transposed panel".

Note:
This service saves us from writing something like this in the algorithm:
```cpp
const auto jj = [=]() {
  const auto idx_cross = distr.template globalTileFromLocalTile<Coord::Col>(j);
  const auto index_diag_local = distr.template localTileFromGlobalTile<Coord::Row>(idx_cross);
  return index_diag_local;
}();

gemmTrailingMatrixTile(executor_np,
    panel.read({Coord::Row, i}),
    // next line is the change in the call
    i_idx == nrtile - 1 ? panel.read({Coord::Row, jj}) : panelT.read({Coord::Col, j}),
    mat_a(LocalTileIndex{i, j}));
```

as opposed to the current code

https://github.com/eth-cscs/DLA-Future/blob/0f286d45f12beb09bc4106039b9cdc7942a30471/include/dlaf/factorization/cholesky/impl.h#L220-L221

albestro:
As pointed out by @rasolca, the problem is not exactly what I reported in my last comment. Indeed, the problem was that I implemented it wrongly, because I was skipping all the last tiles for each rank, while it has to be skipped just the last tile globally.

We opted for improving the panel to make it work with global ranges (see #384), which allow us to fix this PR too.

albestro:
bors try

albestro:
bors try

albestro:
The problem with the `std::prev` has been identified and fixed in #390. I'm going to revert the change and it will work as soon as the other PR will get merged.

include/dlaf/communication/broadcast_panel.h
```diff
@@ -173,7 +173,7 @@ void broadcast(const comm::Executor& ex, comm::IndexT_MPI rank_root,
   const auto owner = dist.template rankGlobalTile<coordT>(last_tile);
   const auto range = dist.rankIndex().get(coordT) == owner
                          ? common::iterate_range2d(*panelT.iteratorLocal().begin(),
-                                                   LocalTileIndex(coordT, panelT.rangeEndLocal() - 1, 1))
+                                                   *std::prev(panelT.iteratorLocal().end()))
```
rasolca, line 176: No it doesn't work:
`*std::prev(panelT.iteratorLocal().end())` gives `LocalTileIndex(coordT, panelT.rangeEndLocal() - 1, 0)`
which results in an empty range.

albestro: You're right! Because the range are defined with the first index outside, not the last one in the range. My error.

The other option would have been to use the overload that accepts as second parameter the size, by specifing the distance between the begin and std::prev(end), but it would require more "work" because the difference of two iterators is a linear index, while the iterate_range2d function requires a 2D size.

I would go for keeping it simple for the moment and just revert (again!) this last change.

rasolca:
bors try

rasolca (approved):

albestro:
bors try-

albestro:
bors try

albestro:
> ## try
> 
> Build failed:
> 
>     * [ci/gitlab/full-pipeline](https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/pipelines/329411237)

I had to manually re-try because the `allocate` step of just one of the builds (cpu-releasetest) failed with
```
salloc: error: Job submit/allocate failed: Socket timed out on send/recv operation
```

After that, all greens.

PR #384, opened 2021-06-11
Improve Panel API and functionality with global range

albestro:
Opening a separate PR for this change, which turns out to be needed for solving the problem in #382.

*Note: This PR has been opened in draft mode, because this new API may allow for some code and logic cleanup.*

Previously the range for the panel had to be specified through local indices, often leaving the burden to the user to convert from the global index to the local one.

Now, the range is specified with global coordinates. In addition to improving its usage from the user point of view, it also opens to a better versatility. For instance, by having access to the global range coordinates, it is possible to retrieve the position of the panel in the global coordinate system of the matrix, which was not possible with just local range coordinates.


include/dlaf/communication/broadcast_panel.h
```diff
@@ -140,57 +140,22 @@ void broadcast(const comm::Executor& ex, comm::IndexT_MPI rank_root,
 
   const auto& dist = panel.parentDistribution();
 
+  DLAF_ASSERT(square_size(dist), dist.size());  // TODO this can be relaxed
```
albestro, line 143: @rasolca since we can shrink the panel from both ends, I think we should relax this and impose that the sub-matrix described by the two panels is a square matrix (see few lines below `DLAF_ASSERT(panelSize == panelTSize, panelSize, panelTSize)`).

does it sound to you?

include/dlaf/communication/broadcast_panel.h
```diff
@@ -140,57 +140,22 @@ void broadcast(const comm::Executor& ex, comm::IndexT_MPI rank_root,
...
+  DLAF_ASSERT(square_blocksize(dist), dist.blockSize());
+
   // Note:
   // This algorithm allow to broadcast panel to panelT using as mirror the parent matrix main diagonal.
   // This means that it is possible to broadcast panels with different axes just if their global offset
   // lie on the diaognal.
```
rasolca, line 149:
```suggestion
  // lie on the diagonal.
```

include/dlaf/communication/broadcast_panel.h
rasolca: Not sure if it is something useful.
As transposing is used with symmetric matrices, they are square and the mirroring is always against the diagonal.
I would keep it simple, and extend it only if we have a use case.

include/dlaf/communication/broadcast_panel.h
```diff
@@ -140,57 +140,22 @@ void broadcast(const comm::Executor& ex, comm::IndexT_MPI rank_root,
...
-        return chances > 0;
-      }(),
-      panel.rangeStart(), panelT.rangeStart(),
-      "broadcast can mirror just on the parent matrix main diagonal");
+  DLAF_ASSERT(panel.rangeStart() == panelT.rangeStart(), panel.rangeStart(), panelT.rangeStart());
+
+  const auto panelSize = panel.rangeEnd() - panel.rangeStart();
+  const auto panelTSize = panelT.rangeEnd() - panelT.rangeStart();
+
+  DLAF_ASSERT(panelSize == panelTSize, panelSize, panelTSize);
```
rasolca, line 155: As start is the same I would simply check that end is the same as well.

rasolca (approved):

Sely85 (approved):
LGTM. I think global indices are much more convenient here.

albestro:
bors try

rasolca (approved):

PR #387, opened 2021-06-18
integration of change for Panel API: use Index instead of Size for range specification

albestro:
As per discussion with @rasolca, here a small proposal (if this will be accepted it will become part of #384).

The main points:
- Panel now accepts Index2D instead of Size2D for ranges (both in c'tor and in setRange functions)
- a new cast operator has been added that allows to convert from Size2D to Index2D if EXPLICITLY requested

About the latter point, I opted for that because the end component of the setRange requires to specify the first index outside the range, and this is often identified by the size of the matrix. Not really the nicest thing, but it is quite handy in the code and it should not be too harmful for the Index/Size semantic separation.

Look forward to your comments and suggestions.

### DIFF
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -267,8 +267,8 @@ Index2D<IndexT, TAG> indexFromOrigin(const Size2D<IndexT, TAG>& size) noexcept {
 }
 
 template <class IndexT, class TAG>
-Size2D<IndexT, TAG> sizeFromOrigin(const Index2D<IndexT, TAG>& size) noexcept {
-  return {size.row(), size.col()};
+Size2D<IndexT, TAG> sizeFromOrigin(const Index2D<IndexT, TAG>& index) noexcept {
+  return {index.row(), index.col()};
 }
 
 /// Compute coords of the @p index -th cell in a row-major ordered 2D grid with size @p dims.
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -134,6 +134,10 @@ protected:
 
 }
 
+// forward declaration due to the cross dependency between Index2D and Size2D
+template <typename IndexT, class Tag>
+class Index2D;
+
 /// A strong-type for 2D sizes.
 ///
 /// @tparam IndexT type for row and column coordinates,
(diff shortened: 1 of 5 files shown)


include/dlaf/common/index2d.h
```diff
@@ -145,6 +149,10 @@ class Size2D : public internal::basic_coords<IndexT> {
 public:
   using BaseT::basic_coords;
 
+  explicit operator Index2D<IndexT, Tag>() const noexcept {
+    return {rows(), cols()};
+  }
```
rasolca, line 154: I would use a method instead of conversion, such that the operation done is eplicitly stated:
e.g.:
`Index indexFromOrigin(size)` i.e. index{0, 0} + size
Moreover I would add also the inverse operation:
`Size sizeFromOrigin(index)` i.e index - Index{0, 0}

albestro: I followed your suggestion. I agree that the cast operator, even if explicitly required, it is not so self-explicative.

About the name of these functions, I'm not really sure. I want to make sure that the user does not interpret it wrongly (e.g. it has to pass an origin?! it doesn't make much sense).

Please consider something slightly more verbose like `indexWrtOriginOf(size)`.

rasolca: The names were just a suggestion... if you have better options feel free to use them.

albestro:
bors try

include/dlaf/common/index2d.h
```diff
@@ -261,6 +261,16 @@ Coords2DType transposed(Coords2DType coords) {
...
 
+template <class IndexT, class TAG>
+Index2D<IndexT, TAG> indexFromOrigin(const Size2D<IndexT, TAG>& size) noexcept {
+  return {size.rows(), size.cols()};
+}
+
+template <class IndexT, class TAG>
+Size2D<IndexT, TAG> sizeFromOrigin(const Index2D<IndexT, TAG>& size) noexcept {
+  return {size.row(), size.col()};
+}
```
Sely85, line 272: Aren't these two swapped? I would have associated `rows()` to `Size2D` while `row()` to `Index2D`.

albestro:
> ## try
> 
> Build failed:
> 
>     * [ci/gitlab/full-pipeline](https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/pipelines/324317299)

It's ok, because this is going to be merged in another PR that will also contain a fix for the problem.

I'm going to merge this into the other PR.

Issue #354, opened 2021-04-16
Work on subtiles


rasolca:
When using the Panel #309 and in general when reusing a Tile as a workspace in an algorithm in some cases the tile sizes might mismatch. (E.g. the width of the last panel of a matrix might be less than the block size).

I can think about 3 ways to solve this problem:

1) Do nothing, and instead of Tile wrappers use Plain BLAS/LAPACK specifying the correct sizes.

2) The panel width is set at the beginning of the iteration (and sets the correct size to the tiles returned).

3) Provide a method to take a sub tile.

Details:

2) A Tile method for resizing is required. It should be private/protected and accessible only from the Panel (friend).
(We don't want users to mess with Tile sizes).
A panel method can be used to set the panel width at the beginning of the iteration (which sets the tile size correctly and stop the execution if a too large (compared to the allocated space) width is requested).
I don't see any complication using this method.

3) To implement this approach is more tricky:
3.1) One solution can be to modify the Tile to include extra information (i.e. current start index and current tile size).
Current index and size are used by the Tile to return the pointer and the size, while when the Tile destructor is called and therefore the tile sets itself in the promise, only the original size is used and the index and size are reset.
This approach is simple, but has two drawbacks:
- In a task the tile can only shrink (i.e. a task cannot work first only on the first half of the tile and then on the full tile)
- Const Tiles referenced from a shared future cannot use this approach, otherwise a race condition between different read operation might occour.

3.2) A sub-tile object which contains a reference to the Tile and the start index and size of the subtile.
The drawback of this approach is that the original Tile lifetime should be managed and guaranteed by the user and therefore the sub-tile can be used only within a task.

I think we should use both 2 and 3.2 for the following reasons:
- 3.2 allows any sub-tiling in tasks.
- The width of the panel is known at the beginning of each iteration, therefore with 2 it is not needed to take a sub-tile each time a panel tile is used.

@albestro @teonnik @msimberg @Sely85 Any comment is welcome.

albestro:
Just to keep everyone up-to-date on this. We discussed with @rasolca about this and we thought about a possible alternative solution (I think @rasolca is going to check feasibility in the next weeks).

Just a few main points to show the general idea, details will come later.
- Use the same concept used in `MatrixView` where given a Future<Tile> from a Matrix, a "DAG loop" is created that it takes care of setting the promise of the Matrix according to the RO/RW access, when the DAG loop is over.
[...]

PR #389, opened 2021-06-18
Tile utility to access subtiles.

rasolca:
First step to address #354.

splitTile returns a vector of tiles which represent the subtiles:

in the case of const tiles all the tiles (including the original full tile) are ready at the same time. The next tile extracted from the matrix will be ready only when all the subtiles and the full tile go out of scope.

in the case of modifiable tiles the subtiles has to be disjoint (to avoid race conditions), therefore all the subtiles will become ready at the same time. Moreover the future passed to splitTile is replaced by a new future containing the full tile that will become ready when all the subtiles will go out of scope. Finally the next tile extracted from the matrix will be ready only when the full tile goes out of scope.


rasolca:
The documentation is still missing, but I would like a first round of reviews.

include/dlaf/matrix/tile.h
```diff
@@ -177,23 +284,33 @@ class Tile : public Tile<const T, device> {
   /// @pre index.isIn(size()).
   T* ptr(const TileElementIndex& index) const noexcept {
     DLAF_ASSERT_HEAVY(index.isIn(size_), index, size_);
```
albestro, line 286: since you are using `linearIndex` you can remove this, right?

include/dlaf/matrix/tile.h
```diff
@@ -177,23 +284,33 @@ class Tile : public Tile<const T, device> {
...
-  using Tile<const T, device>::size_;
-  using Tile<const T, device>::memory_view_;
-  using Tile<const T, device>::ld_;
-  using Tile<const T, device>::p_;
+  // Creates a writable subtile keeping the dependencies.
+  // It calls old_tile.get(), therefore it should be used when old_tile is guaranteed to be ready:
+  // e.g. in dataflow, .then, ...
+  Tile(hpx::shared_future<TileType> tile, TileElementIndex origin, TileElementSize size)
+      : ConstTileType(tile.get(), origin, size) {
+    sf_ = tile;
```
albestro, line 305: you can move in the copy you already got on the call

albestro (commented):
Nice work! I just went quickly through the main parts of the mechanism, I will look more in details later, but my feedback is absolutely positive.

Just a couple of very minor comments.

include/dlaf/matrix/tile.h
```diff
@@ -34,12 +34,22 @@ struct ContinuationException final : public std::runtime_error {
...
 template <class T, Device device>
 class Tile;
 
 template <class T, Device device>
 class Tile<const T, device>;
 
+template <class T, Device D>
+std::vector<hpx::shared_future<Tile<const T, D>>> splitTile(
+    hpx::shared_future<Tile<const T, D>>& tile,
+    std::vector<std::tuple<TileElementIndex, TileElementSize>> specs);
```
msimberg, line 47: Is it maybe worth creating a specific type for `tuple<TileElementIndex, TileElementSize>`?

include/dlaf/matrix/tile.h
```diff
@@ -121,21 +136,113 @@ class Tile<const T, device> {
...
 
   TileElementSize size_;
   memory::MemoryView<ElementType, device> memory_view_;
   SizeType ld_;
 
-  std::unique_ptr<promise_t<Tile<ElementType, device>>> p_;
+  // With C++17 the following objects can be joined in a variant.
+  std::unique_ptr<promise_t<TileType>> p_;
+  hpx::shared_future<TileType> sf_;
+  hpx::shared_future<ConstTileType> sfc_;
```
msimberg, line 178: There's always `union`... But maybe it's better to wait until we require C++17.

include/dlaf/matrix/tile.h
```diff
@@ -202,6 +319,105 @@ auto create_data(const Tile<T, device>& tile) {
...
+        std::get<0>(spec), std::get<1>(spec)));
+  }
+
+  return ret;
+}
+
+/// Create SubTiles:
+/// TODO
+template <class T, Device D>
+std::vector<hpx::future<Tile<T, D>>> splitTile(
```
msimberg, line 343: Would this function be better called `splitTileDisjoint` or something like that? It has stronger preconditions than the read-only version.

rasolca:
Note #391 and #392 have been added to show a use case of this PR.

msimberg (approved):

include/dlaf/matrix/tile.h
```diff
@@ -202,6 +325,151 @@ auto create_data(const Tile<T, device>& tile) {
...
+}
+
+/// Create a writeable subtile of a given tile.
+///
+/// All the returned subtiles will get ready, when the original tile was supposed to get ready
+/// (therefore the returned subtiles should be disjoint, otherwise race conditions might occour).
+/// @p tile is replaced with the (full) tile which will get ready when the all the subtiles go out of scope.
+/// The next dependency in the dependency chain will become ready only when @p tile goes out of scope.
+/// @pre The subtiles described with specs should be disjoint
+///      (i.e. two different subtile cannot access thesame element).
```
albestro, line 421: minor typo
```suggestion
///      (i.e. two different subtile cannot access the same element).
```

test/unit/matrix/test_tile.cpp
```diff
@@ -341,3 +339,300 @@ TYPED_TEST(TileTest, CreateBuffer) {
...
+  testSubtileDisjoint<Type, Device::CPU>("Test Vector 2", {5, 7}, 8,
+                                         {{{4, 3}, {0, 0}}, {{4, 6}, {1, 1}}}, 1);
+  testSubtileDisjoint<Type, Device::CPU>("Test Vector 3", {5, 7}, 8,
+                                         {{{5, 7}, {0, 0}},
+                                          {{1, 2}, {2, 2}},
+                                          {{3, 0}, {2, 7}},
+                                          {{0, 0}, {1, 4}},
+                                          {{0, 4}, {3, 3}}},
+                                         2);
+  testSubtileDisjoint<Type, Device::CPU>("Test Vector 3", {5, 7}, 8,
```
albestro, line 636: minor: I think it's just a message for reference in case of errors, right?
```suggestion
  testSubtileDisjoint<Type, Device::CPU>("Test Vector 4", {5, 7}, 8,
```

albestro (commented):
Few comments, in general very nice job! The mechanism is very well documented.

I would like to check the testing part, which I haven't looked yet, and I would also like to think about the "recursive sub-tile" thing that I quickly shared with you offline.

Tomorrow I will complete my review.

test/unit/matrix/test_tile.cpp
```diff
@@ -341,3 +339,300 @@ TYPED_TEST(TileTest, CreateBuffer) {
...
+
+// TileFutureOrConstTileSharedFuture should be
+// either hpx::future<Tile<T, D>> or hpx::shared_future<Tile<const T, D>>
+// TileFuture should be hpx::future<Tile<T, D>>
+template <class F, class TileFutureOrConstTileSharedFuture, class TileFuture>
+void checkReadyAndDependencyChain(F&& tile_ptr, std::vector<TileFutureOrConstTileSharedFuture>& subtiles,
+                                  const std::vector<SubTileSpec>& specs, std::size_t last_dep,
+                                  TileFuture& next_tile_f) {
+  ASSERT_EQ(subtiles.size(), specs.size());
+  ASSERT_GT(subtiles.size(), last_dep);
```
albestro, line 407: If I got it correctly, `last_dep` should be either `[subtiles.size() - 1` or `subtiles.size()`, right?

albestro (commented):
And here it is my (minor) comments on the test part.

I just want to point out the thing about "recursive" subtiles. For what concerns const tile I think there should be no problem, since you can store the shared_future either to the original or the subtile, and it should work in both cases.

For what concerns the non-const tiles instead, I would do an additional check. From what I got quickly going through your implementation, it should be ok. Indeed, when a subtile becomes a "parent" itself in the recursion, it is transparent that it is a subtile, so the perimeter is limited and race-condition on data will be avoided. For what concerns the management of the shared_future that keeps alive the original, the `splitTileInsertFutureInChain` mechanism should still stand. Indeed, given `original -> subTile -> subsubTile`, you delay the triggering of original from subTile to subsubTile, so you can keep alive the subTile without caring about the original (which is intrinsically managed by the subTile). I shared this thought (maybe the explanation is not so clear, but we can discuss about this if you want) so you can comment about it and verify if it really works or not.

I don't know if we want to allow the recursion (I'm not sure it is needed right now), but in case I would a very simple and basic test for this recursive case.

test/unit/matrix/test_tile.cpp
rasolca: Nope, it can be any subtile or the complete tile: i.e. [0, .size())

rasolca:
bors try

albestro (approved):
LGTM!

rasolca:
bors try

rasolca:
bors try

PR #390, opened 2021-06-22
Fix Range2D iterators

albestro:
The iterator had a few major problems, which I addressed when using `std::prev` which was not working properly (it was not computing the previous iterator and it was just returning exactly the same one).

At the time of writing I added few very basic tests that I used to debug the problems, together with the fixes to the bugs I found out.

Look forward to your comments.

### DIFF
--- a/include/dlaf/common/range2d.h
+++ b/include/dlaf/common/range2d.h
@@ -72,7 +72,9 @@ struct IteratorRange2D {
   }
 
   reference operator[](difference_type n) const noexcept {
-    return computeIndex2D(n);
+    auto it = *this;
+    it += n;
+    return *it;
   }
 
   IteratorRange2D& operator++() noexcept {
@@ -80,7 +82,7 @@ struct IteratorRange2D {
     return *this;
   }
 
-  IteratorRange2D& operator++(int) noexcept {
+  IteratorRange2D operator++(int) noexcept {
     auto tmp = *this;
     operator++();
     return tmp;
@@ -91,18 +93,20 @@ struct IteratorRange2D {
     return *this;
   }
 
-  IteratorRange2D& operator--(int) noexcept {
+  IteratorRange2D operator--(int) noexcept {
     auto tmp = *this;
     operator--();
     return tmp;
   }
 
   IteratorRange2D& operator+=(difference_type n) noexcept {
-    return i_ += n, *this;
+    current_ = (i_ += n, computeIndex2D(i_));
+    return *this;
   }
 
   IteratorRange2D& operator-=(difference_type n) noexcept {
-    return i_ -= n, *this;
+    current_ = (i_ -= n, computeIndex2D(i_));
+    return *this;
   }
 
   friend IteratorRange2D operator+(IteratorRange2D a, difference_type n) noexcept {
--- a/include/dlaf/common/range2d.h
+++ b/include/dlaf/common/range2d.h
@@ -60,7 +60,7 @@ struct IteratorRange2D {
 
   IteratorRange2D(value_type begin, IndexT ld, SizeType i)
       : begin_(begin), i_(i), ld_(std::max(ld, IndexT(1))) {
-    current_ = computeIndex2D(i_);
+    updateIndex2D();
   }
 
   reference operator*() const noexcept {
@@ -78,7 +78,8 @@ struct IteratorRange2D {
   }
 
   IteratorRange2D& operator++() noexcept {
-    current_ = computeIndex2D(++i_);
+    ++i_;
+    updateIndex2D();
     return *this;
   }
 
@@ -89,7 +90,8 @@ struct IteratorRange2D {
   }
 
   IteratorRange2D& operator--() noexcept {
-    current_ = computeIndex2D(--i_);
+    --i_;
+    updateIndex2D();
     return *this;
   }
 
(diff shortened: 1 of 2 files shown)


include/dlaf/common/range2d.h
```diff
@@ -72,37 +72,41 @@ struct IteratorRange2D {
...
   reference operator[](difference_type n) const noexcept {
-    return computeIndex2D(n);
+    auto it = *this;
+    it += n;
+    return *it;
   }
 
   IteratorRange2D& operator++() noexcept {
-    current_ = (++i_, computeIndex2D(i_));
+    current_ = computeIndex2D(++i_);
```
msimberg, line 81: Does `computeIndex2D` ever operate on anything but `i_` (after the changes)? Would it be clearer to change `computeIndex2D` to simply modify `current_` and use `i_`. I know in this case it's actually making things less explicit, but given that it's always used the same way, it might actually be less error-prone to have it modify the state? However, I think the current formulation is also good.

albestro: It's a good point, it is not worth anymore. I fully agree with you that it is better to make the parameters and update value embedded in the method.

It's also true that the problem of the uncoupling between the `i_` and `current_` will still exist, but in this latter case any more complicated solution would look like an overshot to me.

msimberg (approved):

rasolca (approved):

rasolca:
bors try

test/unit/common/test_range2d.cpp
```diff
@@ -113,3 +113,81 @@ TEST(DoubleArgEmptyRange2D, Index2D) {
...
+/// | (1, 1)B | (1, 2) | (1, 3)E |
+/// | (2, 1)  | (2, 2) |         |
+/// | (3, 1)  | (3, 2) |         |
+/// +---------+--------+---------+
+///
+/// {(1, 1) == range.begin(), (2, 1), (3, 1), (1, 2), (2, 2), (3, 2), (1, 3) == range.end()}
+struct IteratorTest : public ::testing::Test {
+  using range2D_t = decltype(iterate_range2d(::Index{0, 0}, ::Size{0, 0}));
+
+  const range2D_t range = iterate_range2d(::Index{1, 1}, ::Size{3, 2});
```
teonnik, line 129: I'm a bit confused about this part, why is (1,3) part of the range?

teonnik (changes requested):
One of the tests fails for me:

```
7: [ RUN      ] IteratorTest.IndexAccess
7: /home/teonnik/code/dlaf/test/unit/common/test_range2d.cpp:135: Failure
7: Expected equality of these values:
7:   *it
7:     Which is: (1, 1)
7:   range.begin()[i]
7:     Which is: (-1137868792, 21913)
7: /home/teonnik/code/dlaf/test/unit/common/test_range2d.cpp:135: Failure
7: Expected equality of these values:
7:   *it
7:     Which is: (2, 1)
7:   range.begin()[i]
7:     Which is: (-1137868792, 21913)
7: /home/teonnik/code/dlaf/test/unit/common/test_range2d.cpp:135: Failure
7: Expected equality of these values:
7:   *it
7:     Which is: (3, 1)
7:   range.begin()[i]
7:     Which is: (-1137868792, 21913)
7: /home/teonnik/code/dlaf/test/unit/common/test_range2d.cpp:135: Failure
7: Expected equality of these values:
7:   *it
7:     Which is: (1, 2)
7:   range.begin()[i]
7:     Which is: (-1137868792, 21913)
7: /home/teonnik/code/dlaf/test/unit/common/test_range2d.cpp:135: Failure
7: Expected equality of these values:
7:   *it
7:     Which is: (2, 2)
7:   range.begin()[i]
7:     Which is: (-1137868792, 21913)
7: /home/teonnik/code/dlaf/test/unit/common/test_range2d.cpp:135: Failure
7: Expected equality of these values:
7:   *it
7:     Which is: (3, 2)
7:   range.begin()[i]
7:     Which is: (-1137868792, 21913)
7: [  FAILED  ] IteratorTest.IndexAccess (0 ms)
```

test/unit/common/test_range2d.cpp
albestro: Let's check together if the assumptions are right, maybe it's just the documentation that it is misleading.

https://github.com/eth-cscs/DLA-Future/blob/1232535cd317667dcda8ecb75495ffff2d641007/test/unit/common/test_range2d.cpp#L129

This defines a 3x2 range2D starting from (1, 1), so it is a range that contains:
```
/// +--------+--------+
/// | (1, 1) | (1, 2) |
/// | (2, 1) | (2, 2) |
/// | (3, 1) | (3, 2) |
/// +--------+--------+
/// 
/// {(1, 1), (2, 1), (3, 1), (1, 2), (2, 2), (3, 2)}
```

But, since we are dealing with iterators, we know that they work with the logic [begin, end), so the end is the first one past the end of the actual range. From an implementation detail (and this may be considered an error), we know that the "end" will be (1, 3).

So, (1, 3) is not part of the range, I was just depicting it in the documentation with the aim of (hopefully) give a better overview of the various arithmetic operations in the next tests.

Hope this helps you and I look forward to your comments/suggestions on how to improve it.

albestro:
bors try

test/unit/common/test_range2d.cpp
teonnik: I see, it's OK to leave it as it is.

teonnik (approved):

PR #394, opened 2021-07-06
Update check format

rasolca:
Improve format check (remove tab in source files also in clang-format disabled regions).


albestro (approved):
LGTM

PR #396, opened 2021-07-14
Fix SubTile asserts

albestro:
The Tile c'tor used for subtiles had ASSERTs checking for the correctness of parameters, but being inside the body of the c'tor they were not executed in case the call to the parent constructor failed (e.g. MemoryView realizes that a given configuration does not fit).


rasolca (approved):

rasolca:
bors try

rasolca:
bors try

PR #388, opened 2021-06-18
Porting to HPX@1.7.0

albestro:
HPX 1.7.0 is not yet released, but we are likely going to need it (e.g. https://github.com/STEllAR-GROUP/hpx/pull/5387)

Here I'll try to keep a report of what is going to change, so that everyone can update their own PRs in the near future.

- `hpx::util::unwrapping` and `hpx::util::unwrap` now are directly under the `hpx::` namespace

PR #398, opened 2021-07-16
Enable warnings in object libraries

msimberg:
Only opening now as a reminder for myself to come back to this; I have not compiled this to check how many of these there are. Inspired by https://github.com/eth-cscs/DLA-Future/pull/388#discussion_r671014072.

Additionally, we should strongly consider setting `-Wall`, with possible exceptions through `-Wno-x` but that may be a bigger task (I don't know, did someone already try it? How bad was it?).

### DIFF
--- a/cmake/DLAF_AddTargetWarnings.cmake
+++ b/cmake/DLAF_AddTargetWarnings.cmake
@@ -24,6 +24,7 @@ macro(target_add_warnings target_name)
       -Wextra
       -Wnon-virtual-dtor
       -Wunused
+      -Wunused-local-typedefs
       -Woverloaded-virtual
       -Wdangling-else
       -Wswitch-enum
--- a/cmake/DLAF_AddTargetWarnings.cmake
+++ b/cmake/DLAF_AddTargetWarnings.cmake
@@ -17,38 +17,39 @@ macro(target_add_warnings target_name)
 
   set(IS_COMPILER_CLANG $<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>)
   set(IS_COMPILER_GCC $<CXX_COMPILER_ID:GNU>)
+  set(IS_CUDA_NVCC $<COMPILE_LANG_AND_ID:CUDA,NVIDIA>)
 
   target_compile_options(${target_name}
     PRIVATE
-      -Wall
-      -Wextra
-      -Wnon-virtual-dtor
-      -Wunused
-      -Wunused-local-typedefs
-      -Woverloaded-virtual
-      -Wdangling-else
-      -Wswitch-enum
-      # Conversions
-      $<${IS_COMPILER_GCC}:
-        -Wsign-conversion
-        -Wfloat-conversion>
-      $<${IS_COMPILER_CLANG}:
-        -Wbitfield-enum-conversion
-        -Wbool-conversion
-        -Wconstant-conversion
-        -Wenum-conversion
-        -Wfloat-conversion
-        -Wint-conversion
-        -Wliteral-conversion
-        -Wnon-literal-null-conversion
-        -Wnull-conversion
-        -Wshorten-64-to-32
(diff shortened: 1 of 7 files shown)


msimberg:
bors try

rasolca:
Just as a remark: `-Wall` is already used, but is different from `-Weverything` which is anyway strongly discouraged.
https://github.com/eth-cscs/DLA-Future/blob/master/cmake/DLAF_AddTargetWarnings.cmake

msimberg:
> Just as a remark: `-Wall` is already used, but is different from `-Weverything` which is anyway strongly discouraged.
> https://github.com/eth-cscs/DLA-Future/blob/master/cmake/DLAF_AddTargetWarnings.cmake

Very good point! I also saw that `-Wunused-local-typedefs` only really applies to typedefs (duh), and not to plain `using a::b::c`. So this would anyway not have caught all of the `using`s in #388. But it is not included in `-Wunused` nor `-Wall` so it should still be an improvement.

rasolca:
According to gcc doc:
> `-Wunused-local-typedefs` (C, Objective-C, C++ and Objective-C++ only)
    Warn when a typedef locally defined in a function is not used. This warning is enabled by -Wall.
    
For both gcc and clang the flag is already included by `-Wall` and it covers typedefs and type aliases, but not `using namespace` and `using a::b::c` cases.

https://godbolt.org/z/G9jqhz7aP

msimberg:
> According to gcc doc:
> 
> > `-Wunused-local-typedefs` (C, Objective-C, C++ and Objective-C++ only)
> > Warn when a typedef locally defined in a function is not used. This warning is enabled by -Wall.
> 
> For both gcc and clang the flag is already included by `-Wall` and it covers typedefs and type aliases, but not `using namespace` and `using a::b::c` cases.
> 
> https://godbolt.org/z/G9jqhz7aP

Interesting... I checked the docs first which does *not* include `-Wunused-local-typedefs` in any of `-Wall`, `-Wextra`, or `-Wunused`, but godbolt is pretty clear on that. Then the question is: why are the warnings not errors on CI? Is this the answer: https://github.com/eth-cscs/DLA-Future/blob/d9344e3390e11f2364752964a75611f2233e5a63/src/CMakeLists.txt#L141? Warnings are only added to the `DLAF` target, not the object libraries...

msimberg:
bors try

msimberg:
I'll try to work through these new warnings-as-errors on the side...

msimberg:
bors try

msimberg:
bors try

msimberg:
By the way, this should actually be ready to go, modulo reviews. I enabled warnings on all targets. I had to disable `-pedantic-errors` for nvcc. Otherwise I've basically only added casts or explicit integral types here and there. `std::uncaught_exception` has been deprecated in favour of `uncaught_exceptions`. That actually seems to be a C++17 thing, which GCC likely just makes available with C++14 as well... I can revert this if you'd like. Finally, I changed the variable-sized arrays `std::vector`s, except the one with bools because `std::vector<bool>` is annoying.

rasolca (approved):

rasolca:
> `std::uncaught_exception` has been deprecated in favour of `uncaught_exceptions`. That actually seems to be a C++17 thing, which GCC likely just makes available with C++14 as well... I can revert this if you'd like.

@msimberg Did you already revert it? I didn't see it in the review.

include/dlaf/matrix/tile.h
```diff
@@ -213,7 +213,7 @@ Tile<const T, device>::Tile(Tile&& rhs) noexcept
 template <class T, Device device>
 Tile<const T, device>::~Tile() {
   if (p_) {
-    if (std::uncaught_exception())
+    if (std::uncaught_exceptions() > 0)
```
msimberg, line 216: @rasolca I didn't revert this change yet. Should I? It may need some pragmas around it to disable the warning.

msimberg:
bors try

PR #402, opened 2021-07-22
set0 for Matrix and Panel

albestro:
Panel enforces some restrictions on tiles access. In particular (until a `reset()` is performed):
- once a tile has been accessed in RW mode (using memory allocated for the panel), it can't be set as external anymore;
- once a tile has been set as external, it can't be accessed in RW anymore.

This is a restriction that may be relaxed/changed/improved in the near future (IIRC talking about TRSM LLT with @rasolca a related problem has been identified).

For these reasons, the `set0(Panel)` have to access tiles bypassing the mechanism and access directly the internal data, otherwise all tile will be thereafter set as "internal" and no external tiles could be linked anymore.

Another implicit assumption made here is that external tiles contains valid data, and in case they have to be set-to-0, the user cannot do it via this panel (because the access on external tiles is RO), but just by accessing the original RW tile.


rasolca (approved):

include/dlaf/matrix/panel.h
```diff
@@ -387,6 +383,14 @@ struct Panel : public Panel<axis, const T, device> {
       return splitTile(tile, {{0, 0}, tileSize(index)});
   }
 
+  /// This functions sets all the elements of the tiles in the range to zero without inhibiting to set the
+  /// tile as external afterward (e.g. the tile is set to zero without changing its state to "already used")
+  template <class ExecutorOrPolicy>
+  friend void set0(ExecutorOrPolicy ex, Panel<axis, T, Device::CPU>& panel) {
```
albestro, line 389: IMHO there is a thing that it is worth highlighting.

`set0` does not touch the status of the tile, i.e. it will still be possible to link external tiles, which is an explicit requirement for the use-case I have in reduction to band.
But this also means that `set0` will always ignore the status of the tile even if it has been previously set as external. Indeed, it just targets internal tiles in the range, and if one of the tile is set as external (so it is RO), it does not care and sets to 0 the internal one (that won't be used at all, since it is hidden by the external tile).

I think that for the moment it is ok to keep this functionality as simple as it is, and  just add this specification to the documentation. What do you think, @rasolca?

test/unit/matrix/test_panel.cpp
```diff
@@ -394,3 +394,45 @@ TYPED_TEST(PanelTest, SetHeight) {
...
+    for (const auto& idx : panel.iteratorLocal())
+      hpx::dataflow(hpx::unwrapping(tile::laset<TypeParam>), lapack::MatrixType::General, 1, 1,
+                    panel(idx));
+
+    set0(hpx::launch::sync, panel);
+
+    // Note:
+    // `set0` does not inhibit from setting a tile as external.
+    //
+    // Unfortunately, this can be tested, because as soon as I access the panel
```
Sely85, line 420:
```suggestion
    // Unfortunately, this can not be tested, because as soon as I access the panel
```
?

include/dlaf/matrix/panel.h
rasolca: Fine with me.

albestro: As quickly discussed offline, we opted for making set0 more compliant with the Panel current constraints.

For this reason, it will just access all the tiles as a user would do it; it means that, since it access all the tiles of the currently active range in the panel, it will be able to work just if all tiles are RW, and after its call it won't be possible anymore to set tile as externals (until the next `reset`).

albestro:
bors try

rasolca (approved):

albestro:
bors try-

rasolca:
bors try

PR #403, opened 2021-07-27
Panel clean state assertions

albestro:
In the current proposal, panel methods can be split in two categories:
- setup: `setRange`, `setRangeStart`, `setRangeEnd`, `setWidth`/`setHeight`
- usage: `setTile`, `read`, `operator ()`

The rationale behind this division is that functions from the first group can be called exclusively just after construction or after a `reset()`, while the others have no restrictions (regarding this).

With this PR, we are going to enforce the above property.

Main points of this PR:
- Add panel member functions property enforcement with `DLAF_ASSERT` (do we want to make it MODERATE? @rasolca)
- Fix panel test that was not respecting the panel property (+ refactoring)
- minor: add `fixedValueTile` helper function for tests

### DIFF
--- a/include/dlaf/matrix/panel.h
+++ b/include/dlaf/matrix/panel.h
@@ -138,7 +138,7 @@ struct Panel<axis, const T, D> {
   /// @pre start <= 1 past the panel last tile
   /// @pre end <= 1 past the panel last tile
   void setRange(GlobalTileIndex start_idx, GlobalTileIndex end_idx) noexcept {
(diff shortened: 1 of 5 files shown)


rasolca (approved):
I would go with MODERATE.

Please also fix the PR summary including setTile.

albestro:
bors try

albestro:
bors try

albestro:
Tests passed correctly, just one task has failed due to a timeout caused by codecov reporting.

PR #404, opened 2021-07-28
Upper version of generalized to standard algorithm

Sely85:
Local and distributed versions of generalized to standard algorithm.


include/dlaf/eigensolver/gen_to_std/impl.h
```diff
@@ -89,11 +90,70 @@ void gemmPanelUpdateTile(Executor&& ex, hpx::shared_future<matrix::Tile<const T,
...
+}
+
+template <class Executor, Device device, class T>
+void gemmPanelUpdateTile(Executor&& ex, hpx::shared_future<matrix::Tile<const T, device>> a_ki,
+                         hpx::shared_future<matrix::Tile<const T, device>> u_ij,
+                         hpx::future<matrix::Tile<T, device>> a_kj) {
+  hpx::dataflow(ex, matrix::unwrapExtendTiles(tile::gemm_o), blas::Op::NoTrans, blas::Op::NoTrans,
+                T(-1.0), a_ki, u_ij, T(1.0), std::move(a_kj));
+}
+}
```
albestro, line 150: IMHO the tradeoff between code duplication and algorithm mapping here has dropped dramatically.

By giving a quick look (I hope I'm not missing something relevant), what changes between the functions in namespaces `gentostd_l` and `gentostd_u` are:
- naming of arguments (`l` for `u`, algorithm mapping for indices)
- `blas::Uplo::Upper` for `blas::Uplo::Lower` for BLAS/LAPACK calls
- some other paramerers of the BLAS/LAPACK calls depends on previous point (e.g. NoTrans/ConjTrans, Left/Right)

IMHO It would be ok to drop the first bullet point of the previous list of changes, i.e. can be a bit more "generic" to suite both variants, and the second point can be addressed differently, e.g. by passing a template parameter and putting the logic for depending parameters inside the function (and the call will become `trsmPanelUpdateTile<Uplo::Lower>(...)`).

To me both solutions does not look perfect, so I look forward to others feedbacks.

@rasolca @msimberg  @teonnik what do you think?

rasolca: I think this approach is the more readable.
This case is still quite simple, but imagine if you have to combine uplo and op cases:
do you prefer:
```
LN:
function(....)
  blas(hardcoded parameters, ...)
LT:
...
UN:
...
UT:
...
```
or
```
template <uplo, op>
function(...) //(or similarly without template function(uplo, op, ...))
  blas_param1 = (uplo == Lower) && (op==NoTrans) || (uplo == Upper) && (op!=NoTrans) ? value1 : value2
  ...
  blas(blas_param1, ...).
```?

Comment (rasolca) [reply include/dlaf/eigensolver/gen_to_std/impl.h:150]: Note: If I would apply your argument (code duplication) for algorithms as well, we should have a single implementation in which indices/panels/communications/computations depends on the value of the parameters.

Comment (albestro) [reply include/dlaf/eigensolver/gen_to_std/impl.h:150]: My note was not generic, but very specific for this case. I agree that in other cases, like the one of your example, the tradeoff changes in favor of having a custom version for each call.

> Note: If I would apply your argument (code duplication) for algorithms as well, we should have a single implementation in which indices/panels/communications/computations depends on the value of the parameters.

This is an extreme view of the thing, I didn't mean that.

However, I'm ok with this solution, as said in my first comment. I was just wondering about possible solutions.

Discussion

**Comment by rasolca** [comment]
bors try

File: include/dlaf/eigensolver/gen_to_std/impl.h [hunk 2]
```diff
@@ -342,6 +403,260 @@ void GenToStd<backend, device, T>::call_L(comm::CommunicatorGrid grid, Matrix<T,
...
+      }
+    }
+
+    broadcast(executor_mpi, kk_rank.row(), u_panel, u_panelT, mpi_row_task_chain, mpi_col_task_chain);
+
+    // continue update previous panels
+    // Note: The tasks of the final huge TRSM of the HEGST step have been reshuffled to avoid extra
+    //       communication of the matrix U.
+    //       During k-th iteration only the tasks involving the k-th panel of U are executed.
+    //       Therefore, all previous panel have to be updated at each step.
```
rasolca, line 533:
```suggestion
    //       Therefore, all previous panels have to be updated at each step.
```
Please fix it in the lower case as well.

albestro (approved):
LGTM

include/dlaf/eigensolver/gen_to_std/impl.h
```diff
@@ -342,6 +403,260 @@ void GenToStd<backend, device, T>::call_L(comm::CommunicatorGrid grid, Matrix<T,
...
+  constexpr std::size_t n_workspaces = 2;
+  common::RoundRobin<matrix::Panel<Coord::Row, T, device>> a_panels(n_workspaces, distr);
+  common::RoundRobin<matrix::Panel<Coord::Col, T, device>> a_panelsT(n_workspaces, distr);
+  common::RoundRobin<matrix::Panel<Coord::Row, T, device>> u_panels(n_workspaces, distr);
+  common::RoundRobin<matrix::Panel<Coord::Col, T, device>> u_panelsT(n_workspaces, distr);
+
+  for (SizeType k = 0; k < nrtile; ++k) {
+    const GlobalTileIndex kk{k, k};
+    const comm::Index2D kk_rank = distr.rankGlobalTile(kk);
+    const auto kt = k + 1;
```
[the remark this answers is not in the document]
Sely85: @albestro @rasolca: can you please check if I've applied this change correctly (both lower and upper version)?

rasolca (approved):

rasolca:
bors try

PR #406, opened 2021-08-04
Use fastcov instead of lcov

rasolca:
lcov is terribly slow.

Switched to fastcov (https://github.com/RPGillespie6/fastcov) which can output lcov format files.

### DIFF
--- a/ci/mpi-ctest
+++ b/ci/mpi-ctest
@@ -24,7 +24,7 @@ if [[ "$ENABLE_COVERAGE" == "YES" ]]; then
 
     fastcov --gcov gcov --process-gcno --branch-coverage --include /DLA-Future/ --exclude /DLA-Future/test/ /DLA-Future/miniapp/ --lcov -o "$LOCAL_REPORTS/run.info" &> /dev/null
 
-    cp "$LOCAL_REPORTS/combined.info" "$SHARED_REPORTS/codecov-$REPORT_NAME.info"
+    cp "$LOCAL_REPORTS/run.info" "$SHARED_REPORTS/codecov-$REPORT_NAME.info"
 
     TZ=CET date +"Done creating codecov reports from rank $SLURM_PROCID at: %H:%M:%S %z"
 fi


rasolca:
bors try

rasolca:
bors try

rasolca:
bors try

rasolca:
bors try

rasolca:
bors try

rasolca:
bors try

rasolca:
Results are very good: (example gpu coverage 6 ranks)
Pre:
Screenshot 2021-08-04 at 22 12 20
Post:
Screenshot 2021-08-04 at 22 11 33

rasolca:
Note: coverage is lower than before as if statements lines are marked as partially covered when only a branch is covered (lcov was instead marking them as fully covered).

albestro (approved):
Very nice job!

I just did a quick check since I don't really have much knowledge on this side, but it LGTM!

msimberg (approved):
Do I understand correctly that `fastcov` is fast because it's parallelized (it uses `gcov` under the hood)?

haampie:
wow, fastcov is faster than grcov too? :thinking: looks good

haampie (approved):

haampie:
The codecov link is not working for me though... :sweat_smile: requires me to login, and when i login, it redirects me to an unrelated page... clicking the link again, it forgets i was logged in already.

rasolca:
> Do I understand correctly that `fastcov` is fast because it's parallelized (it uses `gcov` under the hood)?

Yes, it uses gcov in parallel. But the speedup come from the lcov support for the gcc >= 9 output format, which is very bad and very slow.

> The codecov link is not working for me though... 😅 requires me to login, and when i login, it redirects me to an unrelated page... clicking the link again, it forgets i was logged in already.

I noticed that gitlab links have a lot of problems. On the other hand github links seems to work better now...
Anyway, for me solving this problem is currently a low priority task, and maybe the new uploader will fix them.

PR #407, opened 2021-08-11
algo: backtransfrom tridiagonal -> band (local)

albestro:
TODO:
- [x] Implement a test
- [x] Function name ~and files structure~
- [x] Solve TODOs
- [x] Documentation
- [x] Code cleanup

Changes:
- implement (local) Backtransformation from Tridiagonal to Band
- minor fix: `print(format::numpy{}` now handles nan values correctly
- minor fix: constness for `MatrixBase::tileSize()`
- minor fix: constraints for tau/reflector computation in `test_backtransformation`
- bug fix: missing quick return in computeTFactor


albestro:
bors try

albestro:
bors try-

albestro:
bors try

include/dlaf/eigensolver/backtransformation/impl-t2b.h
```diff
@@ -0,0 +1,317 @@
...
+static constexpr bool is_complex_v = std::is_same_v<T, ComplexType<T>>;
+
+template <class T>
+static SizeType nrSweeps(const SizeType m) {
+  return std::max<SizeType>(0, is_complex_v<T> ? m - 1 : m - 2);
+}
+
+static SizeType nrStepsPerSweep(SizeType sweep, SizeType m, SizeType mb) {
+  return std::max<SizeType>(0, sweep == m - 2 ? 1 : dlaf::util::ceilDiv(m - sweep - 2, mb));
+}
```
albestro, line 40: @rasolca the first one can be probably moved to our type traits in DLAF, but I'm more interested in the second twos, on how to share them, and if you think it is a good thing, between #413 and this PR.

Moreover, the same discussion can be done for the tests, where similar functions exists (without tile concept).

Since our PRs are strictly related, I think it may be a good idea to share also these functions. Moreover, this relationship may be an idea also for structuring differently this side of the library (instead of backtransformation, it may be worth having tridiagonalization and inside there both the "forward & reverse" operators.

albestro: After the call:
- I moved `is_complex_v` to `dlaf/types.h`
- I started moving this algo to its own folder `bt_band_to_tri` (name of the actual algo still to be adapted/changed)
- factor out `nrSweeps` and `nrStepsPerSweep` helpers also from the test

About the last point, I realized I was wrong: the same helpers appears both in the algorithm and in the test, there is no difference. I don't know if it is a good idea to share them between test and algo, but for sure we will share them between algorithms #413 and this one.

Looking forward to your comments.

test/unit/eigensolver/test_bt_band_to_tri.cpp
```diff
@@ -0,0 +1,140 @@
...
+      T& v_head = *mat_i_local.ptr({i_v, j});
+      const T tau = v_head;
+      v_head = 1;
+
+      using blas::Side;
+      lapack::larf(Side::Left, size, n, &v_head, 1, tau, mat_e_local.ptr({i, 0}), mat_e_local.ld());
+    }
+  }
+
+  auto result = [& dist = mat_e.distribution(),
```
rasolca, line 125: See #431.

albestro:
bors try

include/dlaf/eigensolver/bt_band_to_tridiag.h
```diff
@@ -0,0 +1,58 @@
...
+//                   0  0  0  D3
+//
+// where A, B, C and D refers to distinct reflectors, with their components numbered and their taus
+// identified by the letter T.
+//
+// @param mat_i matrix containing reflectors together with taus (compact form see representation above)
+// @param mat_e matrix to which the inverse transformation is applied to
+template <Backend backend, Device device, class T>
+void backTransformationBandToTridiag(matrix::Matrix<T, device>& mat_e,
+                                     matrix::Matrix<const T, device>& mat_i) {
```
rasolca, line 45: Just for curiosity: why the reflector matrix is called `I` instead of the more common `V`?

albestro: I for "input", because I thought it would have been misleading V, because it actually contains the reflectors and the taus in the "compact" form. But I'm open to change it if you think it is better to use V.

albestro:
bors try

albestro:
bors try

PR #408, opened 2021-08-11
algo: backtransform tridiagonal -> band (local) with subblocks (where band is a submultiple of blocksize)

albestro:
TODO:
- [ ] Decide if we want to keep this and #407 implementations separated for some reason
- [ ] Identify major problems with profiler (e.g. check workspace usage if can be improved to remove internal dependencies)
- [ ] Code cleanup


albestro:
This work is delayed and it will be addressed in the future.

PR #409, opened 2021-08-13
algo: triangular solver LLT (distributed)

albestro:
Implementation of the distributed Triangular Solver Left Lower Transposed.

TODO:
- [x] use of hp/np executors
- [x] documentation for extended reduce async machinery (and cleanup code)
- [x] adapt miniapp for LLT case (see #416)

Changelog:
- BUGFIX: `Panel::readwrite_sender` was returning `keep_future` sender, which led to wrong lifetime management by `Transform<GPU>`
- `set0` now works also on GPU tiles
- `set0` for matrix and panel refactored with senders/receivers
- moved `set(mat, elgetter, op)` to `dlaf/util_matrix.h` from `dlaf_test/matrix/util_matrix_blas.h`
- Adapt IReduce for GPU
- Integrate doc for `prepareSendTile`: it copies to contiguous buffer (relevant for GPU)
- ~`getGenericExecutor` workaround for CUDA executor~
- implement `subtract` as "BLAS tile extensions" (temporarily using Axpy for both CPU and GPU)

minor-fix:
- ~LLN was allocating a superfluous Panel~
- ~fix static_assert error message in Transform<GPU>~


albestro:
bors try

albestro:
bors try-

include/dlaf/solver/triangular/impl.h
```diff
@@ -424,6 +425,146 @@ void Triangular<backend, device, T>::call_LLN(comm::CommunicatorGrid grid, blas:
...
+    if (this_rank.col() == rank_kk.col()) {
+      for (SizeType i_loc = kk_offset.row(); i_loc < distr_a.localNrTiles().rows(); ++i_loc) {
+        const LocalTileIndex ik{i_loc, kk_offset.col()};
+        a_panel.setTile(ik, mat_a.read(ik));
+      }
+    }
+    comm::broadcast(executor_mpi, rank_kk.col(), a_panel, mpi_row_task_chain);
+
+    auto& b_panel = b_panels.nextResource();
+    b_panel.setHeight(mat_a.tileSize(kk).cols());  // TODO just if?
```
rasolca, line 523: Yes I would recommend it, such that it avoids creating unneeded continuations.

include/dlaf/communication/rdma.h
```diff
@@ -44,7 +44,7 @@ namespace internal {
 /// Duplicates the tile to CPU memory if CUDA RDMA is not enabled for MPI.
 /// Returns the tile unmodified otherwise.
 template <Device D, typename T>
-auto prepareSendTile(hpx::shared_future<matrix::Tile<const T, D>> tile) {
+auto prepareSendTile(hpx::shared_future<matrix::Tile<T, D>> tile) {
```
albestro, line 47: note-to-self: fix documentation stating that it duplicates to contiguous memory (in case it does)

albestro:
This is ready for a first round of review.

Probably, this is going to be rebased onto #371, which may be interesting to use for refactoring mechanisms for CPU/GPU memory management in MPI reduce wrappers.

include/dlaf/blas/tile_extensions.h
```diff
@@ -0,0 +1,52 @@
...
+namespace tile {
+using matrix::Tile;
+
+template <class T>
+void subtract(const matrix::Tile<T, Device::CPU>& tile_a,
+              const matrix::Tile<const T, Device::CPU>& tile_b) {
+  DLAF_ASSERT(equal_size(tile_a, tile_b), tile_a, tile_b);
+  for (auto j = 0; j < tile_a.size().cols(); ++j)
+    blas::axpy(tile_a.size().rows(), T(-1), tile_b.ptr({0, j}), 1, tile_a.ptr({0, j}), 1);
+}
```
rasolca, line 34: What about a more generic axpy like implementation for tiles?
Note: I would invert the order of the arguments to use the same convention as in BLAS (the modified matrix is the last matrix argument).

rasolca, line 34: What about a more generic "axpy like" implementation for tiles?
Note: I would invert the order of the arguments to use the same convention as in BLAS (the modified matrix is the last matrix argument).

rasolca (commented):
In general it looks good.

albestro:
I've simply rebased the branch on the new master with sender/receivers changes.

I will try to unify LLT implementation to sender/receiver changes and see if I can easily adapt other parts of this PR to the new paradigm (e.g. `add` tile extension), together with a possible simplification of `copyIfNeeded` thanks to C++17.

include/dlaf/util_matrix.h
```diff
@@ -137,27 +137,27 @@ class getter_random<std::complex<T>> : private getter_random<T> {
...
 }
 
 /// Sets all the elements of all the tiles in the active range to zero
-template <class T, Coord axis, class ExecutorOrPolicy>
-void set0(ExecutorOrPolicy ex, Panel<axis, T, Device::CPU>& panel) {
+template <class T, Coord axis, Device D, class ExecutorOrPolicy>
+void set0(ExecutorOrPolicy ex, Panel<axis, T, D>& panel) {
   for (const auto& tile_idx : panel.iteratorLocal())
-    panel(tile_idx).then(ex, hpx::unwrapping(tile::internal::set0_o));
+    hpx::dataflow(ex, matrix::unwrapExtendTiles(tile::internal::set0_o), panel(tile_idx));
```
albestro, line 150: Very minor thing, just for future reference and awareness.

I had to switch from `.then` to `.dataflow`, because as @msimberg said

> the [cuda] executor only customizes for dataflow, not then, and we didn't consider it necessary to make it work with then back then, and even less so now with senders being closer to reality

> it has async_execute which makes hpx::async work, and it has our custom dataflow_finalize which makes hpx::dataflow work correctly, but not then_execute

albestro:
bors try

albestro:
bors try

albestro:
bors try

albestro:
bors try

albestro:
bors -

albestro:
bors try-

albestro:
bors try

rasolca (approved):

include/dlaf/solver/triangular/impl.h
```diff
@@ -519,6 +521,91 @@ void Triangular<backend, device, T>::call_LLN(comm::CommunicatorGrid grid, blas:
...
+        const LocalTileIndex kj(kk_offset.row(), j_loc);
+        // TODO priority
+        const auto& priority = thread_priority::high;
+
+        dlaf::internal::whenAllLift(T(-1), b_panel.read_sender(kj), mat_b.readwrite_sender(kj)) |
+            tile::add(dlaf::internal::Policy<backend>(priority)) |
+            hpx::execution::experimental::detach();
+
+        trsmBPanelTile<backend>(priority, op, diag, alpha, a_panel.read_sender(kk_offset),
+                                mat_b.readwrite_sender(kj));
```
albestro, line 600: @msimberg in this kind of context, would you suggest using a `let_value(mat_b.readwrite_sender)` and do both `tile::add` and the work of `trsmBPanelTile` internally?

IMHO is should be better from a continuation point of view, but knowing the (current?) limitations with `reference_wrapper` and template deduction, the code may become more verbose and uglier (e.g. we cannot use the pipe operator inside, and we have to create a `ex::then()` with a lambda that helps the "helps" the template deduction)

msimberg: I would probably not suggest that, simply because of the big reduction in clarity and separation of the operations. It may indeed lead to a slight improvement in runtimes but I doubt the effect would be huge. If I understand you correctly you're essentially suggesting to "fuse" the access to the tile in `mat_b` into a single operation, correct? It could be beneficial but I would assume there's quite a lot of other higher impact changes that would come before this.

Sely85 (approved):
It looks good to me!

teonnik (approved):
LGTM!

rasolca:
bors try

rasolca:
bors try

rasolca:
Closing and reopening to force run of new format check (#431).

Issue #411, opened 2021-08-17
initialize/finalize RAII wrapper


msimberg:
@albestro discovered a problem in the triangular solver miniapp, where we call `dlaf::finalize` explicitly at the end of `hpx_main`, but the `MatrixMirror` destructor tries to synchronize the matrices after `finalize` is called. This is is an error because the CUDA stream pool has been released by `finalize`.

I think we just need a simple RAII wrapper for finalize to make it easier to avoid these kind of problems. In the simplest case:

```
namespace dlaf {
struct initializer {
    initializer(int argc, char* argc[]) { initialize(argc, argv); }
    ~initializer() { finalize();
};
}
```

This ensures that resources are freed in the right order. Any objections or other ideas?

albestro:
About the solution you are proposing, I think it may not really solve the problem but just shift it.

What we have usually is something like

```cpp
int foo() {
  dlaf::initialize();

  MatrixMirror m;

  dlaf::finalize();
  return hpx::finalize()
}
```

with the order for d'tors being
1. dlaf
2. hpx
3. matrix mirror

The problem is that we would like to have MatrixMirror being destroyed before all the other ones.

If we make our initializer RAII compliant, we would end up having something like
```cpp
int foo() {
  dlaf::initializer dlaf_init;

  MatrixMirror m;

  return hpx::finalize()
}
```

And at this point the d'tors order will be:
1. hpx
2. MatrixMirror
3. dlaf

where the original problem is improved with respect to dlaf, but now it has been moved to `hpx`.

PR #412, opened 2021-08-20
algo: triangular solver LLT (distributed) [adaptations for GPU]

albestro:
This PR is a fork of the main #409, to which adds the needed adaptions for making it work on the GPU too.

Main adaptions:
1. Generic CPU/GPU code for `tile_a -= tile_b` operation (currently carried out column by column via `axpy`)
2. `set0` helper function for GPU
3. Generic executor for `set0`, because `cuSolver`/`cuBlas` executors currently do not dispatch to the basic CUDA one (thanks @msimberg for the support)
4. Adapt MPI async reduce machinery to GPU

In addition to these, there is also a quick workaround in the miniapp for the problem described in #411 (again, thanks @msimberg for the support).

### DIFF
--- a/include/dlaf/communication/rdma.h
+++ b/include/dlaf/communication/rdma.h
@@ -44,7 +44,7 @@ namespace internal {
 /// Duplicates the tile to CPU memory if CUDA RDMA is not enabled for MPI.
 /// Returns the tile unmodified otherwise.
 template <Device D, typename T>
-auto prepareSendTile(hpx::shared_future<matrix::Tile<const T, D>> tile) {
+auto prepareSendTile(hpx::shared_future<matrix::Tile<T, D>> tile) {
   return matrix::duplicateIfNeeded<CommunicationDevice<D>::value>(std::move(tile));
 }
 
--- a/include/dlaf/solver/triangular/impl.h
+++ b/include/dlaf/solver/triangular/impl.h
@@ -442,7 +442,8 @@ struct getGenericExecutor<Backend::GPU> {
 #endif
 
 template <class T>
-void tileSubtractInPlace(matrix::Tile<T, Device::CPU> const & tile_a, matrix::Tile<const T, Device::CPU> const& tile_b) {
+void tileSubtractInPlace(matrix::Tile<T, Device::CPU> const& tile_a,
+                         matrix::Tile<const T, Device::CPU> const& tile_b) {
   DLAF_ASSERT(equal_size(tile_a, tile_b), tile_a, tile_b);
   for (auto j = 0; j < tile_a.size().cols(); ++j)
     blas::axpy(tile_a.size().rows(), T(-1), tile_b.ptr({0, j}), 1, tile_a.ptr({0, j}), 1);
@@ -450,12 +451,14 @@ void tileSubtractInPlace(matrix::Tile<T, Device::CPU> const & tile_a, matrix::Ti
 
 #ifdef DLAF_WITH_CUDA
 template <class T>
-void tileSubtractInPlace(cublasHandle_t handle, matrix::Tile<T, Device::GPU> const & tile_a, matrix::Tile<const T, Device::GPU> const& tile_b) {
+void tileSubtractInPlace(cublasHandle_t handle, matrix::Tile<T, Device::GPU> const& tile_a,
+                         matrix::Tile<const T, Device::GPU> const& tile_b) {
   DLAF_ASSERT(equal_size(tile_a, tile_b), tile_a, tile_b);
   const T alpha = -1;
   for (auto j = 0; j < tile_a.size().cols(); ++j)
     tile::internal::CublasAxpy<T>::call(handle, tile_a.size().rows(), util::blasToCublasCast(&alpha),
-        util::blasToCublasCast(tile_b.ptr({0, j})), 1, util::blasToCublasCast(tile_a.ptr({0, j})), 1);
+                                        util::blasToCublasCast(tile_b.ptr({0, j})), 1,
+                                        util::blasToCublasCast(tile_a.ptr({0, j})), 1);
 }
 #endif
 
@@ -463,8 +466,7 @@ DLAF_MAKE_CALLABLE_OBJECT(tileSubtractInPlace);
 
 template <Backend backend, Device D, class T>
 void Triangular<backend, D, T>::call_LLT(comm::CommunicatorGrid grid, blas::Op op, blas::Diag diag,
-                                              T alpha, Matrix<const T, D>& mat_a,
-                                              Matrix<T, D>& mat_b) {
+                                         T alpha, Matrix<const T, D>& mat_a, Matrix<T, D>& mat_b) {
   constexpr auto NoTrans = blas::Op::NoTrans;
 
   auto executor_hp = dlaf::getHpExecutor<backend>();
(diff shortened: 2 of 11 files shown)


albestro:
bors try

albestro:
bors try-

albestro:
bors try

include/dlaf/solver/triangular/impl.h
```diff
@@ -425,10 +425,48 @@ void Triangular<backend, device, T>::call_LLN(comm::CommunicatorGrid grid, blas:
...
+template <class T>
+void tileSubtractInPlace(cublasHandle_t handle, matrix::Tile<T, Device::GPU> const& tile_a,
+                         matrix::Tile<const T, Device::GPU> const& tile_b) {
+  DLAF_ASSERT(equal_size(tile_a, tile_b), tile_a, tile_b);
+  const T alpha = -1;
+  for (auto j = 0; j < tile_a.size().cols(); ++j)
+    tile::internal::CublasAxpy<T>::call(handle, tile_a.size().rows(), util::blasToCublasCast(&alpha),
+                                        util::blasToCublasCast(tile_b.ptr({0, j})), 1,
+                                        util::blasToCublasCast(tile_a.ptr({0, j})), 1);
+}
```
msimberg, line 462: Smart, I didn't think about batching operations together like this!

There's a potential tradeoff in that submitting each column seperately as a task (i.e. on a potentially separate stream) increases parallelism, but it also means paying the overhead of one future per column instead of one for the whole operation.

Question: do you know if there's a single batched operation that would do this in one blas/cublas call? I guess if the stride were the same as the row/column length you could call a single axpy with length rows * cols but for other strides it won't work.

include/dlaf/solver/triangular/impl.h
```diff
@@ -505,17 +543,13 @@ void Triangular<backend, device, T>::call_LLT(comm::CommunicatorGrid grid, blas:
     if (this_rank.row() == rank_kk.row()) {
       for (SizeType j_loc = 0; j_loc < distr_b.localNrTiles().cols(); ++j_loc) {
         const LocalTileIndex kj(kk_offset.row(), j_loc);
-        auto tile_diff = hpx::unwrapping([](auto tile_a, const auto& tile_b) {
-          DLAF_ASSERT(equal_size(tile_a, tile_b), tile_a, tile_b);
-          for (auto j = 0; j < tile_a.size().cols(); ++j)
-            blas::axpy(tile_a.size().rows(), T(-1), tile_b.ptr({0, j}), 1, tile_a.ptr({0, j}), 1);
-        });
         // TODO executor
```
msimberg, line 546:
```suggestion
```
probably?

include/dlaf/solver/triangular/impl.h
```diff
@@ -425,10 +425,48 @@ void Triangular<backend, device, T>::call_LLN(comm::CommunicatorGrid grid, blas:
...
+};
+
+#ifdef DLAF_WITH_CUDA
+template <>
+struct getGenericExecutor<Backend::GPU> {
+  static auto call() {
+    return dlaf::cuda::Executor{dlaf::internal::getNpCudaStreamPool()};
+  }
+};
+#endif
```
msimberg, line 442: :+1: 

Perhaps a comment saying why this is here instead of in `executors.h`?

include/dlaf/communication/rdma.h
```diff
@@ -44,7 +44,7 @@ namespace internal {
 /// Duplicates the tile to CPU memory if CUDA RDMA is not enabled for MPI.
 /// Returns the tile unmodified otherwise.
 template <Device D, typename T>
-auto prepareSendTile(hpx::shared_future<matrix::Tile<const T, D>> tile) {
+auto prepareSendTile(hpx::shared_future<matrix::Tile<T, D>> tile) {
```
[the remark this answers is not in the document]
albestro: Not really. Following a discussion we had few weeks ago about this kind of stuff, for adapting the communication machinery for the GPU I opted for using `future::share()`, from which I get a `shared_future<Tile<T>>`.

This is potentially problematic because it opens to potential race-conditions, but as discussed, we limit this problem to this specific section of the code.

https://github.com/eth-cscs/DLA-Future/blob/b58654a6ed42734c5f32731a47817fbd1922c587/include/dlaf/communication/kernels/reduce.h#L82-L94

Maybe this code can be improved, so I will try to clarify a bit what I'm doing here (I will add/fix the documentation too).

The code mentioned covers different cases:
- INPUT: CPU vs GPU
- CPU-LAYOUT: contiguous vs non-contiguous

These are addressed in the following phases:
1. (compile-time) ensure that the input is available on CPU
2. (runtime) ensure that the CPU-LAYOUT is contiguous (temporary contiguous)
3. !! COMMUNICATION !!
4. (runtime) if temporary contiguous has been used, copy it back
5. (compile-time) if the input was on GPU, copy it back

The facility responsible for the 1st step is `prepareSendTile`, whose job is to return the same tile OR another temporary one, but whatever branch it takes, the returned tile will be on CPU.

If `prepareSendTile` would accept a `future`, it has to be moved in, so that:
- in the first case it can "pass-through" and get returned as it is;
- in the second case it will be just used as source of the copy.

But, in the latter case it means that we are going to loose the "handle" on the original tile, so we won't have it anymore for the 5th step where we will copy back results.

Sorry for the verbosity, but I wanted to explain (to myself) exactly what is going on in order to see if there may be room for improvements. Otherwise, I will just use this comment as reference for the documentation I'm going to write for the function.

include/dlaf/solver/triangular/impl.h
albestro: As pointed out by @rasolca, the best solution would be checking if MAGMA provides anything useful that operates in block, or just write a custom GPU kernel doing that.

This is some kind of workaround, let me know if it is acceptable for the moment.

include/dlaf/solver/triangular/impl.h
albestro: Is it correct saying that it is a workaround because CuSolver/CuBlas executors do not use the basic CUDA executor as fallback?

include/dlaf/solver/triangular/impl.h
albestro: I left this comment here (and elsewhere) because I still have to verify where I have to use NP/HP executors before merging 😉

include/dlaf/communication/rdma.h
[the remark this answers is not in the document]
msimberg: The explanation makes sense. I was more wondering about the general case, but given that this code might not survive too much longer we can take that when and if that change is needed. I just wanted to make sure I understood the change here.

include/dlaf/solver/triangular/impl.h
msimberg: Yep, that's accurate!

include/dlaf/solver/triangular/impl.h
msimberg: Fine by me at least, we'll have time to update it to something more performant in the future.

albestro:
I still have to update the documentation for IReduce wrappers machinery.

include/dlaf/util_matrix.h
```diff
@@ -137,17 +137,17 @@ class getter_random<std::complex<T>> : private getter_random<T> {
 }
 
 /// Sets all the elements of all the tiles to zero
-template <class T, class ExecutorOrPolicy>
-void set0(ExecutorOrPolicy ex, Matrix<T, Device::CPU>& matrix) {
+template <class T, Device D, class ExecutorOrPolicy>
+void set0(ExecutorOrPolicy ex, Matrix<T, D>& matrix) {
   for (const auto& idx : iterate_range2d(matrix.distribution().localNrTiles()))
-    matrix(idx).then(ex, hpx::unwrapping(tile::set0<T>));
+    matrix(idx).then(ex, matrix::unwrapExtendTiles(tile::set0_o));
```
albestro, line 143: note to self: use `dataflow` instead of `.then` because of the (current) limitation of the CUDA executor

albestro:
I'm going to merge this fork in the main one (#409)

PR #416, opened 2021-09-01
miniapp: add LLT case for triangular solver

albestro:
Extend the miniapp to address, in addition to the currently available LLN case, also the new LLT one, through a CLI option.

- move `set(..., op)` helper to DLAF
- workaround for #411


msimberg (approved):
Looks good to me. The miniapp needs reformatting though.

PR #418, opened 2021-09-13
Reformat python scripts with black

albestro:


msimberg:
Looks good to me, though it looks like you might have to update your version of black (or we install exactly your version in CI?). Might be good to fix the version in any case to avoid surprises when there's a newer version available. I'm fine with downgrading my `black`, but it might make sense to start with the latest version available right now (which seems to be `21.8b0`).

albestro:
> Looks good to me, though it looks like you might have to update your version of black (or we install exactly your version in CI?). Might be good to fix the version in any case to avoid surprises when there's a newer version available. I'm fine with downgrading my `black`, but it might make sense to start with the latest version available right now (which seems to be `21.8b0`).

I was already doing that, I've just seen your comment 😄 You also anticipated my question about which version to use.

Last test I would like to do (because I'm not really sure what I did in github actions) is if the existing formatting checks still work as expected.

msimberg (approved):

teonnik (approved):

Sely85 (approved):
LGTM

PR #423, opened 2021-09-23
Pipeline fixes: non-blocking d'tor

### LINKED ISSUE #417: Pipeline are blocking
I noticed that Pipeline blocks on destruction.

What is the reason behind it?

albestro:
Close #417

(I write here for future reference what we found out with @rasolca)

Removing the blocking call from the `Pipeline` d'tor exposed two main problems:
- ~Resource management with `PromiseGuard`. For instance with `PromiseGuard<Communicator>` used with asynchronous communications, it suffered of the same problem of Tiles, i.e. it has to be kept alive till the completion of the operation and not just for the posting of the operation.~
  EDIT: this is not true, indeed the Communicator has to be kept just for the posting of the operation, not until its completion. Otherwise we end up serializing communications, completely missing the point of asynchronous MPI.
- Shared communicators: it is something we knew, but it shown up in a subtle way. Using the same communicator for multiple algorithms, may end up in mixed/crossed communications resulting in undefined/strange behaviours (e.g. in the case I was debugging it was an "MPI Invalid Communicator"). This happened in `test_gen_to_std` https://github.com/eth-cscs/DLA-Future/blob/41ded01d1f1311a24a80e6ec86c7415df6c7bda4/test/unit/eigensolver/test_gen_to_std.cpp#L135-L142 where the same algorithm is executed sequentially on exactly the same `CommunicatorGrid`. The problem was raised in case one of the rank was able to start the 2nd configuration before other ranks were able to finish the 1st one (e.g. a rank that does not have any tile to work on), so communications for the 2nd were posted mixed up with the 1st one.

Both these problems will need a better management and we can think of redesigning a bit how it works, this is somehow a temporary solution.

CHANGES:
- Pipeline d'tor does not block anymore
- Due to previous point, algorithm call is not "blocking" anymore. This resulted in a deadlock condition in an edge case which has been workaround with by all HPX tasks (see https://github.com/eth-cscs/DLA-Future/pull/423#issuecomment-958735416)
- ~Extend `matrix::unwrapExtendTiles` to extend lifetime also for `PromiseGuard` (and adapt asynchronous communication kernels accordingly)~
- Clone communicators got from `CommunicatorGrid`s inside each algorithm (see https://github.com/eth-cscs/DLA-Future/pull/423#issuecomment-958716222)
- Remove unused `recvBcastAlloc` (and adapt `test_comm_matrix` that was using it)

~Even if this solution is temporary, IMHO it may be worth improving the naming of `unwrapExtendTiles`, e.g. since at this point we are extending the concept of lifetime extension not just to tiles.~

### DIFF
--- a/include/dlaf/communication/kernels/broadcast.h
+++ b/include/dlaf/communication/kernels/broadcast.h
@@ -87,7 +87,7 @@ struct ScheduleRecvBcast {
                    comm::IndexT_MPI root_rank, hpx::future<common::PromiseGuard<Communicator>> pcomm) {
     using hpx::dataflow;
     using matrix::duplicateIfNeeded;
-    using matrix::copy_o;
+    using matrix::internal::copy_o;
 
     // Note:
     //
@@ -104,7 +104,8 @@ struct ScheduleRecvBcast {
     tile_cpu = std::move(hpx::get<0>(hpx::split_future(
         ScheduleRecvBcast<T>::call(ex, std::move(tile_cpu), root_rank, std::move(pcomm)))));
 
-    dataflow(getCopyExecutor<Device::GPU, Device::CPU>(), unwrapExtendTiles(copy_o), tile_cpu, tile_gpu);
+    dataflow(getCopyExecutor<Device::GPU, Device::CPU>(), matrix::unwrapExtendTiles(copy_o), tile_cpu,
+             tile_gpu);
   }
 };
 
--- a/include/dlaf/communication/kernels/broadcast.h
+++ b/include/dlaf/communication/kernels/broadcast.h
@@ -32,6 +32,10 @@ namespace comm {
 template <class T, Device D>
 void sendBcast(const matrix::Tile<const T, D>& tile, const common::PromiseGuard<Communicator>& pcomm,
                MPI_Request* req) {
+#if !defined(DLAF_WITH_CUDA_RDMA)
+  static_assert(D == Device::CPU, "With CUDA_RDMA disabled, MPI accepts just CPU memory.");
+#endif
+
   const auto& comm = pcomm.ref();
   auto msg = comm::make_message(common::make_data(tile));
   DLAF_MPI_CALL(
@@ -43,6 +47,10 @@ DLAF_MAKE_CALLABLE_OBJECT(sendBcast);
 template <class T, Device D>
 void recvBcast(const matrix::Tile<T, D>& tile, comm::IndexT_MPI root_rank,
                const common::PromiseGuard<Communicator>& pcomm, MPI_Request* req) {
+#if !defined(DLAF_WITH_CUDA_RDMA)
+  static_assert(D == Device::CPU, "With CUDA_RDMA disabled, MPI accepts just CPU memory.");
+#endif
+
   auto msg = comm::make_message(common::make_data(tile));
   DLAF_MPI_CALL(MPI_Ibcast(msg.data(), msg.count(), msg.mpi_type(), root_rank, pcomm.ref(), req));
 }
@@ -57,12 +65,41 @@ void scheduleSendBcast(const comm::Executor& ex, Future<matrix::Tile<const T, D>
   hpx::dataflow(ex, unwrapExtendTiles(sendBcast_o), prepareSendTile(std::move(tile)), std::move(pcomm));
 }
 
+namespace internal {
+
+template <class T>
+struct ScheduleRecvBcast {
+  static auto call(const comm::Executor& ex, hpx::future<matrix::Tile<T, Device::CPU>> tile,
+                   comm::IndexT_MPI root_rank, hpx::future<common::PromiseGuard<Communicator>> pcomm) {
+    using hpx::dataflow;
+    using matrix::unwrapExtendTiles;
+
+    return dataflow(ex, unwrapExtendTiles(recvBcast_o), std::move(tile), root_rank, std::move(pcomm));
+  }
+
+  static void call(const comm::Executor& ex, hpx::future<matrix::Tile<T, Device::GPU>> tile,
+                   comm::IndexT_MPI root_rank, hpx::future<common::PromiseGuard<Communicator>> pcomm) {
+    using hpx::dataflow;
+    using matrix::duplicateIfNeeded;
+    using matrix::copy_o;
+
+    auto tile_gpu = tile.share();
+    auto tile_cpu = duplicateIfNeeded<Device::CPU>(tile_gpu);
+
+    tile_cpu = std::move(hpx::get<0>(hpx::split_future(
+        ScheduleRecvBcast<T>::call(ex, std::move(tile_cpu), root_rank, std::move(pcomm)))));
+
+    dataflow(getCopyExecutor<Device::GPU, Device::CPU>(), unwrapExtendTiles(copy_o), tile_cpu, tile_gpu);
+  }
+};
+
+}
+
 template <class T, Device D>
 void scheduleRecvBcast(const comm::Executor& ex, hpx::future<matrix::Tile<T, D>> tile,
                        comm::IndexT_MPI root_rank,
                        hpx::future<common::PromiseGuard<Communicator>> pcomm) {
-  using matrix::unwrapExtendTiles;
-  hpx::dataflow(ex, unwrapExtendTiles(recvBcast_o), std::move(tile), root_rank, std::move(pcomm));
+  internal::ScheduleRecvBcast<T>::call(ex, std::move(tile), root_rank, std::move(pcomm));
 }
 
 }
--- a/include/dlaf/communication/kernels/broadcast.h
+++ b/include/dlaf/communication/kernels/broadcast.h
@@ -29,102 +29,41 @@
 namespace dlaf {
 namespace comm {
 
-// Non-blocking sender broadcast
 template <class T, Device D>
-void sendBcast(matrix::Tile<const T, D> const& tile, common::PromiseGuard<Communicator> pcomm,
+void sendBcast(matrix::Tile<const T, D> const& tile, common::PromiseGuard<Communicator> const& pcomm,
                MPI_Request* req) {
+  const auto& comm = pcomm.ref();
   auto msg = comm::make_message(common::make_data(tile));
-  DLAF_MPI_CALL(MPI_Ibcast(const_cast<T*>(msg.data()), msg.count(), msg.mpi_type(), pcomm.ref().rank(),
-                           pcomm.ref(), req));
+  DLAF_MPI_CALL(
+      MPI_Ibcast(const_cast<T*>(msg.data()), msg.count(), msg.mpi_type(), comm.rank(), comm, req));
 }
 
 DLAF_MAKE_CALLABLE_OBJECT(sendBcast);
 
-// Non-blocking receiver broadcast
 template <class T, Device D>
-matrix::Tile<T, D> recvBcast(matrix::Tile<T, D> tile, comm::IndexT_MPI root_rank,
-                             common::PromiseGuard<Communicator> pcomm, MPI_Request* req) {
+void recvBcast(matrix::Tile<T, D> const& tile, comm::IndexT_MPI root_rank,
+               common::PromiseGuard<Communicator> const& pcomm, MPI_Request* req) {
   auto msg = comm::make_message(common::make_data(tile));
   DLAF_MPI_CALL(MPI_Ibcast(msg.data(), msg.count(), msg.mpi_type(), root_rank, pcomm.ref(), req));
-  return tile;
 }
 
 DLAF_MAKE_CALLABLE_OBJECT(recvBcast);
 
-// Non-blocking receiver broadcast (with Alloc)
-template <class T, Device D>
-matrix::Tile<const T, D> recvBcastAlloc(TileElementSize tile_size, comm::IndexT_MPI root_rank,
-                                        common::PromiseGuard<Communicator> pcomm, MPI_Request* req) {
-  using Tile_t = matrix::Tile<T, D>;
-  using ConstTile_t = matrix::Tile<const T, D>;
-  using MemView_t = memory::MemoryView<T, D>;
-
-  MemView_t mem_view(tile_size.rows() * tile_size.cols());
-  Tile_t tile(tile_size, std::move(mem_view), tile_size.rows());
-
-  auto msg = comm::make_message(common::make_data(tile));
-  DLAF_MPI_CALL(MPI_Ibcast(msg.data(), msg.count(), msg.mpi_type(), root_rank, pcomm.ref(), req));
-  return ConstTile_t(std::move(tile));
-}
-
 template <class T, Device D, template <class> class Future>
 void scheduleSendBcast(const comm::Executor& ex, Future<matrix::Tile<const T, D>> tile,
-                       hpx::future<common::PromiseGuard<comm::Communicator>> pcomm) {
-  hpx::dataflow(ex, matrix::unwrapExtendTiles(sendBcast_o), internal::prepareSendTile(std::move(tile)),
-                std::move(pcomm));
+                       hpx::future<common::PromiseGuard<Communicator>> pcomm) {
+  using matrix::unwrapExtendTiles;
+  using internal::prepareSendTile;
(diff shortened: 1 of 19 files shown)


albestro:
bors try

include/dlaf/matrix/tile.h
```diff
@@ -508,6 +510,15 @@ struct UnwrapFuture<hpx::future<Tile<T, D>>> {
...
 
+template <typename T>
+struct UnwrapFuture<hpx::future<dlaf::common::PromiseGuard<T>>> {
+  template <typename U>
+  static auto call(U&& u) {
+    auto t = u.get();
+    return t;
+  }
+};
+
```
msimberg, line 521: I'd suggest changing the specialization above this to just work for any future.

Also, I have no idea why I did `auto t = u.get(); return t;` instead of just `return u.get();`... I think that should be possible to change.

Finally, agreed on changing the name of `unwrapExtendTiles`. Perhaps `unwrapExtendArgs` or something similarly generic?

albestro:
bors try

albestro:
bors try

albestro:
About cloning `Communicator`s/`CommuicatorGrid`s inside the algorithms: it is a temporary solution we opted for, but the general idea is to change it so that `Communicator`s will provide a `Pipeline` directly, e.g. it internally will have a fixed set of `Pipeline`s that can be used by the different algorithms in a round-robin fashion.

albestro:
About the deadlock and the temporary workaround with `hpx::threads::get_thread_manager().wait()`.

When running tests on PizDaint GPU partition, which run on a single node, with 6 ranks we get `12 cores / 6 ranks = 2 cores/rank`. Considering that 1 thread is dedicated to the MPI pool for asynchronous communication (`mpi` pool), in this configuration we end up having a single core for the `default` pool.

On the `deafult` pool runs both the computations tasks of the algorithms, and also the so called `hpx_main`, the one which we generally use just for scheduling things.

Let's see what was creating the deadlock in the tests.
The algorithm **A0** exits, since the call is not blocking anymore, but after it the test waits for the local tiles for checking the results. Anyway, this does not imply that all algorithm tasks finished, for instance there may be some computation (e.g. on a panel) that are not relevant for the local tiles but does for the other ranks, i.e. they have to be computed and then communicated.
So, we have some algorithm tasks not yet completed/scheduled, but the local check can complete and go to the next test config starting the algorithm **A1**, which as first thing does the `MPI_Comm_dup` in the `hpx_main` task.

And here it is the problem: `MPI_Comm_dup` is a collective AND a blocking call, which means that does not complete until all ranks do the same call. But, we are in a condition where:
- ranks which already completed **A0** are stuck in `MPI_Comm_dup` and cannot do anything else until all other ranks reach the same point; they are stuck because the only core available in `default` is blocked. This implies that any other tasks for **A0** not yet complete cannot be run, and in turn they cannot unlock the related communications that would permit the advancement of other ranks;
- ranks which are still executing **A0** are waiting data, that will be never communicated, and so they will never reach the `MPI_Comm_dup`.

And here it is the deadlock. This is what we found out with `gen_to_std`, but a similar scenario happens for `reduction_to_band`, where instead of having the `MPI_Comm_dup` we have an MPI collective call (i.e. `MPI_AllGather`) for the check phase of the test, that creates exactly the same problem.

The workaround at the moment is to wait for all the tasks before scheduling a collective blocking MPI call. When the algorithm exists, it already scheduled in HPX all the tasks, so we are making the algorithm blocking again.

*Note: the barrier we add with `hpx::threads::get_thread_manager().wait()` is stronger than the Pipeline blocking d'tor. Indeed, the Pipeline blocking d'tor was waiting for the last MPI task to be posted in MPI, not even that it was finished. Moreover, with this kind of barrier now we are waiting for anything to complete before continuing (i.e. we wait also for computations, not just mpi).*

include/dlaf/matrix/tile.h
albestro: An update on this: working on this PR I got misled and I thought that the PromiseGuard lifetime had to be extended. I was wrong (thanks @rasolca), because the PromiseGuard, which in this PR case provides a `Communicator` has to be alive during the posting of the MPI operation, and it is not needed for its completion.

Keeping it alive is not just superfluous, but it's wrong: by keeping it alive till the completion of the operation, it does not allow the pipeline to continue until the MPI operation is fully completed, leading to a sequential execution of async MPI calls (= no sense).

So, I just updated the existing `unwrapExtendTiles` with your suggestion about `return u.get()`.

teonnik (commented):
LGTM so far!

1. The solution in https://github.com/eth-cscs/DLA-Future/pull/423#issuecomment-958716222 seems reasonable
2. Regarding the temporary workaround : `hpx::threads::get_thread_manager().wait().`, was there a discussion for a permanent solution?

albestro:
bors try

albestro:
bors try-

albestro:
bors try

albestro:
bors try-

albestro:
bors try
this should be the good one 🤞🏻 sorry for the spam

include/dlaf/communication/kernels/broadcast.h
```diff
@@ -29,102 +29,93 @@
...
 
-// Non-blocking receiver broadcast
 template <class T, Device D>
-matrix::Tile<T, D> recvBcast(matrix::Tile<T, D> tile, comm::IndexT_MPI root_rank,
-                             common::PromiseGuard<Communicator> pcomm, MPI_Request* req) {
+void recvBcast(const matrix::Tile<T, D>& tile, comm::IndexT_MPI root_rank,
+               const common::PromiseGuard<Communicator>& pcomm, MPI_Request* req) {
+#if !defined(DLAF_WITH_CUDA_RDMA)
+  static_assert(D == Device::CPU, "DLAF_WITH_CUDA_RDMA=off, MPI accepts just CPU memory.");
+#endif
```
msimberg, line 52: Good idea!

include/dlaf/eigensolver/reduction_to_band/mc.h
```diff
@@ -792,8 +792,8 @@ std::vector<hpx::shared_future<common::internal::vector<T>>> ReductionToBand<
   const auto ex_mpi = getMPIExecutor<Backend::MC>();
 
   common::Pipeline<comm::Communicator> mpi_col_chain_panel(grid.colCommunicator().clone());
-  common::Pipeline<comm::Communicator> mpi_row_chain(grid.rowCommunicator());
-  common::Pipeline<comm::Communicator> mpi_col_chain(grid.colCommunicator());
+  common::Pipeline<comm::Communicator> mpi_row_chain(grid.rowCommunicator().clone());
+  common::Pipeline<comm::Communicator> mpi_col_chain(grid.colCommunicator().clone());
```
msimberg, line 796: I'm just curious, is there a case where the communicator should *not* be cloned before sticking it in a pipeline?

msimberg (approved):
Looks good to me.

albestro:
@teonnik 

>     2. Regarding the temporary workaround : `hpx::threads::get_thread_manager().wait().`, was there a discussion for a permanent solution?

What do you think @rasolca? The idea was to wait for a re-design of the pipeline management, but I don't know if we want to keep it like this till we will do it.

include/dlaf/eigensolver/reduction_to_band/mc.h
albestro: It's a good point!

The job of the `Pipeline` is to make the access to a resource serialized to ensure order. Ensuring this across different algorithms would require to move out the Pipeline from the algorithm (like we briefly discussed in the meeting, e.g. by having them in the `Communicator`) so the "continuity" of the order can be maintained.

The quick solution we opted for is to `clone` (`MPI_Comm_dup`) in order to get completely separate `Communicators`, which by construction do not "collide" across different algorithms.

So, the answer I would say is "yes, it always gets a clone". On the other hand, at the moment I cannot think of other quick/nice solutions (? embedding the clone in the `Pipeline<Communicator>` specialization).

msimberg: Yeah, putting the clone inside the construction what I was going for *if* it's always going to be cloned. But having it explicit is also fine by me. Since it'll hopefully be changed soon I don't mind leaving it like this.

rasolca:
Regarding the temporary workaround:
The race condition problem is not simply given by MPI_Comm_dup, but can be triggered from any blocking MPI communication used in tests (e.g gathering a matrix).

A summary:
- the problem: any collective blocking MPI call (collectives communications and communicator creation) called from the "main" thread after a taskbased algorithm can deadlock if using 2 threads, (i.e. 1 mpi thread + 1 compute thread) if not fully synchronized. I would separate the possible occurrence in two groups:
  - correct library usage: due to communicator duplication to allow algorithm overlap -> can be solved by creating pipelines directly in the communicator grid constructor and using them.
  - not recommended application usage but we use it in some tests: due to collective communication in matrix allGather -> 2 solutions possible (syncronization or async gather) which has to be analyzed.

albestro:
Merge of this PR is going to be slightly delayed. Probably after #425 so that I can introduce the same changes in there too.

albestro:
bors try

albestro:
In the last CI run we address another deadlock scenario.

With @rasolca we have a rough idea of what's going on, but we are still investigating the exact reason. Basically, the deadlock shows up in a similar way as previous ones, but where the MPI blocking call is submitted by our custom MPI Listener that at the end of each test collects output from all the ranks (via MPI sync calls).

As soon as I will have more insights, I will comment here. In the meanwhile, I would still not merge this PR.

albestro:
UPDATE: some advancements about the deadlock we were not able to fully explain.

The problem shows up from time to time using the `test_broadcast_panel` with `--hpx:threads=2`, so that we get just one OS thread per pool (default + mpi).

I was able to document few config cases triggering the problem, and they are all the last configurations of a test (the communicator grid configuration varies depending on the rank id, i.e. grouped according to CommunicatorGrid6RanksEnvironment as `(0,1,2)`, `(3,4)` or `(5)`).

This is consistent with our thoughts, which has been also verified by attaching with the debugger, that at the end of the test the collection of MPI ranks results done by `MPIListener` was blocking the default pool and not letting the MPI async operations to start.

The missing explanation was that by implementation, if the MPI blocking call reached, the MPI async `sendBcast` were at least scheduled and so they should have run on the dedicated pool.

```
setPanel +-> sendBcast
         |
         +-> checkSrc -> checkDst | (end-of-test) | MPIListener blocking call
```

Our investigations (thanks @msimberg and @rasolca) revealed that general assumptions were slightly wrong. By adding some log, we were able to see that when the blocking MPI call was about to get scheduled, there still was a `setPanel` for the last config, invalidating our assumptions.

Indeed, now the hypothesis is that Tile destruction mechanism that triggers the next Tile happens inside the `setPanel` task, not when it is terminated.
(see figure) The next future, which is a shared_future, triggers two tasks: `sendBcast` and `checkSource` (in this order). The differences between these two tasks are that `sendBcast` has another dependency, i.e. `Pipeline<Communicator>`, while `checkSource` needs just the tile and it is ready to start, and they run on two different pools, i.e. `mpi` and `default` respectively.

image

For some reason, `checkSource` gets scheduled before `sendBcast` and it start executing without ever yielding, reaching the end of the test and so also the MPI blocking call in `MPIListener`.

About the reason, we are still not sure and @msimberg will try to verifiy. One idea is that since `sendBcast` needs two dependencies, it may happen a situation where the Tile cannot immediately trigger it because the `Pipeline` is doing it (e.g. some kind of lock contention related problem), so HPX may decide to execute the `checkSource` branch before and at that point it will never have the chance to schedule the `sendBcast`.

Anyway, apart the speculations that will hopefully get fully explained/verified, the idea is that the workaround implemented in #429 should also fix this, because the root cause seems confirmed to always be the fact that **blocking calls are problematic**, like said multiple times by the team, and it is way more dangerous when there are just few resources available (i.e. `hpx:threads=2`, which is not exactly our use case, but it may be interesting to use it as a way to test for this edge case condition in order to reveal potential deadlocks.)

msimberg:
After a bit more investigation I think we now have an explanation of the hang.

An important note on the diagram posted by @albestro:

> image

The "CS" task is not a continuation in the traditional sense. It is the main HPX task that calls `future::get`. The "SEND" task is a more traditional continuation in the sense that `dataflow` attempts to trigger it immediately on completion of the `set` task. This distinction is important for the sequence of events. When `set_value` is called on the shared state of a future (through a promise or otherwise) it will:

1. Store the value in the shared state.
2. Mark the shared state as ready. After this any thread calling `future::wait/get` will no longer have to wait.
3. Notify threads waiting with `future::wait/get` through a condition variable.
4. Call the continuations.

The "CS" task, i.e. the main HPX task doing the checking will be ready to run after step 3. The "SEND" task is triggered through a continuation in step 4.

The final piece is how does the continuation end up yielding before the send is actually scheduled. When `dataflow` is called it will "asynchronously traverse" the arguments it was passed. If a future passed to `dataflow` was ready it will continue to the next one synchronously. If a future is not ready it will attach a continuation to the future that will continue the traversal until all arguments are ready. Attaching a continuation requires taking a lock on the shared state. Calling `set_value` on the shared state also requires taking a lock. If dataflow attempts to set a continuation to the `PromiseGuard` future at the same time the previous access to the `PromiseGuard` calls `set_value` one of them may end up yielding while taking the lock. If `dataflow` trying to attach a continuation yields, the main HPX task doing the checking will get to run instead on the main pool before the send has been scheduled on the MPI pool.

PS. To catch the failing case in a debugger I ended up annotating the offending task that calls `set` on the tiles, and then adding (in pseudo-code) `if (task.description == "set_tile") { hpx::util::attach_debugger(); }` in the yield implementation inside HPX.

albestro:
Thanks @msimberg for the explanation!

At this point I think this PR is ready for review/merge.

Note: I gave a quick look if I could improve the code with sender/receivers on the `kernel/broadcast` side, but at the moment a needed functionality is still missing in HPX (https://github.com/STEllAR-GROUP/hpx/pull/5670). Moreover, since it is not strictly related to this PR, this task will be postponed and it will get a dedicated PR.

albestro:
bors try

albestro:
> ## try
> 
> Build failed:
> 
>     * [ci/gitlab/full-pipeline](https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/pipelines/414808644)

Forgot about the newly merged algorithm. I'm going to address that now.

rasolca (approved):

albestro:
bors try

test/unit/common/test_pipeline.cpp
```diff
@@ -45,11 +50,41 @@ TEST(Pipeline, Basic) {
...
+  {
+    Pipeline<int> serial(26);
+    last_task = serial().then(hpx::launch::async, [&is_exited_from_scope](auto) {
+      try_waiting_guard(is_exited_from_scope);
+      EXPECT_TRUE(is_exited_from_scope);
+    });
+  }
+  is_exited_from_scope = true;
+
+  last_task.get();
```
teonnik, line 89: Here wouldn't it be simpler to rewrite it as :

```
hpx::future<void> last_task;
{
Pipeline<int> serial(26);
last_task = serial();
}
ASSERT_TRUE(last_tile.valid()); // if true then `.get()` is not called in the destructor of `Pipeline<>` and it hasn't blocked
```

albestro: As commented during our meeting, this is a "template" test we used also to check matrices destruction. The rationale behind it should be that if the test should fail, it won't block the execution and it will report the failure of the test.

albestro:
This changes introduce the requirement of adding MPI Communicator duplication in algorithms (`clone()`), which is a blocking a call at the beginning of each algorithm. We are all fully aware of the limit of this solution, and as described in some comments above, we are going to think about a re-design of the cooperation between Communicator/CommunicatorGrid/Pipeline. I'm going to open an issue about that.

It seems that #409 is going to be merged just before this, so as soon as that happens I will rebase this PR on the new master and update that algorithm too (adding clone() to it).

Last thing I want to highlight is that after this gets merged, you may need to add "synchronization points" into your tests (https://github.com/eth-cscs/DLA-Future/pull/423#issuecomment-961061554)

albestro:
bors try

albestro:
I did the rebase and fixed the newly merged trsm-LLT with `clone()`.

If anyone of you can give it a quick look to see I didn't introduced anything bad with the rebase, then it can be merged. 😉

PR #426, opened 2021-11-08
Triangular multiplication: LLN LUN RLN and RUN distributed versions

Sely85:
Distributed versions of LLN, LUN, RLN, and RUN triangular multiplication, using sender/receiver approach.


test/unit/multiplication/test_multiplication_triangular.cpp
```diff
@@ -97,6 +108,52 @@ void testTriangularMultiplication(blas::Side side, blas::Uplo uplo, blas::Op op,
...
+  Matrix<T, Device::CPU> mat_ah(std::move(distr_a));
+
+  LocalElementSize size_b(m, n);
+  TileElementSize block_size_b(mb, nb);
+  GlobalElementSize sz_b = globalTestSize(size_b);
+  Distribution distr_b(sz_b, block_size_b, grid.size(), grid.rank(), src_rank_index);
+  Matrix<T, Device::CPU> mat_bh(std::move(distr_b));
+
+  if (side == blas::Side::Left)
+    std::tie(el_op_a, res_b, el_b) =
```
msimberg, line 137: Absolutely not blocking, but this should now be able to use structured bindings: https://en.cppreference.com/w/cpp/language/structured_binding.
```suggestion
    [el_op_a, res_b, el_b] =
```
and same elsewhere.

msimberg (approved):
The sender/receiver stuff looks good to me!

test/unit/multiplication/test_multiplication_triangular.cpp
Sely85: @msimberg This feature makes the code more readable! 😄 
Can you please help me use structured bindings without `auto`? 
Using the above suggestion, compilation dies with an error like:
```
/home/lara/Documents/DLAF/DLA-Future_trmm_distrib/test/unit/multiplication/test_multiplication_triangular.cpp: In lambda function:
/home/lara/Documents/DLAF/DLA-Future_trmm_distrib/test/unit/multiplication/test_multiplication_triangular.cpp:91:28: error: expected ‘{’ before ‘=’ token
   91 |     [el_op_a, res_b, el_b] = getLeftTriangularSystem<GlobalElementIndex, T>(uplo, op, diag, static_cast<T>(1.0) / alpha, m);
```
If I remove the previous declaration of `el_op_a`, `res_b`, and `el_b` and use `auto [el_op_a, res_b, el_b]` it works fine, but then those `std::function` cannot be used outside the `if` statements.

msimberg: Ugh, ignore me. I don't think that's possible. Sorry for the noise. See other comment for a possible alternative, but I'm not convinced it's nicer.

test/unit/multiplication/test_multiplication_triangular.cpp
```diff
@@ -97,6 +108,52 @@ void testTriangularMultiplication(blas::Side side, blas::Uplo uplo, blas::Op op,
...
+        getLeftTriangularSystem<GlobalElementIndex, T>(uplo, op, diag, static_cast<T>(1.0) / alpha, m);
+  else
+    std::tie(el_op_a, res_b, el_b) =
+        getRightTriangularSystem<GlobalElementIndex, T>(uplo, op, diag, static_cast<T>(1.0) / alpha, n);
```
msimberg, line 141:
```suggestion
  auto [el_op_a, res_b, el_b] = []{
      if (side == blas::Side::Left)
          return getLeftTriangularSystem<GlobalElementIndex, T>(uplo, op, diag, static_cast<T>(1.0) / alpha, m);
      else
          return getRightTriangularSystem<GlobalElementIndex, T>(uplo, op, diag, static_cast<T>(1.0) / alpha, n);
    }();
```
and then removing the declaration at the top of the function. But as I said in the other comment, it's arguable whether this is actually better...

Sely85: We have used structured bindings throughout the test (thank you @msimberg for help!). What do you think about this approach @rasolca @teonnik @albestro?

rasolca: I think that everything can be simplified if we provide a `getTriangularSystem(side, uplo, op, ...)` function.
There are 2 ways to do it:
- either combine the two functions in a single one
- add a dispatcher function

teonnik: I prefer the approach without structured bindings in this case, wrapping it in a lambda just to use structured bindings is a bit of a hack. Either of the two approaches suggested by Raffaele seem good to me.

Sely85: I've added something similar to a dispatcher function (https://github.com/eth-cscs/DLA-Future/blob/027558d0eae7c15b0e579ee4f5eed20eaff3a893/test/unit/multiplication/test_multiplication_triangular.cpp#L73-L83) for the first two cases (`get...TriangularSystem`), but then I use it in combination with structured bindings actually. 
On the other hand, I've used structured binginds in the nested loops (like here https://github.com/eth-cscs/DLA-Future/blob/027558d0eae7c15b0e579ee4f5eed20eaff3a893/test/unit/multiplication/test_multiplication_triangular.cpp#L163). 
What do you think about it?

test/unit/multiplication/test_multiplication_triangular.cpp
```diff
@@ -57,11 +68,23 @@ GlobalElementSize globalTestSize(const LocalElementSize& size) {
...
+           std::function<T(const GlobalElementIndex&)>>
+getTriangularSystem(blas::Side side, blas::Uplo uplo, blas::Op op, blas::Diag diag, T alpha, SizeType m,
+                    SizeType n) {
+  if (side == blas::Side::Left)
+    return getLeftTriangularSystem<GlobalElementIndex, T>(uplo, op, diag, static_cast<T>(1.0) / alpha,
+                                                          m);
+  else
+    return getRightTriangularSystem<GlobalElementIndex, T>(uplo, op, diag, static_cast<T>(1.0) / alpha,
+                                                           n);
+}
```
rasolca, line 83: I would put it in `` such that it can be reused anywhere it is needed.
Anyway I would write it in the following way:
```suggestion
template <class ElementIndex, class T, class ...Args>
auto getTriangularSystem(blas::Side side, Args ...args) {
  if (side == blas::Side::Left)
    return getLeftTriangularSystem<ElementIndex, T>(args...);
  else
    return getRightTriangularSystem<ElementIndex, T>(args...);
}
```

rasolca: Note: `1. / alpha` remains on the caller side.

test/include/dlaf_test/matrix/util_generic_blas.h
```diff
@@ -150,6 +150,18 @@ auto getRightTriangularSystem(blas::Uplo uplo, blas::Op op, blas::Diag diag, T a
   return std::make_tuple(el_op_a, el_b, el_x);
 }
 
+/// Dispatcher function to choose between left and right implementation of triangular system generator function
+template <class ElementIndex, class T>
+std::tuple<std::function<T(const ElementIndex&)>, std::function<T(const ElementIndex&)>,
+           std::function<T(const ElementIndex&)>>
```
rasolca, line 156: What about `auto` return type?

rasolca:
bors try

PR #422, opened 2021-09-15
Porting to pika

albestro:
This is the PR in preparation for the next HPX release.

Changelog:
- [x] Update minimum C++ standard requirement to C++17 (changed in #371 and #427)
- [x] update HPX minimum version both in CMake and spack package

TODO:
- [x] Adapt CI to use the actual release branch
- [x] Remove `tag_invoke/dispatch` compatibility macros
  https://github.com/STEllAR-GROUP/hpx/pull/5600
- [x] Use `execution::then` instead of `transform`, and `execution::start_detached` instead of `detach` (renamed)
  https://github.com/STEllAR-GROUP/hpx/pull/5644
- [x] Use `execution::transfer` instead of `on` (renamed)
  https://github.com/STEllAR-GROUP/hpx/pull/5656
- [x] CUDA and C++17 https://github.com/eth-cscs/DLA-Future/pull/422#issuecomment-960633301

Word to @msimberg for any major change (in case there are).

PR #427, opened 2021-11-10
Set C++17 flags in DLA-F

teonnik:
After merging PR #371, HPX requires C++17 and by extension also DLA-F. Currently we still have the C++ standard set to 14. This hasn't caused issues when compiling with HPX built with C++17 because it overrides DLA-F's C++ standards flags. However, we should still use the proper C++ standards flag to avoid confusion. This PR is basically a duplicate of #422 with the difference that we may want to keep the latter one around for other changes required by the upcoming HPX 1.8.0 release.


msimberg (approved):
Thanks @teonnik! Looks good to me.

albestro (approved):
You anticipated me! I was thinking about the same yesterday night. 😄

Thanks!

rasolca (approved):

PR #429, opened 2021-11-16
workaround for MPI Listener with HPX

albestro:
This is a fork from #423, just to test a solution proposal for https://github.com/eth-cscs/DLA-Future/pull/423#issuecomment-965598189

This should solve the problem, and it seems reasonable to wait for all hpx tasks conclusion at the end of each test, but I've not yet investigated why this problem happens in that specific scenario. I'll do and I'll follow up on the comment linked above.

### DIFF
--- a/test/src/gtest_mpihpx_main.cpp
+++ b/test/src/gtest_mpihpx_main.cpp
@@ -41,8 +41,10 @@
 
 #include <gtest/gtest.h>
 
+#include <hpx/include/threadmanager.hpp>
 #include <hpx/init.hpp>
 #include <hpx/program_options.hpp>
+#include <hpx/runtime.hpp>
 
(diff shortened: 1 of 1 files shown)


albestro:
bors try

albestro:
The above CI session has been re-run manually due to a problem related to the CI, but it successfully finished. ✅

msimberg (approved):
It occurs to me that another thing we can consider for the future is using the IO service pool for blocking MPI calls. Or another plain HPX pool for blocking calls. As @albestro said blocking calls are and always will be Bad on HPX threads, and putting them in separate threads helps with that.

teonnik (approved):

rasolca (approved):

PR #433, opened 2021-12-03
senders: refactoring MPI async reduce

albestro:
TODO
- [ ] Requires #422
- [ ] Extend API to `future` and not just `shared_future` for `reduceSend`
- [ ] Refactor common and re-usable parts
- [ ] Code cleanup
- [ ] Various TODOs in the code

Changelog:
- change const-correctness for `matrix::internal::Duplicate`
- minor: fix missing space in error message for `transform<GPU>`


include/dlaf/communication/kernels/reduce.h
```diff
@@ -37,127 +34,179 @@ namespace comm {
...
+  using namespace hpx::execution::experimental;
+  using dlaf::internal::Policy;
+  using dlaf::matrix::internal::CopyBackend;
+
+  auto tile_cpu = dlaf::internal::transform(
+      Policy<CopyBackend<D, Device::CPU>::value>(hpx::threads::thread_priority::high),
+      [](const matrix::Tile<const T, Device::GPU>& tile_gpu, auto... args) {
+        return dlaf::matrix::Duplicate<Device::CPU>{}(tile_gpu, args...);
+      },
+      keep_future(std::move(tile)));
```
msimberg, line 206: This might've been intentional, but I just want to point out that this is not equivalent to `duplicateIfNeeded` since this will unconditionally duplicate the tile. You probably wanted this since CPU tiles won't typically be contiguous?

In general I wonder if these communication helpers would benefit from (another) helper sender adaptor to abstract away a bit of this complexity (not so much a comment about this PR but rather about the complexity I added earlier!). Something that looks a bit like `let_value`, but instead of giving you a reference to the input sender value, it gives you a reference to an already contiguous buffer:
```
with_contiguous_buffer(sender_of_tile, [](contiguous_buffer& buf) {
    return dataflow(mpi_exec, mpi_func, buf, pipeline(), ...);
    // or soon: return when_all(just(buf), pcomm, ...) | transform_mpi(mpi_func); 
});
```
So essentially what you've done now but refactored into separate functions. We can discuss this in more detail offline.

albestro: > This might've been intentional, but I just want to point out that this is not equivalent to duplicateIfNeeded since this will unconditionally duplicate the tile. You probably wanted this since CPU tiles won't typically be contiguous?

Yes, exactly. I wanted to use directly `Duplicate`, because I'm sure it is needed. I thought it was a more clear solution.

> In general I wonder if these communication helpers would benefit from (another) helper sender adaptor to abstract away a bit of this complexity (not so much a comment about this PR but rather about the complexity I added earlier!). Something that looks a bit like let_value, but instead of giving you a reference to the input sender value, it gives you a reference to an already contiguous buffer:

I completely agree with you. It is something I want to do and I have in the back of my mind since https://github.com/eth-cscs/DLA-Future/pull/374#discussion_r636813607 😄

msimberg: Sounds good! I realize now that I never responded to your question on #374. I think you already know the answer but I'll respond anyway.

include/dlaf/communication/kernels/reduce.h
```diff
@@ -37,127 +34,174 @@ namespace comm {
...
+      let_value(hpx::unwrapping([ex, reduce_op, pcomm = std::move(pcomm)](auto& tile_gpu) mutable {
+        using dlaf::internal::Policy;
+        using dlaf::matrix::internal::CopyBackend;
+
+        auto tile_cpu = transform(
+            Policy<CopyBackend<D, Device::CPU>::value>(hpx::threads::thread_priority::high),
+            [](const matrix::Tile<const T, Device::GPU>& tile_gpu, auto... args) mutable {
+              return dlaf::matrix::Duplicate<Device::CPU>{}(tile_gpu, args...);
+            },
+            just(std::cref(tile_gpu)));
```
msimberg, line 155: Another option that might actually be clearer is:
```suggestion
        auto tile_cpu = transform(
            Policy<CopyBackend<D, Device::CPU>::value>(hpx::threads::thread_priority::high),
            std::bind_front(dlaf::matrix::Duplicate<Device::CPU>{}, tile_gpu));
```
I think this works and avoids the use of `reference_wrapper`. On top of that it's shorter. For the more general case (the given argument is not the first argument) there's also `bind`. `std::bind_front` is C++20 but there's `hpx::bind_front` in `hpx/local/functional.hpp`.

Alternatively, it's possible that `just` should support references... but it's quite dangerous so I'm not convinced it's a good idea to allow it.

albestro: It's a nice suggestion, I didn't know about the existence of `bind_front`, and it's a good news it exists in HPX.

IMHO, senders are something from the standard of the future, and we are really trying to use them ahead of time thanks to HPX. We don't have concepts, which will (hopefully) simplify and improve things (e.g. compiler error messages, constraints, ...), but in this case HPX helps us by providing new features that will come, i.e. `bind_front`, and we should make use of it.

What do @teonnik and @rasolca think? Do you like @msimberg proposal?

rasolca: I agree. If some new standard features are already available in HPX feel free to use them.

albestro: After some experiments and checks with @msimberg, we opted for not changing this code to `bind`/`bind_front`.

Main comments/notes (@msimberg feel free to correct or add details):
- the `transform` signature requires a sender: so by bounding the only argument available to the callable, a `just()` (void sender) has to be given to `transform` as 3rd parameter.
- `reference_wrappers` must be used because `just` copies the value, and `Tile` is not copyable
- Use of `reference_wrapper` introduces a template deduction problem, which requires a lambda for "supporting" the template instantiation. This because there are two "unwrapping" going on
i.e. `just(ref(tile)) -> ref(tile) -> &tile`, and the implicit conversion of `reference_wrappers` does not participate to the template deduction process, resulting (without the lambda that "guides" the template deduction) in an error (it tries to fit the `reference_wrapper` itself in the signature of the function)

Till now we haven't mentioned [`bind`](https://en.cppreference.com/w/cpp/utility/functional/bind)/[`bind_front`](https://en.cppreference.com/w/cpp/utility/functional/bind_front). Here some additional notes about them.

Let's start from reporting the main differences between `bind` and `bind_front`:
- `bind` handles in a special way `reference_wrappers`, i.e. it unwraps them calling `.get()`, while `bind_front` has no special treatment for them (apart passing them by reference, note: passing not extracting).
- `bind` requires to specify upfront the full "signature", using placeholders for not bound arguments (_If some of the arguments that are supplied in the call to g() are not matched by any placeholders stored in g, the unused arguments are evaluated and discarded_)

These differences, applied to our problem, turns out in different forms.

### std::bind

With `bind`, the blocking version gets called because without a placeholder, the additional parameter passed by `transform` will get discarded. 
```cpp
transform(
	policy,
	std::bind(dlaf::matrix::Duplicate<Device::CPU>{}, std::cref(tile_gpu)),
	just());
```

In this specific case, we could decide to "loose" generality, by adding the placeholder and so targeting just the gpu asynchronous version of `Duplicate`.
```cpp
transform(
	policy,
	std::bind(dlaf::matrix::Duplicate<Device::CPU>{}, std::cref(tile_gpu), std::placeholders::_1),
	just());
```

Note: @msimberg was wondering if the blocking call of the CUDA memcpy can be removed/hidden. What do you think? @rasolca @teonnik

### std::bind_front
About `bind_front`, we don't have the problem we had with `bind`, but we have again the one about implicit conversion (unwrapping of `reference_wrapper`) that does not participate in template deduction. So we would have to add back a lambda that guides the deduction
```cpp
transform(
	policy,
	pika::bind_front(
		[](const matrix::Tile<const T, Device::GPU>& tile_gpu, auto... args) mutable
			-> decltype(dlaf::matrix::Duplicate<Device::CPU>{}(tile_gpu, args...)) {
				return dlaf::matrix::Duplicate<Device::CPU>{}(tile_gpu, args...);
		}, std::cref(tile_gpu)),
	just());
```

Just for completeness, as just confirmed with @msimberg, it seems that the current implementation of `pika::bind_front` is not conformant with the standard (https://godbolt.org/z/nxEYxYMaT), allowing for
```cpp
transform(
	policy,
	pika::bind_front(dlaf::matrix::Duplicate<Device::CPU>{}, std::cref(tile_gpu)),
	just());
```

We won't pursue this last solution because it will not be supported in the future standard. On the other hand, as @msimberg suggested, we may think of improving `transform` in dlaf to also unwrap `reference_wrapper`s (more or less like what `bind` already does).

include/dlaf/communication/kernels/reduce.h
msimberg: For later: https://github.com/eth-cscs/DLA-Future/issues/485. Thanks @albestro for the issue!

albestro:
As discussed offline, I open this PR to review.

As it can be seen, there are two TODOs not yet addressed, but they are somhow minor things that we can decide to delay (and open related issues).

include/dlaf/communication/kernels/reduce.h
```diff
@@ -37,127 +34,194 @@ namespace comm {
...
-
-  const auto& ex_copy = getHpExecutor<Backend::MC>();
+  using namespace pika::execution::experimental;
+
+  using dlaf::matrix::Tile;
+  using dlaf::internal::Policy;
+  using dlaf::internal::transform;
+  using dlaf::internal::whenAllLift;
+
+  return std::forward<Sender>(tile) | transfer(thread_pool_scheduler{}) |
```
msimberg, line 76: I *think* this is ok since this will most likely always be called from the default pool, but note that `thread_pool_scheduler` will use the pool of the calling thread if on a pika thread and otherwise the default pool. It might be worth being explicit and saying `thread_pool_scheduler{&pika::resource::get_thread_pool("default")}`, or you might want to grab this file from the CUDA scheduler PR: https://github.com/eth-cscs/DLA-Future/blob/6d4f8afc0312a4a27c6f466dea48a514ad386f06/include/dlaf/schedulers.h and use `transfer(getBackendScheduler<Backend::MC>())`.

include/dlaf/communication/kernels/reduce.h
```diff
@@ -37,127 +34,194 @@ namespace comm {
...
-  // TILE_CPU ----> duplicateIfNeeded<CPU> ----> TILE_CPU (no-op)
-  // TILE_GPU ----> duplicateIfNeeded<CPU> ----> TILE_CPU
+  // where: cCPU = contiguous CPU
 
-  pika::shared_future<Tile<T, D>> tile_orig = tile.share();
-  pika::shared_future<Tile<T, Device::CPU>> tile_cpu = duplicateIfNeeded<Device::CPU>(tile_orig);
+  using namespace pika::execution::experimental;
+  using dlaf::internal::whenAllLift;
+
+  return std::forward<Sender>(tile) | transfer(thread_pool_scheduler{}) |
```
msimberg, line 110: The comment from earlier applies here as well.

include/dlaf/communication/kernels/reduce.h
msimberg: Or, in fact, just use `transform(Policy<Backend::MC>(), ...)`.

include/dlaf/communication/kernels/reduce.h
```diff
@@ -37,127 +34,194 @@ namespace comm {
...
+         let_value([ex, reduce_op, pcomm = std::move(pcomm)](Tile<T, Device::CPU>& tile_orig) mutable {
+           using dlaf::common::internal::makeItContiguous;
+
+           return just(makeItContiguous(tile_orig)) |
+                  let_value([ex, reduce_op, &tile_orig,
+                             pcomm = std::move(pcomm)](const auto& cont_buffer) mutable {
+                    return pika::dataflow(ex, internal::reduceRecvInPlace<T>, pcomm.get(), reduce_op,
+                                          std::cref(cont_buffer)) |
+                           then([&]() {
```
msimberg, line 85: Keep in mind that this will be a synchronous continuation. May be intentional, but I just want to point it out. Feel free to close this without action if it's intentional. I guess `copyBack` is a relatively fast operation which might make sense to call synchronously.

include/dlaf/communication/kernels/reduce.h
```diff
@@ -37,127 +34,194 @@ namespace comm {
...
+        return transform(Policy<CopyBackend<Device::CPU, D>::value>(
+                             pika::threads::thread_priority::high),
+                         std::bind(matrix::internal::copy_o, arg::_1, std::cref(tile_gpu), arg::_2),
+                         std::move(tile_reduced));
+      })) |
+      start_detached();
 }
 
-template <class T, Device D, template <class> class Future>
+// TODO scheduleReduceSend with future will require to move the actual value, not the cref
```
msimberg, line 181: Can you expand on this comment? I'm not sure I get it.

msimberg (changes requested):
This looks mostly good, thanks @albestro! Just a couple of questions and I think the only blocker from my perspective is the `SenderElementType` specialization(s) for `reference_wrapper`.

include/dlaf/communication/kernels/reduce.h
```diff
@@ -37,127 +34,194 @@ namespace comm {
...
 
-DLAF_MAKE_CALLABLE_OBJECT(reduceSend);
-
-}
-
-template <class T, Device D>
-void scheduleReduceRecvInPlace(const comm::Executor& ex,
-                               pika::future<common::PromiseGuard<comm::Communicator>> pcomm,
-                               MPI_Op reduce_op, pika::future<matrix::Tile<T, D>> tile) {
+// TODO SenderElementType of a reference_wrapper does not work
```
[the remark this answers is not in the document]
albestro: Yes, it should not be a big deal. What I wanted is to "raise" awareness about the thing that `reference_wrapper`s may become somehow special citizens in our codebase (e.g. we were talking about auto-unwrapping by `transform`).

I'm going to fix this sender trait in this PR!

include/dlaf/communication/kernels/reduce.h
albestro: Good point! I was not exactly sure of what I was doing, but my will is to run the task on the CPU, so I think your last comment is the way to go. Thanks!

include/dlaf/communication/kernels/reduce.h
msimberg: Thanks, makes sense! I suppose this can be refactored afterwards as long as the TODO is not actually a blocker for the other functionality here?

albestro: Yes, exactly. Do you want to keep track of it with an issue? I mean, not just target this specific function, but in general about "allowing a function that requires a readonly tile via a sender, to accept both shared_future and future".

msimberg: I think it's covered by https://github.com/eth-cscs/DLA-Future/issues/465 since that will require quite significant refactoring of these functions anyway. Do you agree?

include/dlaf/communication/kernels/reduce.h
msimberg: > Or, in fact, just use `transform(Policy<Backend::MC>(), ...)`.

As @albestro pointed out, `transform` doesn't actually fit here so ignore that suggestion.

albestro: FIY @rasolca @teonnik 

As discussed, we opted for fixing it (see 4bedad4c824ffb097c2afc5652711cfd0a30b1e7) with the `getBackendScheduler` utility that will be released together with the CUDA scheduler work in a future PR.

For the moment, [include/dlaf/schedulers.h](https://github.com/eth-cscs/DLA-Future/blob/c28f3a8684276881d32f08b3517cfefbb98d2f16/include/dlaf/schedulers.h) file which implements the utility has been cherry-picked, and the part about the GPU Backend has been removed.

include/dlaf/communication/kernels/reduce.h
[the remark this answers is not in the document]
albestro: For what concerns this trait, we opted for treating `reference_wrapper`s the same way we do for `(shared_)future`s.

See https://github.com/eth-cscs/DLA-Future/pull/433/commits/44565fd35b70f6a1671d50081241474af2224828#diff-6735c9ef7e1a74cdd19a0bd8a03735b811d216d92e9042b3ba9c6917d53f99ac

include/dlaf/communication/kernels/reduce.h
albestro: FIY @rasolca @teonnik and for the records.

It’s the copy back after the reduce, which must be completed as soon as possible because we can think of it as “part of the communication”. Previously (before senderification), we were requiring high priority just because of that.

albestro:
bors try

msimberg (approved):

include/dlaf/communication/kernels/reduce.h
msimberg: Created https://github.com/eth-cscs/DLA-Future/issues/492 to keep track of possibly unwrapping `reference_wrapper` automatically.

msimberg (approved):

PR #439, opened 2021-12-15
Remove pointer to promise by variant in Tile.

rasolca:
As the tile promise is optional within a Tile the member was declared as a unique ptr (`std::optional` was not available in C++14). Subtiles added two extra shared futures and therefore `std::variant` became the better option as only one of object can be present at a given time.

However an object cannot contain a `optional/variant` of an object of its type, therefore a small refactoring was needed. 

Note: After this small refactoring, I think that an analysis about which part of the Tile API should be public (mainly dependency tracking part) has to be made.

### DIFF
--- a/include/dlaf/matrix/internal/tile_future_manager.h
+++ b/include/dlaf/matrix/internal/tile_future_manager.h
@@ -29,7 +29,7 @@ hpx::future<ReturnTileType> setPromiseTileFuture(
     std::exception_ptr current_exception_ptr;
 
     try {
-      return ReturnTileType(std::move(NonConstTileType(std::move(fut.get())).setPromise(std::move(p))));
+      return ReturnTileType(std::move(NonConstTileType(fut.get()).setPromise(std::move(p))));
     }
     catch (...) {
       current_exception_ptr = std::current_exception();
--- a/include/dlaf/matrix/internal/tile_future_manager.h
+++ b/include/dlaf/matrix/internal/tile_future_manager.h
@@ -17,15 +17,19 @@ namespace matrix {
 namespace internal {
 
 // Attach the promise to the tile included in old_future with a continuation and returns a new future to it.
-template <class ReturnTileType, class TileType>
-hpx::future<ReturnTileType> setPromiseTileFuture(hpx::future<TileType> old_future,
-                                                 hpx::lcos::local::promise<TileType> p) noexcept {
+template <class ReturnTileType>
+hpx::future<ReturnTileType> setPromiseTileFuture(
+    hpx::future<typename ReturnTileType::TileDataType> old_future,
+    hpx::lcos::local::promise<typename ReturnTileType::TileDataType> p) noexcept {
+  using TileDataType = typename ReturnTileType::TileDataType;
+  using NonConstTileType = typename ReturnTileType::TileType;
+
   DLAF_ASSERT_HEAVY(old_future.valid(), "");
-  return old_future.then(hpx::launch::sync, [p = std::move(p)](hpx::future<TileType>&& fut) mutable {
+  return old_future.then(hpx::launch::sync, [p = std::move(p)](hpx::future<TileDataType>&& fut) mutable {
     std::exception_ptr current_exception_ptr;
 
     try {
-      return ReturnTileType(std::move(fut.get().setPromise(std::move(p))));
+      return ReturnTileType(std::move(NonConstTileType(std::move(fut.get())).setPromise(std::move(p))));
     }
     catch (...) {
       current_exception_ptr = std::current_exception();
@@ -41,10 +45,13 @@ hpx::future<ReturnTileType> setPromiseTileFuture(hpx::future<TileType> old_futur
 
 // Returns a future<ReturnTileType> setting a new promise p to the tile contained in tile_future.
 // tile_future is then updated with the new internal state future (which value is set by p).
-template <class ReturnTileType, class TileType>
-hpx::future<ReturnTileType> getTileFuture(hpx::future<TileType>& tile_future) noexcept {
-  hpx::future<TileType> old_future = std::move(tile_future);
-  hpx::lcos::local::promise<TileType> p;
+template <class ReturnTileType>
+hpx::future<ReturnTileType> getTileFuture(
+    hpx::future<typename ReturnTileType::TileDataType>& tile_future) noexcept {
+  using TileDataType = typename ReturnTileType::TileDataType;
+
+  hpx::future<TileDataType> old_future = std::move(tile_future);
+  hpx::lcos::local::promise<TileDataType> p;
   tile_future = p.get_future();
   return setPromiseTileFuture<ReturnTileType>(std::move(old_future), std::move(p));
 }
@@ -56,10 +63,11 @@ class TileFutureManager {
 public:
   using TileType = Tile<T, device>;
   using ConstTileType = Tile<const T, device>;
+  using TileDataType = internal::TileData<T, device>;
 
   TileFutureManager() {}
 
-  TileFutureManager(TileType tile) : tile_future_(hpx::make_ready_future(std::move(tile))) {}
+  TileFutureManager(TileDataType tile) : tile_future_(hpx::make_ready_future(std::move(tile))) {}
 
   hpx::shared_future<ConstTileType> getReadTileSharedFuture() noexcept {
     if (!tile_shared_future_.valid()) {
@@ -85,7 +93,7 @@ public:
 
 protected:
   // The future of the tile with no promise set.
-  hpx::future<TileType> tile_future_;
+  hpx::future<TileDataType> tile_future_;
 
   // If valid, a copy of the shared future of the tile,
   // which has the promise set to the promise for tile_future_.
--- a/include/dlaf/matrix/tile.h
+++ b/include/dlaf/matrix/tile.h
@@ -170,10 +170,8 @@ public:
 
   Tile(const Tile&) = delete;
 
-  Tile(Tile&& rhs) noexcept
-      : data_(std::move(rhs.data_)), p_(std::move(rhs.p_)), sf_(std::move(rhs.sf_)),
-        sfc_(std::move(rhs.sfc_)) {
-    rhs.p_.reset();
+  Tile(Tile&& rhs) noexcept : data_(std::move(rhs.data_)), dep_tracker_(std::move(rhs.dep_tracker_)) {
+    rhs.dep_tracker_ = std::monostate();
   };
 
   /// Destroys the Tile.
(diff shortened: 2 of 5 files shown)


rasolca:
bors try

include/dlaf/matrix/tile.h
```diff
@@ -282,26 +316,27 @@ class Tile : public Tile<const T, device> {
...
   /// Sets the promise to which this Tile will be moved on destruction.
   ///
   /// @c setPromise can be called only once per object.
   /// @pre this Tile should not be a subtile.
-  Tile& setPromise(promise_t<TileType>&& p) {
-    DLAF_ASSERT(!p_, "setPromise has been already used on this object!");
-    DLAF_ASSERT_HEAVY(!sf_.valid(), "setPromise cannot be used on subtiles!");
-    DLAF_ASSERT_HEAVY(!sfc_.valid(), "setPromise cannot be used on subtiles!");
-    p_ = std::make_unique<promise_t<TileType>>(std::move(p));
+  Tile& setPromise(TilePromise&& p) {
```
msimberg, line 334: Would this be an option?

```suggestion
  Tile&& setPromise(TilePromise&& p) {
```

At least some of the uses wrap `setPromise` in `move`, but then some of the tests just ignore the return value.

msimberg: Or even return type `Tile` and then `return std::move(*this)`. Hmm, not sure. I'm basing this on "can be called only once per object" which would make it fitting that the tile is invalidated after calling this. But I may be misunderstanding...

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -17,15 +17,19 @@ namespace matrix {
...
+  using NonConstTileType = typename ReturnTileType::TileType;
+
   DLAF_ASSERT_HEAVY(old_future.valid(), "");
-  return old_future.then(hpx::launch::sync, [p = std::move(p)](hpx::future<TileType>&& fut) mutable {
+  return old_future.then(hpx::launch::sync, [p = std::move(p)](hpx::future<TileDataType>&& fut) mutable {
     std::exception_ptr current_exception_ptr;
 
     try {
-      return ReturnTileType(std::move(fut.get().setPromise(std::move(p))));
+      return ReturnTileType(std::move(NonConstTileType(std::move(fut.get())).setPromise(std::move(p))));
```
msimberg, line 32: Are all the nested constructors and `move`s really necessary? The inner `move` around `fut.get()` should be unnecessary. See also other comment about the return type of `setPromise`.

include/dlaf/matrix/tile.h
```diff
@@ -35,6 +35,81 @@ struct ContinuationException final : public std::runtime_error {
 };
 
 namespace matrix {
+namespace internal {
+
+template <class T, Device device>
+class TileData {
```
msimberg, line 41: Just checking my understanding: `TileData` is sort of the "dumb" version of `Tile` that doesn't know anything about dependencies, correct?

include/dlaf/matrix/tile.h
```diff
@@ -180,78 +248,44 @@ class Tile<const T, device> {
...
-  if (p_) {
+  if (std::holds_alternative<TilePromise>(dep_tracker_)) {
+    auto& p_ = std::get<TilePromise>(dep_tracker_);
     if (std::uncaught_exceptions() > 0)
-      p_->set_exception(std::make_exception_ptr(ContinuationException{}));
+      p_.set_exception(std::make_exception_ptr(ContinuationException{}));
     else
-      p_->set_value(Tile<ElementType, device>(size_, std::move(memory_view_), ld_));
+      p_.set_value(std::move(this->data_));
   }
```
msimberg, line 268:
```suggestion
  }
  else
  {
    DLAF_ASSERT(std::holds_alternative<std::monostate>(dep_tracker_));
  }
```
or are all the other variants allowed here?

include/dlaf/matrix/tile.h
```diff
@@ -89,11 +163,16 @@ class Tile<const T, device> {
...
+      : data_(size, std::move(memory_view), ld) {}
+
+  Tile(TileDataType&& data) noexcept : data_(std::move(data)) {}
 
   Tile(const Tile&) = delete;
 
-  Tile(Tile&& rhs) noexcept;
+  Tile(Tile&& rhs) noexcept : data_(std::move(rhs.data_)), dep_tracker_(std::move(rhs.dep_tracker_)) {
+    rhs.dep_tracker_ = std::monostate();
+  };
```
msimberg, line 175: Could this and a few of the other constructors/operator= be defaulted instead? From a quick look it seems like they'd be equivalent. I think a moved-from variant will reset to the first variant (`monostate` in this case).

msimberg (commented):
Just a few minor comments. If I understand the changes correctly these are really "only" cosmetic changes in the end, right? The mechanism itself has no significant changes.

include/dlaf/matrix/tile.h
rasolca: Everything is allowed:
monostate <=> no dependent tile has to be set
promise <=> the promise ha to be set to unlock the pipeline
shared_future <=> this is a subtile and the destruction of the shared_future will unlock the pipeline

include/dlaf/matrix/tile.h
rasolca: No, the state of the variant remains the same, just the object inside is moved out.
https://godbolt.org/z/W74rjYb6M

include/dlaf/matrix/tile.h
rasolca: Yes

include/dlaf/matrix/tile.h
rasolca: Setting a dependency doesn't and shouldn't invalidate the object.

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: You are correct, one of the move is not necessary.

include/dlaf/matrix/tile.h
msimberg: Thanks @rasolca. So then if I understand correctly returning `Tile&` from `setPromise` is only to support `move(tile.setPromise(...))`. I'd argue that an explicit
```
tile.setPromise(...);
... move(tile) ...;
```
would be significantly clearer. I found it very surprising that it returns a reference to itself. Is this understanding correct or am I missing some other use of it?

The only place I've seen that kind of an API is if you want to support something like `thing.setA(...).setB(...).setC(...)` but that doesn't seem to be the case here.

msimberg: To add, I realize this is not a change from this PR, so I don't want to block this PR based on this discussion.

rasolca: I agree it is not the best choice, and how I wrote in the note, the API should be re-analyzed and refactored.
However I think it is not worth to do the refactoring now, but to wait for senders/receivers.

msimberg: Yep, agreed, and no worries. Let's come back to this separately.

test/unit/matrix/test_tile.cpp
```diff
@@ -498,7 +505,7 @@ void testSubtilesConst(std::string name, TileElementSize size, SizeType ld,
   tile_p.set_value(std::move(tile));
 
   checkReadyAndDependencyChain(tile_ptr, subtiles, specs, last_dep, next_tile_f);
-  checkFullTile(tile_ptr, next_tile_f.get(), size);
+  checkFullTile(tile_ptr, Tile<T, D>{std::move(next_tile_f.get())}, size);
```
msimberg, line 508:
```suggestion
  checkFullTile(tile_ptr, Tile<T, D>{next_tile_f.get()}, size);
```

test/unit/matrix/test_tile.cpp
```diff
@@ -269,25 +269,28 @@ TYPED_TEST(TileTest, PointerMix) {
...
   {
-    Tile<Type, Device::CPU> tile1 = std::move(tile);
+    TileType tile1 = std::move(tile);
     EXPECT_EQ(false, tile_future.is_ready());
     EXPECT_EQ(TileSizes({0, 0}, 1), getSizes(tile));
   }
 
   ASSERT_EQ(true, tile_future.is_ready());
-  Tile<Type, Device::CPU> tile2 = std::move(tile_future.get());
+  TileType tile2{std::move(tile_future.get())};
```
msimberg, line 293:
```suggestion
  TileType tile2{tile_future.get()};
```

test/unit/matrix/test_tile.cpp
```diff
@@ -296,25 +299,29 @@ TYPED_TEST(TileTest, PromiseToFuture) {
...
   {
-    Tile<const Type, Device::CPU> const_tile = std::move(tile);
+    ConstTileType const_tile = std::move(tile);
     EXPECT_EQ(false, tile_future.is_ready());
     EXPECT_EQ(TileSizes({0, 0}, 1), getSizes(tile));
   }
 
   ASSERT_EQ(true, tile_future.is_ready());
-  Tile<Type, Device::CPU> tile2 = std::move(tile_future.get());
+  TileType tile2{std::move(tile_future.get())};
```
msimberg, line 324:
```suggestion
  TileType tile2{tile_future.get()};
```

msimberg (changes requested):
Just a few more unnecessary moves. Looks good to me otherwise!

rasolca:
bors try

include/dlaf/matrix/tile.h
```diff
@@ -69,17 +144,16 @@ hpx::future<Tile<T, D>> createSubTile(const hpx::shared_future<Tile<T, D>>& tile
 /// the tile of const elements ensure that the memory will not be modified.
 template <class T, Device device>
 class Tile<const T, device> {
+public:
   using TileType = Tile<T, device>;
   using ConstTileType = Tile<const T, device>;
-
-  template <class PT>
-  using promise_t = hpx::lcos::local::promise<PT>;
+  using TileDataType = internal::TileData<T, device>;
```
albestro, line 150: shouldn't this be like this?
```suggestion
  using TileDataType = internal::TileData<const T, device>;
```

albestro (approved):
LGTM

include/dlaf/matrix/tile.h
rasolca: No, as the data member of Tile and Const Tile are the same and should be modifiable.

Note that it was:
https://github.com/eth-cscs/DLA-Future/blob/master/include/dlaf/matrix/tile.h#L187

msimberg (approved):

teonnik (approved):

PR #443, opened 2022-01-04
Add backend flag to miniapps

msimberg:
Fixes https://github.com/eth-cscs/DLA-Future/issues/345.

There are some open questions in the code itself.

At this point I've added only a `--backend {default|mc|gpu}` flag to the Cholesky miniapp. I'll propagate the changes to the other miniapps once everybody is happy with the structure.

Do we want to be able to choose the element type at runtime? I think the answer is quite clearly yes, but I have not added that yet. If yes, should all four types be allowed (is there any reason not to, unless the algorithm doesn't support it?)?

Currently this only allows setting the backend and the device is derived from that. In principle the backend and device are (somewhat) independent, but in practice there's only a one-to-one mapping at the moment. Should it be allowed to choose both independently? Should mismatches be errors or should matrices be copied automatically (and be included in timings)?


include/dlaf/types.h
```diff
@@ -46,6 +57,32 @@ enum class Backend {
...
+    os << "GPU";
+  }
+  return os;
+}
+
+// TODO: Do we need this? Or should we be explicit about which Backend goes with
+// which Device?
+/// Suitable device given a backend.
+template <Backend backend>
+struct BackendDevice;
```
rasolca, line 74: A way to get a default device given a backend is a good idea,
however, I would choose another name (e.g. `DefaultDevice<Backend>` or `PreferredDevice<Backend>`, but feel free to come up with other suggestions).

rasolca (commented):
Currently, as we cannot mix Backends and Devices, I think the `--backend` option is enough.
I also agree that we should add the choice of the type at runtime, but what about Upper/Lower/Trans/Left/Right? Should we add these options too?

include/dlaf/types.h
msimberg: Yep, agreed that the name isn't ideal! I think `DefaultDevice` sounds reasonable. For this PR we don't need it, but should we have the inverse as well, i.e. `DefaultBackend<Device>`?

msimberg:
> I also agree that we should add the choice of the type at runtime, but what about Upper/Lower/Trans/Left/Right? Should we add these options too?

Yeah, I think that's a good idea unless there is something tricky about generating inputs for the different variants.

include/dlaf/types.h
msimberg: I've renamed it to `DefaultDevice` but I'll leave this unresolved to let others comment on the name next week.

msimberg:
bors try

msimberg:
I've pushed a big update that adds a number of helpers for handling options in the miniapps. Now all three (I'm not counting the cublas one) miniapps use the helpers.

Backend, element type, grid rows/cols, number of runs and warmups, and the check flags are considered "common" and can be used through a base options class that can be extended with miniapp-specific options.

The various blas flags are parsed in the same way as `char2X` in blaspp, i.e. with single-character case-insensitive values. I haven't used the `char2X` functions directly because we can give much better error messages this way. The downside is that we have to duplicate the logic for what are valid values.

The element type is also specified with a single character value. This is currently case-sensitive but it would probably make sense to make it insensitive as well.

Without any precedents I would have preferred more verbose names for the blas options and the element types, but given that they are (literally) standardized I suppose this is ok. We can of course accept both short and long names as well. Any preferences here?

Some miniapps obviously don't support all the possible values of the blas enums, but they fail quite gracefully at runtime so I haven't bothered adding any checking into the miniapps themselves. Is that ok?

msimberg:
bors try

msimberg:
bors cancel

scripts/postprocess.py
```diff
@@ -190,9 +190,15 @@ def _calc_metrics(cols, df):
...
+    if bench_name.startswith("chol_dlaf"):
         pstr_arr = []
-        pstr_res = "[{run_index:d}] {time:g}s {perf:g}GFlop/s ({matrix_rows:d}, {matrix_cols:d}) ({block_rows:d}, {block_cols:d}) ({grid_rows:d}, {grid_cols:d}) {:d}"
+        pstr_res = "[{run_index:d}] {time:g}s {perf:g}GFlop/s {} {} ({matrix_rows:d}, {matrix_cols:d}) ({block_rows:d}, {block_cols:d}) ({grid_rows:d}, {grid_cols:d}) {:d} {}"
+    elif bench_name.startswith("trsm_dlaf"):
+        pstr_arr = []
+        pstr_res = "[{run_index:d}] {time:g}s {perf:g}GFlop/s {} {} {} {} {} ({matrix_rows:d}, {matrix_cols:d}) ({block_rows:d}, {block_cols:d}) ({grid_rows:d}, {grid_cols:d}) {:d} {}"
+    elif "dlaf" in bench_name:
+        pstr_arr = []
+        pstr_res = "[{run_index:d}] {time:g}s {perf:g}GFlop/s {} ({matrix_rows:d}, {matrix_cols:d}) ({block_rows:d}, {block_cols:d}) ({grid_rows:d}, {grid_cols:d}) {:d} {}"
```
rasolca, line 201: I'm not sure this is the best way as a comparison of old/new data is not possible without modifying the outputs.
I suggest to use a single compact string to identify the type and the blas options (e.g. dL, zU, ... for Cholesky), and should be used to separate the results (e.g. it should be possible to filter only double precision results).
Moreover, this string should be optional such that old outputs can be loaded.

msimberg: I've updated the miniapps to print a single compact string for the matrix type now. To help with this I've added a `FormatShort` helper struct that simple wraps an existing value, and can be used to overload `operator<<`. This way the default is the long formatting and a more compact representation can be opted into explicitly. The element type is also case insensitive now.

`parse` doesn't support optional fields directly but it does support parsing custom types. Specifying `(|\s+\S+)` as the regex for a custom type matches nothing or whitespace followed by non-whitespace. This seems to work fine at least for the old and new format and I hope there are no funny edge cases...

msimberg:
bors try

scripts/postprocess.py
```diff
@@ -190,9 +202,12 @@ def _calc_metrics(cols, df):
...
 # }
 def _parse_line_based(fout, bench_name, nodes):
-    if "dlaf" in bench_name:
+    if bench_name.startswith("chol"):
         pstr_arr = []
-        pstr_res = "[{run_index:d}] {time:g}s {perf:g}GFlop/s ({matrix_rows:d}, {matrix_cols:d}) ({block_rows:d}, {block_cols:d}) ({grid_rows:d}, {grid_cols:d}) {:d}"
+        # Note that the optional fields must not have a space in front of them.
+        # Otherwise the space is required and parsing the optional field will
+        # fail.
+        pstr_res = "[{run_index:d}] {time:g}s {perf:g}GFlop/s{matrix_type:optional_text} ({matrix_rows:d}, {matrix_cols:d}) ({block_rows:d}, {block_cols:d}) ({grid_rows:d}, {grid_cols:d}) {:d}{backend:optional_text}"
```
albestro, line 210: the `uplo` parameter after `matrix_type` is missing...isn't it?

albestro (approved):
LGTM.

Just a minor reminder: I see you refactored all(?) miniapps, but you updated just the cholesky one for what concerns the parsing of the results in python scripts.

include/dlaf/types.h
```diff
@@ -36,6 +37,16 @@ enum class Device {
...
 
+inline std::ostream& operator<<(std::ostream& os, const Device& device) {
+  if (device == Device::CPU) {
+    os << "CPU";
+  }
+  else if (device == Device::GPU) {
+    os << "GPU";
+  }
+  return os;
+}
```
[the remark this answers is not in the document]
msimberg: Yeah, true! For some of these I found `if-else` easier because the logic was a bit more complicated, but in this case it definitely makes sense to use `switch`. I'll check if there are others as well.

scripts/postprocess.py
msimberg: Not really, see: https://github.com/eth-cscs/DLA-Future/pull/443#discussion_r780154443.

"`matrix_type`" is now a combination the element type and possibly layout options in one string. I wasn't sure if `matrix_type` is descriptive enough but I figured the layout options could fit under that name. Does it make more sense with this explanation?

miniapp/include/dlaf/miniapp/options.h
```diff
@@ -0,0 +1,264 @@
...
+                          const std::string& def = "N") {
+  desc.add_options()("diag", hpx::program_options::value<std::string>()->default_value(def),
+                     "'N' (NonUnit), 'U' (Unit)");
+}
+
+inline void addSideOption(hpx::program_options::options_description& desc,
+                          const std::string& def = "L") {
+  desc.add_options()("side", hpx::program_options::value<std::string>()->default_value(def),
+                     "'L' (Left), 'R' (Right)");
+}
```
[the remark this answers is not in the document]
msimberg: Yeah, I think your suggestion makes sense. I was a bit unsure how much to reuse the `blaspp` functionality.

scripts/postprocess.py
```diff
@@ -190,9 +202,12 @@ def _calc_metrics(cols, df):
 #     "nodes":
 # }
 def _parse_line_based(fout, bench_name, nodes):
-    if "dlaf" in bench_name:
+    if bench_name.startswith("chol"):
```
msimberg, line 205: Note to self: this should be reverted. @albestro this should answer your question in https://github.com/eth-cscs/DLA-Future/pull/443#pullrequestreview-848713690.

I initially separated the parsing for the different miniapps, but didn't revert the condition completely after making the parsing more generic again.

scripts/postprocess.py
albestro: Ah yeah, because there is no space in the output. Got it. thanks

miniapp/include/dlaf/miniapp/options.h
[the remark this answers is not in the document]
albestro: In general I rely on blaspp functionalities because their enums are part of our public interface, so we explicitly rely on them at the moment...and so I fully rely also on the utils 😄

include/dlaf/types.h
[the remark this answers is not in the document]
msimberg: I've changed the two `operator<<` overloads in this file to use `switch`. I've left a number of `if-else` versions in `options.h` that I find more readable that way. Let me know if you think they'd benefit from using `switch` as well.

miniapp/include/dlaf/miniapp/options.h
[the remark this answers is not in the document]
msimberg: This is now using `side2char` and friends (note, not `side2str`). This is to match with the parsing functions which emulate `char2side` and friends (see https://github.com/eth-cscs/DLA-Future/pull/443#issuecomment-1006568214). Does this look ok to you?

msimberg:
bors try

albestro (approved):

include/dlaf/types.h
msimberg: I'm closing this as resolved based on the silence and :+1:s.

scripts/postprocess.py
msimberg: @rasolca are you happy with the update?

rasolca (commented):
A comment about `matrix_type`:
Currently is determined only for dlaf and stored in the data table.
However the plotting utilities do not make use of them.
I think that there are 2 possible solutions:
1) matrix_type is appended to bench_name (https://github.com/eth-cscs/DLA-Future/blob/5218148bea2e5b666f5e336e725f3d16ad6e150c/scripts/postprocess.py#L252).
  This case allows to have different matrix_type in the same plot and filters can be used to show/hide them.
2) all the `gen_***_plot` and `gen_***_plot_weak` functions has to be updated to deal with matrix_type> ( `groupby` calls has to be updated to include matrix_type, and titles and filenames should be updated as well) and other library parsers has to be updated as well.
   This case allows to automatically group results and have different plots according to the `matrix_type`.

scripts/postprocess.py
rasolca: Yes.

miniapp/include/dlaf/miniapp/options.h
```diff
@@ -0,0 +1,270 @@
...
+  std::cout << "Invalid option for --backend. Got '" << backend
+            << "' but expected one of 'default', 'mc', 'gpu'." << std::endl;
+  std::terminate();
+  return Backend::Default;  // unreachable
+}
+
+enum class SupportedTypes { OnlyReal, RealAndComplex };
+enum class ElementType { Single, Double, ComplexSingle, ComplexDouble };
+
+template <SupportedTypes types>
```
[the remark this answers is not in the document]
msimberg: Ok with me if you think it might be needed at some point. If yes, I'd do something like

```
enum class SupportReal { No, Yes };
enum class SupportComplex { No, Yes };
template <SupportReal, SupportComplex> struct MiniappOptions; 
```
so that usage of `MiniappOptions` is `MiniappOptions<SupportReal::Yes, SupportComplex::No>` instead of `MiniappOptions<true, false>`. Does that look ok to you including names?

miniapp/include/dlaf/miniapp/options.h
```diff
@@ -0,0 +1,270 @@
...
+        std::cout << "--type=" << type << " is not supported in this miniapp!" << std::endl;
+        std::terminate();
+      }
+    }
+  }
+
+  std::cout << "Invalid option for --type. Got '" << type << "' but expected one of 's', 'd', 'c', 'z'."
+            << std::endl;
+  std::terminate();
+  return ElementType::Single;  // unreachable
```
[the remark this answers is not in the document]
msimberg: Good point. Are you only talking about the last line? The other messages and the `std::terminate` I find are "normal" checks, i.e. I don't think `DLAF_ASSERT` or `DLAF_UNREACHABLE` semantically fit to replace that, but a new macro would of course be possible.

msimberg:
> A comment about `matrix_type`: Currently is determined only for dlaf and stored in the data table. However the plotting utilities do not make use of them. I think that there are 2 possible solutions:
> 
>     1. matrix_type is appended to bench_name (https://github.com/eth-cscs/DLA-Future/blob/5218148bea2e5b666f5e336e725f3d16ad6e150c/scripts/postprocess.py#L252
>        ).
>        This case allows to have different matrix_type in the same plot and filters can be used to show/hide them.
> 
>     2. all the `gen_***_plot` and `gen_***_plot_weak` functions has to be updated to deal with matrix_type> ( `groupby` calls has to be updated to include matrix_type, and titles and filenames should be updated as well) and other library parsers has to be updated as well.
>        This case allows to automatically group results and have different plots according to the `matrix_type`.

From a quick glance it feels like 2. is the right way forward. I need to have a closer look at what that would mean for the parsing etc. I feel like this may also belong in a separate PR.

miniapp/include/dlaf/miniapp/options.h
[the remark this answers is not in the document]
rasolca: DLAF_ASSERT can replace 67-70 and DLAF_UNREACHABLE 74-77.
Of course the error message produced by the macro are not as descriptive as the custom one.
If they don't fit you can resolve this conversation; I wanted only to make sure that they were considered.

miniapp/include/dlaf/miniapp/options.h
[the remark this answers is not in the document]
rasolca: Not 100% sure that the only complex is needed, but there is a good chance.
Your suggestion of including names is even better.

rasolca:
> > A comment about `matrix_type`: Currently is determined only for dlaf and stored in the data table. However the plotting utilities do not make use of them. I think that there are 2 possible solutions:
> > ```
> > 1. matrix_type is appended to bench_name (https://github.com/eth-cscs/DLA-Future/blob/5218148bea2e5b666f5e336e725f3d16ad6e150c/scripts/postprocess.py#L252
> >    ).
> >    This case allows to have different matrix_type in the same plot and filters can be used to show/hide them.
> > 
> > 2. all the `gen_***_plot` and `gen_***_plot_weak` functions has to be updated to deal with matrix_type> ( `groupby` calls has to be updated to include matrix_type, and titles and filenames should be updated as well) and other library parsers has to be updated as well.
> >    This case allows to automatically group results and have different plots according to the `matrix_type`.
> > ```
> 
> From a quick glance it feels like 2. is the right way forward. I need to have a closer look at what that would mean for the parsing etc. I feel like this may also belong in a separate PR.

OK, but make sure to open an issue such that we don't forget about it.

miniapp/include/dlaf/miniapp/options.h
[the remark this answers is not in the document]
msimberg: I think at least lines 67-70 should be there unconditionally, which is why I said `DLAF_ASSERT` doesn't fit semantically. I'll add macros specifically for these use cases though, so leaving this open.

msimberg:
> OK, but make sure to open an issue such that we don't forget about it.

Of course: https://github.com/eth-cscs/DLA-Future/issues/450. This PR has already strayed from its original purpose which is why I'd like to separate it.

rasolca (approved):

miniapp/include/dlaf/miniapp/options.h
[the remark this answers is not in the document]
msimberg: I've replaced most of these with macros now. Some other places were actually a good fit for `DLAF_UNREACHABLE` so I've used that there. There are only two places where I've left a naked `std::terminate` and that was for checking whether the GPU backend is enabled or not. This was "custom" enough in my opinion that I did not try to come up with a generic macro name for that.

I've also added a (badly named) `DLAF_UNREACHABLE_PLAIN` for cases when the function is `void` and doesn't need anything more than `DLAF_ASSERT(false, "unreachable...")`.

Could you have another look please?

rasolca:
bors try

rasolca:
bors try-

PR #444, opened 2022-01-04
Add RAII helper for `dlaf::initialize/finalize`

msimberg:
Fixes https://github.com/eth-cscs/DLA-Future/issues/411.

This is pretty much as simple as it can be. The wrapper constructor takes the same arguments as the `dlaf::initialize` free function overloads. It initializes in the constructor and finalizes in the destructor. I've changed all uses of `dlaf::initialize` and `dlaf::finalize` to use the the RAII wrapper instead. It's marked `[[nodiscard]]` and the copy and move constructors/assignment operators are conservatively deleted (is there any good use case for having it be movable?). The current name (`ScopedInitializer`) is ok, but not great. I think it would do but suggestions for alternatives are definitely welcome.

### DIFF
--- a/include/dlaf/init.h
+++ b/include/dlaf/init.h
@@ -84,6 +84,20 @@ void initialize(int argc, const char* const argv[], configuration const& user_cf
 /// Should be called once after every dlaf::initialize call.
 void finalize();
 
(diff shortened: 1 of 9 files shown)


msimberg:
bors try

rasolca (approved):

Issue #420, opened 2021-09-14
Compilation issues with Umpire@6.0.0


teonnik:
As pointed out by @albestro :

DynamicPool & DynamicPoolMap seem to be deprecated in 6.0.0:
 https://github.com/LLNL/Umpire/pull/635
 
Changelog : https://github.com/LLNL/Umpire/blob/develop/CHANGELOG.md#changed-1

```
     150    In file included from /tmp/tnikolov/spack-stage/spack-stage-dla-future-develop-tuyamvofcu3ozaxq2azkars6b4gvofyz/spack-src/src/memory/memory_chunk.cpp:4:
  >> 151    /m100/home/userexternal/tnikolov/code/spack/opt/spack/linux-rhel8-power9le/gcc-8.3.1/umpire-6.0.0-jhu5poqrifiukl6nladvccd47sj5dcbn/include/umpire/strategy/DynamicPool.hpp:16:1: error: expected unqual
            ified-id before 'using'
     152     using DynamicPool = DynamicPoolMap;
     153     ^~~~~
     154    /tmp/tnikolov/spack-stage/spack-stage-dla-future-develop-tuyamvofcu3ozaxq2azkars6b4gvofyz/spack-src/src/memory/memory_chunk.cpp: In function 'void dlaf::memory::internal::initializeUmpireHostAllocato
            r(std::size_t)':
  >> 155    /tmp/tnikolov/spack-stage/spack-stage-dla-future-develop-tuyamvofcu3ozaxq2azkars6b4gvofyz/spack-src/src/memory/memory_chunk.cpp:37:46: error: 'DynamicPool' is not a member of 'umpire::strategy'
     156                 .makeAllocator<umpire::strategy::DynamicPool>("PINNED_pool", host_allocator, initial_bytes);
     157                                                  ^~~~~~~~~~~
     158    /tmp/tnikolov/spack-stage/spack-stage-dla-future-develop-tuyamvofcu3ozaxq2azkars6b4gvofyz/spack-src/src/memory/memory_chunk.cpp:37:46: note: suggested alternative: 'DynamicPoolMap'
     159                 .makeAllocator<umpire::strategy::DynamicPool>("PINNED_pool", host_allocator, initial_bytes);
     160                                                  ^~~~~~~~~~~
     161                                                  DynamicPoolMap
[...]

PR #447, opened 2022-01-07
Revert senderified matrix copy

msimberg:
While testing a change for #420 I noticed that the cholesky miniapp occasionally hangs (and probably other miniapps/tests would too if run often enough). My only clue is at the moment is that the senderified matrix copy triggers the hang, but I don't know why (it could be a second order effect as well).

To avoid others being affected while I investigate I'd like to revert that particular change until I understand the problem.


msimberg:
bors try

msimberg:
bors try

rasolca:
@msimberg Test passed already the first time (after some manual rerun), I'm investigating the problem with the CI.

rasolca:
Fixed. @haampie found two independent runners on daint that were conflicting. Should be fine now.

rasolca:
bors try

rasolca (approved):

PR #442, opened 2021-12-23
First attempt CI with spack

rasolca:

PR #449, opened 2022-01-12
Fix GPU warnings with Clang

albestro:
Fix warnings in the GPU part of the codebase, raised by `clang` (and not by `gcc`).

@rasolca do you want to rebase your work in #442 onto this?

### DIFF
--- a/include/dlaf/blas/tile.h
+++ b/include/dlaf/blas/tile.h
@@ -284,10 +284,11 @@ void gemm(cublasHandle_t handle, const blas::Op op_a, const blas::Op op_b, const
           const matrix::Tile<const T, Device::GPU>& a, const matrix::Tile<const T, Device::GPU>& b,
           const T beta, const matrix::Tile<T, Device::GPU>& c) {
   auto s = getGemmSizes(op_a, op_b, a, b, c);
-  CublasGemm<T>::call(handle, util::blasToCublas(op_a), util::blasToCublas(op_b), s.m, s.n, s.k,
-                      util::blasToCublasCast(&alpha), util::blasToCublasCast(a.ptr()), a.ld(),
-                      util::blasToCublasCast(b.ptr()), b.ld(), util::blasToCublasCast(&beta),
-                      util::blasToCublasCast(c.ptr()), c.ld());
+  CublasGemm<T>::call(handle, util::blasToCublas(op_a), util::blasToCublas(op_b), to_int(s.m),
+                      to_int(s.n), to_int(s.k), util::blasToCublasCast(&alpha),
+                      util::blasToCublasCast(a.ptr()), to_int(a.ld()), util::blasToCublasCast(b.ptr()),
+                      to_int(b.ld()), util::blasToCublasCast(&beta), util::blasToCublasCast(c.ptr()),
+                      to_int(c.ld()));
 }
 
 template <class T>
@@ -295,10 +296,10 @@ void hemm(cublasHandle_t handle, const blas::Side side, const blas::Uplo uplo, c
           const Tile<const T, Device::GPU>& a, const Tile<const T, Device::GPU>& b, const T beta,
           const Tile<T, Device::GPU>& c) {
   auto s = getHemmSizes(side, a, b, c);
-  CublasHemm<T>::call(handle, util::blasToCublas(side), util::blasToCublas(uplo), s.m, s.n,
-                      util::blasToCublasCast(&alpha), util::blasToCublasCast(a.ptr()), a.ld(),
-                      util::blasToCublasCast(b.ptr()), b.ld(), util::blasToCublasCast(&beta),
-                      util::blasToCublasCast(c.ptr()), c.ld());
+  CublasHemm<T>::call(handle, util::blasToCublas(side), util::blasToCublas(uplo), to_int(s.m),
+                      to_int(s.n), util::blasToCublasCast(&alpha), util::blasToCublasCast(a.ptr()),
+                      to_int(a.ld()), util::blasToCublasCast(b.ptr()), to_int(b.ld()),
+                      util::blasToCublasCast(&beta), util::blasToCublasCast(c.ptr()), to_int(c.ld()));
 }
 
 template <class T>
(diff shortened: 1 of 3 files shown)


rasolca:
What about cublas calls?
According to the doc they are also using `int` ( https://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemm ), but we pass 64 bit int ( https://github.com/eth-cscs/DLA-Future/blob/master/include/dlaf/blas/tile.h#L287-L290 )

albestro:
@rasolca raised a good point. I applied fixes to warnings I got, but the one he highlighted are not raised by clang.

Since it looked strange, we decided to investigate this in order to understand if we are missing also other warnings without being aware, and possibly the reason behind.

**TL;DR**
_Clang (tested with 10 and 13) doesn't show warnings if they refer to functions defined in header files included with the `-isystem` option, whose name gets replaced via a define macro. On the opposite, they are triggered if the function is called directly and not via its macro_


Here how to reproduce it.

Let's define in an header file a function `baz` together with an alternative name `fuu` for it, provided by a macro define.

```cpp
// fake_include.h
#pragma once

#define fuu baz

int baz(int);
```

Now let's include it in our example, together with the same previous setup (macro + function) as reference (i.e. `foo` -> `bar`).
Then in our example we will target a specific warning about passing a `ptrdiff_t` to a function accepting an `int` (`-Wshorten-64-to-32` on Clang and `-Wconversion` on GNU).

```cpp
// example.cpp
#include <cstddef>

#include "fake_include.h"

#define foo bar

int bar(int);

int main() {
  std::ptrdiff_t n = 13;
  foo(n);      // reference:   foo is translated to bar (current file)
  fuu(n);      // problematic: fuu is translated to baz (header file)
  baz(n);      // proof:       baz is a direct call to the function in header file
}
```

What we are going to show is that when using `-isystem` in Clang, warnings are not triggered if the function is renamed through a macro define (for a real world use case, scroll to the end of this comment).

## Clang
```bash
# clang++ --version
clang version 10.0.0-4ubuntu1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

**With `-isystem`, we get the warnings using the `baz` function directly, but not when we use the macro `fuu`** which results in calling the same exact function.
```bash
root@583d6ed7c9fc:/DLA-Future# clang++ -Wshorten-64-to-32 -c example.cpp -isystem /usr/local/fake
example.cpp:10:7: warning: implicit conversion loses integer precision: 'std::ptrdiff_t' (aka 'long') to 'int' [-Wshorten-64-to-32]
  foo(n);
  ~~~ ^
example.cpp:12:7: warning: implicit conversion loses integer precision: 'std::ptrdiff_t' (aka 'long') to 'int' [-Wshorten-64-to-32]
  baz(n);
  ~~~ ^
2 warnings generated.
```

**Using `-I` instead of `-isystem`, result in all warnings triggered.**
```bash
root@583d6ed7c9fc:/DLA-Future# clang++ -Wshorten-64-to-32 -c example.cpp -I /usr/local/fake
example.cpp:10:7: warning: implicit conversion loses integer precision: 'std::ptrdiff_t' (aka 'long') to 'int' [-Wshorten-64-to-32]
  foo(n);
  ~~~ ^
example.cpp:11:7: warning: implicit conversion loses integer precision: 'std::ptrdiff_t' (aka 'long') to 'int' [-Wshorten-64-to-32]
  fuu(n);
  ~~~ ^
example.cpp:12:7: warning: implicit conversion loses integer precision: 'std::ptrdiff_t' (aka 'long') to 'int' [-Wshorten-64-to-32]
  baz(n);
  ~~~ ^
3 warnings generated.
```

## GCC

```bash
# g++ --version
g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

**Using GCC, the behavior looks more consistent**, both with `-isystem` and the simple `-I`.

```bash
root@583d6ed7c9fc:/DLA-Future# g++ -Wconversion -c example.cpp -I/usr/local/fake
example.cpp: In function 'int main()':
example.cpp:10:7: warning: conversion from 'std::ptrdiff_t' {aka 'long int'} to 'int' may change value [-Wconversion]
   10 |   foo(n);
      |       ^
example.cpp:11:7: warning: conversion from 'std::ptrdiff_t' {aka 'long int'} to 'int' may change value [-Wconversion]
   11 |   fuu(n);
      |       ^
example.cpp:12:7: warning: conversion from 'std::ptrdiff_t' {aka 'long int'} to 'int' may change value [-Wconversion]
   12 |   baz(n);
      |       ^
```

```bash
root@583d6ed7c9fc:/DLA-Future# g++ -Wconversion -c example.cpp -isystem /usr/local/fake
example.cpp: In function 'int main()':
example.cpp:10:7: warning: conversion from 'std::ptrdiff_t' {aka 'long int'} to 'int' may change value [-Wconversion]
   10 |   foo(n);
      |       ^
example.cpp:11:7: warning: conversion from 'std::ptrdiff_t' {aka 'long int'} to 'int' may change value [-Wconversion]
   11 |   fuu(n);
      |       ^
example.cpp:12:7: warning: conversion from 'std::ptrdiff_t' {aka 'long int'} to 'int' may change value [-Wconversion]
   12 |   baz(n);
      |       ^
```

## Use-case
CUBLAS, differently from CUSOLVER, provides an header file `cublas_v2.h` that looks like this

```cpp
#include "cublas_api.h"

#define cublasDgemm          cublasDgemm_v2
```

and then in `cublas_api.h` there is the definition of the function

```cpp
CUBLASAPI cublasStatus_t CUBLASWINAPI cublasDgemm_v2 (cublasHandle_t handle, //...
```

By using directly `cublasDgemm_v2` warnings are triggered, while using the macro `cublasDgemm` they are not.

rasolca (approved):

rasolca:
bors try

PR #451, opened 2022-01-14
Fix Clang warning about unused argument

albestro:
This is a leftover from the previous #449.

Instead of removing names for arguments, I opted for adding a `silenceUnusedWarningFor` function that does nothing but "consuming" what has been passed.

The rationale is that the signature can be "fixed" together with the names for all the variants, so that in the future it will be already there and "unified".

It's a very small and secondary feature. Waiting for your thoughts about it in the review.

### DIFF
--- a/include/dlaf/common/assert.h
+++ b/include/dlaf/common/assert.h
@@ -114,6 +114,17 @@ inline void do_assert(bool expr, const common::internal::source_location& loc, c
 #define DLAF_ASSERT_HEAVY(Expr, ...)
 #endif
 
+namespace dlaf::internal {
+/// Helper function for silencing unused parameter warning.
+///
+/// This allows to not add std::maybe_unused to each argument of an UNIMPLEMENTED function,
+/// but still keeping the name of arguments, so that in the future the signature will be
+/// ready for implementation.
+template <class... Args>
+void silenceUnusedWarningFor(Args&&...) {}
+
+}
+
 #define DLAF_UNIMPLEMENTED(...) DLAF_ASSERT(false, "Not yet implemented!", __VA_ARGS__)
 
 #define DLAF_STATIC_UNIMPLEMENTED(DummyType) \
--- a/include/dlaf/common/assert.h
+++ b/include/dlaf/common/assert.h
@@ -48,6 +48,13 @@ inline void do_assert(bool expr, const common::internal::source_location& loc, c
   }
 }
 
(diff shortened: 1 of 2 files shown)


include/dlaf/common/assert.h
```diff
@@ -114,6 +114,17 @@ inline void do_assert(bool expr, const common::internal::source_location& loc, c
...
+/// Helper function for silencing unused parameter warning.
+///
+/// This allows to not add std::maybe_unused to each argument of an UNIMPLEMENTED function,
+/// but still keeping the name of arguments, so that in the future the signature will be
+/// ready for implementation.
+template <class... Args>
+void silenceUnusedWarningFor(Args&&...) {}
+
+}
+
```
rasolca, line 127: Not sure why the function is defined here with macros. Wouldn't it to be better to define it with the other functions (i.e. after do_assert at line 50)?

include/dlaf/lapack/tile.h
```diff
@@ -396,20 +396,24 @@ void assertExtendInfo(F assertFunc, cusolverDnHandle_t handle, CusolverInfo<T>&&
 }
 
 template <class T>
-dlaf::BaseType<T> lange(cusolverDnHandle_t, const lapack::Norm norm, const Tile<T, Device::GPU>& a) {
+dlaf::BaseType<T> lange(cusolverDnHandle_t handle, const lapack::Norm norm,
+                        const Tile<T, Device::GPU>& a) {
   DLAF_STATIC_UNIMPLEMENTED(T);
+  dlaf::internal::silenceUnusedWarningFor(handle, norm, a);
```
teonnik, line 402: Do you still get the error if you remove the names of the variables? (e.g. handle, norm, a)

If no, I suggest we use that instead of introducing a function (`silenceUnusedWarningFor()`)

teonnik (changes requested):

teonnik (commented):
Have you considered using `void` instead, for example:

```
(void) handle;
(void) type;
...
```

albestro:
> Have you considered using `void` instead? For example:
> 
> ```
> (void) handle;
> (void) type;
> ...
> ```

Yes, it looked a bit more a burden to the developer. While with this the burden is just to remember that the function exist, and IMHO it is a bit more explicit and readable.

teonnik (approved):

rasolca (approved):

PR #454, opened 2022-01-31
Added some debug utilities

rasolca:
- Added the possibility to run the 6 ranks tests with a single rank to debug local tests more easily.
  Note: the check on comm_grids makes sure that distributed tests fails.

- Added a utility that allows to print GPU tiles using the available formats.

### DIFF
--- a/include/dlaf/matrix/print_gpu.h
+++ b/include/dlaf/matrix/print_gpu.h
@@ -32,14 +32,14 @@ template <class Format, class T>
 void print(Format format, const Tile<const T, Device::GPU>& tile, std::ostream& os,
            cudaStream_t stream) {
   const auto size = tile.size();
-  if (size.isEmpty())
-    return;
+  const auto ld = std::max<SizeType>(1, size.rows());
+  Tile<T, Device::CPU> tile_h(size, memory::MemoryView<T, Device::CPU>(size.rows() * size.cols()), ld);
 
-  Tile<T, Device::CPU> tile_h(size, memory::MemoryView<T, Device::CPU>(size.rows() * size.cols()),
-                              size.rows());
+  if (!size.isEmpty()) {
+    internal::copy_o(tile, tile_h, stream);
+    DLAF_CUDA_CALL(cudaStreamSynchronize(stream));
+  }
 
-  internal::copy_o(tile, tile_h, stream);
-  DLAF_CUDA_CALL(cudaStreamSynchronize(stream));
   print(format, tile_h, os);
 }
 
@@ -50,13 +50,14 @@ void print(Format format, const Tile<const T, Device::GPU>& tile, cudaStream_t s
 
 template <class Format, class T>
 void print(Format format, const Tile<const T, Device::GPU>& tile, std::ostream& os = std::cout) {
-  if (tile.size().isEmpty())
-    return;
+  cudaStream_t stream = NULL;
+  if (!tile.size().isEmpty())
+    DLAF_CUDA_CALL(cudaStreamCreate(&stream));
 
-  cudaStream_t stream;
-  DLAF_CUDA_CALL(cudaStreamCreate(&stream));
   print(format, tile, os, stream);
-  DLAF_CUDA_CALL(cudaStreamDestroy(stream));
+
+  if (!tile.size().isEmpty())
+    DLAF_CUDA_CALL(cudaStreamDestroy(stream));
 }
 
 }
--- a/include/dlaf/matrix/print_gpu.h
+++ b/include/dlaf/matrix/print_gpu.h
@@ -11,19 +11,29 @@
 #ifdef DLAF_WITH_CUDA
 #pragma once
 
+#include <iostream>
 #include <cuda_runtime.h>
 
 #include "dlaf/matrix/tile.h"
 
-namespace dlaf {
+namespace dlaf::matrix {
 
-namespace matrix {
-
-/// Print a tile in csv format to standard output
+/// Print a GPU tile in the given format
+///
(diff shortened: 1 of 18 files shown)


rasolca:
bors try

include/dlaf/matrix/print_gpu.h
```diff
@@ -0,0 +1,37 @@
...
+#include <cuda_runtime.h>
+
+#include "dlaf/matrix/tile.h"
+
+namespace dlaf {
+
+namespace matrix {
+
+/// Print a tile in csv format to standard output
+template <class Format, class T>
```
msimberg, line 23: Is this more of a test utility? I.e. does it need to be public in the main dlaf includes? If public the doc should possibly say that this synchronizes forcefully and should not be used in the middle of an algorithm?

msimberg (approved):
Looks good to me, just a namespace question regarding the print utility.

include/dlaf/matrix/print_gpu.h
rasolca: As the CPU print function, the GPU print is a nice to have in any situation.
For debugging, you might want to print a tile (or a matrix (implementation will follow the first time it will be needed)) in the user application as well.
Regarding the synchronization is the same as other functions of tiles:
- if called when scheduling a `.get()` on the future is also needed, and if used with a "clean" stream will not introduce any further synchronization.
- if called in a task, the only difference is that the CPU doesn't only launch the kernels but it does also some work.
  (It can be considered an hybrid task).
  IIRC the cuda tasks are executed from the main pool, therefore even if tasks do not return immediately it shouldn't be a problem.
Anyway as any output operation it introduces some overhead and this should be the main reason to limit its usage.

msimberg: Ok, sounds good. It's just the `cudaStreamSynchronize` that I'm "concerned" about. I don't think you need to do anything about it, but it will block the HPX worker thread (potentially for a long time). I think it's fine for the use cases of this function, but that's why I was wondering if the docs need to mention this behaviour.

msimberg: Thanks the update!

rasolca: Sorry, I was still updating it, now it should be done.

rasolca:
bors try

rasolca:
bors try

include/dlaf/matrix/print_gpu.h
msimberg: My review still stands as approved by the way ;) I don't know if others still have comments though.

test/include/dlaf_test/comm_grids/grids_6_ranks.h
```diff
@@ -24,6 +24,12 @@ class CommunicatorGrid6RanksEnvironment : public ::testing::Environment {
   virtual void SetUp() override {
     if (comm_grids.empty()) {
       comm::Communicator world(MPI_COMM_WORLD);
+
+      // Leave comm_grid empty if invoked with only one rank.
```
albestro, line 28:
```suggestion
      // Leave comm_grids empty if invoked with only one rank.
```

test/unit/matrix/test_matrix_output.cpp
```diff
@@ -104,6 +118,20 @@ struct test_tile_output<std::complex<T>> {
   }
 };
 
+#ifdef DLAF_WITH_CUDA
+template <class T>
+matrix::Tile<T, Device::GPU> gpuTile(const matrix::Tile<const T, Device::CPU>& tile) {
+  const auto size = tile.size();
+  const auto ld = std::max<SizeType>(1, size.rows());
+  matrix::Tile<T, Device::GPU> tile_gpu(size,
+                                        memory::MemoryView<T, Device::GPU>(size.rows() * size.cols()),
```
albestro, line 127:
```suggestion
                                        memory::MemoryView<T, Device::GPU>(size.linear_size()),
```

test/unit/matrix/test_matrix_view.cpp
```diff
@@ -38,6 +38,7 @@ template <typename Type>
 class MatrixViewTest : public ::testing::Test {
 public:
   const std::vector<CommunicatorGrid>& commGrids() {
+    EXPECT_FALSE(comm_grids.empty());
     return comm_grids;
   }
 };
```
albestro, line 44: Minor suggestion: we may decide to create our test type to inherit from.

Something like

```cpp
struct TestWithCommGrids : public ::testing::Test {
  const std::vector<CommunicatorGrid>& commGrids() {
    EXPECT_FALSE(comm_grids.empty());
    return comm_grids;
  }
};
```

and then just have
```cpp
using MatrixViewTest = TestWithCommGrids; // or inherit from it
```

albestro (approved):
LGTM, just some minor comments.

Moreover, I noticed that some tests have not been added, i.e. `CsvFormatMatrix` and `CsvFormatMatrixDist`, am I missing something?

teonnik (approved):
LGTM!

test/unit/matrix/test_matrix_view.cpp
rasolca: Please check the latest commit.

rasolca:
bors try

rasolca:
> LGTM, just some minor comments.
> 
> Moreover, I noticed that some tests have not been added, i.e. `CsvFormatMatrix` and `CsvFormatMatrixDist`, am I missing something?

Yes, that csv format for GPU matrices is not implemented...
(Note: numpy format works as it relies on print tile)

albestro (approved):
LGTM

PR #456, opened 2022-02-02
algo: adapt red2band (local) to sub-tile band size

albestro:
This PR enable to use a band size which evenly divides the block size, for the local version of reduction to band algorithm.

It required a few changes, the main ones are:

- `SubPanelView` and `SubMatrixView`, helpers for `Matrix` sub-tile access, which I quickly described in our last team meeting (feel free to comment or write to me in case something is not clear)
- `Panel` now accepts an offset not tile-aligned
- `Panel` size, respectively width for `Panel<Col>` and height for `Panel<Row>`, is not constrained anymore to be a full-tile
- algorithm test needed some adaptations, mainly changes to `setupHermitianBand` to address the new use-case

In addition to these, I will point out directly in the code some relevant changes that may need additional discussion.

There are still a couple of details to be decided, so as always I look forward to your comments/suggestions/feedbacks, along with proposals for better naming for `Sub{Panel/Matrix}View` helpers.

TODO:
- [x] Add basic documentation for views
- [x] ~Add test cases for TFactor implementation (as pointed out by @rasolca)~ delayed to a future PR where TFactor will work with well formed panels
- [x] Add test cases for Views with non-square blocksize/matrixsize

Changelog:
- bug fix empty size case for reduction to band (both local and distributed)
- add `IterableRange2D::empty` and `isIndexInRange`

### DIFF
--- a/include/dlaf/common/range2d.h
+++ b/include/dlaf/common/range2d.h
@@ -190,6 +190,18 @@ public:
     return iter2d_t(begin_idx_, ld_, i_max_);
   }
 
+  /// Given an index and a range, check if index is in the range
+  ///
+  /// The range is rectangular, with top-left corner set to the beginning of the range, and the bottom
+  /// right corner in the before-last iterator (i.e. std::prev(end)).
+  friend bool isIndexInRange(const index2d_t& index, const IterableRange2D& range) noexcept {
+    const auto tl = *range.begin();
+    const auto br = *std::prev(range.end());
+
+    return tl.row() <= index.row() && index.row() <= br.row() && tl.col() <= index.col() &&
+           index.col() <= br.col();
+  }
+
 private:
   index2d_t begin_idx_;
   IndexT ld_;
--- a/include/dlaf/common/range2d.h
+++ b/include/dlaf/common/range2d.h
@@ -190,13 +190,16 @@ public:
     return iter2d_t(begin_idx_, ld_, i_max_);
   }
 
+  bool empty() const noexcept {
+    return begin() == end();
+  }
+
   /// Given an index and a range, check if index is in the range
   ///
   /// The range is rectangular, with top-left corner set to the beginning of the range, and the bottom
   /// right corner in the before-last iterator (i.e. std::prev(end)).
   friend bool isIndexInRange(const index2d_t& index, const IterableRange2D& range) noexcept {
-    // In case it is empty, it cannot be in the range
-    if (range.begin() == range.end())
+    if (range.empty())
       return false;
 
     const auto tl = *range.begin();
--- a/include/dlaf/common/range2d.h
+++ b/include/dlaf/common/range2d.h
@@ -195,6 +195,10 @@ public:
   /// The range is rectangular, with top-left corner set to the beginning of the range, and the bottom
   /// right corner in the before-last iterator (i.e. std::prev(end)).
   friend bool isIndexInRange(const index2d_t& index, const IterableRange2D& range) noexcept {
+    // In case it is empty, it cannot be in the range
+    if (range.begin() == range.end())
+      return false;
+
     const auto tl = *range.begin();
     const auto br = *std::prev(range.end());
 
--- a/include/dlaf/eigensolver/backtransformation/mc.h
+++ b/include/dlaf/eigensolver/backtransformation/mc.h
@@ -26,9 +26,11 @@
 #include "dlaf/matrix/copy.h"
 #include "dlaf/matrix/copy_tile.h"
 #include "dlaf/matrix/distribution.h"
+#include "dlaf/matrix/index.h"
 #include "dlaf/matrix/layout_info.h"
 #include "dlaf/matrix/matrix.h"
 #include "dlaf/matrix/panel.h"
+#include "dlaf/matrix/views.h"
 #include "dlaf/util_matrix.h"
 
 namespace dlaf {
@@ -149,10 +151,13 @@ void BackTransformation<Backend::MC, Device::CPU, T>::call_FC(
       }
     }
 
+    const GlobalElementIndex v_offset(v_start.row() * mb, v_start.col() * mb);
+    const matrix::SubPanelView panel_view(mat_v.distribution(), v_offset, nr_reflectors);
+
     auto taus_panel = taus[k];
     const LocalTileIndex t_index{Coord::Col, k};
-    dlaf::factorization::internal::computeTFactor<Backend::MC>(nr_reflectors, mat_v, v_start, taus_panel,
-                                                               panelT(t_index));
+    dlaf::factorization::internal::computeTFactor<Backend::MC>(nr_reflectors, mat_v, panel_view,
+                                                               taus_panel, panelT(t_index));
 
     // W = V T
     pika::shared_future<matrix::Tile<const T, Device::CPU>> tile_t = panelT.read(t_index);
--- a/include/dlaf/eigensolver/reduction_to_band/mc.h
+++ b/include/dlaf/eigensolver/reduction_to_band/mc.h
@@ -722,8 +722,9 @@ std::vector<pika::shared_future<common::internal::vector<T>>> ReductionToBand<
   using factorization::internal::computeTFactor;
 
   const auto dist_a = mat_a.distribution();
-  const matrix::Distribution dist({mat_a.size().rows(), band_size}, dist_a.blockSize(),
-                                  dist_a.commGridSize(), dist_a.rankIndex(), dist_a.sourceRankIndex());
+  const matrix::Distribution dist({mat_a.size().rows(), band_size},
+                                  {dist_a.blockSize().rows(), band_size}, dist_a.commGridSize(),
+                                  dist_a.rankIndex(), dist_a.sourceRankIndex());
 
   // Note:
   // Reflector of size = 1 is not considered whatever T is (i.e. neither real nor complex)
--- a/include/dlaf/eigensolver/reduction_to_band/mc.h
+++ b/include/dlaf/eigensolver/reduction_to_band/mc.h
@@ -30,11 +30,13 @@
 #include "dlaf/communication/kernels/reduce.h"
 #include "dlaf/executors.h"
 #include "dlaf/lapack/tile.h"
+#include "dlaf/matrix/copy_tile.h"
 #include "dlaf/matrix/distribution.h"
 #include "dlaf/matrix/index.h"
 #include "dlaf/matrix/matrix.h"
 #include "dlaf/matrix/panel.h"
 #include "dlaf/matrix/tile.h"
+#include "dlaf/matrix/views.h"
 #include "dlaf/util_matrix.h"
 
 #include "dlaf/eigensolver/reduction_to_band/api.h"
@@ -47,13 +49,15 @@ namespace internal {
 template <class T>
 struct ReductionToBand<Backend::MC, Device::CPU, T> {
   static std::vector<pika::shared_future<common::internal::vector<T>>> call(
-      Matrix<T, Device::CPU>& mat_a);
+      Matrix<T, Device::CPU>& mat_a, const SizeType band_size);
   static std::vector<pika::shared_future<common::internal::vector<T>>> call(
       comm::CommunicatorGrid grid, Matrix<T, Device::CPU>& mat_a);
 };
 
 namespace red2band {
 using matrix::Matrix;
+using matrix::SubPanelView;
+using matrix::SubMatrixView;
 using matrix::Tile;
 
 template <class Type>
@@ -144,7 +148,7 @@ std::vector<T> computeWTrailingPanel(const bool has_head, const std::vector<Tile
   bool has_first_component = has_head;
 
   // W = Pt * V
-  for (auto& tile_a : panel) {
+  for (const TileT<const T>& tile_a : panel) {
     const SizeType first_element = has_first_component ? index_el_x0.row() : 0;
 
     TileElementIndex pt_start{first_element, index_el_x0.col() + 1};
@@ -161,6 +165,8 @@ std::vector<T> computeWTrailingPanel(const bool has_head, const std::vector<Tile
       pt_start = pt_start + offset;
       v_start = v_start + offset;
       pt_size = pt_size - offset;
+
+      has_first_component = false;
     }
 
     if (pt_start.isIn(tile_a.size())) {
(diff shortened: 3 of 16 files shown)


include/dlaf/factorization/qr/t_factor_mc.h
```diff
@@ -205,15 +211,18 @@ void QR_Tfactor<Backend::MC, Device::CPU, T>::call(const SizeType k, Matrix<cons
   // 1st step: compute the column partial result `t`
   // First we compute the matrix vector multiplication for each column
   // -tau(j) . V(j:, 0:j)* . V(j:, j)
-  for (const auto& v_i : iterate_range2d(v_start, v_end)) {
-    const bool is_v0 = (v_i.row() == v_start.row());
+  for (const auto& v_i : panel_view.iteratorLocal()) {
+    // TODO improve panel_view API (begin)
+    const bool is_v0 = (v_i.row() == panel_view.begin().row());
```
albestro, line 216: This is wrong in case the diagonal falls across two different tiles.

test/unit/factorization/mc/test_compute_t_factor.cpp
```diff
@@ -205,11 +207,6 @@ std::vector<std::tuple<SizeType, SizeType, SizeType, SizeType, SizeType, GlobalT
     {26, 39, 10, 6, 1, {1, 6}},  // k reflectors (incomplete tile)
 
     {26, 39, 10, 6, 0, {0, 6}},  // 0 reflectors
-
-    // empty matrices
-    {0, 0, 10, 6, 0, {0, 0}},
-    {39, 0, 10, 6, 0, {0, 0}},
-    {0, 26, 10, 6, 0, {0, 0}},
```
albestro, line 212: Empty matrices do not work anymore because of SubPanelView (i.e. it is not possible to define an empty SubPanelView)

include/dlaf/eigensolver/reduction_to_band/mc.h
```diff
@@ -300,32 +313,42 @@ void setupReflectorPanelV(bool has_head, const LocalTileSize& ai_offset, const S
...
   for (auto it = it_begin; it < it_end; ++it) {
-    const LocalTileIndex idx(it->row(), ai_offset.cols());
-    v.setTile(idx, mat_a.read(idx));
+    const LocalTileIndex idx = *it;
+    const matrix::SubTileSpec& spec = panel_view(idx);
+
+    // TODO this is a workaround for the deadlock problem
+    if constexpr (ForceCopy)
+      pika::dataflow(getHpExecutor<Backend::MC>(), pika::unwrapping(matrix::internal::copy_o),
+                     matrix::splitTile(mat_a.read(idx), spec), v(idx));
```
albestro, line 349: This workaround implies copying the panel with reflectors to a temporary memory in Panel, in order to avoid the deadlock problem in matrix update.

Indeed, with the adaption of the algorithm to work with subtiles, it may end up accessing the same tile both in reading mode, for the panel reflectors data, and in write mode, because we are updating the trailing matrix.

include/dlaf/matrix/panel.h
```diff
@@ -266,8 +281,12 @@ struct Panel<axis, const T, D> {
...
     const GlobalTileIndex panel_index(CoordType, panel_coord);
 
-    const auto size_coord = dist_matrix_.tileSize(panel_index).template get<CoordType>();
-    const auto size_axis = dim_ < 0 ? dist_matrix_.blockSize().template get<axis>() : dim_;
+    const bool is_first_global_tile =
+        dist_matrix_.rankGlobalTile<CoordType>(start_) == dist_matrix_.rankIndex().get(CoordType) &&
+        (start_local_ == index.get(CoordType));
+    const auto size_coord = dist_matrix_.tileSize(panel_index).template get<CoordType>() -
+                            (is_first_global_tile ? start_offset_ : 0);
+    const auto size_axis = dim_ < 0 ? dist_matrix_.size().template get<axis>() : dim_;
```
albestro, line 289: As it can be seen, if `dim_` is not set (=-1), it was returning the blocksize, while now it returns the actual size.

This is a somehow major change in the panel, since it will end up being linked to a sub-panel and not just to a panel of tiles.

This change also required a change access operators (i.e. `read` and `operator()`) in order to deal with this new use-case.

albestro:
For the moment I haven't done anything about trying to avoid `splitTile` when the spec returned by `Sub{Panel/Matrix}View` is actually the full-tile.

I have in mind a couple of possible way to do it (e.g. adding a free function that does the check, or adding an additional access operator to matrix that accepts a spec), but for the moment I don't have any clue if it is needed or not.

include/dlaf/eigensolver/reduction_to_band/mc.h
msimberg: "The deadlock problem" being the problem that seemingly unrelated accesses on the panel (view) may in fact point to the same underlying tile?

msimberg:
> proposals for better naming for `Sub{Panel/Matrix}View` helpers.

What about the current names makes you unhappy about them? Naively they sound like fine names, with `View` being a non-owning thing, and `Sub` referring to a subset of the full thing. `Sub` is maybe even redundant, but doesn't hurt I think.

albestro:
bors try

include/dlaf/factorization/qr/t_factor_mc.h
rasolca: It can be fixed when working on the back-transformation.

include/dlaf/matrix/panel.h
rasolca: What is the rationale of this change?

The maximum size of the panel (and therefore its size if it was not limited by setting it) is a block size. It cannot be larger (as matrix size can be).

test/unit/factorization/mc/test_compute_t_factor.cpp
rasolca: What is the reason of the `SubPanelView` limitation?

rasolca (changes requested):
Agree with @msimberg about the `Sub*View` naming.

albestro:
> > proposals for better naming for `Sub{Panel/Matrix}View` helpers.
> 
> What about the current names makes you unhappy about them? Naively they sound like fine names, with `View` being a non-owning thing, and `Sub` referring to a subset of the full thing. `Sub` is maybe even redundant, but doesn't hurt I think.

You give a correct explanation for the rationale behind these first choices. The problem with these names is that they are more "indexing helpers" instead of "matrix views". Moreover, we already have a `MatrixView` (https://github.com/eth-cscs/DLA-Future/pull/154), for which the name correctly represents its functionality.

rasolca:
> > > proposals for better naming for `Sub{Panel/Matrix}View` helpers.
> > 
> > 
> > What about the current names makes you unhappy about them? Naively they sound like fine names, with `View` being a non-owning thing, and `Sub` referring to a subset of the full thing. `Sub` is maybe even redundant, but doesn't hurt I think.
> 
> @rasolca, @msimberg give a correct explanation for the rationale behind these first choices. The problem with these names is that they are more "indexing helpers" instead of "matrix views". Moreover, we already have a `MatrixView` (#154), for which the name correctly represents its functionality.

MatrixView is currently on hold and for the new API we should rethink everything.
Anyway `*ViewIndex` can be another possibility

include/dlaf/matrix/panel.h
albestro: I understand your concerns. I think it's a good starting point for a discussion.

Let me point out a couple of things, so we can decide how to proceed:
- `Panel` allocates a FULL row/column of tiles;
- When we work at sub-tile level, as in this case, we have to `setWidth`/`setHeight` unconditionally; indeed, we always need just part of the full tile

Somehow, they are faces of the same medal. I would consider reducing the allocation to just what needed, because it is a waste of resources, allowing the `Panel` to work in "sub-tile mode", i.e. it knows that is less than a full tile and returns sub-tiles.

What do you think about this?

rasolca: You can just reduce the block size when creating the panel. Isn't it a better solution?

msimberg:
> @rasolca, @msimberg give a correct explanation for the rationale behind these first choices. The problem with these names is that they are more "indexing helpers" instead of "matrix views". Moreover, we already have a `MatrixView` (#154), for which the name correctly represents its functionality.

I might not have thought enough about this, but a view _is_ pretty much an "indexing helper". For prior art there's Kokkos' `SubView` which is an "indexing helper" in the sense that you don't _need_ `SubView` because you could access things directly from the underlying `View` but it makes things easier. The other prior art are views from ranges, which are lazy adaptors on some underlying range. They don't change the range itself, but give alternative "view" on the range. Then there's `std::span` which is kind of similar but is not called "view". Finally, there's `std::string_view`. All in all, I think view is fine in this situation. And @rasolca's `MatrixView` might be a different view to @albestro's views, but they're both views.

albestro:
bors try

albestro:
bors try-

albestro:
bors try

include/dlaf/eigensolver/reduction_to_band/mc.h
msimberg: I'm marking this as resolved since I wasn't really expecting anything to change in this PR regarding this.

I assume this is going to be difficult to fix. Currently all that comes to mind is that something like `transform` would need to know very intimately about how tiles are accessed to be able to detect if overlapping tiles are being used in the same call, but I don't like the idea of this at all.

albestro:
bors try
testing before working on the rebase

albestro:
bors try
rebase on new master and fixed conflicts

include/dlaf/matrix/panel.h
albestro: For reference, this topic has been moved to #478, on which this PR is going to be rebased onto.

test/unit/factorization/mc/test_compute_t_factor.cpp
albestro: I opened to the possibility to define an empty Sub{Matrix/Panel}View, adapted tests and added back these test-cases for computeTFactor.

See changes in https://github.com/eth-cscs/DLA-Future/pull/456/commits/416aa0e9bf871c2acb3c0b806f4ff9500d09e2ec. Briefly, when an empty distribution is given, a SubView with invalid internal indices is created.

As soon as we agree on it, I will document it.

include/dlaf/eigensolver/backtransformation/mc.h
```diff
@@ -149,10 +151,13 @@ void BackTransformation<Backend::MC, Device::CPU, T>::call_FC(
       }
     }
 
+    const GlobalElementIndex v_offset(v_start.row() * mb, v_start.col() * mb);
```
msimberg, line 154: Just checking my understanding of the algorithm here: are the blocks assumed to be square in the current implementation, or can the shape be freely chosen for this part and it's chosen to be square?

include/dlaf/matrix/views.h
```diff
@@ -0,0 +1,136 @@
...
+    const TileElementIndex offset_sub((has_top_tile_ && index == tile_begin_) ? i_sub_offset_ : 0,
+                                      j_sub_offset_);
+
+    const TileElementSize size_tile = dist_.tileSize(dist_.globalTileIndex(index));
+    const TileElementSize size_sub(size_tile.rows() - offset_sub.row(),
+                                   std::min(width_, size_tile.cols() - offset_sub.col()));
+    return {offset_sub, size_sub};
+  }
+
+  SizeType width() const noexcept {
```
msimberg, line 124: What exactly does this represent? Without understanding this fully I'm wondering why it's not `cols` so: what makes it not `cols`?

test/unit/factorization/test_compute_t_factor.cpp
```diff
@@ -257,8 +259,8 @@ void testComputeTFactor(const SizeType a_m, const SizeType a_n, const SizeType m
 
   GlobalElementSize v_size(a_m, a_n);
 
+  const GlobalElementIndex v_start_el = GlobalElementIndex(v_start.row() * mb, v_start.col() * nb);
```
msimberg, line 262:
```suggestion
  const GlobalElementIndex v_start_el(v_start.row() * mb, v_start.col() * nb);
```
?

msimberg (approved):
I can't unfortunately (as usual) comment on the algorithm changes, so only superficial comments from my side. Nothing blocking.

include/dlaf/eigensolver/backtransformation/mc.h
albestro: Yes, exactly. This algorithm assumes square blocksize for `mat_v`. Assumptions valid for the algorithm, and not just for a specific implementation, are enforced by the dispatcher.

https://github.com/eth-cscs/DLA-Future/blob/416aa0e9bf871c2acb3c0b806f4ff9500d09e2ec/include/dlaf/eigensolver/backtransformation.h#L39-L44

include/dlaf/matrix/views.h
albestro: It is actually the number of columns, so it's a good point to keep it coherent with the rest of the library.

Taking the chance to point put one thing: at the moment `SubPanelView` is not generic for Row/Col, but it is just a column one.

rasolca (approved):

include/dlaf/common/range2d.h
```diff
@@ -190,6 +190,18 @@ class IterableRange2D {
     return iter2d_t(begin_idx_, ld_, i_max_);
   }
 
+  /// Given an index and a range, check if index is in the range
+  ///
+  /// The range is rectangular, with top-left corner set to the beginning of the range, and the bottom
+  /// right corner in the before-last iterator (i.e. std::prev(end)).
+  friend bool isIndexInRange(const index2d_t& index, const IterableRange2D& range) noexcept {
+    const auto tl = *range.begin();
+    const auto br = *std::prev(range.end());
```
msimberg, line 199: https://github.com/eth-cscs/DLA-Future/pull/484#discussion_r811817910.

include/dlaf/eigensolver/reduction_to_band/mc.h
```diff
@@ -270,28 +274,37 @@ void updateTrailingPanelWithReflector(const std::vector<TileT<T>>& panel, SizeTy
...
 
 template <class T>
-pika::shared_future<common::internal::vector<T>> computePanelReflectors(
-    MatrixT<T>& mat_a, const common::IterableRange2D<SizeType, matrix::LocalTile_TAG> ai_panel_range,
-    SizeType nrefls) {
-  auto panel_task = pika::unwrapping([nrefls, cols = mat_a.blockSize().cols()](auto fut_panel_tiles) {
+pika::shared_future<common::internal::vector<T>> computePanelReflectors(MatrixT<T>& mat_a,
+                                                                        const SubPanelView& panel_view,
+                                                                        const SizeType nrefls) {
+  auto panel_task = pika::unwrapping([nrefls, width = panel_view.cols()](auto fut_panel_tiles) {
```
msimberg, line 280: https://github.com/eth-cscs/DLA-Future/pull/484#discussion_r811845431.

rasolca:
bors try

albestro:
bors try

PR #458, opened 2022-02-02
GPU implementation of T factor

rasolca:
GPU implementation of computeTfactor.

- Added some utilities helpful to write cuda kernels
- reorg of cublas to simplify its "raw" usage

Currently it is still based on futures, but will work on changing it to senders/receivers.


include/dlaf/cublas/template_cublas.h
```diff
@@ -0,0 +1,91 @@
...
+DLAF_DEFINE_CUBLAS_OP(Gemv, std::complex<float>, Cgemv);
+DLAF_DEFINE_CUBLAS_OP(Gemv, std::complex<double>, Zgemv);
+
+DLAF_DECLARE_CUBLAS_OP(Trmv);
+DLAF_DEFINE_CUBLAS_OP(Trmv, float, Strmv);
+DLAF_DEFINE_CUBLAS_OP(Trmv, double, Dtrmv);
+DLAF_DEFINE_CUBLAS_OP(Trmv, std::complex<float>, Ctrmv);
+DLAF_DEFINE_CUBLAS_OP(Trmv, std::complex<double>, Ztrmv);
+
+// Level 3
```
msimberg, line 52:
```suggestion
}

#define DLAF_MAKE_CUBLAS_OP(Name, f) \
DLAF_DECLARE_CUBLAS_OP(Name); \
DLAF_DEFINE_CUBLAS_OP(Name, float, S##f); \
DLAF_DEFINE_CUBLAS_OP(Name, double, D##f); \
DLAF_DEFINE_CUBLAS_OP(Name, std::complex<float>, C##f); \
DLAF_DEFINE_CUBLAS_OP(Name, std::complex<double>, Z##f)

namespace dlaf::cublas {

// Level 1
DLAF_MAKE_CUBLAS_OP(Axpy, axpy);

// Level 2
DLAF_MAKE_CUBLAS_OP(Gemv, gemv);
DLAF_MAKE_CUBLAS_OP(Trmv, trmv);

// Level 3
```

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -0,0 +1,347 @@
...
+template <class T>
+struct Helpers<Backend::GPU, Device::GPU, T> {
+  static pika::future<matrix::Tile<T, Device::GPU>> set0(pika::future<matrix::Tile<T, Device::GPU>>& t) {
+    return pika::dataflow(dlaf::cuda::Executor{dlaf::internal::getHpCudaStreamPool()},
+                          pika::unwrapping([](auto&& tile, cudaStream_t stream) {
+                            tile::internal::set0<T>(tile, stream);
+                            return std::move(tile);
+                          }),
+                          std::move(t));
+  }
```
msimberg, line 133: This is duplicated only because of the cublas/cusolver executors not working with plain streams, correct? Moving to the sender adaptors should allow you to use only one implementation.

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -0,0 +1,347 @@
...
+
+            if (!va_size.isEmpty()) {
+              cublas::Gemv<T>::call(handle, CUBLAS_OP_C, to_int(va_size.rows()), to_int(va_size.cols()),
+                                    &mtau, util::blasToCublasCast(tile_v.ptr(va_start)),
+                                    to_int(tile_v.ld()), util::blasToCublasCast(tile_v.ptr(vb_start)), 1,
+                                    &one, util::blasToCublasCast(tile_t.ptr(t_start)), 1);
+            }
+          }
+          return std::move(tile_t);
+        });
```
msimberg, line 187: What would it take for this to be merged with the MC version into a generic version? I guess your key requirement here is that you want these scheduled on the same stream sequentially? And the second thing that may make separating those cublas calls are the conditionals? Is there something else? It may not be possible, I'm just asking to understand if there's something in the sender world that would make it easier to merge (parts of) the implementations.

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -8,6 +8,8 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 
+#pragma once
```
msimberg, line 11: Again unrelated to the actual changes here, but we're thinking about making our "inspect" tool usable by others as well: https://github.com/pika-org/pika/issues/6. One of the checks it does is make sure all header files have `#pragma once`. It's definitely low on our priority list, but if you think it'd be useful we could try to do that sooner rather than later.

Alternatively, a low-tech solution is to just put something simple with grep into CI. Our inspect tool is in principle also nothing more than a glorified grep.

msimberg (commented):
Mostly comments unrelated to the actual algorithm, since I'm not familiar with that...

include/dlaf/factorization/qr/t_factor_impl.h
rasolca: Not really. The executor is one reason, on the other hand as I need to define a task, it needs to be duplicated as the lambda has to have a different signatures:
CPU: [](tile)
GPU: [](tile, stream)

include/dlaf/factorization/qr/t_factor_impl.h
rasolca: I think the key is like above: the different signature of the task is the main reason.

include/dlaf/factorization/qr/t_factor_impl.h
msimberg: You can define a helper
```
struct set0_but_also_return_tile {
    template <typename T, Device device, typename... Ts>
    auto operator()(Tile<T, device>&& tile, Ts&&... ts) {
        tile::internal::set0<T>(tile, std::forward<Ts>(ts)...);
        return std::move(tile);
    }
};
```
This is more or less what the `MAKE_CALLABLE_OBJECT` stuff does. However, only with the sender adaptors...

The second case below is clearly more tricky because it uses both plain cuda streams and cublas handles :/

msimberg: It might also be possible to create some sort of adaptor that would make this easier, a bit like `unwrapExtendTiles`. The downside of `unwrapExtendTiles` is that it returns a tuple when often that's unnecessary, or you don't care about the return value of the function that you're wrapping. I'll try to think about this a bit...

rasolca: I considered the variadic template solution, but IMO for these simple use cases I preferred to duplicate it.

Moreover, this solution works with the "stream at the end" case, but not with the "handle at the beginning" case.

rasolca: > It might also be possible to create some sort of adaptor that would make this easier, a bit like `unwrapExtendTiles`. The downside of `unwrapExtendTiles` is that it returns a tuple when often that's unnecessary, or you don't care about the return value of the function that you're wrapping. I'll try to think about this a bit...

One solution can be to return in all the cases the modified tile (e.g. gemm returns c, copy returns the destination tile, ...) but I'm not sure it is the correct approach.

include/dlaf/cublas/template_cublas.h
msimberg: I seem to have left out part of my comment, which was that this change would be unrelated and doesn't have to be done here. It just stood out and I wanted to put it on record...

include/dlaf/factorization/qr/t_factor_impl.h
msimberg: > I considered the variadic template solution, but IMO for these simple use cases I preferred to duplicate it.
> 
> Moreover, this solution works with the "stream at the end" case, but not with the "handle at the beginning" case.

Fair enough!

> One solution can be to return in all the cases the modified tile (e.g. gemm returns c, copy returns the destination tile, ...) but I'm not sure it is the correct approach.

Indeed, I'm also not sure it's the right thing to do. It somewhat goes against the principle of (when possible) accessing tiles through a matrix/panel, but then again it's clearly useful in some situations... Anyway, I'll close this as resolved as a refactoring would anyway require more work elsewhere first.

include/dlaf/cublas/template_cublas.h
rasolca: Too late... It is done.
BTW: I added a macro for symmetric cublas too. What do you think?

test/unit/CMakeLists.txt
```diff
@@ -14,6 +14,12 @@ DLAF_addTest(test_types
   USE_MAIN PLAIN
 )
 
+DLAF_addTest(test_util_cuda
+  SOURCES test_util_cuda.cu
+  LIBRARIES dlaf.core
+  USE_MAIN PLAIN
+)
+
```
albestro, line 22: Shouldn't this be guarded with `DLAF_WITH_CUDA` so that it is built just if CUDA is enabled?

I see there are also `__host__` functions tested there, but those are available just if `DLAF_WITH_CUDA=on`. Moreover, they are in a `.cu` file so a CUDA compiler is required.

albestro (commented):
I did a first quick round, and it looks good.

I still haven't looked at the tfactor part, but I will do it soon.

rasolca:
bors try

src/factorization/qr/tfactor_kernels.cu
```diff
@@ -0,0 +1,109 @@
...
+      const unsigned j = blockIdx.x * kernel_tile_size;
+
+      const unsigned k_max = min(kernel_tile_size, n - j);
+      for (unsigned k = 0; k < k_max; ++k)
+        t[i + (j + k) * ldt] = tile[threadIdx.x][k];
+    }
+  }
+}
+}
+#include <iostream>
```
albestro, line 90: Leftover?

albestro (approved):
LGTM

src/factorization/qr/tfactor_kernels.cu
rasolca: yes... 🤦🏻‍♂️

include/dlaf/matrix/internal/tile_future_manager.h
rasolca: will be addressed in #461.

rasolca:
bors try

PR #459, opened 2022-02-03
Add hip support

aurianer:
This PR adds support for AMD gpus using HIP

- hipSOLVER doesn't have support for `cusolverDnZhegst`, `cusolverDnSsygst`, `cusolverDnDsygst`, `cusolverDnChegst` yet, so I'm using rocSOLVER which has a slightly different interface (`rocsolver_chegst`, which is dense by default instead of `cusolverDnChegst`), this forces the use of `blasToRocblas` utilities temporarily until hipSOLVER fully supports those functions.
- rocSOLVER doesn't have support for the `CUSOLVER_*` error codes so I just undefined this portion and fallback to unknown.

Notes
- The names chosen for the generic blas or solver headers e.g. `gpublas_api.h` could be better so I'm open to suggestions.

TODO:                                                                           
- [ ] Avoid extra allocation of the memoryview workspace with rocsolver (only the info field is needed).
- [ ] Enable the cublas miniapp with hip using rocthrust           .             
                                                                                
Thanks for your patience ^^'


rasolca:
> Notes
> 
>     * The names chosen for the generic blas or solver headers e.g. `gpublas_api.h` could be better so I'm open to suggestions.

I would suggest to rename the `cublas` directory to `gpu/blas` and just call the header `api.h`

Moreover I'm suggesting to use the following tree:
```
gpu/
  blas/
  solver/
  cuda/ (if needed for any cuda specific header)
  cublas/ (any cublas specific header e.g. cublas/hegst.h)
  hip/ (if needed)
  hipblas/ (if needed)
  rocsolver/ (if needed)
```
what do you think?

CMakeLists.txt
```diff
@@ -51,6 +52,16 @@ if (DLAF_WITH_CUDA)
...
+elseif(DLAF_WITH_HIP)
+  if (NOT ${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
+    message(FATAL_ERROR "DLAF_WITH_HIP=ON and the compiler used is not Clang. "
+      "Please specify CMAKE_CXX_COMPILER=hipcc or export CXX=hipcc.")
+  endif()
+  find_package(hipblas HINTS $ENV{HIPBLAS_ROOT} REQUIRED)
+  find_package(rocblas REQUIRED)
+  find_package(rocsolver HINTS $ENV{ROCSOLVER_ROOT} REQUIRED)
+  set(DLAF_WITH_GPU ON)
 endif()
```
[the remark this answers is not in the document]
aurianer: Sorry didn't check carefully, I thought everything was enabled by default
Will do!

include/dlaf/cuda/gpu_api.h
```diff
@@ -0,0 +1,40 @@
...
+#define cudaMemcpyDeviceToDevice         hipMemcpyDeviceToDevice
+#define cudaMemcpyDeviceToHost           hipMemcpyDeviceToHost
+#define cudaMemcpyHostToDevice           hipMemcpyHostToDevice
+#define cudaMemset2DAsync                hipMemset2DAsync
+#define cudaSetDevice                    hipSetDevice
+#define cudaStream_t                     hipStream_t
+#define cudaStreamCreateWithPriority     hipStreamCreateWithPriority
+#define cudaStreamDestroy                hipStreamDestroy
+#define cudaStreamNonBlocking            hipStreamNonBlocking
+#define cudaSuccess                      hipSuccess
```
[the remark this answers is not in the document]
aurianer: Yes I leaved it like that as the first step, but I think it would be nicer to only have those defines in pika

include/dlaf/cusolver/gpusolver_api.h
```diff
@@ -0,0 +1,36 @@
...
+
+#if defined(DLAF_WITH_HIP)
+
+#include <hipblas.h>
+#include <rocblas.h>
+#include <rocsolver.h>
+
+#define cusolverDnChegst    rocsolver_chegst
+#define cusolverDnCreate    rocblas_create_handle
+#define cusolverDnDestroy   rocblas_destroy_handle
```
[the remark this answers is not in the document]
aurianer: it was deprecated, sorry I should have added a comment ^^'

include/dlaf/cusolver/error.h
```diff
@@ -60,10 +60,18 @@ inline std::string cusolverGetErrorString(cusolverStatus_t st) {
   // clang-format on
   return "UNKNOWN";
 }
+#elif defined(DLAF_WITH_HIP)
+/// rocSOLVER does not have support for CUSOLVER_* error codes. This facility,
```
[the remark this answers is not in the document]
aurianer: The use of rocblas is only temporary, so I didn't add the error codes for that, but I will, it's cleaner this way thanks!

include/dlaf/cuda/gpu_api.h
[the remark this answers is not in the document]
msimberg: Right, at least the base functionality. For the actual cuBLAS/cuSOLVER algorithms we might want to think about a variation where we don't hardcode all the possible algorithms but call them instead through something like `<unique namespace>_GPU_BLAS(Ctrsm)` which expands to something appropriate depending on CUDA/HIP. `snake_case` may make that difficult though... In any case, that's not for this PR so I'll close this as resolved (we currently have https://github.com/pika-org/pika/issues/48 to track (approximately) this on the pika side).

spack/packages/dla-future/package.py
```diff
@@ -19,6 +19,7 @@ class DlaFuture(CMakePackage, CudaPackage):
     variant("doc", default=False, description="Build documentation.")
 
     variant("miniapps", default=False, description="Build miniapps.")
+    variant("rocm", default=False, description="Enable HIP support.")
```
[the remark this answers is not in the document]
msimberg: Either you forgot to push a change or my question was unnecessarily indirect :P Let me rephrase: I'm 99% sure that by inheriting from `ROCmPackage` you get this variant for free.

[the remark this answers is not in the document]
aurianer: yes my change is still local, thanks ^^'

spack/packages/dla-future/package.py
```diff
@@ -33,18 +33,23 @@ class DlaFuture(CMakePackage, CudaPackage):
...
     depends_on("umpire~examples")
     depends_on("umpire+cuda~shared", when="+cuda")
+    depends_on("umpire+rocm~shared", when="+rocm")
 
     # https://github.com/eth-cscs/DLA-Future/issues/420
     conflicts("umpire@6:")
 
     depends_on("pika cxxstd=17 +mpi")
-    depends_on("pika@0.4.0")
+    depends_on("pika@0.4.0:")
```
albestro, line 42: Change also this before merge

```suggestion
    depends_on("pika@0.5.0")
```

spack/packages/dla-future/package.py
```diff
@@ -8,7 +8,7 @@
 from spack import *
 
 
-class DlaFuture(CMakePackage, CudaPackage):
+class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
```
albestro, line 11: Should we add a mutual conflict for `+rocm` and `+cuda`?

include/dlaf/blas/tile.h
```diff
@@ -337,6 +340,41 @@ void trmm3(cublasHandle_t handle, const blas::Side side, const blas::Uplo uplo,
...
+#elif defined DLAF_WITH_HIP
+template <class T>
+void trmm(cublasHandle_t handle, const blas::Side side, const blas::Uplo uplo, const blas::Op op,
+          const blas::Diag diag, const T alpha, const matrix::Tile<const T, Device::GPU>& a,
+          const matrix::Tile<T, Device::GPU>& b) {
+  using util::blasToCublas;
+  using util::blasToCublasCast;
+  auto s = tile::internal::getTrmmSizes(side, a, b);
+
+  // TODO: This is almost like the CUDA version except it's missing the last
```
msimberg, line 352: Just marking this as a TODO.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -248,11 +248,46 @@ struct BandBlock {
...
+    //     void* devicePointer;
+    //     void* hostPointer;
+    //     int isManaged;
+    //     unsigned allocationFlags; /* flags specified when memory was allocated*/
+    //     /* peers? */
+    // } hipPointerAttribute_t;
+    //
+    // TODO: What's the right thing to do here?
+    // Not useful documentation:
+    // - https://github.com/RadeonOpenCompute/ROCm_Documentation/blob/master/Programming_Guides/HIP-porting-guide.rst#cu-pointer-attribute-memory-type
```
msimberg, line 287: Fixed link: https://github.com/RadeonOpenCompute/ROCm_Documentation/blob/master/Programming_Guides/HIP-porting-guide.rst#CU-Pointer-Attribute-Memory-Type.

include/dlaf/gpu/assert.cu.h
```diff
@@ -28,7 +31,12 @@ template <class PrintFunc>
 __device__ void gpuAssert(bool expr, PrintFunc&& print_func) {
   if (!expr) {
     print_func();
+#ifdef DLAF_WITH_CUDA
     __trap();
+#else
+    // TODO: Verify what this does...
```
msimberg, line 37: TODO

include/dlaf/gpu/blas/api.h
```diff
@@ -0,0 +1,149 @@
...
+#define CUBLAS_STATUS_EXECUTION_FAILED HIPBLAS_STATUS_EXECUTION_FAILED
+#define CUBLAS_STATUS_INTERNAL_ERROR   HIPBLAS_STATUS_INTERNAL_ERROR
+#define CUBLAS_STATUS_INVALID_VALUE    HIPBLAS_STATUS_INVALID_VALUE
+#define CUBLAS_STATUS_MAPPING_ERROR    HIPBLAS_STATUS_MAPPING_ERROR
+#define CUBLAS_STATUS_NOT_INITIALIZED  HIPBLAS_STATUS_NOT_INITIALIZED
+#define CUBLAS_STATUS_NOT_SUPPORTED    HIPBLAS_STATUS_NOT_SUPPORTED
+#define CUBLAS_STATUS_SUCCESS          HIPBLAS_STATUS_SUCCESS
+
+// Need this macro before the cublasCall definition in error.h
+// TODO: This is dirty. This kind of include structure should not be
```
msimberg, line 79: TODO

include/dlaf/lapack/tile.h
```diff
@@ -423,6 +454,7 @@ class CusolverInfo {
 
 public:
   CusolverInfo(int workspace_size) : workspace_(workspace_size), info_(1) {}
+  CusolverInfo() : info_(1) {}
```
msimberg, line 457: Question related to this and `rocsolver` not requiring a workspace:

Why does `MemoryView` allocate a `MemoryChunk` for the empty state (size 0, default-constructed, and moved-from) instead of just emptying the shared pointer? Is the behaviour relied on anywhere? Looking at the `operator()` overloads for `MemoryView` it seems like it would not need to `make_shared` a `MemoryChunk` since the size is anyways checked explicitly (https://github.com/eth-cscs/DLA-Future/blob/4bb4283fbf0eb344e5302d8ce1e626b84be990b4/include/dlaf/memory/memory_view.h#L130-L143).

The relation is that if we can change `MemoryView` to not allocate in the default constructor then we don't need to add `ifdef`s here to avoid the `MemoryChunk` allocation here in `CusolverInfo`.

src/communication/datatypes.cpp
```diff
@@ -27,8 +27,8 @@ template<> MPI_Datatype mpi_datatype<unsigned long long>   ::type = MPI_UNSIGNED
 template<> MPI_Datatype mpi_datatype<float>                ::type = MPI_FLOAT;
 template<> MPI_Datatype mpi_datatype<double>               ::type = MPI_DOUBLE;
 template<> MPI_Datatype mpi_datatype<bool>                 ::type = MPI_CXX_BOOL;
-template<> MPI_Datatype mpi_datatype<std::complex<float>>  ::type = MPI_CXX_FLOAT_COMPLEX;
-template<> MPI_Datatype mpi_datatype<std::complex<double>> ::type = MPI_CXX_DOUBLE_COMPLEX;
+template<> MPI_Datatype mpi_datatype<std::complex<float>>  ::type = MPI_C_FLOAT_COMPLEX;
+template<> MPI_Datatype mpi_datatype<std::complex<double>> ::type = MPI_C_DOUBLE_COMPLEX;
```
msimberg, line 31: TODO? @rasolca do you know if this will work universally and we could just use this instead everywhere or should we aim to only keep this change for LUMI?

msimberg (commented):
I'll try to take care of a few of the TODOs today. Comments welcome on the other stuff.

msimberg:
I'm done pushing things for today. The other open issues need at least some discussion before addressing them.

src/communication/datatypes.cpp
rasolca: In general it should work as C++ and C complex are compatible.
Would something like `MPI_CXX_DOUBLE_COMPLEX != MPI_TYPE_NULL ? MPI_CXX_DOUBLE_COMPLEX : MPI_C_DOUBLE_COMPLEX` work?

include/dlaf/gpu/blas/api.h
rasolca: I would use lapack instead of solver.
However for CUDA or ROCM specific implementation I would use the library name (e.g. DLAF_DEFINE_CUSOLVER_OP_BUFFER).

test/unit/test_blas_tile.cpp
```diff
@@ -50,6 +50,7 @@ std::vector<std::tuple<SizeType, SizeType, SizeType, SizeType, SizeType, SizeTyp
     {0, 5, 13, 1, 0, 1}, {7, 0, 4, 1, 2, 0},    {3, 11, 0, 0, 1, 0},  // one 0 size
     {1, 1, 1, 0, 3, 0},  {1, 12, 1, 1, 0, 7},   {17, 12, 16, 1, 3, 0}, {11, 23, 8, 0, 3, 4},
     {6, 9, 12, 1, 1, 1}, {32, 32, 32, 0, 0, 0}, {32, 32, 32, 4, 5, 7},
+    {128, 128, 128, 0, 0, 0},
```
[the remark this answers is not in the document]
albestro: I'm going move this in and independent PR, because probably it would need a bit more attention (talking with @rasolca, it may be also useful to re-evaluate how we carry on tests about precision)

test/unit/lapack/gpu/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
-if (DLAF_WITH_CUDA)
+if (DLAF_WITH_HIP)
```
albestro, line 11:
```suggestion
if (DLAF_WITH_GPU)
```

?

rasolca (commented):
I refactored error handling for gpu blas and lapack to improve the readability of error messages (before HIPBLAS error were reported as CUBLAS errors)
there are a couple of thing I would like to suggest:
- have a generic set of GPU methods and types that can bu used in DLAF_WITH_GPU sections. specific HIP and Cuda routines should be used only in DLAF_WITH_CUDA or DLAF_WITH_HIP sections (see e.g. `gpuError_t`)
- reduce the number of macros where possible
- refactor util_cuda, util_cublas, ... in the same way.
 
What do you think of 6b3e73913273cbf6ba30e23a3c197228b3958282 ?
~Please note that the code is not yet updated for the new file structure and function names. I'll work on it in another commit.~ (Edit: Done in 68755fbd414122b92914057dd8d62e2ded01282c)

include/dlaf/lapack/tile.h
```diff
@@ -383,36 +387,75 @@ void scaleCol(T alpha, SizeType col, const Tile<T, Device::CPU>& tile) {
...
+#define DLAF_DEFINE_CUSOLVER_OP_BUFFER(Name, Type, f)                                            \
+  template <>                                                                                    \
+  struct Cusolver##Name<Type> {                                                                  \
+    template <typename... Args>                                                                  \
+    static void call(cusolverDnHandle_t handle, Args&&... args) {                                \
+      std::size_t workspace_size;                                                                \
+      rocblas_start_device_memory_size_query(handle);                                            \
+      rocsolver_##f(handle, std::forward<Args>(args)...);                                        \
+      rocblas_stop_device_memory_size_query(handle, &workspace_size);                            \
+      memory::MemoryView<char, Device::GPU> workspace(to_int(workspace_size));                   \
```
msimberg, line 435: Maybe:
```suggestion
      memory::MemoryView<std::byte, Device::GPU> workspace(to_int(workspace_size));                   \
```

include/dlaf/gpu/blas/error.h
```diff
@@ -0,0 +1,77 @@
...
+    std::cout << "[HIPBLAS ERROR] " << info << std::endl << hipblas::getErrorString(st) << std::endl;
+    std::terminate();
+  }
+#endif
+}
+
+#ifdef DLAF_WITH_HIP
+// Lifted from hipBLAS/library/src/hcc_detail/hipblas.cpp
+// This is not permanent
+inline hipblasStatus_t rocBLASStatusToHIPStatus(rocblas_status_ error) {
```
msimberg, line 45: Remove, replace, do something to clean this up.

include/dlaf/eigensolver/band_to_tridiag/mc.h
msimberg: I've removed most of the comment here. As far as I understand we don't really have a proper way of checking this in HIP, but we can always update it if we find a way in the future.

include/dlaf/lapack/tile.h
msimberg: I've made some tentative changes to not allocate size-0 memory views. I can split that off into a separate PR as well if it's easier to review that way.

test/unit/test_blas_tile.cpp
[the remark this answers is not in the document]
msimberg: I've attempted to add back the 128 blocksizes here which I think will still be fine for most tests in terms of precision. What isn't fine I can revert and will be handled in #557.

include/dlaf/gpu/error.h
```diff
@@ -15,26 +15,28 @@
...
-
 #include "dlaf/common/source_location.h"
+#include "dlaf/gpu/api.h"
 
-namespace dlaf {
-namespace internal {
+#ifdef DLAF_WITH_GPU
 
-#ifdef DLAF_WITH_CUDA
+namespace dlaf::gpu {
```
msimberg, line 23: Is the move to a non-internal namespace intentional?

msimberg:
bors try

include/dlaf/gpu/error.h
rasolca: Yes, but checking again, maybe is not a good idea as the checkError functions should be called through macros...

msimberg:
bors try

msimberg:
bors try

include/dlaf/gpu/assert.cu.h
```diff
@@ -28,7 +32,11 @@ template <class PrintFunc>
 __device__ void gpuAssert(bool expr, PrintFunc&& print_func) {
```
rasolca, line 32: The namespace shouldn't include cuda

msimberg: Also `internal`?

rasolca: Yes, it is a good idea...

include/dlaf/gpu/blas/api.h
msimberg: If I saw correctly @rasolca you've done at least all the file renamings, is that correct? I renamed one macro to be `DLAF_DEFINE_GPULAPACK...`. Some `dlaf::gpu` namespace changes should still be reverted. Is there something else on this point that needs to be done?

include/dlaf/lapack/tile.h
msimberg: Any comments on the general idea here, or the specific changes that I've implemented?

include/dlaf/gpu/error.h
msimberg: I'll update this today.

include/dlaf/gpu/blas/api.h
```diff
@@ -0,0 +1,83 @@
...
+#include <cuComplex.h>
+#include <cublas_v2.h>
+
+namespace dlaf {
+
+// Error handling
+using gpublasStatus_t = cublasStatus_t;
+constexpr gpublasStatus_t GPUBLAS_STATUS_SUCCESS = CUBLAS_STATUS_SUCCESS;
+
+}
```
msimberg, line 81: Are these also meant to be public?

msimberg:
bors try

rasolca:
bors try

rasolca:
bors try

rasolca:
bors try-

include/dlaf/gpu/blas/api.h
rasolca: Yes.

msimberg: Ok, maybe I should've been more explicit in my question :P _why_ are these public? I would not recommend trying to provide a generic HIP/ROCm/CUDA abstraction layer (which is what these definitions are part of), at least at this point. This is far from enough to provide such a layer, and with the HIP/ROCm situation it's far too unstable to consider public in my opinion. It's also far from the core scope of DLA-Future and only required for algorithm development and internals. I understand that eventually such a layer may be useful, but as long as it's not even close to complete and more importantly not needed by any users I would make it internal.

include/dlaf/gpu/blas/api.h
msimberg: Closing this as resolved. Please reopen if you notice something that wasn't changed.

include/dlaf/memory/memory_chunk.h
```diff
@@ -71,7 +67,8 @@ class MemoryChunk {
       ptr_ = static_cast<T*>(internal::getUmpireHostAllocator().allocate(mem_size));
     }
     else {
-      std::cout << "[ERROR] CUDA code was requested but the `DLAF_WITH_CUDA` flag was not passed!";
+      std::cout
+          << "[ERROR] CUDA code was requested but the `DLAF_WITH_CUDA` or `DLAF_WITH_HIP` flags were not passed!";
```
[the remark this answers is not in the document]
msimberg: Well spotted. I updated it to be "GPU _memory_ was requested...".

include/dlaf/gpu/error.h
msimberg: This now internal, closing.

include/dlaf/lapack/tile.h
msimberg: I've `ifdef`ed out the workspace when HIP support is enabled, but I've left the `memory_view` changes for now. I'll move that out to a separate PR if requested.

msimberg:
If I understand correctly the original intention was to only use hipBLAS/hipSOLVER because the APIs are as close to identical to cuBLAS/cuSOLVER as possible. This makes the HIP/CUDA translation layer simpler. However, hipBLAS/hipSOLVER does not yet expose all the required algorithms, whereas rocBLAS/rocSOLVER do, so we're currently using a mix.

There are two directions we can go:
- Once hipBLAS/hipSOLVER have feature parity with cuBLAS/cuSOLVER we could mostly use the HIP wrappers, with the addition of some `roc*` functions for workspace management. This should require a smaller translation layer, but still requires some handling of `roc*` APIs because of the workspace management so it's not quite as minimal as it could be. It's also not possible to do this yet.
- Just use rocBLAS/rocSOLVER directly. This means the translation layer is not quite one-to-one between CUDA and HIP, but we have to deal with that anyway in some form. The benefit of this is that it requires fewer dependencies and less abstraction between DLA-Future and the GPU (we don't currently and likely never will use `hipBLAS/hipSOLVER` to compile for CUDA, so the abstraction is unnecessary).

I think on balance the second option is the better option. For the first option there is no possibility to do anything about it at this point in time. The second option is already doable. However, I'd prefer to do it in a separate PR if that is the preferred option. I'll open an issue in any case for whichever option we prefer.

msimberg:
Still needs changes in pika (task in PR description), comments on the `memory_view` changes, and last sanity checks. In other words, this is not done, but marking ready for review as the rest is minor cleanup.

rasolca:
> If I understand correctly the original intention was to only use hipBLAS/hipSOLVER because the APIs are as close to identical to cuBLAS/cuSOLVER as possible. This makes the HIP/CUDA translation layer simpler. However, hipBLAS/hipSOLVER does not yet expose all the required algorithms, whereas rocBLAS/rocSOLVER do, so we're currently using a mix.
> 
> There are two directions we can go:
> 
>     * Once hipBLAS/hipSOLVER have feature parity with cuBLAS/cuSOLVER we could mostly use the HIP wrappers, with the addition of some `roc*` functions for workspace management. This should require a smaller translation layer, but still requires some handling of `roc*` APIs because of the workspace management so it's not quite as minimal as it could be. It's also not possible to do this yet.
> 
>     * Just use rocBLAS/rocSOLVER directly. This means the translation layer is not quite one-to-one between CUDA and HIP, but we have to deal with that anyway in some form. The benefit of this is that it requires fewer dependencies and less abstraction between DLA-Future and the GPU (we don't currently and likely never will use `hipBLAS/hipSOLVER` to compile for CUDA, so the abstraction is unnecessary).
> 
> 
> I think on balance the second option is the better option. For the first option there is no possibility to do anything about it at this point in time. The second option is already doable. However, I'd prefer to do it in a separate PR if that is the preferred option. I'll open an issue in any case for whichever option we prefer.

I was planing to propose option 2 as well after the merge of this PR.

include/dlaf/gpu/blas/api.h
rasolca: We cannot just distinguish between user and developer. We should consider advanced users which might want to setup matrices or post-process the result using tasks. Therefore we should not hide our basic CUDA/HIP+cublas/rocblas abstraction layer but let it accessible to users.

include/dlaf/lapack/tile.h
rasolca: Yes, better to move it in a different PR.

include/dlaf/gpu/blas/api.h
rasolca: I still don't see the reason why we should make them internal now and move them again later, but fine, make them internal.

msimberg: Because is this even something that DLAF needs to provide? pika is having to do the same translation as DLAF. A separate library would perhaps even be more appropriate. This is essentially doing what HIP does without pulling in HIP.

include/dlaf/memory/memory_view.h
```diff
@@ -53,6 +54,7 @@ class MemoryView {
   MemoryView(T* ptr, SizeType size)
       : memory_(std::make_shared<MemoryChunk<ElementType, device>>(const_cast<ElementType*>(ptr), size)),
         offset_(0), size_(size) {
+    // TODO: ptr == nullptr? deleter?
```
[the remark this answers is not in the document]
msimberg: The `memory_view` changes have been removed, this is no longer relevant.

rasolca (approved):

spack/packages/dla-future/package.py
```diff
@@ -33,18 +33,26 @@ class DlaFuture(CMakePackage, CudaPackage):
...
     depends_on("umpire~examples")
     depends_on("umpire+cuda~shared", when="+cuda")
+    depends_on("umpire+rocm~shared", when="+rocm")
 
     # https://github.com/eth-cscs/DLA-Future/issues/420
     conflicts("umpire@6:")
 
     depends_on("pika cxxstd=17 +mpi")
-    depends_on("pika@0.4:")
+    depends_on("pika@0.5:")
```
msimberg, line 42: To do before merging:
```suggestion
    depends_on("pika@0.6:")
```

CMakeLists.txt
```diff
@@ -94,7 +108,7 @@ else()
...
 
 # ----- pika
-find_package(pika 0.4.0 REQUIRED)
+find_package(pika 0.5.0 REQUIRED)
```
msimberg, line 111: To do before merging:
```suggestion
find_package(pika 0.6.0 REQUIRED)
```

CMakeLists.txt
```diff
@@ -94,7 +108,7 @@ else()
 endif()
 
 # ----- pika
-find_package(pika 0.4.0 REQUIRED EXACT)
+find_package(pika REQUIRED)
```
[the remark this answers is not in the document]
msimberg: Requires 0.6.0.

spack/packages/dla-future/package.py
```diff
@@ -39,7 +39,7 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("umpire@6:")
 
     depends_on("pika cxxstd=17 +mpi")
-    depends_on("pika@0.5:")
+    depends_on("pika@main")
```
msimberg, line 42:
```suggestion
    depends_on("pika@0.6:")
```

msimberg:
bors try

msimberg:
bors try

rasolca:
bors try

ci/docker/amdgpu-release-rocm510.yaml
```diff
@@ -0,0 +1,83 @@
...
+      - build_type=Release
+      - cxxstd=17
+    blaspp:
+      variants:
+      - ~cuda
+      - ~rocm
+      - ~openmp
+    pika:
+      version:
+      - main
```
rasolca, line 31: TODO

rasolca:
bors try

rasolca:
bors try

rasolca:
bors try

rasolca:
I runned the tests on ault using the image built by the CI and found:
An error:
```
[----------] 2 tests from BacktransformationT2BTestGPU/2, where TypeParam = std::complex<float>
[ RUN      ] BacktransformationT2BTestGPU/2.CorrectnessLocal
/DLA-Future/test/unit/eigensolver/test_bt_band_to_tridiag.cpp:161: Failure
Failed
Error at index (16, 0): expected (0.280768,0.0382712) == (0.373067,0.0711032) (Relative diff: 0.25795 > 1.90735e-05, Absolute diff: 0.0979648 > 1.90735e-05)

[  FAILED  ] BacktransformationT2BTestGPU/2.CorrectnessLocal, where TypeParam = std::complex<float> (88 ms)
```
a test killed by OOM
```
+ srun -n 4 -c 10 -i none sarus run --mount=type=bind,src=/dev/kfd,dst=/dev/kfd --mount=type=bind,src=/dev/dri,dst=/dev/dri art.cscs.ch/contbuild/cscs-ci/eth-cscs/dla-future/release-amdgpu-clang10/deploy:e0b20662a1b8cb91884aa36a9790d21f2b2fe352 /DLA-Future-build/test/unit/communication/test_broadcast_matrix
Running main() from gtest_mpipika_main.cpp
Running main() from gtest_mpipika_main.cpp
Running main() from gtest_mpipika_main.cpp
Running main() from gtest_mpipika_main.cpp
[==========] Running 8 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 8 tests from BroadcastMatrixTest
[ RUN      ] BroadcastMatrixTest.Matrix2Workspace
[       OK ] BroadcastMatrixTest.Matrix2Workspace (184 ms)
[ RUN      ] BroadcastMatrixTest.ConstMatrix2Workspace
[       OK ] BroadcastMatrixTest.ConstMatrix2Workspace (0 ms)
[ RUN      ] BroadcastMatrixTest.Matrix2Matrix
slurmstepd: error: Detected 1 oom-kill event(s) in StepId=124064.34 cgroup. Some of your processes may have been killed by the cgroup out-of-memory handler.
srun: error: ault20: task 1: Out Of Memory
[       OK ] BroadcastMatrixTest.Matrix2Matrix (0 ms)
[ RUN      ] BroadcastMatrixTest.ConstMatrix2Matrix
```
And the test that kills the nodes
```
+ srun -n 6 -c 10 -i none sarus run --mount=type=bind,src=/dev/kfd,dst=/dev/kfd --mount=type=bind,src=/dev/dri,dst=/dev/dri art.cscs.ch/contbuild/cscs-ci/eth-cscs/dla-future/release-amdgpu-clang10/deploy:e0b20662a1b8cb91884aa36a9790d21f2b2fe352 /DLA-Future-build/test/unit/communication/test_broadcast_panel
Running main() from gtest_mpipika_main.cpp
Running main() from gtest_mpipika_main.cpp
Running main() from gtest_mpipika_main.cpp
Running main() from gtest_mpipika_main.cpp
Running main() from gtest_mpipika_main.cpp
Running main() from gtest_mpipika_main.cpp
[==========] Running 16 tests from 4 test suites.
[----------] Global test environment set-up.
[----------] 4 tests from PanelBcastTest/0, where TypeParam = float
[ RUN      ] PanelBcastTest/0.BroadcastCol
```

Issue #448, opened 2022-01-11
Unrevert matrix copy senderification


msimberg:
Opening this as a reminder to go back to using senders in the matrix copy after it was reverted in #447.

It seems like the problem is fundamentally unrelated to senders or futures, but to the termination checking we do in HPX. HPX needs to be updated to:
1. Not require taking a lock when checking work counts (both CUDA and MPI polling).
2. Release the polling lock sooner (callbacks should ideally not be called under a lock).

Point 1. already avoids hangs in the Cholesky miniapp. It's not clear to me why the dataflow version of the matrix copy does not trigger hangs.

PR #460, opened 2022-02-03
Use `cuda_scheduler` from pika to replace custom GPU transform

msimberg:
This replaces the custom transform sender implementation for `Backend::GPU` with one based on CUDA utilities in pika. This needs https://github.com/pika-org/pika/pull/37 which will hopefully be available in pika 0.2.0.

Specifically, this uses the sender adaptors `then_with_stream/cublas/cusolver` through a helper adaptor `then_with_any_cuda` (comments about the name on https://github.com/pika-org/pika/pull/37 please!).

The existing CUDA executors stay for now because there are still a few places using them, but once those places have been senderified (https://github.com/msimberg/DLA-Future/commits/more-senders) they can be deleted.

~More details about the new CUDA scheduler and pool are on [that](https://github.com/pika-org/pika/pull/37) but the main difference is that the pool of streams is used round-robin by all worker threads, instead of each worker thread having its own small set of streams.~ See https://github.com/pika-org/pika/pull/37 and https://github.com/pika-org/pika/pull/128 for more details. The cuBLAS/cuSOLVER handles are lazily initialized as `thread_local` handles. ~I did this mainly to leave the door open for using the CUDA scheduler independently of the pika runtime (the current executors need the worker thread count from the runtime).~

Performance plots to follow but the short story is that performance is mostly the same, though it's somewhat sensitive to the configuration (i.e. number of streams) used.

TODO: 
- [x] Add pika 0.2.0 to spack package (and possibly an rc)
- [x] Update pika requirements in dla-future spack package
- [x] Rerun GPU _and_ CPU benchmarks (the latter for all the other changes in pika)


msimberg:
bors try

include/dlaf/common/assert.h
```diff
@@ -126,6 +126,8 @@ void silenceUnusedWarningFor(Args&&...) {}
 #define DLAF_STATIC_UNIMPLEMENTED(DummyType) \
   static_assert(sizeof(DummyType) == 0, "Not yet implemented!")
 
+#define DLAF_STATIC_FAIL(DummyType, Msg) static_assert(sizeof(DummyType) == 0, Msg)
```
msimberg, line 129: The name is not great, but it is a generic compile-time failure. It's definitely not `UNIMPLEMENTED` where I've used it. `UNREACHABLE` is close, but since it's compile-time it's not really right either. Hence I just used `FAIL`, but better suggestions are welcome.

include/dlaf/communication/kernels/broadcast.h
```diff
@@ -104,8 +102,11 @@ struct ScheduleRecvBcast {
     tile_cpu = std::move(pika::get<0>(pika::split_future(
         ScheduleRecvBcast<T>::call(ex, std::move(tile_cpu), root_rank, std::move(pcomm)))));
 
-    dataflow(getCopyExecutor<Device::GPU, Device::CPU>(), matrix::unwrapExtendTiles(copy_o), tile_cpu,
-             tile_gpu);
+    pika::execution::experimental::when_all(dlaf::internal::keepIfSharedFuture(std::move(tile_cpu)),
```
msimberg, line 105: This closes https://github.com/eth-cscs/DLA-Future/issues/448. This seems to not be an issue anymore.

include/dlaf/executors.h
```diff
@@ -108,8 +75,15 @@ inline auto getHpExecutor<Backend::GPU>() {
...
 decltype(auto) getCopyExecutor() {
-  return internal::GetCopyExecutor<S, D>::call();
+  if constexpr (S == Device::CPU && D == Device::CPU) {
+    return pika::execution::parallel_executor{&pika::resource::get_thread_pool("default"),
+                                              pika::threads::thread_priority::normal};
+  }
+  else {
+    DLAF_STATIC_FAIL(Dummy,
+                     "Do not use getCopyExecutor for copying anymore. Prefer sender adaptors with CopyBackend.");
+  }
```
msimberg, line 87: This is due for removal but there is still one CPU-CPU copy that uses the executor rather than the sender adaptors. I've disallowed GPU copies with this one to discourage anyone from even thinking about using this anymore. This should be completely removed in #475.

include/dlaf/matrix/copy.h
```diff
@@ -41,9 +41,10 @@ void copy(Matrix<const T, Source>& source, Matrix<T, Destination>& dest) {
 
   for (SizeType j = 0; j < local_tile_cols; ++j) {
     for (SizeType i = 0; i < local_tile_rows; ++i) {
-      pika::dataflow(dlaf::getCopyExecutor<Source, Destination>(),
-                     unwrapExtendTiles(dlaf::matrix::internal::copy_o),
-                     source.read(LocalTileIndex(i, j)), dest(LocalTileIndex(i, j)));
+      ex::when_all(source.read_sender(LocalTileIndex(i, j)),
+                   dest.readwrite_sender(LocalTileIndex(i, j))) |
+          copy(dlaf::internal::Policy<internal::CopyBackend_v<Source, Destination>>{}) |
+          ex::start_detached();
```
msimberg, line 47: _This_ closes https://github.com/eth-cscs/DLA-Future/issues/448 (not https://github.com/eth-cscs/DLA-Future/pull/460/files#r812685238). This seems to not be an issue anymore.

include/dlaf/communication/kernels/broadcast.h
msimberg: Actually, this does: https://github.com/eth-cscs/DLA-Future/pull/460/files#r812688112.

msimberg: And note, I changed this to sender adaptors in this PR to make sure that the old CUDA pools were not used in the benchmarks anymore (it turned out not to really make a difference, but it needed to be done anyway).

include/dlaf/schedulers.h
```diff
@@ -0,0 +1,42 @@
...
+#include <pika/thread.hpp>
+
+#ifdef DLAF_WITH_CUDA
+#include <pika/modules/async_cuda.hpp>
+#endif
+
+#include <dlaf/init.h>
+#include <dlaf/types.h>
+
+namespace dlaf {
```
msimberg, line 24: TODO: This might not need to be public... it's currently only used inside `transform` which is internal. Opinions?

include/dlaf/sender/transform.h
```diff
@@ -25,210 +26,81 @@
...
+  if constexpr (B == Backend::MC) {
+    return then(std::move(transfer_sender), std::move(f_unwrapping));
+  }
+  else if constexpr (B == Backend::GPU) {
+#if defined(DLAF_WITH_CUDA)
+    using pika::cuda::experimental::then_with_cublas;
+    using pika::cuda::experimental::then_with_cusolver;
+    using pika::cuda::experimental::then_with_stream;
+
+    if constexpr (std::is_invocable_v<decltype(then_with_stream), decltype(std::move(transfer_sender)),
```
msimberg, line 69: Note to @auriane (and FYI for others), this used `then_with_any_cuda` for a moment but because of the problems with some of the HIP/ROCm handles being `void*` and this needing explicit dispatching I've reverted the choice of CUDA stream/cuBLAS handle/cuSOLVER handle back to DLA-Future. This will be something like `if constexpr (dispatch_type == cuda_stream)` after #459.

spack/packages/dla-future/package.py
```diff
@@ -36,7 +36,7 @@ class DlaFuture(CMakePackage, CudaPackage):
     conflicts("umpire@6:")
 
     depends_on("pika cxxstd=17 +mpi")
-    depends_on("pika@0.1.0")
+    depends_on("pika@cuda-scheduler-2")
```
msimberg, line 39: To do:
```suggestion
    depends_on("pika@0.3.0")
```

src/init.cpp
```diff
@@ -245,6 +266,10 @@ pika::program_options::options_description getOptionsDescription() {
 
   desc.add_options()("dlaf:help", "Print help message");
   desc.add_options()("dlaf:print-config", "Print the DLA-Future configuration");
+  desc.add_options()("dlaf:num-np-cuda-streams", pika::program_options::value<std::size_t>(),
+                     "Number of normal priority CUDA streams");
+  desc.add_options()("dlaf:num-hp-cuda-streams", pika::program_options::value<std::size_t>(),
+                     "Number of high priority CUDA streams");
```
msimberg, line 272: To do: remove.

include/dlaf/schedulers.h
msimberg: I've moved it now to `dlaf::internal` but can move it back if someone prefers that.

msimberg:
Outside of pika release/CI/version bump changes this is now ready for review.

teonnik (approved):

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

include/dlaf/schedulers.h
msimberg: Feel free to reopen this if someone feels differently.

msimberg:
bors try

rasolca (approved):

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

include/dlaf/eigensolver/bt_reduction_to_band/impl.h
```diff
@@ -60,7 +60,7 @@ template <>
 struct Helpers<Backend::GPU> {
   template <class T>
   static void copyAndSetHHUpperTiles(SizeType j_diag, const matrix::Tile<const T, Device::GPU>& src,
-                                     matrix::Tile<T, Device::GPU>& dst, cudaStream_t stream) {
+                                     const matrix::Tile<T, Device::GPU>& dst, cudaStream_t stream) {
```
msimberg, line 63: This did not compile with a non-const reference. I will check separately why that happens. I've for now changed it to a const reference (which is what we use in other tile operations) and I'm tracking it here: https://github.com/pika-org/pika/issues/136.

msimberg:
This is now ready to be merged.

PR #462, opened 2022-02-08
Eigensolver + Gen Eigensolver Implementation + Test

albestro:
This is what @rasolca worked on few months ago when we tested the (almost) full pipeline for both generalized and standard eigensolver.

This PR adds the implementation of them together with their tests.

- [x] Fix test (it seems a matter of numeric error allowed)
- [x] Add documentation

I will highlight a couple of things inline that may be worth being commented and discussed.

Changelog:
- ReductionToBand API was not coherent with its backtransform: now both deals with `common::internal::vector<shared_future<common::internal::vector<T>>>`

### DIFF
--- a/include/dlaf/eigensolver/eigensolver.h
+++ b/include/dlaf/eigensolver/eigensolver.h
@@ -31,7 +31,7 @@ namespace eigensolver {
 /// @param uplo specifies if upper or lower triangular part of @p mat will be referenced
 /// @param mat contains the Hermitian matrix A
 template <Backend backend, Device device, class T>
-ReturnEigensolverType<T, device> eigensolver(blas::Uplo uplo, Matrix<T, device>& mat) {
+EigensolverResult<T, device> eigensolver(blas::Uplo uplo, Matrix<T, device>& mat) {
   DLAF_ASSERT(matrix::local_matrix(mat), mat);
   DLAF_ASSERT(square_size(mat), mat);
   DLAF_ASSERT(square_blocksize(mat), mat);
--- a/include/dlaf/eigensolver/eigensolver.h
+++ b/include/dlaf/eigensolver/eigensolver.h
@@ -18,7 +18,18 @@
 namespace dlaf {
 namespace eigensolver {
 
-/// TODO
+/// Standard Eigensolver.
+///
+/// It solves the standard eigenvalue problem A * x = lambda * x.
+///
+/// On exit, the lower triangle or the upper triangle (depending on @p uplo) of @p mat_a,
+/// including the diagonal, is destroyed.
+///
+/// Implementation on local memory.
+///
+/// @return struct ReturnEigensolverType with eigenvalues, as a vector<T>, and eigenvectors as a Matrix
+/// @param uplo specifies if upper or lower triangular part of @p mat will be referenced
+/// @param mat contains the Hermitian matrix A
 template <Backend backend, Device device, class T>
 ReturnEigensolverType<T, device> eigensolver(blas::Uplo uplo, Matrix<T, device>& mat) {
   DLAF_ASSERT(matrix::local_matrix(mat), mat);
--- a/include/dlaf/eigensolver/eigensolver.h
+++ b/include/dlaf/eigensolver/eigensolver.h
@@ -1,0 +1,31 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2021, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+#pragma once
+
+#include <blas.hh>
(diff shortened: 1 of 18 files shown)


test/unit/eigensolver/CMakeLists.txt
```diff
@@ -12,19 +12,31 @@ add_subdirectory(mc)
 
 DLAF_addTest(test_band_to_tridiag
   SOURCES test_band_to_tridiag.cpp
-  LIBRARIES dlaf.eigensolver dlaf.core
+  LIBRARIES dlaf.core dlaf.eigensolver dlaf.factorization dlaf.solver
```
albestro, line 15: This is the problem I was talking about few weeks ago during one of our meeting, about CMake dependencies and the fact that OBJECT libraries are not propagated as dependencies.

I don't know if I addressed the problem as you actually suggested, but in this way I had to manually propagate the new dependencies to all existing targets that were already using `dlaf.eigensolver`. Indeed, by adding the new implementation to this library, now it also has new dependencies (i.e. `dlaf.factorization`, `dlaf.solver`).

test/unit/eigensolver/test_eigensolver.cpp
```diff
@@ -0,0 +1,98 @@
...
+using namespace testing;
+
+template <typename Type>
+class EigensolverTestMC : public ::testing::Test {};
+TYPED_TEST_SUITE(EigensolverTestMC, MatrixElementTypes);
+
+const std::vector<blas::Uplo> blas_uplos({blas::Uplo::Lower});
+
+const std::vector<std::tuple<SizeType, SizeType>> sizes = {
+    //    {0, 2},                              // m = 0
```
albestro, line 35: what about the test cases? are we ok? should we add back also more empty cases?

include/dlaf/eigensolver/eigensolver.h
```diff
@@ -0,0 +1,31 @@
...
+#include <blas.hh>
+#include "dlaf/eigensolver/eigensolver/mc.h"
+#include "dlaf/matrix/matrix.h"
+#include "dlaf/types.h"
+#include "dlaf/util_matrix.h"
+
+namespace dlaf {
+namespace eigensolver {
+
+/// TODO
```
msimberg, line 21: TODO what?

include/dlaf/eigensolver/eigensolver/api.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include "dlaf/common/vector.h"
+#include "dlaf/matrix/matrix.h"
+#include "dlaf/types.h"
+
+namespace dlaf {
+namespace eigensolver {
+
+template <class T, Device device>
+struct ReturnEigensolverType {
```
msimberg, line 20:
```suggestion
struct EigensolverReturnType {
```
?

include/dlaf/eigensolver/eigensolver/api.h
```diff
@@ -0,0 +1,32 @@
...
+  common::internal::vector<BaseType<T>> eigenvalues;
```
msimberg, line 21: Presumably in the real version this can't be part of the return type, right? (Assuming it's supposed to be async.)

rasolca: Yes, it will be replaced by the async return type used in the tridiagonal eigensolver.

include/dlaf/eigensolver/eigensolver.h
rasolca: As TODO is preceded by 3 `\` it refers to the missing documentation... 😁

msimberg: Ah, thanks :heavy_check_mark:

include/dlaf/eigensolver/eigensolver/api.h
albestro: @rasolca @msimberg what about `EigensolverResultType`/`EigensolverResult`? any preference among all proposals?

msimberg: Weak preference:
1. `EigensolverResult`
2. `EigensolverResultType`
...
100. (<- is actually 100 in markdown but github renders it as a 3) `ReturnEigensolverType`

test/unit/eigensolver/test_eigensolver.cpp
rasolca: The empty case should be enabled. Not sure why it is currently not.
What do you mean with `more` empty cases?

albestro:
Just realized that this depends on #456 for a fix about empty cases with reduction to band. 

After that, I will fix the conflicts and then the last missing point is the TODO above about the numeric error

image

what do you think we should do about it @rasolca?

Just for the records, it happens just with the std eigensolver, not with the gen one. And as it can be seen, it happens just with `T=double` and the error seems negligible (IMHO).

albestro:
bors try

msimberg (approved):

rasolca (approved):

PR #463, opened 2022-02-08
Update the default plot names to have `strong`/`weak` in them

msimberg:
Deviating from my suggestion yesterday to put it in the default suffix, I've changed the base filenames directly to always include `strong`/`weak`. `gen_chol_plots` and friends don't have `strong` in them, but that's what they're used for. So my only question is: is there a use case where it doesn't produce a strong scaling plot? If no, I'd rename the functions as well to have `strong` in them.


rasolca (approved):
They are strong scaling plotter.
Originally we had only them and when we added the weak scaling plotters we didn't rename them.

msimberg:
> They are strong scaling plotter. Originally we had only them and when we added the weak scaling plotters we didn't rename them.

Ah, indeed, I forgot about that. That means that the only reason not to rename the functions would be to keep old plotting scripts working without changes. Is that something we care about? (They're easy to update...)

rasolca:
You can rename them.

msimberg:
> You can rename them.

Done.

Issue #466, opened 2022-02-08
Ensure no more `future`-based adaptors are used


msimberg:
Replace the following with sender-based adaptors:
- [ ] `dataflow`
- [ ] `async`
- [ ] `future::then`
- [ ] `future::get`
- [ ] `wait_all`
- [ ] `when_all`

PR #475, opened 2022-02-08
Replace more future-based adaptors with sender equivalents

### LINKED ISSUE #457: Add partial versions of `transform` and friends
`transform`, `transformLift` and so on can currently only be used with the sender supplied as an argument, not partially applied. Add versions that can be used with `operator|`.

msimberg:
Part of #466. Fixes #457.

Definitely work in progress, so I ask you to not waste too much time on reviewing. Some tests are definitely still failing, hopefully simply due to my silly mistakes. I just want to (ab)use the CI to get feedback on what is missing.

Style question: I've out of habit used a lot of
```
... | ex::then(...) | ... // etc.
```
but generally we tend to use
```
using pika::execution::experimental::then;
... | then(...) | ... // etc.
```
Do you prefer the latter? If yes, I'll update the former to the latter. I think it fits better with the style used elsewhere.

I've kept the possibly conflicting changes to algorithms that are in progress. I'll drop those if/when necessary. It's useful to see that things are working correctly with those changes so I'm keeping them on the branch for the moment.

To do:
- [ ] benchmarks
- [x] prefer using `operator|` as much as possible


msimberg:
bors try

include/dlaf/auxiliary/norm/mc.h
```diff
@@ -78,14 +79,19 @@ dlaf::BaseType<T> Norm<Backend::MC, Device::CPU, T>::max_L(comm::CommunicatorGri
...
   }
 
   // than it is necessary to reduce max values from all ranks into a single max value for the matrix
 
   // TODO unwrapping can be skipped for optimization reasons
+  // TODO this requires a new sender adaptor to handle a vector of
+  // senders/futures before dataflow can be replaced
   NormT local_max_value = pika::dataflow(unwrapping([](const auto&& values) {
                                            if (values.size() == 0)
                                              return std::numeric_limits<NormT>::min();
```
msimberg, line 97: @albestro this is one of the places that still requires dataflow for a vector of futures.

misc/synchronization.md
```diff
@@ -36,6 +36,7 @@ however they cannot be compiled since they are simplified to improve readability
 ```cpp
   Matrix m;
 
+  // TODO: These should probably be updated to use sender algorithms?
   pika::dataflow(Task1, m({0, 0}));
```
msimberg, line 40: What would you like to do about the documentation? Eventually this needs be updated.

test/unit/matrix/test_matrix.cpp
```diff
@@ -1423,6 +1580,23 @@ TEST(MatrixExceptionPropagation, RWPropagatesInRWAccess) {
...
+  namespace ex = pika::execution::experimental;
+
+  auto matrix = createMatrix<TypeParam>();
+
+  auto s = matrix.readwrite_sender(LocalTileIndex(0, 0)) | ex::then([](auto&& f) {
+             // TODO: This reproduces the behaviour with futures, but is it a must?
+             auto f_local = std::move(f);
+             throw CustomException{};
+           }) |
+           ex::ensure_started();
```
msimberg, line 1593: FYI, this is due to a difference in how/when stored values are released in `future::then`/`execution::then` (and all the other sender adaptors). I have to check if it's the `unwrapping` in the future-version that produces the difference in behaviour or if it's something else.

include/dlaf/eigensolver/reduction_to_band/mc.h
```diff
@@ -287,12 +297,16 @@ pika::shared_future<common::internal::vector<T>> computePanelReflectors(
 
   auto panel_tiles = pika::when_all(matrix::select(mat_a, ai_panel_range));
 
+  // TODO: This needs a when_all equivalent that can take a range of senders.
   return pika::dataflow(getHpExecutor<Backend::MC>(), std::move(panel_task), std::move(panel_tiles));
```
msimberg, line 301: @albestro this is another of the places that still requires dataflow for a vector of futures.

include/dlaf/eigensolver/reduction_to_band/mc.h
```diff
@@ -532,6 +570,7 @@ pika::shared_future<common::internal::vector<T>> computePanelReflectors(
 
   auto panel_tiles = pika::when_all(matrix::select(mat_a, ai_panel_range));
 
+  // TODO: This needs a when_all equivalent that can take a range of senders.
   return pika::dataflow(getHpExecutor<Backend::MC>(), std::move(panel_task), std::move(panel_tiles),
```
msimberg, line 574: @albestro this is the last of the places that still requires dataflow for a vector of futures.

misc/synchronization.md
rasolca: I think we can add a note on outdated docs, and wait when the sender/receiver port is completed to update it.

msimberg: https://github.com/eth-cscs/DLA-Future/issues/476

rasolca:
> Style question: I've out of habit used a lot of
> 
> ```
> namespace ex = pika::execution::experimental;
> ... | ex::then(...) | ... // etc.
> ```
> 
> but generally we tend to use
> 
> ```
> using pika::execution::experimental::then;
> ... | then(...) | ... // etc.
> ```
> 
> Do you prefer the latter? If yes, I'll update the former to the latter. I think it fits better with the style used elsewhere.

Not sure if we should prefer one over the other in general.
Your approach is IMO better if you have to use different functions/classes of the namespace only once with the short namespace alias. The other approach is better if you use one or few functions several times.

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

msimberg:
bors try

CMakeLists.txt
```diff
@@ -93,7 +93,8 @@ else()
 endif()
 
 # ----- pika
-find_package(pika 0.1.0 REQUIRED EXACT)
+# TODO: 0.2.0 EXACT
+find_package(pika 0.1.0 REQUIRED)
```
msimberg, line 97: TODO

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -100,6 +105,11 @@ pika::shared_future<matrix::Tile<const T, Device::CPU>> computeTFactor(
 
     return matrix::Tile<const T, Device::CPU>(std::move(tile_t));
   };
+  // TODO: What's wrong here? This hangs.
+  // return ex::make_future(
+  //     dlaf::internal::transform(dlaf::internal::Policy<Backend::MC>(), pika::unwrapping(tfactor_task),
+  //                               ex::when_all(ex::keep_future(std::move(tile_taus)),
+  //                                            ex::keep_future(std::move(tile_v)), std::move(mat_t))));
   return pika::dataflow(pika::unwrapping(tfactor_task), std::move(tile_taus), std::move(tile_v),
                         std::move(mat_t));
```
msimberg, line 114: TODO

include/dlaf/eigensolver/reduction_to_band/mc.h
```diff
@@ -299,13 +309,22 @@ pika::shared_future<common::internal::vector<T>> computePanelReflectors(MatrixT<
...
   }
 
+  // TODO: Is when_all_vector still buggy?
+  // return pika::execution::experimental::when_all_vector(std::move(panel_tiles)) |
+  //        dlaf::internal::transform(dlaf::internal::Policy<Backend::MC>(
+  //                                      pika::threads::thread_priority::high),
+  //                                  std::move(panel_task)) |
+  //        pika::execution::experimental::make_future();
   return pika::dataflow(getHpExecutor<Backend::MC>(), std::move(panel_task),
                         pika::when_all(std::move(panel_tiles)));
```
msimberg, line 319: TODO

include/dlaf/eigensolver/reduction_to_band/mc.h
```diff
@@ -561,6 +603,14 @@ pika::shared_future<common::internal::vector<T>> computePanelReflectors(
...
+  // return pika::execution::experimental::when_all(pika::execution::experimental::when_all_vector(
+  //                                                    matrix::select(mat_a, ai_panel_range)),
+  //                                                std::move(mpi_col_chain_panel), std::move(trigger)) |
+  //        dlaf::internal::transform(dlaf::internal::Policy<Backend::MC>(
+  //                                      pika::threads::thread_priority::high),
+  //                                  std::move(panel_task)) |
+  //        pika::execution::experimental::make_future();
   auto panel_tiles = pika::when_all(matrix::select(mat_a, ai_panel_range));
 
   return pika::dataflow(getHpExecutor<Backend::MC>(), std::move(panel_task), std::move(panel_tiles),
```
msimberg, line 616: TODO

misc/synchronization.md
```diff
@@ -36,6 +36,7 @@ however they cannot be compiled since they are simplified to improve readability
 ```cpp
   Matrix m;
 
+  // TODO: These should probably be updated to use sender algorithms?
```
msimberg, line 39: Remove comment, #476 covers this.

spack/packages/dla-future/package.py
```diff
@@ -36,7 +36,8 @@ class DlaFuture(CMakePackage, CudaPackage):
     conflicts("umpire@6:")
 
     depends_on("pika cxxstd=17 +mpi")
-    depends_on("pika@0.1.0")
+    # TODO: 0.2.0
+    depends_on("pika@main")
```
msimberg, line 40: TODO

test/unit/matrix/test_util_matrix.cpp
```diff
@@ -236,13 +240,15 @@ void testSet0(const config_t& cfg, const comm::CommunicatorGrid& comm_grid) {
     panel.setRange(GlobalTileIndex(coord1D, head), GlobalTileIndex(coord1D, tail));
 
     for (const auto& idx : panel.iteratorLocal())
-      pika::dataflow(pika::unwrapping(tile::internal::laset_o), blas::Uplo::General, TypeParam(1),
-                     TypeParam(1), panel(idx));
+      // TODO: don't use transform, but when_all | laset(policy) | start_detached();
+      dlaf::internal::transformLiftDetach(dlaf::internal::Policy<dlaf::Backend::MC>(),
+                                          tile::internal::laset_o, blas::Uplo::General,
+                                          TypeParam(1), TypeParam(1), panel.readwrite_sender(idx));
```
msimberg, line 246: TODO

msimberg:
Most of the uses of dataflow and friends have now been replaced. The remaining TODOs are pieces of code that lead to test failures when using the new adaptors. I'll be focusing on those this week.

CMakeLists.txt
msimberg: To do: require 0.2.1 or 0.3.0.

msimberg:
bors try

test/unit/matrix/test_matrix.cpp
msimberg: It's `unwrapping` that makes the difference. I've added `unwrapping` to the sender tests as well and the behaviour is now the same.

msimberg:
bors try

msimberg:
bors try

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -432,24 +462,32 @@ template <class T>
...
-  const auto ex = getHpExecutor<Backend::MC>();
+  using pika::execution::experimental::keep_future;
+  using pika::threads::thread_priority;
 
   // Note:
   // Not all ranks in the column always hold at least a tile in the panel Ai, but all ranks in
   // the column are going to participate to the reduce. For them, it is important to set the
   // partial result W2 to zero.
-  pika::dataflow(ex, unwrapExtendTiles(tile::internal::set0_o), w2(LocalTileIndex(0, 0)));
+  tile::set0(dlaf::internal::Policy<Backend::MC>(thread_priority::high),
```
msimberg, line 472: To do: create a variable `policy_hp = Policy<Backend::MC>(thread_priority::high)`.

include/dlaf/lapack/tile.h
```diff
@@ -388,8 +387,10 @@ void assertExtendInfo(F assertFunc, cusolverDnHandle_t handle, CusolverInfo<T>&&
   DLAF_CUSOLVER_CALL(cusolverDnGetStream(handle, &stream));
   assertFunc(stream, info.info());
   // Extend info scope to the end of the kernel execution
-  pika::cuda::experimental::detail::get_future_with_event(stream)  //
-      .then(pika::launch::sync, [info = std::move(info)](pika::future<void>&&) {});
+  auto extend_info = [info = std::move(info)]() {};
+  pika::cuda::experimental::detail::get_future_with_event(stream) |
+      pika::execution::experimental::then(std::move(extend_info)) |
+      pika::execution::experimental::start_detached();
```
msimberg, line 393: To do: use `pika::cuda::detail::add_event_callback`? If I'm using `detail` might as well use it. However, would be best not to use `detail` in the first place... This might need some refactoring.

msimberg:
bors try

msimberg:
bors try

msimberg:
bors cancel

msimberg:
bors try-

msimberg:
bors try

include/dlaf/util_matrix.h
```diff
@@ -224,13 +229,17 @@ void set_random(Matrix<T, Device::CPU>& matrix) {
     GlobalTileIndex tile_wrt_global = dist.globalTileIndex(tile_wrt_local);
     auto tl_index = dist.globalElementIndex(tile_wrt_global, {0, 0});
     auto seed = tl_index.col() + tl_index.row() * matrix.size().cols();
-    auto rnd_f = pika::unwrapping([seed](auto&& tile) {
+
+    using TileType = typename std::decay_t<decltype(matrix)>::TileType;
+    auto rnd_f = pika::unwrapping([seed](TileType&& tile) {
```
albestro, line 234: is `pika::unwrapping` needed?

include/dlaf/matrix/tile.h
```diff
@@ -394,8 +400,7 @@ pika::shared_future<Tile<T, D>> splitTileInsertFutureInChain(pika::future<Tile<T
 
     return pika::make_tuple(std::move(tile), std::move(dep_tracker));
   };
-  auto tmp =
-      pika::split_future(tile.then(pika::launch::sync, pika::unwrapping(std::move(swap_promise))));
+  auto tmp = pika::split_future(std::move(tile) | ex::then(std::move(swap_promise)) | ex::make_future());
```
albestro, line 403: what about using C++17 structured binding?

include/dlaf/matrix/tile.h
```diff
@@ -362,12 +364,16 @@ namespace internal {
 template <class T, Device D>
 pika::future<Tile<T, D>> createSubTile(const pika::shared_future<Tile<T, D>>& tile,
                                        const SubTileSpec& spec) {
-  return pika::dataflow(
-      pika::launch::sync, [](auto tile, auto spec) { return Tile<T, D>(tile, spec); }, tile, spec);
+  namespace ex = pika::execution::experimental;
+  auto f = [](auto tile, auto spec) { return Tile<T, D>(tile, spec); };
```
albestro, line 368: What about something like this?

```suggestion
  auto f = [](const Tile<T, D>& tile, const SubTileSpec spec) { return Tile<T, D>(tile, spec); };
```

include/dlaf/matrix/panel.h
```diff
@@ -93,10 +93,12 @@ struct Panel<axis, const T, D> {
...
-      new_tile_fut.then(pika::launch::sync, pika::unwrapping([panel_tile_size](const auto& tile) {
-                          DLAF_ASSERT_MODERATE(panel_tile_size == tile.size(), panel_tile_size,
-                                               tile.size());
-                        }));
+      auto assert_tile_size = pika::unwrapping([panel_tile_size](ConstTileType const& tile) {
+        DLAF_ASSERT_MODERATE(panel_tile_size == tile.size(), panel_tile_size, tile.size());
+      });
+      pika::execution::experimental::keep_future(new_tile_fut) |
+          pika::execution::experimental::then(std::move(assert_tile_size)) |
+          pika::execution::experimental::start_detached();
```
albestro, line 101: Is there any reason for not using `dlaf::internal::transformDetach`? In case, you could also avoid explicit unwrapping in `assert_tile_size` definition.

msimberg:
bors try

albestro (commented):
First partial round of (random) comments 😄

spack/packages/dla-future/package.py
```diff
@@ -38,7 +38,7 @@ class DlaFuture(CMakePackage, CudaPackage):
     conflicts("umpire@6:")
 
     depends_on("pika cxxstd=17 +mpi")
-    depends_on("pika@0.2.0")
+    depends_on("pika@main")
```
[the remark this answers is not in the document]
msimberg: Yes, waiting for the next release.

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -418,8 +415,8 @@ void hemmComputeX(matrix::Panel<Coord::Col, T, D>& x, const SubMatrixView& view,
       const auto& tile_a = splitTile(a.read(ij), view(ij));
 
       if (is_diagonal_tile) {
-        hemmDiag<B>(thread_priority::high, ex::keep_future(tile_a), w.read_sender(ij),
-                    x.readwrite_sender(ij));
+        hemmDiag<B, T>(thread_priority::normal, ex::keep_future(tile_a), w.read_sender(ij),
```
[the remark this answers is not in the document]
msimberg: None! It's a mistake. Thanks for spotting it!

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -678,8 +678,8 @@ void hemmComputeX(comm::IndexT_MPI reducer_col, PanelT<Coord::Col, T>& x, PanelT
       const bool is_diagonal_tile = (ij.row() == ij.col());
 
       if (is_diagonal_tile) {
-        hemmDiag<B>(thread_priority::high, a.read_sender(ij_local), w.read_sender(ij_local),
-                    x.readwrite_sender(ij_local));
+        hemmDiag<B, T>(thread_priority::normal, a.read_sender(ij_local), w.read_sender(ij_local),
```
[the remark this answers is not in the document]
msimberg: Also a mistake...

include/dlaf/util_matrix.h
msimberg: No, it can go.

include/dlaf/matrix/panel.h
msimberg: Only to keep the `launch::sync` semantics, though I'm not sure how important that is.

msimberg:
bors try

include/dlaf/matrix/tile.h
msimberg: Note, it's `[](pika::shared_future<Tile<T, D>>&& tile, ...)` not just `Tile<T, D>`, so slightly more verbose than what you suggested. I still like being explicit, but let me know if you'd prefer the `auto` after all.

msimberg:
bors try

msimberg:
bors try

include/dlaf/matrix/internal/tile_future_manager.h
```diff
@@ -23,16 +23,21 @@ template <class ReturnTileType>
...
+  using pika::execution::experimental::then;
+
   using TileDataType = typename ReturnTileType::TileDataType;
   using NonConstTileType = typename ReturnTileType::TileType;
 
   DLAF_ASSERT_HEAVY(old_future.valid(), "");
-  return old_future.then(pika::launch::sync, [p = std::move(p)](
-                                                 pika::future<TileDataType>&& fut) mutable {
+
+  // This uses keep_future because we want to handle exceptions in a special way
```
albestro, line 35: Can you quickly explain this?

albestro (commented):
Mainly comments for checking my understanding of senders.

Thanks in advance for your explanations 😄 

Just a reminder about something you already know: there is a mix of styles (pipelining vs nesting).

test/unit/matrix/test_util_matrix.cpp
```diff
@@ -132,7 +139,7 @@ void check_is_hermitian(Matrix<const T, Device::CPU>& matrix, comm::Communicator
         continue;
 
       if (current_rank == owner_original) {
-        const auto& tile_original = matrix.read(index_tile_original).get();
+        const auto& tile_original = sync_wait(matrix.read_sender(index_tile_original)).get();
```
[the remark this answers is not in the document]
msimberg: Exactly correct. The extra `.get` never actually waits for the future in this situation; it will always be ready. But it's needed for unwrapping. This is something that needs to be cleaned up. This situation doesn't actually seem to come up in the library itself, but I wanted to keep the test as close as possible to what it used to be.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -458,7 +457,9 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
     // Create the first max_workers workers and then reuse them.
     auto& w_pipeline = workers[sweep % max_workers];
 
-    auto dep = pika::dataflow(executor_hp, unwrapping(init_sweep), sweep, w_pipeline(), deps[0]).share();
+    auto dep = ex::make_future(
+                   dlaf::internal::transformLift(policy_hp, init_sweep, sweep, w_pipeline(), deps[0]))
+                   .share();
```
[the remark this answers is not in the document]
msimberg: Yes, exactly like `dataflow`. However, this comment is a good reminder to check if it actually needs to be a shared future... From a quick glance it looks like it might not need to be one, but I may be missing something.

include/dlaf/matrix/tile.h
albestro: I like being explicit too.

My tendency in `auto` usage is to use it when the type is already clear in the line.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -467,9 +468,11 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
...
       const GlobalElementIndex index_v((sweep / b + step) * b, sweep);
 
-      pika::dataflow(pika::launch::sync, unwrapping(store_tau_v), w_pipeline(),
-                     mat_v(dist_v.globalTileIndex(index_v)), dist_v.tileElementIndex(index_v));
-      deps[step] = pika::dataflow(executor_hp, unwrapping(cont_sweep), w_pipeline(), deps[dep_index]);
+      dlaf::internal::whenAllLift(w_pipeline(), mat_v.readwrite_sender(dist_v.globalTileIndex(index_v)),
+                                  dist_v.tileElementIndex(index_v)) |
+          ex::then(store_tau_v) | ex::start_detached();
+      deps[step] = ex::make_future(
+          dlaf::internal::transformLift(policy_hp, cont_sweep, w_pipeline(), deps[dep_index]));
```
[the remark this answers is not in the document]
msimberg: Correct. `keep_future` is in a sense a lifetime helper, but it's also required for another technical reason: a `shared_future<T>` sends a `T const&` which requires copying the value to the receiver, which we can't do with tiles.

include/dlaf/matrix/internal/tile_future_manager.h
msimberg: This is partly to keep the current behaviour with futures and partly because I found it too awkward/impossible to express the same behaviour without `keep_future`.

If this didn't use `keep_future` exceptions would go through a separate channel and would not be handled by the lambda `set_promise`, and the promise would not get the exception set. Not necessarily wrong, but different, and I did not want to change the semantics at the moment.

A naive implementation could look a bit like:
```
return std::move(old_future) |
  let_error([&](std::exception_ptr ep) { p.set_exception(ep); }) |
  then([&](TileDataType&& tile) { return ReturnTileType(std::move(NonConstTileType(std::move(tile)).setPromise(std::move(p)))); });
```
but this does not handle the lifetime of the promise `p`. Seeing this again, it's possible this could be expressed with an additional `let_value` around this.

Alternatively, the semantics could also be changed, but I wouldn't want to do that here and without some additional discussion.

test/unit/matrix/test_matrix.cpp
```diff
@@ -1275,6 +1280,29 @@ TEST(MatrixDestructorFutures, NonConstAfterRead) {
...
+  {
+    auto matrix = createMatrix<TypeParam>();
+
+    auto shared_future = matrix.read_sender(LocalTileIndex(0, 0));
+    last_task = dlaf::internal::transform(
+                    dlaf::internal::Policy<dlaf::Backend::MC>(),
+                    [&is_exited_from_scope](auto&&) {
+                      try_waiting_guard(is_exited_from_scope);
+                      EXPECT_TRUE(is_exited_from_scope);
+                    },
```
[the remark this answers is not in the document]
msimberg: I'm not quite sure what you're proposing. Could maybe roughly sketch the code that you're imagining?

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -209,40 +209,36 @@ void updateTrailingPanel(const bool has_head, const std::vector<TileT<T>>& panel
   }
 }
 
-template <Backend B, typename ASender, typename WSender, typename XSender>
+template <Backend B, typename T, typename ASender, typename WSender, typename XSender>
 void hemmDiag(pika::threads::thread_priority priority, ASender&& tile_a, WSender&& tile_w,
               XSender&& tile_x) {
-  using T = dlaf::internal::SenderElementType<ASender>;
```
rasolca, line 215: When senders were first introduced we discussed about it and chose not to add an extra template parameter.
What did change?

test/unit/matrix/test_matrix.cpp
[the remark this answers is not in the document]
albestro: I was thinking about creating a function that accepts a reference to the guard, so that we can reuse it all over our tests.

Just a draft for sharing the idea, I'm pretty sure compiler won't be on my side.
```cpp
void waitAndSet(auto&&, auto& guard) {
  try_waiting_guard(guard);
  EXPECT_TRUE(guard);
}

last_task =
  dlaf::internal::whenAllLift(
    shared_future,
    std::reference_wrapper(is_exited_from_scope))
  | dlaf::internal::transform(dlaf::internal::Policy<dlaf::Backend::MC>(), waitAndSet)
  | make_future();
```

miniapp/miniapp_cholesky.cpp
```diff
@@ -237,13 +237,14 @@ void setUpperToZeroForDiagonalTiles(Matrix<T, Device::CPU>& matrix) {
     if (distribution.rankIndex() != distribution.rankGlobalTile(diag_tile))
       continue;
 
-    auto tile_set = unwrapping([](auto&& tile) {
+    auto tile_set = [](typename Matrix<T, Device::CPU>::TileType&& tile) {
```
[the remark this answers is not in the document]
msimberg: In some cases it's required. One-argument callables get instantiated internally in pika in an unrelated path with an unrelated type and fails to compile with `auto`. It's on my todo list to see if we can avoid that instantiation (though now I'm reminded to open an issue). In other cases it's just because I find it nice to be explicit but it's not strictly required.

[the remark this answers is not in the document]
msimberg: For reference, this is the corresponding issue: https://github.com/pika-org/pika/issues/181.

miniapp/miniapp_reduction_to_band.cpp
```diff
@@ -105,7 +105,7 @@ struct reductionToBandMiniapp {
       {
         GlobalTileIndex last_tile(matrix.nrTiles().rows() - 1, matrix.nrTiles().cols() - 2);
         if (matrix.rankIndex() == distribution.rankGlobalTile(last_tile))
-          matrix(last_tile).get();
+          pika::this_thread::experimental::sync_wait(matrix.readwrite_sender(last_tile));
```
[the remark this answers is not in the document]
msimberg: Merged latest master. This is resolved.

test/unit/matrix/test_matrix.cpp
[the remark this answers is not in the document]
msimberg: What do you think about
```
struct Waiter {
  std::atomic<bool>& is_exited_from_scope;
  template <typename T>
  void operator()(T&& t) {
    try_waiting_guard(is_exited_from_scope);
    EXPECT_TRUE(is_exited_from_scope);
  }
};

// ...

last task = shared_future |
  transform(Policy<Backend::MC>(), Waiter{is_exited_from_scope}) |
  make_future();
```
? Also not compiled... but `Waiter` would also work as a drop-in replacement in the `future::then` versions of the tests as well. They could of course also be changed to use `when_all` with your version.

I'm not set on either of them so I'll change this to whatever you prefer.

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

test/unit/matrix/test_matrix.cpp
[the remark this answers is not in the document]
albestro: Yeah, it was just an idea, and not a real request.

I think you should go for whatever it makes more sense for the code that will use it. If `then` is a use-case, I'm completely fine with the callable that does not require the guard as argument. Indeed, my point was about having re-usable code.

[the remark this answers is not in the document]
msimberg: I think it's a good idea. Somewhat unrelated, but small enough that I'll happily do the change here. I'll try to push something in a bit.

msimberg:
Note: I pushed some minor cosmetic changes yesterday in an attempt to be at least more consistent. I changed a few more places to use the pipeline operator, and added more `namespace ex = ...` where I felt it was worth it.

include/dlaf/eigensolver/reduction_to_band/impl.h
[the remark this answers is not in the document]
rasolca: IMO adding the template is an extra possible point of failure... Mismatch of types in senders generates huge compiler error messages and are hard to track, so even reducing by one the places where it can occour is a good step forward.
We realized that checking everything makes the code completely unreadable, but at least we kept the simple things...
Looking forward to use C++20 concepts... 😉

teonnik (approved):
LGTM!

msimberg:
bors try

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -467,9 +468,11 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
...
       const GlobalElementIndex index_v((sweep / b + step) * b, sweep);
 
-      pika::dataflow(pika::launch::sync, unwrapping(store_tau_v), w_pipeline(),
-                     mat_v(dist_v.globalTileIndex(index_v)), dist_v.tileElementIndex(index_v));
-      deps[step] = pika::dataflow(executor_hp, unwrapping(cont_sweep), w_pipeline(), deps[dep_index]);
+      dlaf::internal::whenAllLift(w_pipeline(), mat_v.readwrite_sender(dist_v.globalTileIndex(index_v)),
+                                  dist_v.tileElementIndex(index_v)) |
+          ex::then(store_tau_v) | ex::start_detached();
+      deps[step] = dlaf::internal::whenAllLift(w_pipeline(), deps[dep_index]) |
+                   dlaf::internal::transform(policy_hp, cont_sweep) | ex::make_future();
```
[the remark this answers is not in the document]
msimberg: The reason is the same as in https://github.com/eth-cscs/DLA-Future/pull/475#discussion_r841616786, i.e. only to keep the `launch::sync` behaviour unchanged.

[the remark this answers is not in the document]
albestro: IMO he preferred it because it was closer to the original code. Indeed, you can spot the `dataflow(launch::sync, ...)` in the old code, which means that we wanted to have a continuation, which is explicit in the new code with the `then`.

I anticipated @msimberg 😄 just for checking my understanding.

include/dlaf/eigensolver/reduction_to_band/impl.h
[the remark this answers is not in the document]
msimberg: > Looking forward to use C++20 concepts... wink

Yes, they can't come soon enough...

I changed this back to using `SenderElementType` now.

test/unit/matrix/test_matrix.cpp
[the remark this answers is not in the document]
msimberg: I moved this out into a helper function/struct. Please have a look to see if you agree with the implementation (and the name!).

rasolca (approved):

include/dlaf/auxiliary/norm/mc.h
```diff
@@ -78,21 +80,25 @@ dlaf::BaseType<T> Norm<Backend::MC, Device::CPU, T>::max_L(comm::CommunicatorGri
...
+
+  auto max_element = [](std::vector<NormT>&& values) {
+    DLAF_ASSERT(!values.empty(), "");
+    return *std::max_element(values.begin(), values.end());
+  };
+  NormT local_max_value =
+      tiles_max.empty() ? std::numeric_limits<NormT>::min()
+                        : sync_wait(ex::when_all_vector(std::move(tiles_max)) |
+                                    dlaf::internal::transform(dlaf::internal::Policy<Backend::MC>(),
+                                                              std::move(max_element)));
```
albestro, line 101: I don't know if it was deliberate or not, but `sync_wait` is not using the "pipeline style".

test/unit/matrix/test_matrix.cpp
```diff
@@ -1422,15 +1575,38 @@ TEST(MatrixExceptionPropagation, RWPropagatesInReadAccess) {
   EXPECT_THROW(f.get(), CustomException);
 }
 
+TEST(MatrixExceptionPropagation, RWPropagatesInReadAccessWithSenderAdaptors) {
+  namespace ex = pika::execution::experimental;
+
+  auto matrix = createMatrix<TypeParam>();
+
+  auto s = matrix.readwrite_sender(LocalTileIndex(0, 0)) |
+           then(unwrapping([](auto&&) { throw CustomException{}; })) | ensure_started();
```
albestro, line 1584: super minor: probably we could take the chance to refactor also this 😛 your choice 😄

albestro (approved):
Nice work! I really like the pipeline style.

Next comments are very minor, it can absolutely be merged even without addressing them.

test/unit/matrix/test_matrix.cpp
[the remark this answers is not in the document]
albestro: super nice! about the name it can be left like this. thanks

include/dlaf/auxiliary/norm/mc.h
msimberg: It is actually deliberate! I haven't checked if it's "non-pipeline" everywhere now, but changed the ones that were touched by this PR otherwise.

`pika::this_thread::sync_wait` (as you've noticed) can be used at the end of a pipeline. However, P2300 actually doesn't allow that, so I thought I'd start changing them to "non-pipeline" already now.

test/unit/matrix/test_matrix.cpp
msimberg: Sorry, nope :stuck_out_tongue: won't do that here.

include/dlaf/eigensolver/band_to_tridiag/mc.h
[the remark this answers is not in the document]
msimberg: This one I didn't check yet... next up on my list.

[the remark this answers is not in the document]
msimberg: I changed it to a plain future now. I also made `copy_diag` generic so that it takes `future` or `shared_future` (it only took `shared_future` previously, that was the reason for me still using `future::share` here).

rasolca:
bors try

albestro:
bors merge

albestro:
bors try

albestro:
bors merge-

albestro:
bors try

PR #477, opened 2022-02-10
Panel sub-tile improvements

albestro:
This is a fork of #456, with a change proposal for improving Panel when dealing with sub-tile.
Discussion started from https://github.com/eth-cscs/DLA-Future/pull/456#discussion_r802942287.

Mainly, it avoids allocating full-tiles even when it is known it will work with sub-tiles.

Before merging:
- [ ] Add doc for new Panel constructor
- [ ] Additional check for avoiding `splitTile` if first global tile is a "full sub-tile"


rasolca:
As commented in the other PR. I don't see a reason to have a specific treatment of this case.
If you wand to work with a `mb x b` panel you just allocate a panel with a `mb x b` block size. Why you need to allocate a panel with blocksize`mb x nb` forced to have `b` columns or less?

PR #478, opened 2022-02-11
Panel: add setRangeStart for non-tile aligned offset

albestro:
This change is extracted from the work going on in #456 with sub-tiles, for which it is needed.

In particular, for what concerns the Panel, the sub-tile requires that:
- first global tile of the panel may be incomplete due to a sub-tile offset;
- panel size along axis dimension has to be "constrained" to sub-tile size;

This PR targets just the first point, since for the second one the choice has been to exploit the non-strictness of the API.

We used to give the actual Distribution of the matrix which the panel supports to construct it. Now, in order to work with sub-tiles, we will pass a Distribution object created ad-hoc, with a blocksize customized to represent the sub-tile size.

It is a **TEMPORARY SOLUTION**, because it will be problematic with the distributed version (e.g. panel transposed broadcast). Anyway the plan is to fix it later when we will address it, so that now we can prioritize the completion of the local version.

- [x] Add documentation

### DIFF
--- a/include/dlaf/matrix/panel.h
+++ b/include/dlaf/matrix/panel.h
@@ -22,6 +22,7 @@
 #include "dlaf/communication/kernels.h"
 #include "dlaf/lapack/tile.h"
 #include "dlaf/matrix/distribution.h"
+#include "dlaf/matrix/index.h"
 #include "dlaf/matrix/matrix.h"
 #include "dlaf/matrix/matrix_base.h"
 #include "dlaf/types.h"
@@ -119,7 +120,7 @@ struct Panel<axis, const T, D> {
       internal_.insert(internal_linear_idx);
       auto tile = data_.read(fullIndex(index));
 
-      if (dim_ < 0)
+      if (dim_ < 0 && (isFirstGlobalTile(index) && isFirstGlobalTileFull()))
         return tile;
       else
         return splitTile(tile, {{0, 0}, tileSize(index)});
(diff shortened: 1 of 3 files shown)


test/unit/matrix/test_panel.cpp
```diff
@@ -42,14 +45,31 @@ struct PanelTest : public TestWithCommGrids {};
...
   const GlobalElementSize sz;
   const TileElementSize blocksz;
-  const GlobalTileIndex offset;
+  const GlobalTileIndex offset = {0, 0};
 };
 
 std::vector<config_t> test_params{
-    {{0, 0}, {3, 3}, {0, 0}},  // empty matrix
+    {{0, 0}, {3, 3}},  // empty matrix
+    {{8, 5}, {3, 3}},
```
albestro, line 72: This test-case, when distributed, resulted in edge cases, in particular in case of empty local matrices or just incomplete tiles available locally, for the next two tests:
- ExternalTiles
- Shrink

I opted for adding a check for this edge-cases and just skip them without over-complicating the implementation of the test.

Do you think it is ok?

albestro:
bors try

test/unit/matrix/test_panel.cpp
rasolca: It is fine for now.

rasolca (approved):
>It is a TEMPORARY SOLUTION, because it will be problematic with the distributed version (e.g. panel transposed broadcast). Anyway the plan is to fix it later when we will address it, so that now we can prioritize the completion of the local version.

Note: The problem is not this solution, but it is the simplified model of the transposed panel (that only works with square tiles) we are using in the broadcast.
This is something we should keep in mind when designing the new matrix API.

msimberg (approved):
No important comments, this looks good to me.

rasolca:
bors try

include/dlaf/matrix/panel.h
```diff
@@ -164,11 +165,26 @@ struct Panel<axis, const T, D> {
   /// @pre (just the index relevant for the axis of the panel)
   /// @pre start <= current end range of the panel
   /// @pre panel offset on construction <= start
-  void setRangeStart(GlobalTileIndex start_idx) noexcept {
+  void setRangeStart(const GlobalTileIndex& start_idx) noexcept {
     DLAF_ASSERT_MODERATE(!hasBeenUsed(), hasBeenUsed());
 
     start_ = start_idx.get(CoordType);
```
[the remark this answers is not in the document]
albestro: I will fix it in a separate PR 😉

PR #481, opened 2022-02-17
algo: backtransform tridiagonal to band, adaptation to sub-band

albestro:
This PR introduces the possibility to compute the backtransformation from tridiagonal to a band matrix, where the band size is a divisor of the blocksize.

This removes the constraint about having the band size equal to the blocksize, which will be mainly useful with the next GPU version of the algorithm. Indeed, with GPU block sizes will be bigger, but with this it is possible to keep the band size smaller, reducing the overload on the entire eigensolver pipeline.

- [x] Refactor `Helper`, from spaghetti code to something a bit more structured (and add some notes in there)
- [x] Add more test-cases for sub-band
- [x] ~minor: Improve allocation size for W panel (minor optimization)~ it's not worth it
- [x] minor: check tau calculation for complex case

Changelog:
- https://github.com/eth-cscs/DLA-Future/commit/aba53a5c0b1d74ea1f7ccfb2392d6ee5633b9a9c fixes a bug in the test, where tau was computed not using all reflector components (1 less). It didn't affect the test results because there is no check about orthogonality of the result (@rasolca already knew the limitation of this test).

### DIFF
--- a/include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
+++ b/include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
@@ -193,11 +193,11 @@ struct Helper {
 
     if (acrossTiles_) {
       const auto part0_nrows = b - 1;
-      parts_[0] = part_t(sub_offset.row() + 1, part0_nrows, nrefls);
-      parts_[1] = part_t(0, std::min(b, rows_below - b), nrefls, part0_nrows);
+      part_top_ = part_t(sub_offset.row() + 1, part0_nrows, nrefls);
+      part_bottom_ = part_t(0, std::min(b, rows_below - b), nrefls, part0_nrows);
     }
     else {
-      parts_[0] = part_t(sub_offset.row() + 1, std::min(fullsize, rows_below - 1), nrefls);
+      part_top_ = part_t(sub_offset.row() + 1, std::min(fullsize, rows_below - 1), nrefls);
     }
   };
 
@@ -213,14 +213,16 @@ struct Helper {
 
   // Return SubTileSpec to use for accessing Householder reflectors in well formed form
   matrix::SubTileSpec specHH() const noexcept {
-    return {{0, 0}, {parts_[0].rows() + parts_[1].rows(), nrefls_}};
+    return {{0, 0}, {part_top_.rows() + part_bottom_.rows(), nrefls_}};
   }
 
-  // Return requested part, which gives access to specs (see part_t for more info)
-  const auto& operator[](const SizeType index) const noexcept {
-    DLAF_ASSERT_MODERATE(index >= 0 && index < (affectsMultipleTiles() ? 2 : 1), index,
-                         affectsMultipleTiles());
-    return parts_[to_sizet(index)];
+  const auto& topPart() const noexcept {
+    return part_top_;
+  }
+
+  const auto& bottomPart() const noexcept {
+    DLAF_ASSERT_MODERATE(affectsMultipleTiles(), affectsMultipleTiles());
+    return part_bottom_;
   }
 
 private:
@@ -257,7 +259,8 @@ private:
   matrix::SubTileSpec input_spec_;
 
   bool acrossTiles_;
-  std::array<part_t, 2> parts_;
+  part_t part_top_;
+  part_t part_bottom_;
 };
 
 template <class T>
@@ -360,15 +363,17 @@ void BackTransformationT2B<Backend::MC, Device::CPU, T>::call(const SizeType ban
         const auto sz_e = mat_e.tileSize({ij.row(), j_e});
         auto tile_e = mat_e.read(LocalTileIndex(ij.row(), j_e));
 
-        computeW2<backend>(thread_priority::normal, keep_future(splitTile(tile_v, helper[0].specHH())),
-                           keep_future(splitTile(tile_e, helper[0].specEV(sz_e.cols()))), T(0),
+        computeW2<backend>(thread_priority::normal,
+                           keep_future(splitTile(tile_v, helper.topPart().specHH())),
+                           keep_future(splitTile(tile_e, helper.topPart().specEV(sz_e.cols()))), T(0),
                            mat_w2.readwrite_sender(LocalTileIndex(0, j_e)));
 
         if (helper.affectsMultipleTiles()) {
           auto tile_e = mat_e.read(LocalTileIndex(ij.row() + 1, j_e));
-          computeW2<backend>(thread_priority::normal, keep_future(splitTile(tile_v, helper[1].specHH())),
-                             keep_future(splitTile(tile_e, helper[1].specEV(sz_e.cols()))), T(1),
-                             mat_w2.readwrite_sender(LocalTileIndex(0, j_e)));
+          computeW2<backend>(thread_priority::normal,
+                             keep_future(splitTile(tile_v, helper.bottomPart().specHH())),
+                             keep_future(splitTile(tile_e, helper.bottomPart().specEV(sz_e.cols()))),
+                             T(1), mat_w2.readwrite_sender(LocalTileIndex(0, j_e)));
         }
       }
 
@@ -395,14 +400,14 @@ void BackTransformationT2B<Backend::MC, Device::CPU, T>::call(const SizeType ban
         const auto& tile_w2 = mat_w2.read_sender(idx_e);
 
         updateE<backend>(pika::threads::thread_priority::normal,
-                         keep_future(splitTile(tile_w, helper[0].specHH())), tile_w2,
-                         splitTile(tile_e, helper[0].specEV(sz_e.cols())));
+                         keep_future(splitTile(tile_w, helper.topPart().specHH())), tile_w2,
+                         splitTile(tile_e, helper.topPart().specEV(sz_e.cols())));
 
         if (helper.affectsMultipleTiles()) {
           auto tile_e = mat_e(LocalTileIndex{ij.row() + 1, j_e});
           updateE<backend>(pika::threads::thread_priority::normal,
-                           keep_future(splitTile(tile_w, helper[1].specHH())), tile_w2,
-                           splitTile(tile_e, helper[1].specEV(sz_e.cols())));
+                           keep_future(splitTile(tile_w, helper.bottomPart().specHH())), tile_w2,
+                           splitTile(tile_e, helper.bottomPart().specEV(sz_e.cols())));
         }
       }
 
--- a/include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
+++ b/include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
@@ -40,6 +40,8 @@ struct BackTransformationT2B<Backend::MC, Device::CPU, T> {
                    Matrix<const T, Device::CPU>& mat_hh);
 };
 
+namespace bt_tridiag {
+
 template <class T>
 pika::shared_future<matrix::Tile<const T, Device::CPU>> setupVWellFormed(
     const SizeType b, pika::shared_future<matrix::Tile<const T, Device::CPU>> tile_i,
@@ -138,9 +140,9 @@ auto updateE(pika::threads::thread_priority priority, WSender&& tile_w, W2Sender
       pika::execution::experimental::start_detached();
 }
 
-struct Helper {
-  Helper(const SizeType b, const SizeType nrefls, matrix::Distribution dist,
-         const GlobalElementIndex offset)
+struct TileAccessHelper {
+  TileAccessHelper(const SizeType b, const SizeType nrefls, matrix::Distribution dist,
+                   const GlobalElementIndex offset)
       : nrefls_(nrefls), input_spec_{dist.tileElementIndex(offset),
                                      {std::min(b, dist.size().rows() - offset.row()),
                                       std::min(b, dist.size().cols() - offset.col())}} {
@@ -262,6 +264,7 @@ private:
   part_t part_top_;
   part_t part_bottom_;
 };
+}
 
 template <class T>
 void BackTransformationT2B<Backend::MC, Device::CPU, T>::call(const SizeType band_size,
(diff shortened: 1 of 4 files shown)


include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -135,8 +138,103 @@ auto updateE(pika::threads::thread_priority priority, WSender&& tile_w, W2Sender
...
+        mode_ = Mode::DOUBLE_FULL;
+        const SizeType mb2 = dist.size().rows() - offset.row() - b;
+        parts_[0] = part_t{b - 1, nrefls, sub_offset.row() + 1};
+        parts_[1] = part_t{std::min(b, mb2), nrefls, 0, b - 1};
+      }
+    }
+
+    input_spec_ = {sub_offset,
+                   {std::min(b, dist.size().rows() - offset.row()),
+                    std::min(b, dist.size().cols() - offset.col())}};
```
rasolca, line 173: It is really hard to follow the logic here. Maybe a comment would be helpful.

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -135,8 +138,103 @@ auto updateE(pika::threads::thread_priority priority, WSender&& tile_w, W2Sender
...
+      : nrefls_(nrefls), input_spec_{dist.tileElementIndex(offset),
+                                     {std::min(b, dist.size().rows() - offset.row()),
+                                      std::min(b, dist.size().cols() - offset.col())}} {
+    const TileElementIndex& sub_offset = input_spec_.origin;
+
+    const SizeType tile_row = dist.globalTileFromGlobalElement<Coord::Row>(offset.row());
+    const bool isLastRow = tile_row == dist.nrTiles().rows() - 1;
+
+    const SizeType max_size = dist.size().rows() - offset.row() - 1;
+    if (isLastRow && max_size < 2 * b - 1) {
```
rasolca, line 153: If it is the last row, `max_size` should be <= `b`,  so the second part of the condition has effect only for b=1.
Is this correct?

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -135,8 +138,103 @@ auto updateE(pika::threads::thread_priority priority, WSender&& tile_w, W2Sender
...
+      }
+      else {
+        mode_ = Mode::DOUBLE_FULL;
+        const SizeType mb2 = dist.size().rows() - offset.row() - b;
+        parts_[0] = part_t{b - 1, nrefls, sub_offset.row() + 1};
+        parts_[1] = part_t{std::min(b, mb2), nrefls, 0, b - 1};
+      }
+    }
+
+    input_spec_ = {sub_offset,
```
rasolca, line 171: Why do you set it again with the same values?

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -149,13 +247,16 @@ void BackTransformationT2B<Backend::MC, Device::CPU, T>::call(Matrix<T, Device::
...
-  const SizeType mb = mat_e.blockSize().rows();
-  const SizeType b = mb;
+  // Note: if no householder reflectors are going to be applied (in case of trivial matrix)
+  if (mat_hh.size().rows() <= (dlaf::isComplex_v<T> ? 1 : 2))
+    return;
+
+  const SizeType b = band_size;
   const SizeType nsweeps = nrSweeps<T>(mat_hh.size().cols());
 
   const auto& dist_i = mat_hh.distribution();
```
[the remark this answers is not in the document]
albestro: I went for a generic I for INPUT, just because it is the matrix with reflectors in compact form. If you have any better suggestion, I'm ready to address that in this PR too 😉

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -168,175 +269,113 @@ void BackTransformationT2B<Backend::MC, Device::CPU, T>::call(Matrix<T, Device::
...
-
-  const SizeType dist_w_rows = [&]() {
-    if (last_w_tile_sz != 0)
-      return (mat_e.nrTiles().rows() - 1) * w_tile_sz.rows() + last_w_tile_sz;
-    return (mat_e.nrTiles().rows() - 2) * w_tile_sz.rows() + (b - 1) + last_tile_size;
-  }();
-
-  const matrix::Distribution dist_w({dist_w_rows, w_tile_sz.cols()}, w_tile_sz, dist_i.commGridSize(),
-                                    dist_i.rankIndex(), dist_i.sourceRankIndex());
+  const SizeType dist_w_rows = [&]() { return mat_e.nrTiles().rows() * w_tile_sz.rows(); }();
```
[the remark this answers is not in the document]
albestro: Good point, it was a leftover of a change we opted for not pursuing anymore 👍🏻

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
albestro: Good catch, leftover of the refactoring 😄

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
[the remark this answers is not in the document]
rasolca: as the matrix is called `mat_hh` what about using `dist_hh`?

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
albestro: I tried to (hopefully) simplify the logic and document it.

See https://github.com/eth-cscs/DLA-Future/pull/481/commits/3e03aca7a5a052a4a29264b4b7ca86ee9821b32c

albestro:
bors try

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -135,8 +138,131 @@ auto updateE(pika::threads::thread_priority priority, WSender&& tile_w, W2Sender
       pika::execution::experimental::start_detached();
 }
 
+struct Helper {
+  Helper(const SizeType b, const SizeType nrefls, matrix::Distribution dist,
+         const GlobalElementIndex offset)
```
teonnik, line 143: Perhaps a more descriptive name can be found?

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -135,8 +138,131 @@ auto updateE(pika::threads::thread_priority priority, WSender&& tile_w, W2Sender
...
+
+    acrossTiles_ = [&]() {
+      // Note:
+      // A single tile is involved if:
+      // - it is the last row tile, so by construction reflectors will be applied to a single tile;
+      // - applying reflectors falls entirely withing a single tile
+      const SizeType i_tile = dist.globalTileFromGlobalElement<Coord::Row>(offset.row());
+      const bool is_last_row_tile = i_tile == dist.nrTiles().rows() - 1;
+      return !(is_last_row_tile || sub_offset.row() + 1 + fullsize <= mb);
+    }();
```
teonnik, line 192: Quite a lot of setup seems do be done in the constructor of this struct. Is it possible to refactor this into separate functions with names?

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -135,8 +138,131 @@ auto updateE(pika::threads::thread_priority priority, WSender&& tile_w, W2Sender
...
+    SizeType nrows_ = 0;
+    SizeType nrefls_ = 0;
+    SizeType offset_;
+  };
+
+  SizeType nrefls_;
+  matrix::SubTileSpec input_spec_;
+
+  bool acrossTiles_;
+  std::array<part_t, 2> parts_;
```
teonnik, line 260: If there are only two parts, why not have two `part_t` members with different names like `top/bottom` or `left/right` (not sure if these make sense)?

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -168,175 +297,113 @@ void BackTransformationT2B<Backend::MC, Device::CPU, T>::call(Matrix<T, Device::
...
+      const LocalTileIndex ij(dist_hh.localTileIndex(dist_hh.globalTileIndex(ij_e)));
+
+      // Note:  reflector with size = 1 must be ignored, except for the last step of the last sweep
+      //        with complex type
+      const SizeType nrefls = [&]() {
+        const bool allowSize1 = isComplex_v<T> && j == j_last_sweep && step == steps - 1;
+        const GlobalElementSize delta(dist_hh.size().rows() - ij_e.row() - 1,
+                                      std::min(b, dist_hh.size().cols() - ij_e.col()));
+        return std::min(b, std::min(delta.rows() - (allowSize1 ? 0 : 1), delta.cols()));
+      }();
```
teonnik, line 334: Why is this wrapped in a lambda?

teonnik (approved):
I am only vaguely familiar with the algorithm, I am afraid can't comment on the correctness in depth (except that the unit tests run fine :) ).

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
albestro: Yes, it makes sense (top/bottom). I'm going to do it 😉

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
albestro: It's a preference of mine. A `[&]() {}` lambda allows to:
- create a scope for variables (generally useful to keep clean the namespace)
- allow to initialize a `const` value in some complex cases (it's not this case, but [here](https://herbsutter.com/2013/04/05/complex-initialization-for-a-const-variable/) there is a good one).

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
albestro: Yes! Open to suggestions! 😄

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
albestro: IMHO it is a big comment chunk, but in the end it does two (linked) things:
- determine if there is just 1 or 2 parts (`acrossTiles_`)
- create descriptors for existing part(s)

Do you think it would be worth splitting them?

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
teonnik: In general, I am a bit wary of lambdas with `[&]` in asynchronous code because I have to then check if something with complicated lifetime is used in the lambda and if that may cause an issue. However, that doesn't seem to be the case here.

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
albestro: See https://github.com/eth-cscs/DLA-Future/pull/481/commits/e2b97d04d8aaf2e4f973a0f58db8fd96e1132c2b

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
albestro: It's a very good point, but it's not the case here as you said.

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
teonnik: I think it's not a blocker. Perhaps we can do it in the future.

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
rasolca: Just some ideas:
TileShapeHelper
TileAccessHelper
TileAccessSpecs
SubAccessSpecs

albestro: Currently, the namespace enclosing the full implementation (together with helpers) is `dlaf::eigensolver::internal`, which is not specific to this algorithm.

I fully agree with @teonnik about looking for a better naming for `Helper`. In order to do that I think we should think about what the name should do and if it has to identify a specific helper for this algorithm or if it has also to distinguish this helper from other helpers available for other algorithms.

My proposal would be to enclose helpers inside an additional namespace, e.g. `dlaf::eigensolver::internal::bt_tridiag`, so that we can use a name that indicates what the helper does, leaving the job of discriminating it from other algorithms to the additional namespace.

At that point, name proposed by @rasolca are perfectly fine: my personal preference is to have Helper inside the name, so I would go for `TileShapeHelper`/`TileAccessHelper`/`TileSpecHelper`.

rasolca: > My proposal would be to enclose helpers inside an additional namespace, e.g. `dlaf::eigensolver::internal::bt_tridiag`, so that we can use a name that indicates what the helper does, leaving the job of discriminating it from other algorithms to the additional namespace.

Fully agree.

include/dlaf/eigensolver/bt_band_to_tridiag/mc.h
```diff
@@ -135,8 +138,131 @@ auto updateE(pika::threads::thread_priority priority, WSender&& tile_w, W2Sender
...
+
+  // Return requested part, which gives access to specs (see part_t for more info)
+  const auto& operator[](const SizeType index) const noexcept {
+    DLAF_ASSERT_MODERATE(index >= 0 && index < (affectsMultipleTiles() ? 2 : 1), index,
+                         affectsMultipleTiles());
+    return parts_[to_sizet(index)];
+  }
+
+private:
+  struct part_t {
```
msimberg, line 229:
```suggestion
  struct Part {
```
?

albestro:
@rasolca please give a look at 4647db92ba3bdfe5ca54e9311f51bac7d222fcff and if it is ok, we can start the CI test.

albestro:
bors try

PR #488, opened 2022-02-23
Refactor tfactor to work only with well formed panels.

### LINKED ISSUE #486: TFactor: change API to accept well formed panel
Currently TFactor accepts a panel in the original matrix, that is not well formed,
i.e.
- reflectors does not start with a 1 as first component
- the part above each reflector (which progressively decrease in size, resulting in an unused part above them) contains noise (it is generally not equal to 0)

Considering that TFactor is not a public algorithm, and that where we use it we already prepare a well formed panel, passing it will ease the implementation of TFactor, which will not have anymore to manage the "special" case (which is somehow a code duplication).

It will indirectly help fixing also the instance of the problem shown in #483 (trivially, because that code won't be required anymore), but the general problem will still require a fix.

rasolca:
Closes #486.

Note: it depends on the merge of #484, hence please neglect the commit `Temp #484` when reviewing this PR.


include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -248,31 +227,16 @@ struct Helpers<Backend::GPU, Device::GPU, T> {
...
-
-  const auto panel_width = panel_view.cols();
-
-  t = splitTile(t, {{0, 0}, {k, k}});
-
-  DLAF_ASSERT(k <= panel_width, k, panel_width);
-
-  const GlobalTileIndex v_end{v.nrTiles().rows(), std::min(v.nrTiles().cols(), v_start_tile.col() + 1)};
-
-  // TODO S/R
```
albestro, line 267: Just a note in case you removed this TODO line inadvertently.

test/unit/factorization/test_compute_t_factor.cpp
```diff
@@ -399,39 +379,39 @@ void testComputeTFactor(comm::CommunicatorGrid grid, const SizeType a_m, const S
...
+  for (const auto& [m, k, mb, nb, v_start] : configs) {
+    testComputeTFactor<TypeParam, Backend::MC, Device::CPU>(m, k, mb, nb, v_start);
   }
 }
 
 TYPED_TEST(ComputeTFactorTestMC, CorrectnessDistributed) {
-  for (auto comm_grid : this->commGrids()) {
-    for (const auto& [a_m, a_n, mb, nb, k, v_start] : configs) {
-      testComputeTFactor<TypeParam>(comm_grid, a_m, a_n, mb, nb, k, v_start);
+  for (auto comm_grid : {this->commGrids()[0]}) {
```
albestro, line 404: I imagine it is a debug leftover...

```suggestion
  for (auto comm_grid : this->commGrids()) {
```

test/unit/factorization/test_compute_t_factor.cpp
```diff
@@ -92,8 +92,9 @@ void is_orthogonal(const MatrixLocal<const T>& matrix) {
 
 template <class T>
 std::tuple<dlaf::common::internal::vector<T>, MatrixLocal<T>> computeHAndTFactor(
-    const SizeType k, const MatrixLocal<const T>& v) {
-  const SizeType m = v.size().rows();
+    const SizeType k, const MatrixLocal<const T>& v, GlobalElementIndex v_start) {
+  // PRE: m >= k
```
albestro, line 96: Why not making it more explicit with an `ASSERT_*`?

test/unit/factorization/test_compute_t_factor.cpp
```diff
@@ -148,13 +149,14 @@ std::tuple<dlaf::common::internal::vector<T>, MatrixLocal<T>> computeHAndTFactor
 
 template <class T>
 MatrixLocal<T> computeHFromTFactor(const SizeType k, const Tile<const T, Device::CPU>& t,
-                                   const MatrixLocal<const T>& v) {
-  const SizeType m = v.size().rows();
+                                   const MatrixLocal<const T>& v, GlobalElementIndex v_start) {
+  // PRE: m >= k
```
albestro, line 153: Same comment as before.

test/unit/factorization/test_compute_t_factor.cpp
```diff
@@ -245,61 +239,55 @@ std::vector<std::tuple<SizeType, SizeType, SizeType, SizeType, SizeType, GlobalT
...
-
-  const GlobalTileSize v_offset{v_start.row(), v_start.col()};
-  for (const auto& ij_tile : iterate_range2d(v.nrTiles())) {
-    // copy only the panel
-    const auto& source_tile = v_input_h.read(ij_tile + v_offset).get();
-    matrix::internal::copy(source_tile, v.tile(ij_tile));
-    if (ij_tile.row() == 0)
-      tile::internal::laset<T>(blas::Uplo::Upper, 0.f, 1.f, v.tile(ij_tile));
-  }
+  auto v_local = dlaf::matrix::test::allGather<T>(blas::Uplo::General, v_h);
```
albestro, line 265: If I'm not wrong, you could enforce `const`-correctness on `v_local`.

```suggestion
  auto v_local = dlaf::matrix::test::allGather<const T>(blas::Uplo::General, v_h);
```

test/unit/factorization/test_compute_t_factor.cpp
```diff
@@ -310,85 +298,77 @@ void testComputeTFactor(const SizeType a_m, const SizeType a_n, const SizeType m
...
-    MatrixLocal<T> v(v_size, a.blockSize());
-
-    // copy only the panel
-    const GlobalTileSize v_offset{v_start.row(), v_start.col()};
-    for (const auto& ij : iterate_range2d(v.nrTiles())) {
-      matrix::internal::copy(a.tile_read(ij + v_offset), v.tile(ij));
-      if (ij.row() == 0)
-        tile::internal::laset<T>(blas::Uplo::Upper, 0.f, 1.f, v.tile(ij));
-    }
+  auto v_local = matrix::test::allGather<T>(blas::Uplo::General, v_h, grid);
```
albestro, line 341: In case, above comment about `const`-correctness applies here too.

albestro (approved):
LGTM in general.

Just as reminder, as you also suggested, add tests for the `Panel` extensions.

test/unit/factorization/test_compute_t_factor.cpp
rasolca: The function is not void. Anyway the assert is present in the parent function

include/dlaf/factorization/qr/t_factor_impl.h
rasolca: Sender and receivers are there...

include/dlaf/matrix/panel.h
```diff
@@ -262,6 +272,18 @@ struct Panel<axis, const T, D> {
     dim_ = height;
   }
 
+  /// Get the current width of the col panel.
+  template <Coord A = axis, std::enable_if_t<A == axis && Coord::Col == axis, int> = 0>
+  SizeType getWidth() const noexcept {
```
msimberg, line 277: Is this again `cols`? I'm not necessarily asking you to change, but just checking. This might be something that will be difficult to have consistent... I suppose the reason for this naming is that the setters are `setHeight/Width`.

msimberg (approved):
I have no useful comments, and only one question. This looks good to me.

include/dlaf/matrix/panel.h
rasolca: Exactly, to be consistent with the setters.

The setter were called setWidth/setHeight because they have a specific behaviour and they don't match any size.cols()/rows() you can get from the panel (or its parent distribution).

Anyway, in my plans for the API refactoring these methods will disappear, as I want to unify Vector, Panel and Matrix.

rasolca:
bors try

Issue #471, opened 2022-02-08
GPU reduction to band


rasolca:
Panel computation can remain on CPU.

PR #491, opened 2022-03-03
Generalize library structure for Reduction to band (GPU-local)

albestro:
This is a preliminary byproduct of the ongoing work for #471

For the moment GPU implementation is kept separated and unimplemented.


msimberg:
Do I understand correctly that this "only" restructures the reduction to band files to match the other algorithms that already have GPU/MC implementations? Is there anything else to keep an eye on here?

albestro:
> Do I understand correctly that this "only" restructures the reduction to band files to match the other algorithms that already have GPU/MC implementations? Is there anything else to keep an eye on here?

Correct. I have a branch where I did the actual work (on top of this), but I kept them in two separate steps with the hope to ease the review process.

test/unit/eigensolver/test_reduction_to_band.cpp
```diff
@@ -48,10 +50,20 @@ ::testing::Environment* const comm_grids_env =
     ::testing::AddGlobalTestEnvironment(new CommunicatorGrid6RanksEnvironment);
 
 template <class T>
-struct ReductionToBandTestMC : public TestWithCommGrids {};
+struct ReductionToBandTest : public TestWithCommGrids {};
+
+template <class T>
+using ReductionToBandTestMC = ReductionToBandTest<T>;
```
teonnik, line 56: Isn't this supposed to be `using ReductionToBandTestMC = ReductionToBandTest<Backend::MC>` ?

test/unit/eigensolver/test_reduction_to_band.cpp
```diff
@@ -48,10 +50,20 @@ ::testing::Environment* const comm_grids_env =
...
 
 TYPED_TEST_SUITE(ReductionToBandTestMC, MatrixElementTypes);
 
+#ifdef DLAF_WITH_CUDA
+template <class T>
+using ReductionToBandTestGPU = ReductionToBandTest<T>;
```
teonnik, line 62: and this : `ReductionToBandTestGPU = ReductionToBandTest<Backend::GPU> ?

test/unit/eigensolver/test_reduction_to_band.cpp
```diff
@@ -328,17 +344,35 @@ TYPED_TEST(ReductionToBandTestMC, CorrectnessLocal) {
...
+}
+
+TYPED_TEST(ReductionToBandTestGPU, CorrectnessLocalSubBand) {
+  for (const auto& config : configs_subband) {
+    const auto& [size, block_size, band_size] = config;
+
+    testReductionToBandLocal<TypeParam, Backend::GPU, Device::GPU>(size, block_size, band_size);
   }
 }
+#endif
```
rasolca, line 375: I would keep this tests out until you have an implementation.

test/unit/eigensolver/test_reduction_to_band.cpp
albestro: ReductionToBandTest is a GTest test parameterized over T, which is the type of the value used in test.

For us T is an element of `MatrixElementTypes`, on the other end we are going to have for each `Backend` an indepdendent test (because we also want to turn them off when `DLAF_WITH_CUDA=off`).

The `using` is just for giving a "self-documenting" name to the generic test, together with respecting the DRY principle by re-using exactly the same generic test (very simple in this case) for both Backend::MC and Backend::GPU.

I don't know if I was clear enough 😅

test/unit/eigensolver/test_reduction_to_band.cpp
albestro: Ok, so we can merge this independently from the implementation work.

In the meanwhile `master` will give a "not implemented" error in case this specific gpu function is going to be called.

teonnik (approved):

albestro:
I'm doing some test on my own, as soon as I have the green light I will start the CI 😉

albestro:
bors try

PR #493, opened 2022-03-08
algo: Reduction to Band (local GPU/hybrid)

### LINKED ISSUE #471: GPU reduction to band

albestro:
Close #471 

Implementation of hybrid GPU/CPU reduction to band (panel computation of HH reflectors runs on CPU, rest on GPU).

Main points:
- helper for code unification between CPU and GPU
- add `ElementType` to `Panel` API, going towards having a `MatrixLike` "concept"
- add test for reduction to band GPU
- basic needed senderification

TODO:
- [x] GPU panel computation uses a full blocksize panel also for sub-band
  It's a limitation because we are using a Panel as a Matrix (e.g. MatrixLike "concept"), in order to be able to apply `SubTileSpec` to both Matrix (CPU) and Panel (GPU).
- [x] Check what's the best thing for `copy + laset` in `setupReflectorPanelV`


include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -293,9 +304,13 @@ pika::shared_future<common::internal::vector<T>> computePanelReflectors(MatrixT<
                         pika::when_all(std::move(panel_tiles)));
 }
 
-template <class T, bool ForceCopy = false>
+template <Backend B, Device D, class T, bool ForceCopy = false>
```
rasolca, line 307:
```suggestion
template <Backend B, Device D, class T, bool force_copy = false>
```

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -680,33 +712,96 @@ void her2kUpdateTrailingMatrix(const LocalTileSize& at_start, MatrixT<T>& a,
...
+                        a.readwrite_sender(ij_local));
       }
     }
   }
 }
 }
 }
 
+template <Backend B, Device D, class T>
+struct Helper;
```
rasolca, line 737: I think a better name should be found.
In general, I'm fine with a generic name if it is a struct with just static members that is never initialized.
(the member function name is always visible when used)
When a instance of the struct is needed, the line of code with construction should be understandable and not a simple `Helper<...> helper(...)`;

rasolca (approved):
I would check all the TODOs:
 - are all of them relaay TODOs (maybe they are just implementation comments)
 - address it if it can be addressed easily

rasolca (approved):
Looks good.

msimberg (approved):

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -320,9 +330,12 @@ void setupReflectorPanelV(bool has_head, const SubPanelView& panel_view, const S
...
 
-    auto tile_v = v(i);
-    pika::dataflow(getHpExecutor<Backend::MC>(), std::move(setupV0), std::move(tile_v),
-                   splitTile(mat_a.read(i), spec));
+    const auto p = dlaf::internal::Policy<B>(thread_priority::high);
+    dlaf::internal::whenAllLift(ex::keep_future(splitTile(mat_a.read(i), spec)), v.readwrite_sender(i)) |
+        matrix::copy(p) | ex::start_detached();
+
+    dlaf::internal::whenAllLift(blas::Uplo::Upper, T(0), T(1), v.readwrite_sender(i)) | tile::laset(p) |
+        ex::start_detached();
```
albestro, line 338: Here there are two separate tasks when it could theoretically be a single one.

@msimberg @rasolca do you have any suggestion? if not, do you want me to add a note/issue about this?

msimberg: I think this is ok as it is. But an issue or a note doesn't hurt. It can always be ignored in the future!

albestro: note just added in the code 😉

albestro:
bors try

albestro:
Added `#ifdef DLAF_WITH_CUDA` guards around GPU helper specialization (clang complained, gcc didn't). In the meanwhile, I also rebased this branch onto last master.

Before re-running CI I have to investigate a test error. I will start it as soon I will solve the problem.

albestro:
bors try

PR #495, opened 2022-03-10
Add option to use precompiled headers

### LINKED ISSUE #452: Use precompiled headers
This has worked fairly well in HPX to speed up compile times. They don't magically fix compile times and don't affect instantiation times, but do give a nice speedup. With `pika` one of the main offenders (`PIKA_APPLICATION_NAME`) against using precompiled headers will be removed/made optional which should allow using them everywhere.

This should be considered a low-priority quality of life improvement. The required work for this is not very high.

msimberg:
Fixes #452.

While testing on daint this removes up to 5 seconds per translation unit. The initial compilation of the precompiled headers takes ~15 seconds.

Currently the default for the option is off, and I think at least CI should not have this enabled since this can hide the lack of inclusion of headers in the precompiled headers. My suggestion is that we keep it off by default for now, and you all test it in your own setups and if we later conclude that everyone can use this without problems we can enable it by default.


src/CMakeLists.txt
```diff
@@ -87,6 +87,43 @@ target_compile_definitions(dlaf.prop
...
 
+# Precompiled headers
+if(DLAF_WITH_PRECOMPILED_HEADERS)
+  add_library(dlaf.pch_lib OBJECT dummy.cpp)
+  target_link_libraries(dlaf.pch_lib PRIVATE dlaf.prop)
+  target_add_warnings(dlaf.pch_lib)
+
+  add_executable(dlaf.pch_exe dummy.cpp)
+  target_link_libraries(dlaf.pch_exe PRIVATE dlaf.prop)
+  target_add_warnings(dlaf.pch_exe)
```
msimberg, line 98: There are two precompiled headers because executables and libraries may be compiled with different flags making e.g. a library precompiled header incompatible with executables. In my case executables were compiled with `-fpie` and libraries not.

src/CMakeLists.txt
```diff
@@ -87,6 +87,43 @@ target_compile_definitions(dlaf.prop
...
+    <pika/future.hpp>
+    <pika/mpi.hpp>
+    <pika/mutex.hpp>
+    <pika/program_options.hpp>
+    <pika/runtime.hpp>
+    <pika/thread.hpp>
+    <pika/unwrap.hpp>
+
+    <blas.hh>
+    # <lapack.hh>
```
msimberg, line 114: This is disabled because of the `I` definition that would get pulled in and break everything (gtest uses `I` as a template parameter). The header doesn't add much to compilation times so I've simply excluded it. The only workaround I can think of is to include a header last in this list which undefines `I`. What would you prefer? I can add a comment if we decide to keep it excluded.

msimberg:
bors try

src/CMakeLists.txt
rasolca: I think we can keep it excluded.

rasolca (commented):
Will the option to use prcompiled headers be exposed in the spack package as well, or it will be used only during development?

cmake/DLAF_AddPrecompiledHeaders.cmake
```diff
@@ -0,0 +1,20 @@
...
+function(DLAF_addPrecompiledHeaders target_name)
+  if(DLAF_WITH_PRECOMPILED_HEADERS)
+    get_target_property(target_type ${target_name} TYPE)
+    if(target_type STREQUAL "EXECUTABLE")
+      target_precompile_headers(${target_name} REUSE_FROM dlaf.pch_exe)
+    else()
+      target_precompile_headers(${target_name} REUSE_FROM dlaf.pch_lib)
+    endif()
+  endif()
+endfunction()
```
albestro, line 20: [`target_precompile_headers`](https://cmake.org/cmake/help/latest/command/target_precompile_headers.html?highlight=precompiled%20headers) is a feature from CMake 3.16.

We have to update our requirements both in cmake and spack package.

src/CMakeLists.txt
albestro: Have you considered using an INTERFACE library instead of a classic library with a `dummy.cpp`?

Reporting also from the [doc](https://cmake.org/cmake/help/v3.16/command/target_precompile_headers.html?highlight=target_precompile_header), it looks like our use case
> [...] A notable exception to this is where an [interface library](https://cmake.org/cmake/help/v3.16/manual/cmake-buildsystem.7.html#interface-libraries) is created to define a commonly used set of precompile headers in one place and then other targets link to that interface library privately. In this case, the interface library exists specifically to propagate the precompile headers to its consumers and the consumer is effectively still in control, since it decides whether to link to the interface library or not.

Moreover, I'm missing the point of having `add_executable(dlaf.pch_exe dummy.cpp)` and then use it as a library. Why not just `add_library` here too?

albestro (changes requested):
Interesting work! Thanks @msimberg!

src/CMakeLists.txt
msimberg: Thanks for pointing that out! I had a vague memory of it not working as expected with interface libraries (probably because of wrong flags again) but I wasn't aware of this part of the docs and I might have done it wrong so I'll have another look. That would definitely be nicer.

The separate target for executables is there also for the correct flags but if using an interface library gets rid of that need then it's also not necessary to have a separate target.

Note that `dlaf.pch_exe` is _not_ used like a library which might be causing some of the confusion. Or I'm interpreting "use it as a library" differently than you mean.

msimberg: I will add some more comments into the code about this, but in short we can't (really) use interface libraries here.

This CMake issue is relevant: https://gitlab.kitware.com/cmake/cmake/-/issues/20288. Particularly this comment: https://gitlab.kitware.com/cmake/cmake/-/issues/20288#note_689092.

In short: one can add precompiled headers to interface libraries but they will also only be interface precompiled headers and no precompiled header will be generated for the interface target. This means one has to link to the interface library, not "reuse" the precompiled header from the library and all targets linking to this will have a separate precompiled header which is unnecessary.

src/CMakeLists.txt
msimberg: I've added a comment explaining this. Please have a look and close if you're happy with it.

src/CMakeLists.txt
msimberg: Added a comment about this. Please have a look.

msimberg:
> Will the option to use prcompiled headers be exposed in the spack package as well, or it will be used only during development?

I think it should not be a variant since compilation is supposed to have the same result. If it works well after us all using it for a while we can turn it on unconditionally, but the benefit is much smaller in the spack package since it's a one-off compilation not recompiled all the time.

rasolca (approved):

albestro (approved):

Issue #485, opened 2022-02-22
Add senderified communication helper for contiguous buffer


albestro:
In general I wonder if these communication helpers would benefit from (another) helper sender adaptor to abstract away a bit of this complexity (not so much a comment about this PR but rather about the complexity I added earlier!). Something that looks a bit like `let_value`, but instead of giving you a reference to the input sender value, it gives you a reference to an already contiguous buffer:
```
with_contiguous_buffer(sender_of_tile, [](contiguous_buffer& buf) {
    return dataflow(mpi_exec, mpi_func, buf, pipeline(), ...);
    // or soon: return when_all(just(buf), pcomm, ...) | transform_mpi(mpi_func); 
});
```
So essentially what you've done now but refactored into separate functions.

_Originally posted by @msimberg in https://github.com/eth-cscs/DLA-Future/pull/433#discussion_r762028518_

Issue #467, opened 2022-02-08
Remove executors


msimberg:
Replace with P2300 schedulers.
- [ ] MPI
- [ ] CUDA/cuBLAS/cuSOLVER

PR #497, opened 2022-03-14
Replace MPI executor with `transformMPI` sender adaptor

### LINKED ISSUE #465: Senderify MPI calls

msimberg:
Fixes #465, #466, and #467.

This first iteration simply replaces uses of the MPI executor with the `transform_mpi` sender adaptor. I've also locally refactored the communication helpers to `let_value` for temporary buffers/tiles, as @albestro already started in #433. I will further refactor them with something like #485 in a separate PR.

Depends on #475. That allows this PR to remove all the executors, including the various CUDA executors. Direct link to what's changed from #485: https://github.com/msimberg/DLA-Future/compare/more-senders...msimberg:transform-mpi.

To do:
- [ ] benchmarks


msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

src/init.cpp
```diff
@@ -57,10 +50,9 @@ template <>
 struct Init<Backend::MC> {
   static void initialize(configuration const& cfg) {
     memory::internal::initializeUmpireHostAllocator(cfg.umpire_host_memory_pool_initial_bytes);
-    // TODO: Consider disabling polling in finalize()
-    if (cfg.mpi_mech == comm::MPIMech::Polling) {
-      pika::mpi::experimental::init(false, cfg.mpi_pool);
-    }
+    // TODO: Is there any need to do polling on the MPI pool? Blocking MPI
+    // communication is also there and keeps polling from happening.
```
rasolca, line 54: Can you please explain what you mean exactly?

include/dlaf/communication/kernels/broadcast.h
```diff
@@ -58,67 +59,88 @@ void recvBcast(const matrix::Tile<T, D>& tile, comm::IndexT_MPI root_rank,
...
-        ScheduleRecvBcast<T>::call(ex, std::move(tile_cpu), root_rank, std::move(pcomm)))));
-
-    pika::execution::experimental::when_all(dlaf::internal::keepIfSharedFuture(std::move(tile_cpu)),
-                                            dlaf::internal::keepIfSharedFuture(std::move(tile_gpu))) |
-        dlaf::matrix::copy(
-            dlaf::internal::Policy<dlaf::matrix::internal::CopyBackend_v<Device::CPU, Device::GPU>>()) |
-        pika::execution::experimental::start_detached();
+    // TILE_GPU -+-> duplicate<CPU> ---> TILE_CPU ---> recvBcast ---> TILE_CPU -+-> copy
+    //           |                                                              |
+    //           +--------------------------------------------------------------+
```
rasolca, line 96: This diagram doesn't look correct, but it was already before the changes...
Shouldn't it be  
```
TILE_GPU -+-> duplicate<CPU> ---> TILE_CPU ---> recvBcast ---> TILE_CPU -+-> copy -+->
          +-----------------------------------> recvBcast -------------------------+ 
```
?

rasolca (commented):
Just reviewed the changes on top of #475 and it looks good.
Only a few minor comments.

test/unit/communication/test_comm_sender.cpp
```diff
@@ -33,37 +38,33 @@ void test_exec() {
...
   int tag = 0;
 
-  auto send_fut = pika::async(ex, MPI_Isend, send_buf.data(), size, dtype, send_rank, tag, comm);
-  auto recv_fut = pika::async(ex, MPI_Irecv, recv_buf.data(), size, dtype, recv_rank, tag, comm);
-  pika::wait_all(send_fut, recv_fut);
+  auto send = just(send_buf.data(), size, dtype, send_rank, tag, comm) | transform_mpi(MPI_Isend);
+  auto recv = just(recv_buf.data(), size, dtype, recv_rank, tag, comm) | transform_mpi(MPI_Irecv);
+  when_all(std::move(send), std::move(recv)) | then([](int e1, int e2) {
+    DLAF_MPI_CALL(e1);
+    DLAF_MPI_CALL(e2);
```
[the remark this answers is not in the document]
msimberg: Agree. We call the CUDA-equivalent of this in pika simply `check_cuda_error`. The difference is that it's not a macro (and thus can't do source location like in DLAF at the moment, until C++20). Would a rename to `DLAF_CHECK_MPI_ERROR` or similar be in place? Doesn't even have to be a rename, but can be a new macro. Note that I have a similar use on #475 (https://github.com/eth-cscs/DLA-Future/blob/79ea83cea5727c7c3ba993ec68c174c998fb8fa2/include/dlaf/lapack/tile.h#L390).

src/init.cpp
msimberg: I think I was either confused or just unclear when I wrote this. This is related to the second TODO that I've put in the description ("check if polling actually needs to run on MPI pool"). It's a TODO for the future, that I will leave out of this PR.

In any case, what I mean by this is that polling could run on the default pool (like the CUDA polling) and the MPI pool removed _if_ we make sure that no blocking calls ever happen on the default pool (cf. #512). The MPI pool is not fully utilized at the moment. However, moving the polling to the default pool may also increase latencies or have other bad effects.

include/dlaf/communication/kernels/broadcast.h
msimberg: Hmm, in your diagram the GPU tile is passed into `recvBcast` together with the copied CPU tile? Or what is the difference you're thinking of? I think the original diagram is accurate in that the GPU tile is first duplicated, and then used again after the `recvBcast` is done to be filled with the received data.

rasolca: Ok I understand it now...
I misinterpreted the diagram... thinking it represented the two possible paths with and without direct GPU comm...

test/unit/communication/test_comm_sender.cpp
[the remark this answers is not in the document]
rasolca: I missed the case in #475... For me both a rename or a new macro is ok.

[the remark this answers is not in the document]
msimberg: See the later commits in #475. I ended up renaming the macros to `DLAF_X_CHECK_ERROR`. I first added new macros, but found the look of
```
#define DLAF_CUDA_CALL(...) ...
#define DLAF_CUDA_CHECK_ERROR(cuda_err) DLAF_CUDA_CALL(cuda_err)
```
a bit silly.

I'll update this PR to use the new macros if you're happy with this when I rebase/merge.

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -14,65 +14,74 @@
...
-                       pika::future<common::PromiseGuard<comm::Communicator>> pcomm, MPI_Op reduce_op,
+void scheduleAllReduce(pika::future<common::PromiseGuard<comm::Communicator>> pcomm, MPI_Op reduce_op,
                        pika::shared_future<matrix::Tile<const T, Device::CPU>> tile_in,
                        pika::future<matrix::Tile<T, Device::CPU>> tile_out) {
-  using pika::dataflow;
+  using pika::execution::experimental::keep_future;
+  using pika::execution::experimental::let_value;
+  using pika::execution::experimental::start_detached;
+  using pika::execution::experimental::transfer;
+  using pika::execution::experimental::when_all;
```
msimberg, line 74:
```suggestion
  namespace ex = pika::execution::experimental;
```
?

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -83,37 +92,35 @@ void scheduleAllReduce(const comm::Executor& ex,
...
-    const comm::Executor& ex, pika::future<common::PromiseGuard<comm::Communicator>> pcomm,
-    MPI_Op reduce_op, pika::future<matrix::Tile<T, Device::CPU>> tile) {
-  using pika::dataflow;
-  using pika::unwrapping;
+    pika::future<common::PromiseGuard<comm::Communicator>> pcomm, MPI_Op reduce_op,
+    pika::future<matrix::Tile<T, Device::CPU>> tile) {
+  using pika::execution::experimental::let_value;
+  using pika::execution::experimental::make_future;
+  using pika::execution::experimental::then;
+  using pika::execution::experimental::transfer;
```
msimberg, line 116:
```suggestion
  namespace ex = pika::execution::experimental;
```
?

test/unit/communication/test_comm_sender.cpp
```diff
@@ -33,37 +39,33 @@ void test_exec() {
   int recv_rank = (rank != 0) ? rank - 1 : nprocs - 1;
   int tag = 0;
 
-  auto send_fut = pika::async(ex, MPI_Isend, send_buf.data(), size, dtype, send_rank, tag, comm);
-  auto recv_fut = pika::async(ex, MPI_Irecv, recv_buf.data(), size, dtype, recv_rank, tag, comm);
-  pika::wait_all(send_fut, recv_fut);
+  auto send = just(send_buf.data(), size, dtype, send_rank, tag, comm) | transform_mpi(MPI_Isend);
```
msimberg, line 42:
```suggestion
  auto send = just(send_buf.data(), size, dtype, send_rank, tag, comm) | dlaf::internal::transformMPI(MPI_Isend);
```
?

src/init.cpp
msimberg: To do: remove the comment...

test/unit/communication/test_comm_sender.cpp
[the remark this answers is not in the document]
msimberg: This has been updated.

msimberg:
bors try

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -83,37 +92,35 @@ void scheduleAllReduce(const comm::Executor& ex,
...
+      transfer(getBackendScheduler<Backend::MC>()) |
+      let_value(unwrapping(
+          [pcomm = std::move(pcomm), reduce_op](const matrix::Tile<const T, Device::CPU>& tile_in,
+                                                matrix::Tile<T, Device::CPU>& tile_out) mutable {
+            return whenAllLift(std::move(pcomm), reduce_op, makeItContiguous(tile_in),
+                               makeItContiguous(tile_out)) |
+                   transformMPI(internal::allReduce_o) |
+                   transform(Policy<Backend::MC>(thread_priority::high),
+                             std::bind(copyBack_o, std::placeholders::_1, std::cref(tile_out)));
+          })) |
```
teonnik, line 105: To make it a little easier to parse, do you mind refactoring the lambda out?

msimberg: Yeah, that's a good idea and I'll make the change here. Also keep in mind that the hope is that https://github.com/eth-cscs/DLA-Future/issues/485 will hopefully make this significantly easier to parse as well. I just don't want to mix that change into this PR. I'm fully aware that the current changes in this PR are not as readable as they could be...

test/unit/communication/test_comm_sender.cpp
teonnik: What is the difference?

msimberg: `dlaf::internal::transformMPI` is the DLAF helper adaptor which does unwrapping and, for the moment, does `yield_while` on the MPI request. I.e. this is what is currently actually used for communication on this branch. `pika::mpi::experimental::transform_mpi` is the pika facility that does the same with scheduler polling but no unwrapping. This is not currently used because of the performance drop that I mentioned earlier.

Eventually I want DLAF's wrapper `transformMPI` to call `transform_mpi` (potentially with unwrapping if required). Since `transformMPI` will be around in the future even if the implementation changes I'm just asking myself if this should use `transformMPI`. I'm 99% certain the answer is yes.

include/dlaf/sender/transform_mpi.h
```diff
@@ -0,0 +1,164 @@
...
+/// Helper for moving a PromiseGuard<Communicator> into the MPI function being
+/// called by transformMPI. Callables passed to transformMPI have their
+/// arguments passed by reference, but doing so with PromiseGuard would keep the
+/// guard alive until the completion of the MPI operation, whereas we are only
+/// looking to guard the submission of the MPI operation. We therefore use this
+/// wrapper to move PromiseGuard<Communicator> into a transformMPI callables so
+/// that the guard is released on return from the callable.
+template <typename T>
+decltype(auto) movePromiseGuard(T&& t) {
+  if constexpr (std::is_same_v<std::decay_t<T>, dlaf::common::PromiseGuard<Communicator>>) {
```
msimberg, line 31: Maybe add some sort of static assert here. Everything should be coming in as a reference already...

src/init.cpp
msimberg: Comment removed.

include/dlaf/communication/kernels/all_reduce.h
msimberg: I replaced this and a few other places of `using namespace pika::execution::experimental` with `namespace ex = pika::execution::experimental`.

include/dlaf/communication/kernels/all_reduce.h
msimberg: I moved the lambda to a separate variable. However, I couldn't think of a good name for it so it's now called `f` :stuck_out_tongue: If you have suggestions for a better name let me know.

include/dlaf/sender/transform_mpi.h
msimberg: I changed this to only take references with `T&`.

msimberg:
bors try

msimberg:
This _may_ be related to the performance of pika's own `transform_mpi`: https://github.com/pika-org/pika/issues/184. For future investigation...

msimberg:
CI seems to be happy. Starting benchmarks.

albestro (approved):
Very good work. Senders are really nice!!

include/dlaf/communication/kernels/reduce.h
```diff
@@ -214,19 +214,18 @@ void scheduleReduceSend(const comm::Executor& ex, comm::IndexT_MPI rank_root,
...
-  auto tile_cpu = dlaf::internal::transform(
-      Policy<CopyBackend<D, Device::CPU>::value>(pika::threads::thread_priority::high),
-      [](const matrix::Tile<const T, Device::GPU>& tile_gpu, auto... args) {
-        return dlaf::matrix::Duplicate<Device::CPU>{}(tile_gpu, args...);
-      },
-      keep_future(std::move(tile)));
+  auto tile_cpu = dlaf::internal::transform(Policy<CopyBackend<D, Device::CPU>::value>(
+                                                pika::threads::thread_priority::high),
+                                            dlaf::matrix::Duplicate<Device::CPU>{},
+                                            ex::keep_future(std::move(tile)));
```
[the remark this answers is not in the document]
msimberg: Thanks! Yes, we're going for it anywhere possible now I think. I'm just not being consistent yet.

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -83,37 +89,32 @@ void scheduleAllReduce(const comm::Executor& ex,
...
-    tile_out = std::move(pika::get<0>(wrapped.second));
-  }
-
-  cont_buf_out = getUnwrapReturnValue(dataflow(ex, unwrapExtendTiles(internal::allReduce_o),
-                                               std::move(pcomm), reduce_op, std::move(cont_buf_in),
-                                               std::move(cont_buf_out), tile_in));
-
-  dataflow(ex_copy, unwrapping(copyBack_o), std::move(cont_buf_out), std::move(tile_out));
+  auto f =
+      unwrapping([pcomm = std::move(pcomm), reduce_op](const matrix::Tile<const T, Device::CPU>& tile_in,
```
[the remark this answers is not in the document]
msimberg: Exactly, and `let_value` doesn't do unwrapping by itself.

include/dlaf/communication/kernels/broadcast.h
```diff
@@ -58,67 +59,87 @@ void recvBcast(const matrix::Tile<T, D>& tile, comm::IndexT_MPI root_rank,
...
+    using dlaf::matrix::internal::copy_o;
+
+    // TODO: std::bind currently serves as a reference_wrapper unwrapper until
+    // https://github.com/eth-cscs/DLA-Future/issues/492 is resolved.
+    return std::move(tile_gpu) |
+           // Start an asynchronous scope for keeping the GPU tile alive until
+           // data has been copied back into it.
+           ex::let_value([=, pcomm = std::move(pcomm)](Tile<T, Device::GPU>& tile_gpu) mutable {
+             // Create a CPU tile with the same dimensions as the GPU tile.
+             return transform(Policy<Backend::GPU>(),
```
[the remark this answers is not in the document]
msimberg: Good point. I doubt there was a specific reason, but high priority seems reasonable here.

src/CMakeLists.txt
```diff
@@ -106,7 +106,6 @@ if(DLAF_WITH_PRECOMPILED_HEADERS)
     $<$<BOOL:${DLAF_WITH_CUDA}>:<pika/cuda.hpp$<ANGLE-R>>
     <pika/execution.hpp>
     <pika/future.hpp>
```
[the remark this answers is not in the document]
msimberg: It's not actually used included/used anywhere at the moment because of the previously mentioned performance difference. The `transformMPI` implementation in `transform_mpi.h` is self-contained and all there is to it at the moment.

msimberg:
> About this, a thing I was wondering is: I see you cleaned up the code from executors, but the cuda pool executor is still there. What is it needed for? Isn't cuda "management" completely inside pika nowadays?

Unsure if I'm misunderstanding... The CUDA pool is implemented in pika, but instantiated in DLA-Future. And there is no CUDA pool executor.

albestro:
> > About this, a thing I was wondering is: I see you cleaned up the code from executors, but the cuda pool executor is still there. What is it needed for? Isn't cuda "management" completely inside pika nowadays?
> 
> Unsure if I'm misunderstanding... The CUDA pool is implemented in pika, but instantiated in DLA-Future. And there is no CUDA pool executor. Edit: unless I've forgotten to remove some files...

I think it was my oversight, thanks for the feedback 😉

include/dlaf/sender/transform_mpi.h
```diff
@@ -0,0 +1,179 @@
...
+/// Helper type for wrapping MPI calls.
+///
+/// Wrapper type around calls to MPI functions. Provides a call operator that
+/// creates an MPI request and passes it as the last argument to the provided
+/// callable. The wrapper then waits for the the request to complete with
+/// yield_while.
+///
+/// This could in theory be a lambda inside transformMPI.  However, clang at
+/// least until version 12 fails with an internal compiler error with a trailing
+/// decltype for SFINAE. GCC has no problems with a lambda.
```
msimberg, line 47: FYI, this seems to be a close to minimal reproducible example of the failure: https://github.com/pika-org/pika/issues/192.

msimberg:
bors try

rasolca (approved):

PR #499, opened 2022-03-15
algo: Backtransform Tridiagonalization (local GPU/hybrid)

### LINKED ISSUE #474: GPU Back tranformation (tridiagonalization)
conflicts:
#473

albestro:
Close #474

Implementation of backtransfom tridiagonalization on GPU. It's an hybrid solution with LARFT (computeT) computed on CPU.

Legend:
- RED: CPU related things
- BLU: GPU related things

image

TODO:
- [x] Evaluate helper refactoring (there is one for tile access of sub-band and one for gpu/cpu code unification)
- [x] Rename helper
- [x] Cleanup helper code

### DIFF
--- a/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
+++ b/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
@@ -260,6 +260,12 @@ private:
   Part part_bottom_;
 };
 
+template <class T, Device D>
+auto splitTileAndDiscard(pika::future<matrix::Tile<T, D>> tile,
+                         const matrix::SubTileSpec spec) noexcept {
+  return splitTile(tile, spec);
+}
+
 template <Backend B, Device D, class T>
 struct Helper;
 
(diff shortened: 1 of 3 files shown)


include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
```diff
@@ -259,13 +259,101 @@ struct TileAccessHelper {
   Part part_top_;
   Part part_bottom_;
 };
+
+template <class T, Device D>
+auto splitTileAndDiscard(pika::future<matrix::Tile<T, D>> tile,
+                         const matrix::SubTileSpec spec) noexcept {
+  return splitTile(tile, spec);
+}
+
```
albestro, line 268: This helper resulted useful whenever I wanted to split a readwrite tile, but I don't care about the "joined" future after the splitted parts are consumed.

It is very specific and targets exactly the use case I need here.

Waiting for your comments before moving it to a more generic place, and possibly extending it to other API signatures (if you think it is worth).

msimberg: How does this work differently than just calling `splitTile`? Is the difference that `splitTile` takes a reference to a tile and therefore you can't call it directly? And finally, is it still appropriate to have `splitTile` in the name if the original is discarded?

albestro: > Is the difference that `splitTile` takes a reference to a tile and therefore you can't call it directly?

You're right, it's exactly the point!



> And finally, is it still appropriate to have `splitTile` in the name if the original is discarded?

Probably it's not the best name, but IMHO *splitTile* still makes sense, because it refers to the action of spltting a tile in subtiles, not to what happens to the coming next future, for which I suffixed "discard" to describe this last behavior.

Just to point out that, it actually splits the tile, but it does not give you back the access to the immediately next tile ("discard" refers to this behavior), because its ownership is not kept by "anyone". However, it does not mean we won't be able to access anymore that tile, because we can always access the original source of the tile, i.e. the chain correctness is ensured.


I'm going to open a small PR with this change isolated, because it may need a bit of discussion. I will link that here too, as soon as I open it 😉

albestro: Here it is #504

albestro: It has been rebased and adapted since #504 got merged.

rasolca (approved):

include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
```diff
@@ -259,13 +259,95 @@ struct TileAccessHelper {
   Part part_top_;
   Part part_bottom_;
 };
+
+template <Backend B, Device D, class T>
+struct Helper;
+
+template <class T>
+struct Helper<Backend::MC, Device::CPU, T> {
```
msimberg, line 267: I guess this still needs a renaming. Is it also a `ComputePanelHelper` like in https://github.com/eth-cscs/DLA-Future/pull/493?

albestro: I fully agree, we have to find a better name before merging.

This helper, given an input tile with "compact" form reflectors, extracts and return them well-formed together with the T factor useful for their application in block.

Any name suggestion? @msimberg @rasolca @teonnik

rasolca: Some random suggestions:
`HHManager`, `HHMatricesHelper`
You can also replace HH with HHR if you find it easier to understand...

rasolca:
bors try

rasolca:
bors try

msimberg (approved):

rasolca:
bors try-

rasolca:
bors try

PR #504, opened 2022-03-17
splitTile discard

albestro:
I often address the situation where a subtile is needed, but the immediately after full tile, unlocked when the subtile goes out of scope, is not.

Dealing with readonly tiles is ok, I can do:

```cpp
auto ro_subtile = splitTile(matrix.read(ij), spec);
```

But when a readwrite one is needed, there is a downside. Indeed, it is required to do

```cpp
auto rw_fulltile = matrix(ij);
auto rw_subtile = splitTile(rw_fulltile, spec);
```

and it is not possible to simply do

```cpp
// this doesn't work, because splitTile requires a left-value reference `&`
auto rw_subtile = splitTile(matrix(ij), spec);
```

In my first proposal here, I used an additional name because of the "clash" that would provoke overloading `splitTile(future&)` (accepting lvalue ref) with a `splitTile(future)` (by value).

We may even evaluate if keeping `spitTile(future&)` is needed. The use-cases of `splitTile` as it is now, i.e. returing the immedaiately after tile, for which I think the main uses-cases are:
- accessing subtile of subtiles (recursive use-case), otherwise we would always had to start from the full tile everytime
- when we know we will attach something to the full tile as soon as we finish working on the subtiles; this removes the overhead introduced by an additional future "hop" it would otherwise require asking for the same tile to the "source"

### Minor changes:
- apply newly introduced split tile function where it is already useful in codebase
- fix build with apple-clang (which requires `.template foo<>()` syntax in some cases
- use structured bindings for `test_tile` replacing `tmp` + `std::get<>`.

### DIFF
--- a/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
+++ b/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
@@ -375,8 +375,7 @@ void BackTransformationT2B<B, D, T>::call(const SizeType band_size, Matrix<T, D>
 
       // Note:
       // And we use it also for computing the T factor
-      auto tile_t =
-          computeTFactor(tile_hh, tile_v, splitTileDiscard(mat_t(ij), {{0, 0}, {nrefls, nrefls}}));
+      auto tile_t = computeTFactor(tile_hh, tile_v, splitTile(mat_t(ij), {{0, 0}, {nrefls, nrefls}}));
 
       // W = V . T
       // Note:
@@ -396,13 +395,13 @@ void BackTransformationT2B<B, D, T>::call(const SizeType band_size, Matrix<T, D>
 
         updateE<B>(pika::threads::thread_priority::normal,
                    keep_future(splitTile(tile_w, helper.topPart().specHH())), tile_w2,
-                   matrix::splitTileDiscard(mat_e(idx_e), helper.topPart().specEV(sz_e.cols())));
+                   matrix::splitTile(mat_e(idx_e), helper.topPart().specEV(sz_e.cols())));
 
         if (helper.affectsMultipleTiles()) {
           updateE<B>(pika::threads::thread_priority::normal,
                      keep_future(splitTile(tile_w, helper.bottomPart().specHH())), tile_w2,
-                     matrix::splitTileDiscard(mat_e(LocalTileIndex{ij.row() + 1, j_e}),
-                                              helper.bottomPart().specEV(sz_e.cols())));
+                     matrix::splitTile(mat_e(LocalTileIndex{ij.row() + 1, j_e}),
+                                       helper.bottomPart().specEV(sz_e.cols())));
         }
       }
 
--- a/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
+++ b/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
@@ -375,8 +375,8 @@ void BackTransformationT2B<B, D, T>::call(const SizeType band_size, Matrix<T, D>
 
       // Note:
       // And we use it also for computing the T factor
-      auto tile_t_full = mat_t(LocalTileIndex(ij.row(), 0));
-      auto tile_t = computeTFactor(tile_hh, tile_v, splitTile(tile_t_full, {{0, 0}, {nrefls, nrefls}}));
+      auto tile_t =
+          computeTFactor(tile_hh, tile_v, splitTileDiscard(mat_t(ij), {{0, 0}, {nrefls, nrefls}}));
 
       // W = V . T
       // Note:
@@ -392,18 +392,17 @@ void BackTransformationT2B<B, D, T>::call(const SizeType band_size, Matrix<T, D>
         const LocalTileIndex idx_e(ij.row(), j_e);
         const auto sz_e = mat_e.tileSize({ij.row(), j_e});
 
-        auto tile_e = mat_e(idx_e);
         const auto& tile_w2 = mat_w2.read_sender(idx_e);
 
         updateE<B>(pika::threads::thread_priority::normal,
                    keep_future(splitTile(tile_w, helper.topPart().specHH())), tile_w2,
-                   splitTile(tile_e, helper.topPart().specEV(sz_e.cols())));
+                   matrix::splitTileDiscard(mat_e(idx_e), helper.topPart().specEV(sz_e.cols())));
 
         if (helper.affectsMultipleTiles()) {
-          auto tile_e = mat_e(LocalTileIndex{ij.row() + 1, j_e});
           updateE<B>(pika::threads::thread_priority::normal,
                      keep_future(splitTile(tile_w, helper.bottomPart().specHH())), tile_w2,
-                     splitTile(tile_e, helper.bottomPart().specEV(sz_e.cols())));
+                     matrix::splitTileDiscard(mat_e(LocalTileIndex{ij.row() + 1, j_e}),
+                                              helper.bottomPart().specEV(sz_e.cols())));
         }
       }
 
--- a/include/dlaf/matrix/tile.h
+++ b/include/dlaf/matrix/tile.h
@@ -454,6 +454,20 @@ pika::future<Tile<T, D>> splitTile(pika::future<Tile<T, D>>& tile, const SubTile
   return internal::createSubTile(old_tile, spec);
 }
 
+/// Create a writeable subtile of a given tile, without replacing given tile with the (full) tile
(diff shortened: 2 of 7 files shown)


msimberg:
Wouldn't this use case call for something like:
```
rw_subtile = matrix(ij, spec);
```
as the API instead of going via `splitTile`. The above could even internally just use `splitTile`, but it feels strange to me to have to "split a tile" when one just wants to directly access a subtile and nothing else. I may, however, be misunderstanding. Would something like `matrix(ij, spec)` still fit your use case?

include/dlaf/matrix/tile.h
```diff
@@ -454,6 +454,20 @@ pika::future<Tile<T, D>> splitTile(pika::future<Tile<T, D>>& tile, const SubTile
   return internal::createSubTile(old_tile, spec);
 }
 
+/// Create a writeable subtile of a given tile, without replacing given tile with the (full) tile
+/// that will come after the subtile will go out of scope.
+///
+/// The returned subtile will get ready, when the original tile was supposed to get ready.
+/// The next dependency in the dependency chain will become ready only when @p tile goes out of scope.
+template <class T, Device D>
+pika::future<Tile<T, D>> splitTileDiscard(pika::future<Tile<T, D>> tile, const SubTileSpec& spec) {
```
rasolca, line 463: You can hide the different function name need behind
```
pika::future<Tile<T, D>> splitTile(pika::future<Tile<T, D>>&& tile, const SubTileSpec& spec) {
  return splitTileDiscard(std::move(tile), spec);
}
```

albestro:
> Wouldn't this use case call for something like:
> 
> ```
> rw_subtile = matrix(ij, spec);
> ```
> 
> as the API instead of going via `splitTile`. The above could even internally just use `splitTile`, but it feels strange to me to have to "split a tile" when one just wants to directly access a subtile and nothing else. I may, however, be misunderstanding. Would something like `matrix(ij, spec)` still fit your use case?

Your objection is absolutely valid and the API `matrix(ij, spec)` could make sense.

My main point about this decision can be labelled as *separation of concerns*. Matrix gives access to a set of tiles, which may have different shapes depending on the overall size of the matrix. But when you want to work with a subtile, it is a tile duty to give you access to its subtiles. That's why a free function (or a function member of the tile itself) looks to me as the best choice from a design perspective.

Let me be clear, it is not completely wrong what you are proposing. In the future, if matrix will provide an API to access subviews of it (i.e. panel, submatrix), it will make use of the tile facilities to give access to sub-parts.

Another point of view of this aspect is about API maintenance. Let's see the options where an API providing splitTile functionalities can be:
- just in Tile (like it is now)
- just in Matrix
- both in Tile and in Matrix

In terms of design, IMHO, it is reasonable to have it for tiles, I consider it fundamental. Having it in Matrix looks like a good option, but it can't be exclusive. So I would say that we can look at option 1 and 3.

The downside of option 3, which is also what you are proposing, is that it easily becomes a burden to maintain. Indeed, we will have to replicate the same API we have for tiles, to be accessed via Matrix (e.g. for shared_future, for vectors of [spec]).

You may argue that maintainability in this case it is not a strong point, because we can probably do some trick with template meta-programming...but is it really worth?! 😄

include/dlaf/matrix/tile.h
albestro: After some research with @rasolca we opted for not pursuing this road, since it would introduce an additional cost (compiler cannot apply copy-elision).

So, we would go for keeping the two separate functions with different names without any helper.

albestro:
The new function has been applied where it shown to be useful in the codebase.

Pointing out two points where it was not worth using it.

https://github.com/eth-cscs/DLA-Future/blob/4754f309956d722128dd4962132b9ddf1d623bf3/include/dlaf/eigensolver/reduction_to_band/impl.h#L468-L485

https://github.com/eth-cscs/DLA-Future/blob/4754f309956d722128dd4962132b9ddf1d623bf3/include/dlaf/matrix/panel.h#L462-L466

msimberg:
> Your objection is absolutely valid and the API `matrix(ij, spec)` could make sense.
> 
> My main point about this decision can be labelled as _separation of concerns_. Matrix gives access to a set of tiles, which may have different shapes depending on the overall size of the matrix. But when you want to work with a subtile, it is a tile duty to give you access to its subtiles. That's why a free function (or a function member of the tile itself) looks to me as the best choice from a design perspective.
> 
> Let me be clear, it is not completely wrong what you are proposing. In the future, if matrix will provide an API to access subviews of it (i.e. panel, submatrix), it will make use of the tile facilities to give access to sub-parts.
> 
> Another point of view of this aspect is about API maintenance. Let's see the options where an API providing splitTile functionalities can be:
> 
>     * just in Tile (like it is now)
> 
>     * just in Matrix
> 
>     * both in Tile and in Matrix
> 
> 
> In terms of design, IMHO, it is reasonable to have it for tiles, I consider it fundamental. Having it in Matrix looks like a good option, but it can't be exclusive. So I would say that we can look at option 1 and 3.
> 
> The downside of option 3, which is also what you are proposing, is that it easily becomes a burden to maintain. Indeed, we will have to replicate the same API we have for tiles, to be accessed via Matrix (e.g. for shared_future, for vectors of [spec]).
> 
> You may argue that maintainability in this case it is not a strong point, because we can probably do some trick with template meta-programming...but is it really worth?! smile

All good points, agreed. My main concern is the name, and whether it's a free function, member function, or call operator is secondary in my opinion.

The existing `splitTile` functions "split" the tile in that it gives you two futures to the same tile. The "sub-tile"-functionality is important but not part of the name. Is my understanding here correct?

You now have a function (with pseudo-syntax for concepts) `sender_of<SubTile> f(sender_of<Tile>, spec)` where no splitting happens and creating a sub-tile is the main purpose, so I think it would be appropriate to have `sub` somewhere in the name.
- `subTile`
- `makeSubTile`
- `getSubTile`
- or... `createSubTile` (https://github.com/eth-cscs/DLA-Future/blob/4754f309956d722128dd4962132b9ddf1d623bf3/include/dlaf/matrix/tile.h#L362-L367) ;)

Isn't `createSubTile` exactly what you want, except for `future`, not `shared_future`? I'm sorry if it's not and I'm making a big fuss about nothing...

msimberg:
@albestro thanks for the response! I think we agree, I may just have been unclear in my explanation.

> Isn't `createSubTile` exactly what you want, except for `future`, not `shared_future`?

I'd still be interested in hearing your answer to this one ;) because I think it affects both the naming of the function and its implementation. Or maybe you implicitly answered yes and I didn't get it.

albestro:
> @albestro thanks for the response! I think we agree, I may just have been unclear in my explanation.

With `"split" the tile in that it gives you two futures to the same tile` it seemed to me that for you the "next" tile (`tile'` in the image below) was a direct result of split; instead in my view the split action operates on the source tile, creating as result one or more subtiles (`sub`), and just as byproduct we get the next tile (`tile'`), that we may want to keep or discard.

This is why I thought we were looking at the split keyword in two different ways 😄 

image

> Isn't `createSubTile` exactly what you want, except for `future`, not `shared_future`?

Yes, the functionality is exactly the same!

albestro:
> Isn't `createSubTile` exactly what you want, except for `future`, not `shared_future`?

Actually this comment was a very good one... 😁 We were looking at it with @rasolca and now I updated the code of `splitTile` with a more simplified version (it is not named discard anymore). Give it a look 😉

include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
```diff
@@ -375,8 +375,8 @@ void BackTransformationT2B<B, D, T>::call(const SizeType band_size, Matrix<T, D>
 
       // Note:
       // And we use it also for computing the T factor
-      auto tile_t_full = mat_t(LocalTileIndex(ij.row(), 0));
-      auto tile_t = computeTFactor(tile_hh, tile_v, splitTile(tile_t_full, {{0, 0}, {nrefls, nrefls}}));
+      auto tile_t =
+          computeTFactor(tile_hh, tile_v, splitTileDiscard(mat_t(ij), {{0, 0}, {nrefls, nrefls}}));
```
msimberg, line 379: Still `splitTileDiscard`? Maybe you have the changes locally... Same for all the other `splitTileDiscard`s.

msimberg (changes requested):
Thanks @albestro, I like it much better now!

albestro:
bors try

msimberg (approved):

rasolca (approved):

Issue #501, opened 2022-03-15
GPU SEP pipeline


albestro:
Implement GPU backend for SEP pipeline.

- Manage input to tridiagonalization
- Manage output from D&C tridiagonal eigensolver

The former one will probably remain, but the last one is just a transitional management until #472 will be merged.

Depends on #493 #499

PR #508, opened 2022-03-29
GPU kernel: LACPY 

albestro:
Implementation of `lacpy` GPU kernel (preparatory work for #501).

It's a basic implementation that uses `cudaMemcpy2DAsync` for the general case and a custom kernel for the lower and upper cases (note: the kernel implemented supports also the general case, even if it is not used).

IMHO it is a reasonable basic implementation that I can use for completing tasks related to the local GPU eigensolver deadline (specifically this kernel is required for tridiagonalization input management).

If you agree, as soon as them will be completed, we can iterate over this and benchmark different solutions/options.

Changes:
- `gpulapack::` kernels (currently `laset` and `lacpy`) now uses `blas::Uplo` instead of `cublasFillMode_t` (uniformly with cpu routines/wrappers)
- move lower/upper/all predicates to `util_cublas.h`
- minor fix doxygen documentation in `lapack/tile.h`

### DIFF
--- a/include/dlaf/lapack/gpu/lacpy.h
+++ b/include/dlaf/lapack/gpu/lacpy.h
@@ -1,0 +1,36 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2022, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+
+#ifdef DLAF_WITH_CUDA
+
+#include <cublas_v2.h>
+#include <cuda_runtime.h>
+
+#include "dlaf/types.h"
+
+namespace dlaf::gpulapack {
+
+template <class T>
+void lacpy(cublasFillMode_t uplo, SizeType m, SizeType n, const T* a, SizeType lda, T* b, SizeType ldb,
+           cudaStream_t stream);
+
+#define DLAF_CUBLAS_LACPY_ETI(kword, Type)                                                              \
+  kword template void lacpy(cublasFillMode_t uplo, SizeType m, SizeType n, const Type* a, SizeType lda, \
+                            Type* b, SizeType ldb, cudaStream_t stream)
+
+DLAF_CUBLAS_LACPY_ETI(extern, float);
+DLAF_CUBLAS_LACPY_ETI(extern, double);
+DLAF_CUBLAS_LACPY_ETI(extern, std::complex<float>);
+DLAF_CUBLAS_LACPY_ETI(extern, std::complex<double>);
+}
+
+#endif
--- a/include/dlaf/lapack/gpu/lacpy.h
+++ b/include/dlaf/lapack/gpu/lacpy.h
@@ -14,17 +14,18 @@
 
 #include <cublas_v2.h>
 #include <cuda_runtime.h>
+#include <blas.hh>
 
 #include "dlaf/types.h"
 
 namespace dlaf::gpulapack {
 
 template <class T>
-void lacpy(cublasFillMode_t uplo, SizeType m, SizeType n, const T* a, SizeType lda, T* b, SizeType ldb,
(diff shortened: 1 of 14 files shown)


include/dlaf/lapack/gpu/lacpy.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <cublas_v2.h>
+#include <cuda_runtime.h>
+
+#include "dlaf/types.h"
+
+namespace dlaf::gpulapack {
+
+template <class T>
+void lacpy(cublasFillMode_t uplo, SizeType m, SizeType n, const T* a, SizeType lda, T* b, SizeType ldb,
```
albestro, line 23: With the aim of unifying our API, probably it would be better to use `blaspp::Uplo` instead of `cublasFillMode_t`, and convert it internally.

include/dlaf/util_cublas.h
```diff
@@ -48,6 +48,21 @@ inline constexpr cublasFillMode_t blasToCublas(const blas::Uplo uplo) {
   }
 }
 
+/// Predicate returning true for all coordinates lying in the lower part or on the diagonal of a matrix
+__device__ inline constexpr bool isLower(unsigned i, unsigned j) noexcept {
+  return i >= j;
+}
```
msimberg, line 54: As mentioned already to @albestro these could just be `inline constexpr` with the `--expt-relaxed-constexpr` flag to `nvcc`. Basically it allows using constexpr functions anywhere. But this is also fine since the CUDA headers define `__device__` to nothing for non-CUDA compilers.

src/lapack/gpu/lacpy.cu
```diff
@@ -0,0 +1,136 @@
...
+  const unsigned i = blockIdx.x * kernel_tile_size_rows + threadIdx.x;
+  const unsigned j = blockIdx.y * kernel_tile_size_cols;
+
+  if (i >= m)
+    return;
+
+  const unsigned k_max = min(j + kernel_tile_size_cols, n);
+
+  for (unsigned k = j; k < k_max; ++k)
+    if (i == k or (comp(i, k)))
```
rasolca, line 62: why the i==k comparison? (already included by comp)

src/lapack/gpu/lacpy.cu
```diff
@@ -0,0 +1,136 @@
...
+      break;
+    case CUBLAS_FILL_MODE_UPPER:
+      if (i == j)
+        copyDiag<dlaf::util::isUpper>(m, n, a, lda, b, ldb);
+      else if (i < j)
+        copyAll(m, n, a, lda, b, ldb);
+      break;
+    case CUBLAS_FILL_MODE_FULL:
+      copyAll(m, n, a, lda, b, ldb);
+      break;
```
rasolca, line 100: This code branch is never used.

albestro: yep, it was there just in case we wanted to do an iteration of benchmarks and compare it with `cudaMemcpy2DAsync`.

> (note: the kernel implemented supports also the general case, even if it is not used).

If you prefer to remove it (sidenote: I tested it), I understand it. Just let me know 😉

rasolca: > (sidenote: I tested it)
Currently is tested, but as no test executes it, it might break without noticing...

albestro: See 0b32ac385beefe2aa5ec73bec7b8164561fc824a

src/lapack/gpu/lacpy.cu
```diff
@@ -97,7 +97,9 @@ __global__ void lacpy(cublasFillMode_t uplo, const unsigned m, const unsigned n,
         copyAll(m, n, a, lda, b, ldb);
       break;
     case CUBLAS_FILL_MODE_FULL:
-      copyAll(m, n, a, lda, b, ldb);
+      // Note: it seems more appropriate to use cudaMemcpy2DAsync in this case
```
rasolca, line 100:
```suggestion
      // Note: it is more appropriate to use cudaMemcpy2DAsync in this case
```

rasolca (approved):

albestro:
bors try

fixed after new master

PR #511, opened 2022-03-31
Change barrier in `cholesky` and `reduction_to_band` miniapps

msimberg:
Wait for all tiles first to avoid blocking the pika thread with `MPI_Barrier`. This is similar to how it's done in the other miniapps.

### DIFF
--- a/miniapp/miniapp_cholesky.cpp
+++ b/miniapp/miniapp_cholesky.cpp
@@ -142,15 +142,10 @@ struct choleskyMiniapp {
         dlaf::factorization::cholesky<backend, DefaultDevice_v<backend>, T>(comm_grid, opts.uplo,
                                                                             matrix.get());
 
-        // wait for last task and barrier for all ranks
-        {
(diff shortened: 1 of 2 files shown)


rasolca:
I just noticed another barrier problem in the miniapp.
During the matrix setup there is no need to wait for matrix_host to be ready, but it is important that the MatrixMirror is ready on all the ranks.
https://github.com/eth-cscs/DLA-Future/blob/3b050e1e916c88c13c8ae489d8aa51ebc4960099/miniapp/miniapp_cholesky.cpp#L130-L139

So line 131 should be removed and the barrier moved after line 139.

msimberg:
@rasolca I've moved the barrier and removed the wait for local CPU tiles. Does it look better to you?

rasolca (approved):

rasolca:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try-

PR #513, opened 2022-03-31
algo: making Band to Tridiagonal accept input from both CPU and GPU

albestro:
This work is part of #501.

This algorithm is still on MC backend, but it now accepts input from either CPU or GPU matrix.

The code for CPU and GPU, specifically `copyDiag` and `copyOffDiag`, has been duplicated because at the moment we don't have a final solution for lapack kernel wrappers (I will open an issue about that trying to describe the current status of the art).

### DIFF
--- a/include/dlaf/eigensolver/band_to_tridiag/mc.h
+++ b/include/dlaf/eigensolver/band_to_tridiag/mc.h
@@ -10,6 +10,8 @@
 
 #pragma once
 
+#include "dlaf/eigensolver/band_to_tridiag/api.h"
+
 #include <pika/future.hpp>
 #include <pika/unwrap.hpp>
 
@@ -17,8 +19,9 @@
 #include "dlaf/common/index2d.h"
 #include "dlaf/common/pipeline.h"
 #include "dlaf/common/vector.h"
-#include "dlaf/eigensolver/band_to_tridiag/api.h"
 #include "dlaf/executors.h"
+#include "dlaf/lapack/gpu/lacpy.h"
+#include "dlaf/lapack/gpu/laset.h"
 #include "dlaf/lapack/tile.h"
 #include "dlaf/matrix/matrix.h"
 #include "dlaf/memory/memory_view.h"
@@ -83,13 +86,11 @@ void applyHHRight(const SizeType m, const SizeType n, const T tau, const T* v, T
 }
 
 template <class T>
-class BandBlock {
-  using MatrixType = Matrix<T, Device::CPU>;
-  using ConstTileType = typename MatrixType::ConstTileType;
-
-public:
+struct BandBlock {
   BandBlock(SizeType n, SizeType band_size)
-      : size_(n), band_size_(band_size), ld_(2 * band_size_ - 1), mem_(n * (ld_ + 1)) {}
+      : size_(n), band_size_(band_size), ld_(2 * band_size_ - 1), mem_(n * (ld_ + 1)) {
+    // TODO add assert about memory accessible from gpu
+  }
 
   T* ptr(SizeType offset, SizeType j) noexcept {
     DLAF_ASSERT_HEAVY(0 <= offset && offset < ld_ + 1, offset, ld_);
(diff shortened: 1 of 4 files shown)


include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -101,55 +100,160 @@ class BandBlock {
...
+              gpulapack::lacpy(General, band_size_ + 1, index, source.ptr(), source.ld() + 1, ptr(0, j),
+                               ld() + 1, stream);
+            }
+            const auto size = std::min(band_size_, source.size().cols());
+            gpulapack::lacpy(Lower, size, size, source.ptr({index, index}), source.ld(),
+                             ptr(0, j + index), ld(), stream);
+          },
+          std::move(source));
+    }
+#endif
```
msimberg, line 172: Commenting only as a reminder to add a link to the issue about removing this duplication.

test/unit/eigensolver/test_band_to_tridiag.cpp
```diff
@@ -43,22 +44,24 @@ const std::vector<std::tuple<SizeType, SizeType, SizeType>> sizes = {
...
 
-  Matrix<T, Device::CPU> mat_a(size, block_size);
-  matrix::util::set_random_hermitian(mat_a);
+  Matrix<T, Device::CPU> mat_a_h(size, block_size);
+  matrix::util::set_random_hermitian(mat_a_h);
 
-  auto [mat_trid, mat_v] = eigensolver::bandToTridiag<Backend::MC>(uplo, band_size, mat_a);
+  MatrixMirror<T, D, Device::CPU> mat_a(mat_a_h);
+  auto [mat_trid, mat_v] = eigensolver::bandToTridiag<Backend::MC>(uplo, band_size, mat_a.get());
+  mat_a.copyTargetToSource();
```
albestro, line 58: Do you like this? IMHO it is more "correct" and readable, but I look forward to your comments.

```suggestion
auto [mat_trid, mat_v] = [&](){
  MatrixMirror<T, D, Device::CPU> mat_a(mat_a_h);
  return eigensolver::bandToTridiag<Backend::MC>(uplo, band_size, mat_a.get());
}();
```

include/dlaf/eigensolver/band_to_tridiag/mc.h
albestro: Yes, I'm going to document it today. Thanks for the reminder!

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -101,55 +100,160 @@ class BandBlock {
...
+              const auto size2 = band_size_ - size;
+              ::lapack::lacpy(General, source.size().rows(), size2, source.ptr({0, index + size}),
+                              source.ld(), dest + ld() * size, ld());
+            }
+          },
+          std::move(source));
+    }
+#ifdef DLAF_WITH_CUDA
+    else if constexpr (D == Device::GPU) {
+      DLAF_ASSERT_HEAVY(isAccessibleFromGPU(), "BandBlock memory should be accessible from GPU");
```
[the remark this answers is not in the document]
albestro: It's a good point. This assertion has been added to ensure that CPU memory allocated by BandBlock for the workspace matrix is accessible from the GPU, so we can work with lacpy GPU kernel to copy directly from GPU to CPU.

GPU can access CPU memory iff it is pinned memory.

Let's check how we allocate host memory in DLAF.

https://github.com/eth-cscs/DLA-Future/blob/a285bd501c852e0b275e77826bb95884a69f2d10/include/dlaf/memory/memory_chunk.h#L63-L65

https://github.com/eth-cscs/DLA-Future/blob/a285bd501c852e0b275e77826bb95884a69f2d10/src/memory/memory_chunk.cpp#L20-L24

Since BandBlock internal CPU memory is pinned, it is accessible from GPU.

As already said, the assertion is just for ensuring this fact, since it is not a given fact that CPU memory is accessible from GPU, and the place where this memory has been made "valid" is unrelated to this code.

For the assertion I used [cudaPointerGetAttributes](https://docs.nvidia.com/cuda/archive/11.0/cuda-runtime-api/group__CUDART__UNIFIED.html#group__CUDART__UNIFIED_1gd89830e17d399c064a2f3c3fa8bb4390). Whatever type of memory it returns that is not `cudaMemoryTypeUnregistered` should be ok for the purpose, since:
- `cudaMemoryTypeHost`: it is registered host memory, so it is "page-locked"/"pinned"
- `cudaMemoryTypeDevice`: it is already on gpu, so it is accessible trivially
- `cudaMemoryTypeManaged`: it is managed memory, i.e. unified memory accessible by all processors

About first point with registered memory, I'm not sure how to deal with that in case of multiple GPUs (see [cudaHostRegister](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ge8d5c17670f16ac4fc8fcb4181cb490c)). Probably in this case won't be enough.

Do you have any suggestion?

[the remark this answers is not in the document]
msimberg: Today I learned! I only knew pinned host memory was required for asynchronous transfers, but not that it could actually be accessed directly on the device. Are there any special synchronization requirements for this to work? Surely reads and writes are not visible immediately on the other side?

No suggestions on how to deal with multiple GPUs, but I'd say it's unlikely that we'll support that any other way than with multiple MPI ranks for quite some time. We could perhaps add an assertion somewhere in the initialization to avoid surprises, but I don't know if that's necessary.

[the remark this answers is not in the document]
albestro: > Today I learned! I only knew pinned host memory was required for asynchronous transfers, but not that it could actually be accessed directly on the device. Are there any special synchronization requirements for this to work? Surely reads and writes are not visible immediately on the other side?

Wait, I'm not entirely sure at this point. I remember of something like zero-copy/unified memory, but it may be a specific use case I had when I was working on a Jetson (which is quite "special" in terms of hardware architecture).

Let me check and I will write here.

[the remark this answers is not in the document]
albestro: From [CUDA doc](https://docs.nvidia.com/cuda/archive/11.0/cuda-runtime-api/group__CUDART__UNIFIED.html#group__CUDART__UNIFIED) about Unified Virtual Addressing (UVA)

> Automatic Mapping of Host Allocated Host Memory
>
> All host memory allocated through all devices using cudaMallocHost() and cudaHostAlloc() is always directly accessible from all devices that support unified addressing. This is the case regardless of whether or not the flags cudaHostAllocPortable and cudaHostAllocMapped are specified.
>
> The pointer value through which allocated host memory may be accessed in kernels on all devices that support unified addressing is the same as the pointer value through which that memory is accessed on the host. It is not necessary to call cudaHostGetDevicePointer() to get the device pointer for these allocations.

From [Unified Memory or Unified Virtual Addressing?](https://developer.nvidia.com/blog/unified-memory-in-cuda-6/)
> UVA enables “Zero-Copy” memory, which is pinned host memory accessible by device code directly, over PCI-Express, without a memcpy. Zero-Copy provides some of the convenience of Unified Memory, but none of the performance, because it is always accessed with PCI-Express’s low bandwidth and high latency.

We use UVA (its support is currently implicitly required in DLAF), so in the case of this PR zero-copy is used. Since we are accessing memory just one time and not multiple times, it may be a good solution, but we will do some benchmark in the near future to confirm this.

**TL;DR** CUDA Unified Virtual Addressing (UVA) enables pinned host memory to be accessed by the GPU (via zero-copy)

test/unit/eigensolver/test_band_to_tridiag.cpp
rasolca: What about (if it works) `MatrixMirror<const T, ...`

rasolca (approved):

test/unit/eigensolver/test_band_to_tridiag.cpp
msimberg: Is the `copyTargetToSource` important in the original? Currently `MatrixMirror<const T, ...>` applies the constness to both target and source, i.e. it will not copy anything in either direction. In theory this could be expanded to have different constness for source and target, and if source is non-const even if target is const, the target could be copied back to the source.

That is to say, `MatrixMirror<const T, ...> should work, but it might not do everything you expect it to do at the moment.

rasolca: I don't see a reason for `copyTargetToSource`... the argument of bandToTridiag is a matrix const T, therefore the matrix cannot be changed...

albestro: > What about (if it works) `MatrixMirror<const T, ...`

It's a nice idea!


> That is to say, `MatrixMirror<const T, ...> should work, but it might not do everything you expect it to do at the moment.

I was not aware of this feature! I'm not sure I fully understood your point @msimberg (specifically _"i.e. it will not copy anything in either direction"_, probably you meant just with function members). However looking at the code it seems that by having a `MatrixMirror<const T, ...>` we get a nice helper that duplicates, just if needed, a matrix on the target device.

This matrix provided by `MatrixMirror<const T, ...>` will be a `Matrix<const T, TargetDevice>`, and in this case it perfectly fits, because the algorithm requires a const matrix as @rasolca already said.

However, in next PRs I will open today, there is a slightly different use-case, where I need to copy a matrix to the device just if needed, so the helper unifying the code between CPU and GPU versions will be useful, but at the same time I need this as a `Matrix<T>`.

I don't know if it is too much overload for this class and it would be preferable to have a dedicated helper just for this. For the moment it is ok like this, but in the next hours we may have to take a decision also on this side.

albestro:
bors try

include/dlaf/eigensolver/band_to_tridiag/mc.h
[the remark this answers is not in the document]
teonnik: > Today I learned! I only knew pinned host memory was required for asynchronous transfers, but not that it could actually be accessed directly on the device. Are there any special synchronization requirements for this to work? Surely reads and writes are not visible immediately on the other side?

I think this is discussed in the Mapped Memory section: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#mapped-memory . The following is stated:

> A block of page-locked host memory can also be mapped into the address space of the device by passing flag cudaHostAllocMapped to `cudaHostAlloc()`

So I am not sure we can assume it's available on the GPU unless the flag is passed.

test/unit/eigensolver/test_band_to_tridiag.cpp
msimberg: > I was not aware of this feature! I'm not sure I fully understood your point @msimberg (specifically _"i.e. it will not copy anything in either direction"_, probably you meant just with function members).

Yep, sorry. It will of course copy it to the target on construction. But after that no copies will be made. The non-const version copies back to the source on destruction. But that could also be extended to be controlled by some flag if needed in other situations.

> However, in next PRs I will open today, there is a slightly different use-case, where I need to copy a matrix to the device just if needed, so the helper unifying the code between CPU and GPU versions will be useful, but at the same time I need this as a `Matrix<T>` (EDIT: and I don't need them to be copied back).

Yeah, so this would be a situation where you don't want the automatic copying back. I wasn't entirely sure about making it automatic when I added it, and it might even be better to make it explicit instead. Not sure... As for extending `MatrixMirror` or having another utility, I think the answer will depend on what exactly "if needed" means in your case.

include/dlaf/eigensolver/band_to_tridiag/mc.h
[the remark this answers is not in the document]
rasolca: > So I am not sure we can assume it's available on the GPU unless the flag is passed.

From the doc:
>__host__ ​cudaError_t cudaMallocHost ( void** ptr, size_t size )
> Allocates size bytes of host memory that is page-locked and accessible to the device.

This means that the flag is passed.

teonnik (approved):

albestro:
bors try-

test/unit/eigensolver/test_band_to_tridiag.cpp
albestro: With "iff needed" I mean that since the code is shared between CPU and GPU, we have

`MatrixMirror<T, D, CPU>`

that could either become

`MatrixMirror<T, CPU, CPU>` where you don't want to have a copy, it acts as an "alias"

or

`MatrixMirror<T, GPU, CPU>` where you want to allocate on the device and get a clone of the source

albestro: Yes, the use case you are looking at is the first one, i.e. "the alias" for `MatrixMirror<T, CPU, CPU>`.

But, when the generic code translates to `MatrixMirror<T, GPU, CPU>`, it ends up also copying back on destruction, which is not needed in my next use-case.

This PR is going to be closed soon, I will highlight this problem in my next PR and link to this comment here. 😉

PR #516, opened 2022-04-01
SEP eigensolver GPU

### LINKED ISSUE #501: GPU SEP pipeline

albestro:
Close #501

### DIFF
--- a/test/unit/eigensolver/test_eigensolver.cpp
+++ b/test/unit/eigensolver/test_eigensolver.cpp
@@ -70,7 +70,6 @@ void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb)
   auto ret = [&]() {
     MatrixMirror<T, D, Device::CPU> mat_a(mat_a_h);
     return eigensolver::eigensolver<B>(uplo, mat_a.get());
-    // TODO there is a superflous copy back
   }();
 
   if (mat_a_h.size().isEmpty())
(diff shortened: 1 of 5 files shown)


test/unit/eigensolver/test_eigensolver.cpp
```diff
@@ -55,16 +64,23 @@ void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb)
...
-  copy(reference, mat_a);
+  Matrix<T, Device::CPU> mat_a_h(reference.distribution());
+  copy(reference, mat_a_h);
 
-  auto ret = eigensolver::eigensolver<B>(uplo, mat_a);
+  auto ret = [&]() {
+    MatrixMirror<T, D, Device::CPU> mat_a(mat_a_h);
+    return eigensolver::eigensolver<B>(uplo, mat_a.get());
+    // TODO there is a superflous copy back
+  }();
```
albestro, line 74: @msimberg follow up https://github.com/eth-cscs/DLA-Future/pull/513#discussion_r840617468

msimberg:
test cscs

test/unit/eigensolver/test_eigensolver.cpp
rasolca: IMO is not superfluous in this case. As the matrix is changed you want to copy back the new values, such that the effect in both cases.
There is no test here, but it is better to keep it consistent.

rasolca (approved):

PR #531, opened 2022-04-08
miniapp: reduction to band - adaptation to GPU

albestro:
Adapt the existing miniapp to run on GPU backend (just local).

Moreover, introduce also `band-size` option for controlling the size of the band.

### DIFF
--- a/miniapp/miniapp_reduction_to_band.cpp
+++ b/miniapp/miniapp_reduction_to_band.cpp
@@ -44,6 +44,8 @@ struct Options
   Options(const pika::program_options::variables_map& vm)
       : MiniappOptions(vm), m(vm["matrix-size"].as<SizeType>()), mb(vm["block-size"].as<SizeType>()),
         b(vm["band-size"].as<SizeType>()) {
+    const bool isDistributed = (grid_rows * grid_cols) > 1;
+
     DLAF_ASSERT(m > 0, m);
     DLAF_ASSERT(mb > 0, mb);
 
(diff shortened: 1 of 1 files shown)


miniapp/miniapp_reduction_to_band.cpp
```diff
@@ -153,6 +183,7 @@ int main(int argc, char** argv) {
   desc_commandline.add_options()
     ("matrix-size",  value<SizeType>()   ->default_value(4),      "Matrix rows")
     ("block-size",   value<SizeType>()   ->default_value(2),      "Block cyclic distribution size")
+    ("band-size",    value<SizeType>()   ->default_value(-1),     "Band size")
```
rasolca, line 186: I would add in the help message that negatives value means block-size?

miniapp/miniapp_reduction_to_band.cpp
```diff
@@ -153,6 +183,7 @@ int main(int argc, char** argv) {
   desc_commandline.add_options()
     ("matrix-size",  value<SizeType>()   ->default_value(4),      "Matrix rows")
     ("block-size",   value<SizeType>()   ->default_value(2),      "Block cyclic distribution size")
```
rasolca, line 185: Please use some more realistic default params...

miniapp/miniapp_reduction_to_band.cpp
```diff
@@ -38,17 +39,26 @@ struct Options
...
+      : MiniappOptions(vm), m(vm["matrix-size"].as<SizeType>()), mb(vm["block-size"].as<SizeType>()),
+        b(vm["band-size"].as<SizeType>()) {
     DLAF_ASSERT(m > 0, m);
     DLAF_ASSERT(mb > 0, mb);
 
-    DLAF_ASSERT(do_check == dlaf::miniapp::CheckIterFreq::None,
-                "Error! At the moment result checking is not implemented. Please rerun with --check-result=none.");
+    if (b < 0)
+      b = mb;
+    DLAF_ASSERT(b > 0 && b <= mb, b, mb);
```
rasolca, line 52: Isn't the constrain `mb % b == 0` missing?
```suggestion
    DLAF_ASSERT(b > 0 && mb % b == 0, b, mb);
```

albestro: Yeah, I was not sure about adding it or not.

My concern was that it is a limitation of the algorithm and I didn't want to relay it here, but it is also true that in miniapps we tend to constrain runs to what is possible to run currently.

I'm going to add it. 😄

miniapp/miniapp_reduction_to_band.cpp
albestro: I would go for our classic `matrix-size=4096` and `block-size=256`.

About `band-size`? Do you like having it defaulted to `block-size`? or would you prefer to have a smaller number just to have a more "reasonable" use-case?

rasolca: we can leave it as nb for now and change later when we have a good choice for the eigensolver.
Anyway, does the distributed case handle nb != b, otherwise I would add a fallback in the options check.

teonnik (commented):
LGTM, apart from some comments @rasolca already made.

rasolca (approved):

rasolca:
bors try

PR #534, opened 2022-04-15
Fix: group taus from band chunks to tile chunks for reduction to band

albestro:
This fixes a misalignment between reduction to band and its backtransformation.

Reduction to band is a blocked algorithm working with blocks of size `band_size`, while its backtransformation works with blocks equals to the full tile size.

These algorithms interact via Householder reflectors and related taus: former ones are stored in the matrix, while latter ones are stored in a vector as they were produced, so in chunks of `band_size` taus. However `backtransformation` expects them to be in chunks of `tile_size`, so there was a discrepancy.

For this reason, this PR proposes to trivially merge taus produced by reduction to band as if they were produced in `tile_size` chunks, according to how its backtransformation expects them.

### DIFF
--- a/include/dlaf/eigensolver/reduction_to_band.h
+++ b/include/dlaf/eigensolver/reduction_to_band.h
@@ -12,12 +12,55 @@
 #include "dlaf/common/vector.h"
 #include "dlaf/communication/communicator_grid.h"
 #include "dlaf/matrix/matrix.h"
+#include "dlaf/sender/when_all_lift.h"
 #include "dlaf/util_matrix.h"
 
 #include "dlaf/eigensolver/reduction_to_band/impl.h"
 
-namespace dlaf {
-namespace eigensolver {
+namespace dlaf::eigensolver {
+
+namespace internal {
+
+template <class T>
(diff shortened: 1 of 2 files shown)


albestro:
Just for additional reference, in case someone of you is willing to propose a better solution.

Reduction to band provides taus in a data structure which looks like `vector<shared_future<vector>>>`, where the `shared_future<vector>` represents the chunks of taus, and the wrapping vector is the one containing the different chunks produced.

Till now there is no size mentioned, since the data-structure is exactly the same no matter the size of the chunks.

Reduction to band at each step, working with a panel `band_size` wide, produces a `shared_future<vector>[band_size]`, which gets pushed to the external vector with all other chunks. The problem is that chunks of taus are future values, not actual values, so whatever I'd like to do to merge them, I would have to "extract" them from the future, before managing their values and storing them in "another format".

This is the reason why I went for having some kind of post-processing instead of changing how the algorithm stores in the first place them. If you have any better idea, feel free to suggest it.

msimberg (approved):

teonnik (approved):

PR #436, opened 2021-12-06
Tridiagonal eigensolver (local)

teonnik:
# TODO

- [x] z-vector formation: matrix-vecrtor multiplication of orthogonal matrix (eigenvectors) of the sub-tridiagonal system with Cuppen's vector (i.e. the rank-1 update vector in Cuppen's decomposition)
- [x] Use the recently provided laed4 API from lapackpp for the solution of the secular equation of the rank-1 updated diagonal system (instead of trying to implement your own...)
- [x] Implement deflation for the rank-1 updated diagonal system: DLA-F APIs for Given's rotation
- [x] Matrix-multipy (GEMM) the eigenvectors of the sub-tridiagonal systems with the eigenvectors of the rank-1 updated diagonal system
- [x] Improve and expand unit-testing of various components

Currently the algorithm is not in a working state. Most components are WIP. I will continue pushing changes in this PR as remaining parts are getting ready.

PR #541, opened 2022-05-02
Introduce gemm (local)

albestro:
This PR has its root in a very small part of the work going on in #436, where a constrained use-case of a `gemm` is used.

The specific use-case is a gemm between sub-matrices, which have the characteristic of being placed in the same exact spot in all matrices, so both inputs and output.
For this reason, given a linear range `[s, e]`, the operations is `C[s:e, s:e] = beta * C[s:e, s:e] + alpha * A[s:e, s:e] * B[s:e, s:e]`.
In particular, elements outside these 2D ranges are not referenced/modified in any of the matrices involved.

It is a specific use-case but, as soon as we will change Matrix API to manage sub-matrix not as a special case, it will become a normal `gemm`. In the meanwhile, I think we can go a bit more "easier" on its API.

Looking forward to suggestions/feedbacks.

### DIFF
--- a/include/dlaf/multiplication/general/impl.h
+++ b/include/dlaf/multiplication/general/impl.h
@@ -10,6 +10,39 @@
 
 #pragma once
 
+#include "dlaf/multiplication/general/api.h"
+
+#include "dlaf/blas/enum_output.h"
+#include "dlaf/blas/tile.h"
+#include "dlaf/common/index2d.h"
+#include "dlaf/sender/transform.h"
+#include "dlaf/sender/when_all_lift.h"
+
 namespace dlaf::multiplication {
-namespace internal {}
+namespace internal {
+
+template <Device D, class T>
+void GeneralSub<D, T>::call(const SizeType a, const SizeType b, const blas::Op opA, const blas::Op opB,
+                            const T alpha, Matrix<const T, D>& mat_a, Matrix<const T, D>& mat_b,
+                            const T beta, Matrix<T, D>& mat_c) {
+  namespace ex = pika::execution::experimental;
+
+  if (opA != blas::Op::NoTrans)
+    DLAF_UNIMPLEMENTED(opA);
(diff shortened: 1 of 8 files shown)


test/include/dlaf_test/matrix/util_generic_blas.h
```diff
@@ -15,13 +15,64 @@
...
+    if (kj.row() < a || kj.row() > b || kj.col() < a || kj.col() > b)
+      return TypeUtilities<T>::polar(13, -26);
+
+    const double k = kj.row();
+    const double j = kj.col();
+
+    return TypeUtilities<T>::polar((k + .5) / (j + 2), k + j);
+  };
+
+  auto elC = [a, b](const GlobalElementIndex) { return TypeUtilities<T>::polar(0, 0); };
```
teonnik, line 62: `a` and `b` are not used in the lambda, warnings are generated.

teonnik (changes requested):
LGTM, some minor changes needed!

include/dlaf/multiplication/general/impl.h
```diff
@@ -0,0 +1,48 @@
...
+#include "dlaf/blas/tile.h"
+#include "dlaf/common/index2d.h"
+#include "dlaf/sender/transform.h"
+#include "dlaf/sender/when_all_lift.h"
+
+namespace dlaf::multiplication {
+namespace internal {
+
+template <Device D, class T>
+void GeneralSub<D, T>::call(const SizeType a, const SizeType b, const blas::Op opA, const blas::Op opB,
```
rasolca, line 25: Missing `NN` in the name of the function...

albestro:
bors try

teonnik (approved):

msimberg (approved):
No comments.

rasolca (approved):

PR #546, opened 2022-05-02
Change shape of tridiagonalized matrix

albestro:
#436 requires the tridiagonal matrix to be a 2 columns matrix instead of a 2 rows matrix like it is currently.

(credits @rasolca)

PR #547, opened 2022-05-02
Introduce MC tridiagonal solver in SEP

albestro:
This PR is about changes currently required to introduce #436 in the eigensolver pipeline, replacing the placeholder.

It requires:
- [x] #436 - (actually it is a fork of it)
- [x] #546 
- [x] #632 

TODO:
- [x] Complex values are not yet supported (disabled tests, miniapps and ETIs)
- [x] Empty matrices tests are disabled

### DIFF
--- a/include/dlaf/eigensolver/eigensolver/api.h
+++ b/include/dlaf/eigensolver/eigensolver/api.h
@@ -19,7 +19,7 @@ namespace dlaf::eigensolver {
 
 template <class T, Device D>
 struct EigensolverResult {
-  common::internal::vector<BaseType<T>> eigenvalues;
+  Matrix<BaseType<T>, D> eigenvalues;
   Matrix<T, D> eigenvectors;
 };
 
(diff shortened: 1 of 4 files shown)


albestro:
About empty matrices tests

https://github.com/eth-cscs/DLA-Future/blob/fa5ef85793db4de75329fa58c72a18a9fc04d8a6/test/unit/eigensolver/test_eigensolver.cpp#L52

https://github.com/eth-cscs/DLA-Future/blob/fa5ef85793db4de75329fa58c72a18a9fc04d8a6/test/unit/eigensolver/test_gen_eigensolver.cpp#L50

The problem is that here

https://github.com/eth-cscs/DLA-Future/blob/fa5ef85793db4de75329fa58c72a18a9fc04d8a6/include/dlaf/eigensolver/tridiag_solver/mc.h#L92-L98

`i_end` becomes negative when there are no tiles, i.e. with empty matrices, which is not valid for `reserve` (with DLAF_ASSERTs enabled the `to_sizet` correctly catches it).

It should be an easy fix, but probably it is even easier for you @teonnik to add the quick return in the right point of the code.

I can open an issue for that, since probably it would be better to add a test case for it in tridiagonal solver.

albestro:
bors try

rasolca (approved):

teonnik (approved):

PR #548, opened 2022-05-04
Fix compile time bug when mixing CHECK_TILE_NEAR and CHECK_MATRIX_NEAR

teonnik:
The following results in a compile-time error:

```
auto tile = ...
CHECK_TILE_NEAR(tile, tile, error, error);
```

but the following doesn't:

```
const auto& tile = ...
CHECK_TILE_NEAR(tile, tile, error, error);
```

when CHECK_MATRIX_NEAR is also used in the translation unit.

### DIFF
--- a/test/include/dlaf_test/matrix/util_tile.h
+++ b/test/include/dlaf_test/matrix/util_tile.h
@@ -146,8 +146,8 @@ namespace internal {
 /// @pre The second argument of err_message should be either T, T& or const T&.
 template <class T, class ElementGetter, class ComparisonOp, class ErrorMessageGetter,
           std::enable_if_t<!std::is_convertible_v<ElementGetter, T>, int> = 0>
-void check(ElementGetter&& expected, const Tile<const T, Device::CPU>& tile, ComparisonOp comp,
-           ErrorMessageGetter err_message, const char* file, const int line) {
+void checkTile(ElementGetter&& expected, const Tile<const T, Device::CPU>& tile, ComparisonOp comp,
+               ErrorMessageGetter err_message, const char* file, const int line) {
   for (SizeType j = 0; j < tile.size().cols(); ++j) {
     for (SizeType i = 0; i < tile.size().rows(); ++i) {
       TileElementIndex index(i, j);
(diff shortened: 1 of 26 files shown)


test/include/dlaf_test/matrix/util_tile.h
```diff
@@ -265,8 +265,10 @@ void checkNear(ElementGetter&& expected, const Tile<const T, D>& tile, BaseType<
   };
   internal::check(expected, tile, comp, err_message, file, line);
 }
+DLAF_MAKE_CALLABLE_OBJECT(checkNear);
+
 #define CHECK_TILE_NEAR(expected, tile, rel_err, abs_err) \
-  ::dlaf::matrix::test::checkNear(expected, tile, rel_err, abs_err, __FILE__, __LINE__)
+  ::dlaf::matrix::test::checkNear_o(expected, tile, rel_err, abs_err, __FILE__, __LINE__)
```
msimberg, line 271: Do you have an example of the compilation failure? This is not really what `DLAF_MAKE_CALLABLE_OBJECT` should be used (more importantly, needed) for. I don't understand why this would fix anything :/

teonnik: Yes, this was commented out a few commits back because of a compile-time error : https://github.com/eth-cscs/DLA-Future/pull/436/files#diff-b5a15fd19989a0843c5344662f5ce7e61e8f2110d28e7ff6ab785aca63bcc44dR40-R65

I used it because the issue appeared to be related to a compile-time failure in the templated and overloaded `checkNear()` function.

msimberg: Do you have an error message? Why does this fix it? `DLAF_MAKE_CALLABLE_OBJECT` _should_ create a transparent wrapper that ends up calling `checkNear` with the same arguments and _should_  have the same compilation failure. If it doesn't, there may be a bug in the wrapper as well...

teonnik: Yes, it's easy to reproduce, just insert the following:

```
TYPED_TEST(MatrixLocalTest, Dummy) {    
  constexpr auto error = TypeUtilities<TypeParam>::error;    
  auto tile = createTile<TypeParam>([](TileElementIndex idx) { return idx.row() + idx.col(); },    
                                    TileElementSize(5, 5), 5);    
  CHECK_TILE_NEAR(tile, tile, error, error);    
}  
```

at the bottom of `test_matrix_local.cpp` and remove the `_o` from checkNear in the macro.

Error message: https://pastebin.com/ebjezAnR

teonnik:
bors try

teonnik:
@msimberg I was wondering the same but in the interest of making the least amount of changes, I left them as they are. However, if anyone would like to have the names modified, there is no issue to do it here.

I recently noticed that the generic form of the `checkNear()` overloads for Matrix is to support the types `Matrix<T, D>` and `MatrixLocal<T>` (defined under dlaf_test).

teonnik:
bors try

rasolca (approved):

PR #554, opened 2022-05-10
CI: going to ubuntu22.04

rasolca:
Move mc tests to ubuntu22.04.

Note: nvidia images are not available yet with ubuntu22.04, therefore gpu tests were not touched.

PR #555, opened 2022-05-12
Fix dangling reference in `test_panel`

msimberg:
A few todos in the comments that also explain the problem (briefly...).

### DIFF
--- a/test/unit/matrix/test_panel.cpp
+++ b/test/unit/matrix/test_panel.cpp
@@ -91,7 +91,8 @@ TYPED_TEST(PanelTest, AssignToConstRef) {
       EXPECT_EQ(exp_indices, ref_indices);
 
       for (const auto& idx : exp_indices) {
-        const auto& exp_tile = panel.read(idx).get();
+        auto exp_tile_f = panel.read(idx);
+        const auto& exp_tile = exp_tile_f.get();
         auto get_element_ptr = [&exp_tile](const TileElementIndex& index) {
           return exp_tile.ptr(index);
         };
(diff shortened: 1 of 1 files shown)


test/unit/matrix/test_panel.cpp
```diff
@@ -378,17 +379,31 @@ void testShrink(const config_t& cfg, const comm::CommunicatorGrid& comm_grid) {
...
 
     counter = 0;
     for (const auto& idx : panel.iteratorLocal()) {
+      auto panel_tile_f = panel.read(idx);
+      const auto& panel_tile = panel_tile_f.get();
+      auto matrix_tile_f = matrix.read(idx);
+      const auto& matrix_tile = matrix_tile_f.get();
+
+      // TODO: Was it intentional to do two panel.reads, on for CHECK_TILE_EQ
+      // and one for EXPECT_EQ?
```
rasolca, line 403: I don't see a reason to be intentional. In both it behaves the same.

test/unit/matrix/test_panel.cpp
```diff
@@ -378,17 +379,31 @@ void testShrink(const config_t& cfg, const comm::CommunicatorGrid& comm_grid) {
...
-      const auto& tile = panel.read(idx).get();
+
+      auto panel_tile_f = panel.read(idx);
+      const auto& panel_tile = panel_tile_f.get();
+      auto matrix_tile_f = matrix.read(idx);
+      const auto& matrix_tile = matrix_tile_f.get();
+
+      // TODO: Comment about why the gets are up there above the SCOPED_TRACE?
+      // This was again a case of gtest expecting SCOPED_TRACE to be destroyed
+      // on the same thread it was constructed.
```
rasolca, line 390: Yes, a comment is welcomed.

rasolca (approved):

test/unit/matrix/test_panel.cpp
msimberg: I've removed the comment.

test/unit/matrix/test_panel.cpp
msimberg: Please have a look, I've added a comment both about the dangling reference and `SCOPED_TRACE`. I didn't reproduce the comment in the for loop just below which does the same but can add a note there as well if you'd like.

rasolca: Good! Isn't it worth to mention gtest in the second part?
I think is better to add at least a small note in the for loop as well.

msimberg: I think both the dangling reference and the gtest comment apply in the second part as well. What about I add just a short comment to see the comment above since it applies there as well?
```
// See comment in for loop above. The same problems about dangling references and SCOPED_TRACE apply here as well.
```
?

msimberg: Sorry for the delay, updated! I think this can be merged for testing on #554 now.

rasolca:
bors try

PR #560, opened 2022-05-24
algo: BackTransform from Tridiagonal - attempt to reduce micro-tasking overhead

albestro:
This is an attempt at reducing the problem related to micro-tasking, where there is a lot of overhead due to small tasks scheduled.

This algorithm was structured to re-use memory for multiple computations, which constrained algorithmic steps order together, together with a minor reduction of parallelisation chances (W2 and W could not be computed in parallel).

At the cost of using slightly more memory, having two independent memory areas for V and W, this PR proposes to:
- [x] group V, T and W computations tasks*
- [x] group W2 and E computations

*note: on GPU there will be two tasks, one for the computation on CPU of V and T, and one on GPU for copying (V,T) to the device and computing W

TODO
- [x] ~cleanup and refactoring (e.g. `split_future` can be avoided in favour of passing tuples)~
- [x] verify build problem 6c55a5acb1c13ead4552b9cc1a7ba0eab9513553 (it seems related to `friend`-ness of a lambda, but it's just a speculation at the moment)
- [x] verify difference between passing by-ref or by-value between CPU and GPU

### DIFF
--- a/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
+++ b/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
@@ -495,11 +495,12 @@ void BackTransformationT2B<B, D, T>::call(const SizeType band_size, Matrix<T, D>
 
       auto [tile_v, tile_w] = helperBackend.computeVW(ij, nrefls, helper, mat_hh, mat_v, mat_t, mat_w);
 
-      // W2 = V* . E
       for (SizeType j_e = 0; j_e < mat_e.nrTiles().cols(); ++j_e) {
+        const LocalTileIndex idx_e(ij.row(), j_e);
         const auto sz_e = mat_e.tileSize({ij.row(), j_e});
-        auto tile_e = mat_e.read(LocalTileIndex(ij.row(), j_e));
+        auto tile_e = mat_e.read(idx_e);
 
+        // W2 = V* . E
         computeW2<B>(thread_priority::normal,
                      ex::keep_future(splitTile(tile_v, helper.topPart().specHH())),
                      ex::keep_future(splitTile(tile_e, helper.topPart().specEV(sz_e.cols()))), T(0),
@@ -512,15 +513,10 @@ void BackTransformationT2B<B, D, T>::call(const SizeType band_size, Matrix<T, D>
                        ex::keep_future(splitTile(tile_e, helper.bottomPart().specEV(sz_e.cols()))), T(1),
                        mat_w2.readwrite_sender(LocalTileIndex(0, j_e)));
         }
-      }
-
-      // E -= W . W2
-      for (SizeType j_e = 0; j_e < mat_e.nrTiles().cols(); ++j_e) {
-        const LocalTileIndex idx_e(ij.row(), j_e);
-        const auto sz_e = mat_e.tileSize({ij.row(), j_e});
 
         const auto& tile_w2 = mat_w2.read_sender(idx_e);
 
+        // E -= W . W2
         updateE<B>(pika::threads::thread_priority::normal,
                    ex::keep_future(splitTile(tile_w, helper.topPart().specHH())), tile_w2,
                    splitTile(mat_e(idx_e), helper.topPart().specEV(sz_e.cols())));
--- a/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
+++ b/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
@@ -113,13 +113,15 @@ pika::shared_future<matrix::Tile<const T, Device::CPU>> computeTFactor(
          ex::make_future();
 }
 
-template <Backend backend, class TSender, class VSender>
-auto computeW(pika::threads::thread_priority priority, TSender&& tile_t, VSender&& tile_v) {
+template <Backend backend, class TSender, class VSender, class WSender>
+auto computeW(pika::threads::thread_priority priority, TSender&& tile_t, VSender&& tile_v,
+              WSender&& tile_w) {
   using namespace blas;
-  using T = dlaf::internal::SenderElementType<VSender>;
+  using T = dlaf::internal::SenderElementType<WSender>;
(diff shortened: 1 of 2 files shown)


include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
```diff
@@ -38,111 +43,158 @@ namespace dlaf::eigensolver::internal {
...
 
-    return matrix::Tile<const T, Device::CPU>(std::move(tile_t));
-  };
-  return dlaf::internal::whenAllLift(ex::keep_future(std::move(tile_taus)),
-                                     ex::keep_future(std::move(tile_v)), std::move(mat_t)) |
-         dlaf::internal::transform(dlaf::internal::Policy<Backend::MC>(), tfactor_task) |
-         ex::make_future();
-}
+template <class T>
+struct single_tile_f<Backend::MC, T> {
```
albestro, line 131: note to self: find a better name

include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
```diff
@@ -38,111 +43,158 @@ namespace dlaf::eigensolver::internal {
...
-             ESender&& tile_e) {
-  using blas::Op;
-  using T = dlaf::internal::SenderElementType<ESender>;
-  dlaf::internal::whenAllLift(Op::NoTrans, Op::NoTrans, T(-1), std::forward<WSender>(tile_w),
-                              std::forward<W2Sender>(tile_w2), T(1), std::forward<ESender>(tile_e)) |
-      tile::gemm(dlaf::internal::Policy<backend>(priority)) |
-      pika::execution::experimental::start_detached();
-}
+template <class T>
+struct double_tile_f<Backend::MC, T> {
```
albestro, line 162: note-to-self: find a better name

include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
```diff
@@ -38,111 +43,158 @@ namespace dlaf::eigensolver::internal {
...
+template <class T>
+struct single_tile_f<Backend::GPU, T> {
+  void operator()(cublasHandle_t handle, const matrix::Tile<const T, Device::GPU>& tile_v,
+                  const matrix::Tile<const T, Device::GPU>& tile_w,
+                  matrix::Tile<T, Device::GPU>& tile_w2, matrix::Tile<T, Device::GPU>& tile_e) {
+    using namespace blas;
+    using tile::internal::gemm;
+    // W2 = V* . E
+    gemm(handle, Op::ConjTrans, Op::NoTrans, T(1), tile_v, tile_e, T(0), tile_w2);
+    // E -= W . V2
```
msimberg, line 153:
```suggestion
    // E -= W . W2
```
? If yes, I think there's another comment with `V2` as well.

More importantly (and not for this PR), would it be useful to be able to express this as
```
gemm(Policy<B>(), ...) |
  gemm(Policy<B>(), ...);
```
, assuming it can be done equivalently (i.e. with the two gemms submitted immediately on the same stream)? The details are obviously in the `...`...

albestro:
small update about these todos

> * verify build problem [6c55a5a](https://github.com/eth-cscs/DLA-Future/commit/6c55a5acb1c13ead4552b9cc1a7ba0eab9513553) (it seems related to `friend`-ness of a lambda, but it's just a speculation at the moment)
> * verify difference between passing by-ref or by-value between CPU and GPU

The build problem was due to the fact that in a wip version of this PR I was dealing with a "non-standard" `pika::future<matrix::Tile<const T, D>>` (future must be rw, while this is "hybrid"). The problem is that `splitTileInsertFutureInChain(Tile<T, D>&)` is meant to work with `Tile<T, D>` and not `Tile<const T, D>`, and for this reason is defined as `friend` for the non-const one, but not for the const one. This triggered the access to protected member.

About the second point, the difference between passing by-ref or by-value is something already known from the past: async GPU task has to manage the lifetime of the tile, so it just passes a reference to the actual function. In this case I opted for having both MC and GPU helpers to use `const&`, just to have them more similar.

albestro:
> [ ] cleanup and refactoring (e.g. `split_future` can be avoided in favour of passing tuples)

We had a chat with @msimberg about this. Currently, it seems we can't get rid off futures because this algorithm needs `splitTile`, which is still `future`-based. Indeed, we could migrate to sender for the rest, e.g. by having a sender of `tuple<tile, tile>` and not using `make_future`, but then we would end up needing `splitTile` to work with sender of tiles.

It looks reasonable to temporarily keep this as it is, delaying refactoring to when the tile-mechanism will be completely senderified.

albestro:
bors try

albestro:
bors try

albestro:
@rasolca if CI gives green, please check change in last commit https://github.com/eth-cscs/DLA-Future/pull/560/commits/5fe009c436ec2fcda787f4efa96b0d4e559edcbb, that should fix a leftover from #561

include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
msimberg: Sorry for leaving this open. This comment does not block the PR.

albestro:
bors try

msimberg (approved):

rasolca:
> @rasolca if CI gives green, please check change in last commit [5fe009c](https://github.com/eth-cscs/DLA-Future/commit/5fe009c436ec2fcda787f4efa96b0d4e559edcbb), that should fix a leftover from #561

Not sure what you mean...

albestro:
> > @rasolca if CI gives green, please check change in last commit [5fe009c](https://github.com/eth-cscs/DLA-Future/commit/5fe009c436ec2fcda787f4efa96b0d4e559edcbb), that should fix a leftover from #561
> 
> Not sure what you mean...

Nevermind, it was an old comment before you fixed the same problem in #570 😉

PR #561, opened 2022-06-02
miniapp: Backtransform Band To Tridiag

### LINKED ISSUE #524: Miniapp backtransformation tridiagonalization

albestro:
Close #524

### DIFF
--- a/miniapp/miniapp_bt_band_to_tridiag.cpp
+++ b/miniapp/miniapp_bt_band_to_tridiag.cpp
@@ -1,0 +1,198 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2022, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#include <iostream>
+#include <limits>
+#include <string>
+
+#include <pika/init.hpp>
+#include <pika/program_options.hpp>
+#include <pika/runtime.hpp>
+
+#include "dlaf/common/format_short.h"
+#include "dlaf/common/index2d.h"
+#include "dlaf/common/range2d.h"
+#include "dlaf/common/timer.h"
+#include "dlaf/communication/communicator_grid.h"
+#include "dlaf/communication/init.h"
+#include "dlaf/eigensolver/bt_band_to_tridiag.h"
+#include "dlaf/init.h"
+#include "dlaf/matrix/copy.h"
+#include "dlaf/matrix/matrix.h"
+#include "dlaf/matrix/matrix_mirror.h"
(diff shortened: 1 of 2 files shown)


miniapp/miniapp_bt_band_to_tridiag.cpp
```diff
@@ -0,0 +1,198 @@
...
+        DLAF_MPI_CHECK_ERROR(MPI_Barrier(world));
+
+        elapsed_time = timeit.elapsed();
+      }
+
+      double gigaflops;
+      {
+        double n = mat_e_host.size().rows();
+        double b = opts.b;
+        auto add_mul = 3 * n * n * b;
```
albestro, line 139: note-to-self: fix this

miniapp/miniapp_bt_band_to_tridiag.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+int pika_main(pika::program_options::variables_map& vm) {
+  {
+    dlaf::ScopedInitializer init(vm);
+    const Options opts(vm);
+
+    dlaf::miniapp::dispatchMiniapp<BacktransformBandToTridiagMiniapp>(opts);
+  }
+
+  return pika::finalize();
```
msimberg, line 169: Not a must for this PR as this can be updated elsewhere as well, but:
```suggestion
  pika::scoped_finalize fin;
  dlaf::ScopedInitializer init(vm);
  const Options opts(vm);

  dlaf::miniapp::dispatchMiniapp<BacktransformBandToTridiagMiniapp>(opts);

  return EXIT_SUCCESS;
```

rasolca: Is this a new pika feature?

rasolca (approved):

miniapp/miniapp_bt_band_to_tridiag.cpp
msimberg: > Is this a new pika feature?

Yes, and I realize this was never mentioned in the release notes so no wonder you wouldn't know about it... It was added in 0.2.0.

albestro: Small proposal: since this applies to almost all miniapps, I would open a separate PR with all required changes.

Moreover, in this future PR, I would propose also to have something like this

https://github.com/eth-cscs/DLA-Future/blob/89ca01fe3f7c365b34ce1d6d2adb78d79fde7c0e/miniapp/miniapp_bt_band_to_tridiag.cpp#L178-L179

where the name of the mini app is not hard-coded. It may be an overkill, in case I would just check that all names of miniapps are correct (I've seen at least one wrong name).

msimberg: Proposal accepted :wink:

PR #562, opened 2022-06-02
Use correct assert function in GPU potrf

msimberg:
While I remember...


msimberg:
bors try

albestro:
Can we take the chance in this PR to do also the other minor fix in the doc?

Specifically, it should not refer to `potrf` since it is a generic lapack call.

https://github.com/eth-cscs/DLA-Future/blob/4d52be37c347523967d25c8e3f8e1c379815ce61/test/include/dlaf_test/lapack/invoke.h#L59-L63

rasolca (approved):

PR #564, opened 2022-06-07
Add local matrix permutations on CPU and GPU

teonnik:
This PR implements permutations on both CPUs and GPUs for local matrices. The kernel is used in the tridiagonal solver.

### DIFF
--- a/test/unit/permutations/test_permutations.cpp
+++ b/test/unit/permutations/test_permutations.cpp
@@ -52,7 +52,7 @@ void testPermutations(SizeType n, SizeType nb, SizeType i_begin, SizeType i_end)
     return index_finish - 1 - i.row();
   });
   dlaf::matrix::util::set(mat_in_h, [](GlobalElementIndex i) {
-    return T(i.get<C>() - i.get<orthogonal(C)>()) / T(8);
+    return T(i.get<C>()) - T(i.get<orthogonal(C)>()) / T(8);
   });
   dlaf::matrix::util::set0<Backend::MC>(pika::threads::thread_priority::normal, mat_out_h);
 
@@ -70,7 +70,7 @@ void testPermutations(SizeType n, SizeType nb, SizeType i_begin, SizeType i_end)
       GlobalElementIndex i_in(i.get<orthogonal(C)>(), index_finish + index_start - 1 - i.get<C>());
       if constexpr (C == Coord::Row)
         i_in.transpose();
-      return T(i_in.get<C>() - i_in.get<orthogonal(C)>()) / T(8);
+      return T(i_in.get<C>()) - T(i_in.get<orthogonal(C)>()) / T(8);
     }
     return T(0);
   };


include/dlaf/permutations/general/impl.h
```diff
@@ -0,0 +1,201 @@
...
+//             sz.cols()
+//
+//
+// [1]: The portion of each input column or row is defined by the interval [in_offset, in_offset +
+//      sz.col()) or the interval [in_offset, in_offset + sz.row()) respectively.
+// [2]: The input submatrix is defined by `begin_tiles`, `ld_tiles`, `distr` and `in_tiles`
+// [3]: The subregion is defined by `begin` and `sz`
+// [4]: The output submatrix is defined by `begin_tiles`, `ld_tiles`, `distr` and `out_tiles`
+//
+// Note: `in_tiles` should be `const T` but to avoid extra allocations necessary for unwrapping
```
msimberg, line 74: Note, I don't think this would lead to additional allocations, but definitely more clunky usage.

include/dlaf/permutations/general/impl.h
```diff
@@ -0,0 +1,201 @@
...
+//       `std::vector<shared_future<matrix::Tile<const T, D>>>` it is left as non-const
+template <class T, Coord coord>
+void applyPermutations(GlobalElementIndex out_begin, GlobalElementSize sz, SizeType in_offset,
+                       const matrix::Distribution& distr, const SizeType* perm_arr,
+                       const std::vector<matrix::Tile<T, Device::CPU>>& in_tiles,
+                       const std::vector<matrix::Tile<T, Device::CPU>>& out_tiles) {
+  constexpr Coord ocrd = orthogonal(coord);
```
msimberg, line 81: `ocrd` = `Orthogonal CooRD`? Maybe `orth_coord` or something more explicit would be nice?

include/dlaf/permutations/general/impl.h
```diff
@@ -0,0 +1,201 @@
...
+#include "dlaf/lapack/tile.h"
+#include "dlaf/matrix/copy_tile.h"
+#include "dlaf/matrix/matrix.h"
+#include "dlaf/schedulers.h"
+#include "dlaf/sender/transform.h"
+#include "dlaf/sender/when_all_lift.h"
+#include "dlaf/types.h"
+#include "dlaf/util_matrix.h"
+
+#include "pika/parallel/algorithms/for_each.hpp"
```
msimberg, line 27:
```suggestion
#include "pika/algorithm.hpp"
```
We may provide a smaller header in the future just for `for_each` but `pika/parallel/algorithms/for_each.hpp` will at least not be the final location for that header, if it appears.

include/dlaf/permutations/general/impl.h
```diff
@@ -0,0 +1,201 @@
...
+  constexpr Coord ocrd = orthogonal(coord);
+  std::vector<SizeType> splits =
+      util::interleaveSplits(sz.get<ocrd>(), distr.blockSize().get<ocrd>(),
+                             distr.distanceToAdjacentTile<ocrd>(in_offset),
+                             distr.distanceToAdjacentTile<ocrd>(out_begin.get<ocrd>()));
+
+  // Parallelized over the number of permutated columns or rows
+  std::vector<SizeType> loop_arr(to_sizet(sz.get<coord>()));
+  std::iota(std::begin(loop_arr), std::end(loop_arr), 0);
+  pika::for_each(pika::execution::par, std::begin(loop_arr), std::end(loop_arr), [&](SizeType i_perm) {
```
msimberg, line 90:
```suggestion
  pika::for_loop(pika::execution::par, to_sizet(0), to_sizet(sz.get<coord>()), [&](SizeType i_perm) {
```

msimberg: Alternatively a `counting_iterator` type-thing from boost or self-made would be more appropriate to use with `for_each` to avoid having to allocate a vector.

include/dlaf/permutations/general/impl.h
```diff
@@ -0,0 +1,201 @@
...
+  namespace ex = pika::execution::experimental;
+
+  const matrix::Distribution& distr = mat_in.distribution();
+  TileElementSize sz_last_tile = distr.tileSize(GlobalTileIndex(i_end, i_end));
+  SizeType m = distr.globalTileElementDistance<Coord::Row>(i_begin, i_end) + sz_last_tile.rows();
+  SizeType n = distr.globalTileElementDistance<Coord::Col>(i_begin, i_end) + sz_last_tile.cols();
+  matrix::Distribution subm_distr(LocalElementSize(m, n), distr.blockSize());
+  SizeType ntiles = i_end - i_begin + 1;
+
+  // Note: there is an issue compiling with dlaf::internal::transform()
```
msimberg, line 164: For this I'd try replacing https://github.com/eth-cscs/DLA-Future/blob/60c620713feaaf6590e823947f412b5254609685/include/dlaf/sender/transform.h#L80 with `pika::unwrapping_n<1>(...)` (in the hope that we don't have other places that depend on having nested futures unwrapped automatically).

include/dlaf/util_matrix.h
```diff
@@ -387,6 +387,45 @@ void set_random_hermitian_positive_definite(Matrix<T, Device::CPU>& matrix) {
...
+auto collectReadTileSenders(GlobalTileIndex begin, GlobalTileSize sz, Matrix<const T, D>& mat) {
+  using pika::execution::experimental::keep_future;
+
+  std::vector<decltype(keep_future(pika::shared_future<matrix::Tile<const T, D>>()))> tiles;
+  tiles.reserve(to_sizet(sz.linear_size()));
+  for (auto idx : iterate_range2d(begin, sz)) {
+    tiles.push_back(mat.read_sender(idx));
+  }
+  return tiles;
+}
```
msimberg, line 427: It looks to me like you don't need the non-sender versions of these, so I would just make the "default" use `Matrix::read_sender` and `Matrix::readwrite_sender`. For read-write `Matrix::operator()` and `Matrix::readwrite_sender` are equivalent, but I'm trying to phase out the non-sender accesses (and then we can eventually just remove the `_sender` members).
```suggestion
template <class T, Device D>
auto collectReadWriteTile(GlobalTileIndex begin,
                           GlobalTileSize sz,
                           Matrix<T, D>& mat) {
  std::vector<decltype(mat.readwrite_sender(std::declval<LocalTileIndex>()))> tiles;
  tiles.reserve(to_sizet(sz.linear_size()));
  for (auto idx : iterate_range2d(begin, sz)) {
    tiles.push_back(mat(idx));
  }
  return tiles;
}

/// The tiles are returned in column major order
template <class T, Device D>
auto collectReadTile(GlobalTileIndex begin, GlobalTileSize sz, Matrix<const T, D>& mat) {
  std::vector<decltype(mat.read_sender(std::declval<LocalTileIndex>()))> tiles;
  tiles.reserve(to_sizet(sz.linear_size()));
  for (auto idx : iterate_range2d(begin, sz)) {
    tiles.push_back(mat.read_sender(idx));
  }
  return tiles;
}
```

msimberg (commented):
Only minor cleanup comments, looks good to me otherwise!

include/dlaf/util_math.h
```diff
@@ -18,8 +18,47 @@
...
+// o2      b ─┬──┘              l
+// │          │                │
+// └─►  │   │ ▼ │   │   │      │
+//    ──┴───┴───┴───┴───┴──  ◄─┘
+//
+inline std::vector<SizeType> interleaveSplits(SizeType l, SizeType b, SizeType o1, SizeType o2) {
+  DLAF_ASSERT(l > 0, l);
+  DLAF_ASSERT(b > 0, b);
+  DLAF_ASSERT(o1 >= 0, o1);
+  DLAF_ASSERT(o2 >= 0, o2);
```
rasolca, line 40: I think (according to tests and implementation) that offset should be smaller than b.
Please complete the assertions or add corresponding test cases.

rasolca (commented):
Looks good, however GPU test segfaulted...

include/dlaf/permutations/general/impl.h
teonnik: To unwrap `std::vector<pika::shared_future<matrix::Tile<const T, D>>>` we need to do something like:

```
template <class U>
std::vector<matrix::Tile<const U, Device::CPU>> unwrapConstTile(
    const std::vector<pika::shared_future<matrix::Tile<const U, Device::CPU>>>& tiles_fut) {
  std::vector<matrix::Tile<const U, Device::CPU>> tiles;
  tiles.reserve(tiles_fut.size());
  for (const auto& tile_fut : tiles_fut) {
    const auto& tile = tile_fut.get();
    SizeType len = tile.size().isEmpty() ? 0 : tile.ld() * (tile.size().cols() - 1) + tile.size().rows();
    tiles.emplace_back(tile.size(), MemoryView(tile.ptr(), len), tile.ld());
  }
  return tiles;
}
```

so I guess you are write in a sense because there is no new heap memory allocated from `MemoryView`. The allocations for the `std::vector<matrix::Tile<const T, D>>` is not a problem.

include/dlaf/permutations/general.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include "dlaf/common/assert.h"
+#include "dlaf/matrix/matrix.h"
+#include "dlaf/permutations/general/api.h"
+#include "dlaf/util_matrix.h"
+
+namespace dlaf::permutations {
+
+template <Backend B, Device D, class T, Coord coord>
+void permutate(SizeType i_begin, SizeType i_end, Matrix<const SizeType, D>& perms, Matrix<T, D>& mat_in,
```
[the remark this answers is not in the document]
teonnik: I'm not sure which is better tbh... https://wikidiff.com/permutate/permute

include/dlaf/permutations/general/impl.h
teonnik: I tried this but it still resulted in a compile-time error.

teonnik:
bors try

include/dlaf/permutations/general.h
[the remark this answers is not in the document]
msimberg: I would go for permute because:
1. It's shorter
2. At least according to https://en.wiktionary.org/wiki/permutate "Permute is much more commonly used".

include/dlaf/permutations/general/impl.h
msimberg: Ah, well, yes the `vector` allocation is indeed a new allocation. I was imagining it being possible to work directly with the `shared_future::get` references from the first vector avoiding the second vector completely. As I said, more clunky so I wouldn't do it, but I imagine it should be possible.

TL;DR: I'm only commenting on the comment. I have no problems with the decision to use non-const `T`.

include/dlaf/permutations/general/impl.h
msimberg: Hmm, ok. I will try to see what's causing problems there.

teonnik:
bors try

rasolca (approved):

include/dlaf/permutations/general/impl.h
msimberg: This is not a blocker. I'll look into this as soon as possible.

msimberg: So if I understand the `unwrapping` implementation correctly, the "depth" actually refers to the depth of unwrapping specifically futures which means that the `shared_future` in `std::vector<pika::shared_future<...>>` is considered to be at depth 1 and will be unwrapped.

As a short term solution I would suggest something like https://github.com/msimberg/DLA-Future/commit/e7297d3ede222df11f1138ba64dba1979e99e1e5, which simply adds an option to tell `transform` not to unwrap. That commit only adds it to the basic `transform` overload and nowhere else.

albestro (approved):
A couple of minor things. LGTM in general

include/dlaf/permutations/general/impl.h
```diff
@@ -0,0 +1,202 @@
...
+      }
+
+      TileElementIndex i_subtile_in = distr.tileElementIndex(i_split_gl_in);
+      auto& tile_in = in_tiles[to_sizet(distr.globalTileLinearIndex(i_split_gl_in))];
+      TileElementIndex i_subtile_out = distr.tileElementIndex(i_split_gl_out);
+      auto& tile_out = out_tiles[to_sizet(distr.globalTileLinearIndex(i_split_gl_out))];
+
+      dlaf::gpulapack::lacpy<T>(blas::Uplo::General, region.rows(), region.cols(),
+                                tile_in.ptr(i_subtile_in), tile_in.ld(), tile_out.ptr(i_subtile_out),
+                                tile_out.ld(), stream);
```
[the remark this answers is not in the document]
msimberg: @teonnik are you planning to try to combine the MC and GPU kernels here as well in this PR? I'd happily help if you have any problems with it.

include/dlaf/permutations/general.h
```diff
@@ -0,0 +1,60 @@
...
+namespace dlaf::permutations {
+
+/// Permutes the columns or rows of an input sub-matrix mat_in[i_begin:i_end][i_begin:i_end] into an
+/// output sub-matrix mat_out[i_begin:i_end][i_begin:i_end] using the index map of permutations
+/// perms[i_begin:i_end].
+///
+/// @param perms is the index map of permutations represented as a tiled column vector. Indices are in
+///        the range [0, n) where `n` is the size of the subproblem (i.e. the indices are local to the
+///        subproblem, they are not global). Only tiles whose row tile coords are in the closed range
+///        [i_begin,i_end] are accessed in read-only mode.
```
albestro, line 29: Not sure "subproblem" in this context is the right word. If I'm not wrong, indices refers to the local part of the matrix.

include/dlaf/util_matrix.h
```diff
@@ -387,6 +387,28 @@ void set_random_hermitian_positive_definite(Matrix<T, Device::CPU>& matrix) {
...
+template <class T, Device D>
+auto collectReadTiles(GlobalTileIndex begin, GlobalTileSize sz, Matrix<const T, D>& mat) {
+  std::vector<decltype(mat.read_sender(std::declval<LocalTileIndex>()))> tiles;
+  tiles.reserve(to_sizet(sz.linear_size()));
+  for (auto idx : iterate_range2d(begin, sz)) {
+    tiles.push_back(mat.read_sender(idx));
+  }
+  return tiles;
+}
+
```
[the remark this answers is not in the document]
teonnik: Thanks for pointing this out! They are quite similar but there are differences:

`selec()` is pretty much the same as `collectReadWriteTiles()` except that it calls `mat.read()` instead of `mat.read_sender()` and the 2D range is passed instead of constructed within the function.

`selectRead()` and `collectReadTiles()` have one more difference, in order to prevent unwrapping `std::vector<pika::shared_future<>>` which would lead to compile-time errors we need `decltype((mat.readwrite_sender(...))` so that the shared future is kept.

I suggest we try to refactor these functions in a separate PR.

test/unit/permutations/test_permutations.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+  for (auto [i_tile, i_el, val] : expected_entries) {
+    EXPECT_EQ(out_tiles[i_tile](i_el), val);
+  }
+}
+
+// Permute columns or rows in reverse order.
+// Each column or row of the input matrix has its index as a value.
+template <Backend B, Device D, class T, Coord C>
+void testPermutations(SizeType n, SizeType nb) {
```
albestro, line 208: Could it be useful to test this with non-square matrices?

include/dlaf/permutations/general/impl.h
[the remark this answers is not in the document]
teonnik: Is this OK? : https://github.com/eth-cscs/DLA-Future/pull/564/commits/38d85d3a93d8bb32ad87e0ed4eb46aaddd209df1

teonnik:
bors try

test/unit/permutations/test_permutations.cpp
teonnik: In principle, yes, it should work for non-square matrices as well. I haven't tested non-square matrices because the current use-case within the tridiagonal solver is square matrices.

include/dlaf/permutations/general/impl.h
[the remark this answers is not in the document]
msimberg: Yes, with the current `lacpy` situation... `lacpy` should have a wrapper for calling it with or without a stream which would avoid having you choose the overload at the callsite. However, it seems that's for some reason missing. I've opened an issue to track that: https://github.com/eth-cscs/DLA-Future/issues/575. Feel free to still implement it here, but I'd be fine with doing that in a follow-up PR.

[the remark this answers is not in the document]
msimberg: Please reopen this comment for tracking if you would still implement #575 in this PR.

msimberg:
Still no additional comments from me. I'll leave the test changes to @albestro to comment on.

teonnik:
bors try

test/unit/permutations/test_permutations.cpp
```diff
@@ -19,6 +19,8 @@
 #include "dlaf_test/matrix/util_tile.h"
 #include "dlaf_test/util_types.h"
 
+#include "dlaf/matrix/print_csv.h"
+
```
rasolca, line 23: Is it needed?

test/unit/permutations/test_permutations.cpp
```diff
@@ -204,66 +206,94 @@ void testApplyRowPermutations(SizeType n, SizeType nb) {
...
+  SizeType index_start = distr.globalElementFromGlobalTileAndTileElement<C>(i_begin, 0);
+  SizeType index_finish = distr.globalElementFromGlobalTileAndTileElement<C>(i_end, 0) +
+                          distr.tileSize(GlobalTileIndex(i_end, i_end)).get<C>();
+  dlaf::matrix::util::set(perms, [index_start, index_finish](GlobalElementIndex i) {
+    if (index_start > i.row() || i.row() >= index_finish)
+      return SizeType(0);
+
+    return index_finish - 1 - i.row();
+  });
   dlaf::matrix::util::set(mat_in_h, [](GlobalElementIndex i) { return T(i.get<C>()); });
```
rasolca, line 58: The elements of the columns (resp. the rows depending on C) of the matrix have the same value.
I'm wondering if it would be better to choose a linear combination of row and col index.

test/unit/permutations/test_permutations.cpp
teonnik: https://github.com/eth-cscs/DLA-Future/pull/564/commits/51ccf2923500485f08269c2a132f13660a99a51a

test/unit/permutations/test_permutations.cpp
rasolca: I'm worried about misalignment of rows/cols during the swap.
I suggest something like `return T(i.get<C>()) - T(i.get<OrthogonalToC>()) / 8;` which shouldn't overcomplicate the expected result.

albestro (approved):
LGTM

test/unit/permutations/test_permutations.cpp
teonnik: Is this OK? https://github.com/eth-cscs/DLA-Future/pull/564/commits/1c7ed5e5262c4a3ba957e5b5efd1b1a01da2c3d5

test/unit/permutations/test_permutations.cpp
```diff
@@ -0,0 +1,113 @@
...
+  SizeType index_finish = distr.globalElementFromGlobalTileAndTileElement<C>(i_end, 0) +
+                          distr.tileSize(GlobalTileIndex(i_end, i_end)).get<C>();
+  dlaf::matrix::util::set(perms, [index_start, index_finish](GlobalElementIndex i) {
+    if (index_start > i.row() || i.row() >= index_finish)
+      return SizeType(0);
+
+    return index_finish - 1 - i.row();
+  });
+  dlaf::matrix::util::set(mat_in_h, [](GlobalElementIndex i) {
+    return T(i.get<C>() - i.get<orthogonal(C)>()) / T(8);
```
rasolca, line 55:
```suggestion
    return T(i.get<C>()) - T(i.get<orthogonal(C)>()) / T(8);
```
I suggested `c - o / 8` such that to get the same sequence of elements in two nearest rows/cols you should move by 8 cols/rows. Using `(c - o) / 8` increasing the rows and cols by the same amount gives the same sequence.

test/unit/permutations/test_permutations.cpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  auto expected_out = [i_begin, i_end, index_start, index_finish, &distr](const GlobalElementIndex i) {
+    GlobalTileIndex i_tile = distr.globalTileIndex(i);
+    if (i_begin <= i_tile.row() && i_tile.row() <= i_end && i_begin <= i_tile.col() &&
+        i_tile.col() <= i_end) {
+      GlobalElementIndex i_in(i.get<orthogonal(C)>(), index_finish + index_start - 1 - i.get<C>());
+      if constexpr (C == Coord::Row)
+        i_in.transpose();
+      return T(i_in.get<C>() - i_in.get<orthogonal(C)>()) / T(8);
```
rasolca, line 73:
```suggestion
      return T(i_in.get<C>()) - T(i_in.get<orthogonal(C)>()) / T(8);
```

test/unit/permutations/test_permutations.cpp
teonnik: https://github.com/eth-cscs/DLA-Future/pull/564/commits/f41ebd0f32551a8e0d6726d94b0345f45cf7a99e

test/unit/permutations/test_permutations.cpp
teonnik: https://github.com/eth-cscs/DLA-Future/pull/564/commits/f41ebd0f32551a8e0d6726d94b0345f45cf7a99e

rasolca:
bors try

rasolca:
bors try-

PR #567, opened 2022-06-10
miniapps: install to bin/

haampie:

PR #569, opened 2022-06-13
Add CMake helper function for adding miniapps

msimberg:
The new function `DLAF_addMiniapp` is identical to `DLAF_addTest` with the removal of the `MPIRANKS` and `USE_MAIN` arguments (cf. https://github.com/eth-cscs/DLA-Future/pull/567#discussion_r894471351).

Flyby: fix compilation of `miniapp_bt_band_to_tridiag` with CUDA enabled (reminder to always run `bors try` ;) https://github.com/eth-cscs/DLA-Future/pull/561).


miniapp/CMakeLists.txt
```diff
@@ -8,80 +8,61 @@
...
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
 )
 
-add_executable(miniapp_cholesky miniapp_cholesky.cpp)
-target_link_libraries(miniapp_cholesky PRIVATE dlaf.core dlaf.auxiliary dlaf.factorization)
-target_link_libraries(miniapp_cholesky PRIVATE DLAF_miniapp)
-target_add_warnings(miniapp_cholesky)
-DLAF_addPrecompiledHeaders(miniapp_cholesky)
-install(TARGETS miniapp_cholesky RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+DLAF_addMiniapp(miniapp_cholesky
```
teonnik, line 18: Where is `DLAF_addMiniapp` defined?

msimberg: Ugh, in a file that I forgot to add ;)

msimberg: In `cmake/DLAF_AddMiniapp.cmake`.

rasolca:
bors try

teonnik (approved):

miniapp/CMakeLists.txt
```diff
@@ -8,80 +8,61 @@
...
+  SOURCES miniapp_cholesky.cpp
+  LIBRARIES dlaf.core dlaf.auxiliary dlaf.factorization
```
albestro, line 20: Just realised that in this miniapp we still rely on "internal" targets, but by adding `DLAF_miniapp` we are implicitly depending on the full `DLAF` dependency.

albestro (approved):
Some minor comments, but for the rest it's ok.

Nice and clean work!

cmake/DLAF_AddMiniapp.cmake
```diff
@@ -0,0 +1,67 @@
...
+#     PRIVATE
+#       boost::boost
+#       include/
+# )
+
+function(DLAF_addMiniapp miniapp_target_name)
+  set(options "")
+  set(oneValueArgs MPIRANKS USE_MAIN)
+  set(multiValueArgs SOURCES COMPILE_DEFINITIONS INCLUDE_DIRS LIBRARIES ARGUMENTS)
+  cmake_parse_arguments(DLAF_AT "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
```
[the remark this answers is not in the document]
msimberg: Ah, yes, that makes a lot more sense. I'll change that.

miniapp/CMakeLists.txt
msimberg: Good catch. I think `DLAF_miniapp` probably doesn't need all the individual libraries. Will update it to depend only on a subset if that's the case.

msimberg: Hrm, so `dlaf.core` would be enough for `DLAF_miniapp`. However, we use a mix in the miniapps. The one above depends on individual sub-libraries, but most others depend on `DLAF`.

Do we want to keep depending on individual libraries only? If yes, I'll change the ones depending on `DLAF` to more fine-grained dependencies.

rasolca: I think is better to link with DLAF, at least until we find a way to workaround the problem of dependencies of object libraries.

msimberg: Everything links to `DLAF` now. The miniapps link explicitly to `DLAF` even though they would get it through `DLAF_miniapp` as well. Is that ok with you?

rasolca:
bors try

PR #571, opened 2022-06-14
Add p2p communication

albestro:
This PR revamps a byproduct of @rasolca and "senderifise" it.

This is needed by distributed versions of BandToTridiagonal and its BackTransformation.

Additional changes:
- fix cmake-format problems on `master`
- fix build problem (missing header)

### DIFF
--- a/include/dlaf/communication/kernels/p2p.h
+++ b/include/dlaf/communication/kernels/p2p.h
@@ -31,6 +31,10 @@ namespace internal {
 template <class T, Device D>
 void send(const Communicator& comm, const matrix::Tile<const T, D>& tile, IndexT_MPI dest,
           IndexT_MPI tag, MPI_Request* req) {
+#if !defined(DLAF_WITH_CUDA_RDMA)
+  static_assert(D == Device::CPU, "DLAF_WITH_CUDA_RDMA=off, MPI accepts just CPU memory.");
+#endif
+
   auto msg = comm::make_message(common::make_data(tile));
   DLAF_MPI_CHECK_ERROR(
       MPI_Isend(const_cast<T*>(msg.data()), msg.count(), msg.mpi_type(), dest, tag, comm, req));
@@ -42,6 +46,10 @@ DLAF_MAKE_CALLABLE_OBJECT(send);
 template <class T, Device D>
 auto recv(const Communicator& comm, const matrix::Tile<T, D>& tile, IndexT_MPI source, IndexT_MPI tag,
           MPI_Request* req) {
+#if !defined(DLAF_WITH_CUDA_RDMA)
+  static_assert(D == Device::CPU, "DLAF_WITH_CUDA_RDMA=off, MPI accepts just CPU memory.");
+#endif
+
   auto msg = comm::make_message(common::make_data(tile));
   DLAF_MPI_CHECK_ERROR(MPI_Irecv(msg.data(), msg.count(), msg.mpi_type(), source, tag, comm, req));
 }
@@ -54,8 +62,8 @@ void scheduleSend(IndexT_MPI dest, CommSender&& pcomm, IndexT_MPI tag, Sender&& 
   using dlaf::internal::whenAllLift;
   using pika::execution::experimental::start_detached;
 
-  whenAllLift(std::forward<CommSender>(pcomm), std::forward<Sender>(tile), dest, tag) |
-      internal::transformMPI(internal::send_o) | pika::execution::experimental::start_detached();
+  start_detached(whenAllLift(std::forward<CommSender>(pcomm), std::forward<Sender>(tile), dest, tag) |
+                 internal::transformMPI(internal::send_o));
 }
 
 template <class CommSender, class Sender>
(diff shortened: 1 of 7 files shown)


include/dlaf/communication/kernels/p2p.h
```diff
@@ -0,0 +1,72 @@
...
+#include "dlaf/common/pipeline.h"
+#include "dlaf/communication/communicator.h"
+#include "dlaf/communication/message.h"
+#include "dlaf/matrix/tile.h"
+#include "dlaf/sender/traits.h"
+#include "dlaf/sender/transform_mpi.h"
+#include "dlaf/sender/when_all_lift.h"
+
+namespace dlaf {
+namespace comm {
```
msimberg, line 27: I'm changing these in the other `communication/kernels` files at the moment and can do this later as well, but it could be updated here as well:
```suggestion
namespace dlaf::comm {
```

include/dlaf/communication/kernels/p2p.h
```diff
@@ -0,0 +1,72 @@
...
+  auto msg = comm::make_message(common::make_data(tile));
+  DLAF_MPI_CHECK_ERROR(MPI_Isend(const_cast<T*>(msg.data()), msg.count(), msg.mpi_type(), receiver, tag,
+                                 pcomm.ref(), req));
+}
+
+DLAF_MAKE_CALLABLE_OBJECT(send);
+
+// Non-blocking point to point receive
+template <class T, Device D>
+auto recv(const matrix::Tile<T, D>& tile, IndexT_MPI sender, IndexT_MPI tag,
```
msimberg, line 44: Possibly...
```suggestion
void send(const matrix::Tile<const T, D>& tile, IndexT_MPI dest, IndexT_MPI tag,
          common::PromiseGuard<Communicator> pcomm, MPI_Request* req) {
  auto msg = comm::make_message(common::make_data(tile));
  DLAF_MPI_CHECK_ERROR(MPI_Isend(const_cast<T*>(msg.data()), msg.count(), msg.mpi_type(), dest, tag,
                                 pcomm.ref(), req));
}

DLAF_MAKE_CALLABLE_OBJECT(send);

// Non-blocking point to point receive
template <class T, Device D>
auto recv(const matrix::Tile<T, D>& tile, IndexT_MPI source, IndexT_MPI tag,
```
I know `sender` and `receiver` are already overloaded and the names would fit here as well, but considering that various MPI implementations call them `source` and `dest` those might be better names (https://www.mpich.org/static/docs/latest/www3/MPI_Isend.html).

msimberg (changes requested):
Looks good to me in general, just a few minor requests.

test/unit/communication/test_comm_p2p.cpp
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  const auto& tile = matrix.read(idx).get();
+  SCOPED_TRACE(test_name);
+  CHECK_TILE_EQ(input_tile, tile);
+}
+
+TEST_F(P2PTest, SendRecv) {
+  testSendRecv(world, newBlockMatrixContiguous<T, device>(), "Contiguous");
+  testSendRecv(world, newBlockMatrixStrided<T, device>(), "Strided");
```
albestro, line 90: For P2P communications it is not really needed to check that it works for both contiguous and strided memory, since that's a problem related to collective MPI operations.

In addition, it may be useful to test a more articulated use-case, e.g. using tagged communications inside the same communicator.

include/dlaf/communication/kernels/p2p.h
```diff
@@ -0,0 +1,70 @@
...
+  auto msg = comm::make_message(common::make_data(tile));
+  DLAF_MPI_CHECK_ERROR(
+      MPI_Irecv(msg.data(), msg.count(), msg.mpi_type(), source, tag, pcomm.ref(), req));
+}
+
+DLAF_MAKE_CALLABLE_OBJECT(recv);
+}
+
+template <class CommSender, class Sender>
+void scheduleSend(IndexT_MPI dest, CommSender&& pcomm, IndexT_MPI tag, Sender&& tile) {
```
rasolca, line 54: How do you plan to use it if communications are not ordered?
with a shared_future like sender?

albestro: If you are talking about the problem that currently we implicitly require in our internal communication kernels to get a `PromiseGuard` instead of the actual `Communicator`, I agree with you and it is something I'd like to discuss with @msimberg in these days.

On the contrary, if it is not that, I'd ask you to explain a bit more in detail the problem you are thinking about.

albestro: @rasolca confirmed to me that the point was exactly that one. This is going to be addressed in #573.

test/unit/communication/test_comm_p2p.cpp
rasolca: contiguous/strided is not related only to collectives...

Anyway testing mixing tags is a good idea.

albestro: just added 😉 feel free to resolve this comment if you think it is enough.

rasolca (approved):

rasolca:
bors try

include/dlaf/communication/kernels/p2p.h
```diff
@@ -0,0 +1,76 @@
...
+DLAF_MAKE_CALLABLE_OBJECT(recv);
+}
+
+template <class CommSender, class Sender>
+[[nodiscard]] auto scheduleSend(IndexT_MPI dest, CommSender&& pcomm, IndexT_MPI tag, Sender&& tile) {
+  using dlaf::internal::whenAllLift;
+  using pika::execution::experimental::start_detached;
+
+  return whenAllLift(std::forward<CommSender>(pcomm), std::forward<Sender>(tile), dest, tag) |
+         internal::transformMPI(internal::send_o);
```
msimberg, line 66: I'm happy to change this in #485 as well since I did similar changes there, but I'd try to have the `CommSender` as the first argument in `scheduleSend` and have the same order of arguments in `send` and `scheduleSend`.

Please feel free to leave this to me if your work is blocked on getting this PR in.

msimberg (approved):

PR #572, opened 2022-06-14
Add helpers for handling communication tiles

### LINKED ISSUE #485: Add senderified communication helper for contiguous buffer

msimberg:
Fixes #485.

This is open to renamings, refactorings, etc. This is a starting point.

There are (at least) three independent factors that need to be taken into account when creating "communication buffers":
- Should the input be copied into the temporary buffer, or is it enough to have a "similarly shaped" temporary buffer?
- Does the temporary buffer have to be contiguous or not?
- Should the temporary buffer be copied back into the input at the end?

I've currently put the first two factors into the function names (`with(Contiguous)(Similar)CommTile`), and the last is a separate helper function. This is not fixed and these decisions were mostly made to keep things simple while implementing these helpers. Some of these could be template parameters instead of separate functions but I kind of like the separate functions (unless someone sees a use case for choosing a particular variant at compile time). Creating contiguous tiles is a runtime option because it is not encoded in the type.

This now uses a single implementation again for `CPU` and `GPU` communication.

I'm proposing to change all the communication "kernels" to only take tiles, for consistency with each other (some took tiles before, some `ContiguousBufferHolder`s), and to keep the kernels oblivious of how a tile lifetime is managed. The kernels may still have prerequisites (contiguous, CPU-only), but that is only asserted in the kernels and made sure to hold outside the kernels. The change to take tiles in all kernels means that this only supports sending real-valued tiles, i.e. no integral tiles. `lacpy` (currently?) only supports real-valued inputs and with the setup in this PR that becomes a requirement for the `scheduleX` functions as well. One of the unit tests used `int` tiles which works on master but not with this PR. Is this acceptable? Do we need to support anything else than what `lacpy` supports?

Related changes:
- `namespace a { namespace b {` to `namespace a::b {`.
- changed `scheduleX` communication helper functions to return a sender instead of starting the work internally. This matches the use of the word "schedule" in P2300 better.
- Added a `static constexpr` member to `Tile` with the used `Device`.
- Added an `is_contiguous` member to `Tile`.

A few questions:
- ~Is there any good reason not to provide `operator==` (and `!=`) for `Tile`?~ Not required after refactoring. The meaning of `==` is unclear as well.
- Do we expect the MPI reduction functions to require CPU inputs even if CUDA RDMA would otherwise work? I'm currently assuming that reductions require CPU memory.

Definitely to do:
- [x] naming
- [x] clean up comments and diagrams (if needed)
- [x] remove now unneeded helpers (copy/duplicate/contiguous buffer holder etc.)
- [x] add `make_(unique_)any_sender` to pika (will require 0.6.0) (https://github.com/pika-org/pika/pull/259)
- [x] add `drop_values` adaptor to pika (https://github.com/pika-org/pika/pull/262)
- [x] remove duplication in `withTemporaryTile`
- [x] move `with_communication_tile.h` to `include/dlaf/sender/with_temporary_tile.h`?
- [x] unit tests
- [x] wait for #583
- [x] https://github.com/eth-cscs/DLA-Future/pull/571#discussion_r917914324


include/dlaf/common/callable_object.h
```diff
@@ -17,12 +17,14 @@
...
-#define DLAF_MAKE_CALLABLE_OBJECT(fname)                                                 \
-  constexpr struct fname##_t {                                                           \
-    template <typename... Ts>                                                            \
-    auto operator()(Ts&&... ts) const noexcept(noexcept(fname(std::forward<Ts>(ts)...))) \
-        -> decltype(fname(std::forward<Ts>(ts)...)) {                                    \
-      return fname(std::forward<Ts>(ts)...);                                             \
-    }                                                                                    \
-  } fname##_o {                                                                          \
+///
+/// TODO: Aaargh, (f)(...) turns off ADL!? That's great, but too obscure?
```
msimberg, line 21: This was an interesting one... The wrapper for `copy` would in the changes that I've made attempt to call `std::copy` because of ADL. Wrapping a function name in parentheses seems to be a valid way to disable ADL and that is exactly what we want here. However, is it too obscure? It definitely needs a comment if this will be the way we disable ADL. Other wrappers likely also need to disable ADL just to be on the safe side. The other option is qualifying the function name which is simple, but it may not be as obvious at the callsite (of the macro) why the qualified name is required.

msimberg:
bors try

rasolca:
> Do we need to support anything else than what `lacpy` supports?

The use case is to communicate indices (in our case `SizeType`). I think it it is needed in the tridiagonal solver. @teonnik can you confirm it?

>     * Is there any good reason not to provide `operator==` (and `!=`) for `Tile`?

Not sure, what would == mean for you?

msimberg:
> > Do we need to support anything else than what `lacpy` supports?
> 
> The use case is to communicate indices (in our case `SizeType`). I think it it is needed in the tridiagonal solver. @teonnik can you confirm it?

Thanks, that's good to know. If they indeed need to be communicated (and not just passed around in local tiles) I would perhaps still prefer to keep the `Tile` interface since we have `copy` defined for `CPU-CPU`, `CPU-GPU`, etc. copies. The `data` copy only works with CPU memory. Perhaps a specialization of non-floating point types would be enough (or even calling the CPU-only `data` copy). I have to check if any of the CUDA copy functions require only floating point types (and all my previous mentions of real types were of course meant to be floating point...), but I don't see why it wouldn't support any type.

> > ```
> > * Is there any good reason not to provide `operator==` (and `!=`) for `Tile`?
> > ```
> 
> Not sure, what would == mean for you?

Same underlying buffer and dimensions for use here: https://github.com/eth-cscs/DLA-Future/blob/d48262017c4c6313793718cffaad2cd65cdc4924/include/dlaf/communication/with_communication_tile.h#L146. However, this is far from a must since combining `copyBack` into the `withX` helpers I would already know if the tile was the same or not and I wouldn't need to check equality like this.

To add the "combined API", it would perhaps look something like this:
```
withTemporaryTile<DestinationDevice, RequireContiguous or AllowDiscontiguous, CopyToDestination, CopyToSource>(tile_sender, [](const tile& temp_tile) {...})
```
This would have the benefit of not exposing the input tile in the callable parameters (not really bad, but it's not necessary other than to be able to copy the tile back). If this were the API I would also always return the original tile in a sender from the `withTemporaryTile` adaptor, and the user can discard or use it as it sees fit. My only worry with this is that it's a bit of a kitchen sink, but at the moment I don't see a reason for it to expand, but if someone already now sees a future extension of this I'd be wary to combine all the functionality in one adaptor to avoid it becoming huge.

rasolca:
> > > Do we need to support anything else than what `lacpy` supports?
> > 
> > 
> > The use case is to communicate indices (in our case `SizeType`). I think it it is needed in the tridiagonal solver. @teonnik can you confirm it?
> 
> Thanks, that's good to know. If they indeed need to be communicated (and not just passed around in local tiles) I would perhaps still prefer to keep the `Tile` interface since we have `copy` defined for `CPU-CPU`, `CPU-GPU`, etc. copies. The `data` copy only works with CPU memory. Perhaps a specialization of non-floating point types would be enough (or even calling the CPU-only `data` copy). I have to check if any of the CUDA copy functions require only floating point types (and all my previous mentions of real types were of course meant to be floating point...), but I don't see why it wouldn't support any type.

Cuda memcpys are type independent. It just need the correct sizes in bytes.

> > > ```
> > > * Is there any good reason not to provide `operator==` (and `!=`) for `Tile`?
> > > ```
> > 
> > 
> > Not sure, what would == mean for you?
> 
> Same underlying buffer and dimensions for use here:
> 
> https://github.com/eth-cscs/DLA-Future/blob/d48262017c4c6313793718cffaad2cd65cdc4924/include/dlaf/communication/with_communication_tile.h#L146
> . However, this is far from a must since combining `copyBack` into the `withX` helpers I would already know if the tile was the same or not and I wouldn't need to check equality like this.

I would not use the == operator (or !=) for this. I think a member (or a friend function) is better such that it is clear what you are checking. My fear is that user will expect the ==operator to perform a element by element comparison.

> To add the "combined API", it would perhaps look something like this:
> 
> ```
> withTemporaryTile<DestinationDevice, RequireContiguous or AllowDiscontiguous, CopyToDestination, CopyToSource>(tile_sender, [](const tile& temp_tile) {...})
> ```
> 
> This would have the benefit of not exposing the input tile in the callable parameters (not really bad, but it's not necessary other than to be able to copy the tile back). If this were the API I would also always return the original tile in a sender from the `withTemporaryTile` adaptor, and the user can discard or use it as it sees fit. My only worry with this is that it's a bit of a kitchen sink, but at the moment I don't see a reason for it to expand, but if someone already now sees a future extension of this I'd be wary to combine all the functionality in one adaptor to avoid it becoming huge.

msimberg:
> I would not use the == operator (or !=) for this. I think a member (or a friend function) is better such that it is clear what you are checking. My fear is that users will expect the ==operator to perform a element by element comparison.

Yep, I think that's a valid concern. I'll try to implement the second version today for comparison and that one wouldn't need `operator==`.

msimberg:
The implementation is not complete, but see https://github.com/eth-cscs/DLA-Future/pull/572/commits/b476321e82f8aa5726e94159c53b6aa845176931 for an idea of what the combined adaptor would look like. There's a ton of refactoring required in the implementation of `withTemporaryTile`, so only have a look at the usage in https://github.com/eth-cscs/DLA-Future/blob/b476321e82f8aa5726e94159c53b6aa845176931/include/dlaf/communication/kernels/broadcast.h#L76-L77 and https://github.com/eth-cscs/DLA-Future/blob/b476321e82f8aa5726e94159c53b6aa845176931/include/dlaf/communication/kernels/broadcast.h#L99-L100. Assuming I don't hit any nasty surprises for `reduce.h` and `all_reduce.h` I think I do actually prefer this version.

What do you think?

rasolca:
It's a nice interface, but I would like to ask you if and how it can handle the case in which
the RequireContiguous parameter depends on the device used.
Use case: when I tried some time ago MPI direct on daint non contiguous buffers had a huge performance problem.
Therefore a contiguous buffer would be always required in that case.

msimberg:
bors try

msimberg:
bors try

msimberg:
I'm not looking for approvals of the whole PR since this is still very much a work in progress, but I'd like some more thumbs up if you think the `withTemporaryTile` API looks good or comments if you think it would need other capabilities before I go and clean this up properly.

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

include/dlaf/matrix/copy_tile.h
```diff
@@ -190,6 +190,22 @@ struct Duplicate {
...
+    auto source_size = source.size();
+    dlaf::memory::MemoryView<T, Destination> mem_view(source_size.linear_size());
+    Tile<T, Destination> destination(source_size, std::move(mem_view), source_size.rows());
+    return Tile<T, Destination>(std::move(destination));
+  }
+};
+
+// TODO?
+// template <typename Destination>
+// inline constexpr DuplicateNoCopy<Destination> duplicate_no_copy(_o){};
```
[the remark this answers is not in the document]
msimberg: Not doing this here.

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -89,52 +72,60 @@ void scheduleAllReduce(CommSender&& pcomm, MPI_Op reduce_op,
...
+  using dlaf::internal::whenAllLift;
+  using dlaf::internal::withTemporaryTile;
+
+  auto all_reduce_in_place = [reduce_op,
+                              pcomm = std::forward<CommSender>(pcomm)](auto const& tile_comm) mutable {
+    return whenAllLift(std::move(pcomm), reduce_op, std::cref(tile_comm)) |
+           transformMPI(allReduceInPlace_o);
+  };
+
+  // TODO: Can reductions happen on CPU only?
```
msimberg, line 126: @rasolca do you happen to know if reductions could be made on GPU buffers if CUDA RDMA is enabled? I think to be on the safe side we should probably stay with explicitly using `Device::CPU` here until we get to test CUDA RDMA with a working implementation, but I'm wondering if I should update this comment in some way.

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -89,52 +72,60 @@ void scheduleAllReduce(CommSender&& pcomm, MPI_Op reduce_op,
...
+[[nodiscard]] auto scheduleAllReduceInPlace(CommSender&& pcomm, MPI_Op reduce_op, TileSender&& tile) {
   // Note:
   //
   // TILE ---> makeContiguous --+--> CONT_BUF ----> mpi_call ---> CONT_BUF --+
   //                            |                                            |
   //                            +------------------> TILE -------------------+-> copyBack ---> TILE
   //
-  //
   // The last TILE after the copyBack is returned so that other task can be attached to it,
   // AFTER the asynchronous MPI_AllReduce has completed
```
msimberg, line 111: How would you prefer to see these comments updated?

My preference would be to change the comments in the `scheduleX` functions to explain why the template parameters are as they are in the call to `withTemporaryTile` and remove the diagrams in `scheduleX`. I would instead add one or more diagrams to the `withTemporaryTile` documentation instead explaining the general idea of creating a temporary tile. :+1: if you're happy with this or comment if you prefer something else.

include/dlaf/communication/kernels/all_reduce.h
rasolca: They work with GPU pointers too, however the implementation I checked was reducing on CPU and copying back to GPU.

msimberg: Thanks. In that case it sounds like I can add to the general comment for the reduction functions that we explicitly choose to do the reduction on `Device::CPU` so that we can manage the copies, instead of having MPI do it for us. Sounds good?

msimberg:
bors try

I've reordered the function parameters for the communication helpers to be a bit more consistent. They are now `(communicator(guard/sender), root rank (if available), reduce op (if available), tile(s))`.

include/dlaf/communication/kernels/all_reduce.h
msimberg: I've updated this. I'm marking this as resolved, but please feel free to reopen/comment if you'd like to see changes.

include/dlaf/communication/kernels/all_reduce.h
msimberg: Updated the comment.

include/dlaf/common/callable_object.h
msimberg: I moved this to #580 since it no longer seems to be required specifically for this PR, but I still find it a useful change.

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

msimberg:
I've restarted the CI jobs. It seems like the scratch quota issues are (temporarily?) solved...

msimberg:
This should be ready to go. However, as mentioned earlier I'd be happy to have https://github.com/eth-cscs/DLA-Future/pull/571 go in before this if needed.

albestro (approved):
Very nice and elegant work...I really like it!

Just a couple of minor comments and one last question: about `CUDA_RDMA`, with the `static_assert` asking the Device to be CPU, it won't work right? not asking for this PR, probably it would be worth an issue in case (afaik we never tested it yet)

include/dlaf/matrix/tile.h
```diff
@@ -222,9 +223,13 @@ class Tile<const T, device> {
...
   }
 
+  bool is_contiguous() const noexcept {
+    return data_.ld() == data_.size().rows();
+  }
+
   /// Prints information about the tile.
   friend std::ostream& operator<<(std::ostream& out, const Tile& tile) {
-    return out << "size=" << tile.size() << ", ld=" << tile.ld();
+    return out << "size=" << tile.size() << ", ld=" << tile.size().rows();
```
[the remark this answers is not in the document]
msimberg: Yeah, I don't remember what this is about... but it certainly isn't required. I think it's a leftover from when I was adding the `is_contiguous` helper to `Tile`. Will remove.

include/dlaf/matrix/copy_tile.h
```diff
@@ -56,14 +56,31 @@ struct CopyBackend<Device::GPU, Device::GPU> {
...
+// TODO: Move to some common util header?
+template <typename T>
+struct IsFloatingPointOrComplex : std::is_floating_point<T> {};
+
+template <typename T>
+struct IsFloatingPointOrComplex<std::complex<T>> : IsFloatingPointOrComplex<T> {};
+
+template <typename T>
+inline constexpr bool IsFloatingPointOrComplex_v = IsFloatingPointOrComplex<T>::value;
+
```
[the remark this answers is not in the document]
msimberg: Yeah, sounds good to me. @rasolca, @teonnik fine with you?

include/dlaf/sender/with_temporary_tile.h
```diff
@@ -0,0 +1,199 @@
...
+                      }();
+                      // Send the input tile to continuations if the tile is
+                      // non-const.
+                      return std::move(copy_sender) | ex::then(moveNonConstTile{in});
+                    });
+           };
+
+           // One of the helpers may be unused depending on which branch is taken
+           dlaf::internal::silenceUnusedWarningFor(helper_nocopy);
+           dlaf::internal::silenceUnusedWarningFor(helper_copy);
```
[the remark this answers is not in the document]
msimberg: Yep, that's accurate and a good idea. I like `helper_{no,with}temp` as alternative names as well. Any other suggestions?

albestro:
> > about CUDA_RDMA, with the static_assert asking the Device to be CPU, it won't work right? not asking for this PR, probably it would be worth an issue in case (afaik we never tested it yet)
> 
> That's orthogonal, though it of course needs testing once we get a system with working RDMA. `withTemporaryTile` doesn't use `CommunicationDevice` internally. The caller of `withTemporaryTile` chooses what `Device` should be used for the temporary tile. They can explicitly say `Device::CPU` or `CommunicationDevice<...>`. The cases where the communication requires CPU memory (reductions) you call `withTemporaryTile<Device::CPU, ...>` and where it doesn't matter you use `withTemporaryTile<CommunicationDevice<...>, ...>`. As far as I can tell I've been consistent with the `static_assert`s and explicitly using `Device::CPU`, but please let me know if you noticed any inconsistencies.

Nope, sorry, you're fully right, I got mislead.

include/dlaf/sender/with_temporary_tile.h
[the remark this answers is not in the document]
msimberg: I've renamed the `helper_{no,with}temp`. Please mark this as resolved if you're happy with that.

include/dlaf/matrix/copy_tile.h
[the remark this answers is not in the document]
msimberg: Moved to `dlaf/types.h`. There's a naming style issue that I'm not sure I want to start solving here without more discussion about e.g. #576. I named the variable `IsFloatingPointOrComplex_v` but the guide says that variables should be `snake_case`. I've previously added other `inline constexpr` variables that are also `PascalCase` in that file. There's also one that is `camelCase`. Are you happy with the current inconsistency or should I make some change right now?

msimberg:
Modulo things that I've messed up in the latest commits this is now again ready to go (though I'm not quite sure CI is ready for it...).

rasolca (approved):

msimberg:
Ok, so I did break `withTemporaryTile` with the latest commits. Back to testing for a moment...

msimberg:
So I was making assumptions that I should not have done about the context that a function was running in which lead to deadlocks in the test. The actual adapter itself is unchanged with the latest commit. I was calling `sync_wait` which in some cases ended up being called inside the pika polling loop, and the `sync_wait` itself was waiting for the polling loop to make progress.

bors try

PR #573, opened 2022-06-17
change transformMPI to pass Communicator instead of PromiseGuard

albestro:
This PR aims at proposing a way to make internal communication kernels deal with `Communicator`s instead of `PromiseGuard<Communicator>`s.

Let's consider the internal reduce communication kernel as an example:
https://github.com/eth-cscs/DLA-Future/blob/5a23516e72a2cff20a8672e5a5bb8f71809ae092/include/dlaf/communication/kernels/reduce.h#L40-L48

which currently accepts as parameter a `PromiseGuard<Communicator>`.

For what concerns the kernel implementation, it is interesting just the `Communicator`, but currently it has the burden of unwrapping it from the `PromiseGuard` that contains it before using it.

Apart from the (minor) burden of the unwrapping, passing a `PromiseGuard` implicitly enforces a superfluous constraint: in order to use it, we cannot use a `Communicator` as it is, but we have to wrap it inside a `Pipeline`.

We may want to drop this requirement, since we may soon have use-cases with algorithms that do not require ordered communications, i.e. by means of a `Pipeline`, which might also represent an over-constraint to the parallelisation.

After a quick chat with @msimberg about this, we thought about exploring the solution of adapting `transformMPI` so that it still "consumes" a `PromiseGuard` just after the kernel finishes (i.e. waiting just for the posting, not for its completion), and at the same time it also unwraps the `PromiseGuard<Communicator>` by calling `.ref()`. This is a "conditional" management, which means that if the `Communicator` is passed as it is without a `PromiseGuard`, it will go through and it won't get altered.

There might be drawbacks, and this PR is for sharing with the team so that we can find and implement the best solution.

Looking forward to your comments/feedbacks!

### DIFF
--- a/include/dlaf/sender/transform_mpi.h
+++ b/include/dlaf/sender/transform_mpi.h
@@ -10,6 +10,7 @@
 #pragma once
 
 #include <pika/unwrap.hpp>
+#include <type_traits>
 
 #include "dlaf/common/pipeline.h"
 #include "dlaf/communication/communicator.h"
@@ -17,6 +18,9 @@
 #include "dlaf/sender/when_all_lift.h"
 
 namespace dlaf::comm::internal {
+
+template <class T>
+struct IsPromiseGuardCommunicator : std::is_same<T, dlaf::common::PromiseGuard<Communicator>> {};
 
 /// Helper for moving a PromiseGuard<Communicator> into the MPI function being
 /// called by transformMPI. Callables passed to transformMPI have their
@@ -27,11 +31,23 @@ namespace dlaf::comm::internal {
 /// that the guard is released on return from the callable.
 template <typename T>
 decltype(auto) movePromiseGuard(T& t) {
-  if constexpr (std::is_same_v<std::decay_t<T>, dlaf::common::PromiseGuard<Communicator>>) {
+  if constexpr (IsPromiseGuardCommunicator<std::decay_t<T>>::value) {
     return std::move(t);
   }
   else {
     return static_cast<T&>(t);
+  }
+}
+
+/// Helper for unwrapping a PromiseGuard and get a reference to its content, while
+/// letting through anything else.
+template <typename T>
+decltype(auto) getPromiseGuardRef(T&& t) {
+  if constexpr (IsPromiseGuardCommunicator<std::decay_t<T>>::value) {
+    return t.ref();
+  }
+  else {
+    return std::forward<T>(t);
   }
 }
 
@@ -50,7 +66,8 @@ struct MPICallHelper {
   std::decay_t<F> f;
   template <typename... Ts>
   auto operator()(Ts&&... ts) -> decltype(pika::unwrapping(std::move(f))(
-      movePromiseGuard(dlaf::internal::getReferenceWrapper(ts))..., std::declval<MPI_Request*>())) {
+      getPromiseGuardRef(movePromiseGuard(dlaf::internal::getReferenceWrapper(ts)))...,
+      std::declval<MPI_Request*>())) {
     MPI_Request req;
     auto is_request_completed = [&req] {
       int flag;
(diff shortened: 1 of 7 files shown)


include/dlaf/sender/transform_mpi.h
```diff
@@ -50,23 +66,25 @@ struct MPICallHelper {
...
     };
 
-    using result_type = decltype(pika::unwrapping(
-        std::move(f))(movePromiseGuard(dlaf::internal::getReferenceWrapper(ts))..., &req));
+    using result_type = decltype(pika::unwrapping(std::move(
+        f))(getPromiseGuardRef(movePromiseGuard(dlaf::internal::getReferenceWrapper(ts)))..., &req));
     if constexpr (std::is_void_v<result_type>) {
-      pika::unwrapping(std::move(f))(movePromiseGuard(dlaf::internal::getReferenceWrapper(ts))..., &req);
+      pika::unwrapping(std::move(
+          f))(getPromiseGuardRef(movePromiseGuard(dlaf::internal::getReferenceWrapper(ts)))..., &req);
```
msimberg, line 82: Something like this:
```suggestion
      pika::unwrapping(std::move(
          f))(getPromiseGuardRef(dlaf::internal::getReferenceWrapper(ts))..., &req);
      consumePromiseGuard(ts...);
```
where `consumePromiseGuard` would be something like this:
```
template <typename T>
void consumePromiseGuard(T& t) {
  if constexpr (IsPromiseGuardCommunicator<std::decay_t<T>>::value) {
     auto t_local = std::move(t);
  }
```

msimberg (changes requested):
@albestro just putting down in writing what we discussed on Friday. I think your idea of separating consuming the promise guard from the call to `f`. If you indeed had some idea for verifying this behaviour with a test I think that'd be great, even if it may be somewhat complicated...

include/dlaf/sender/transform_mpi.h
msimberg: That's a good point. I think adding `Communicator` to the name is a good idea.

msimberg (approved):

msimberg:
bors try

albestro:
bors try

test/unit/communication/test_transform_mpi.cpp
```diff
@@ -0,0 +1,131 @@
...
+  Communicator world(MPI_COMM_WORLD);
+  const IndexT_MPI rank = world.rank();
+
+  if (rank == 0) {
+    using dlaf::comm::internal::transformMPI;
+    using dlaf::internal::whenAllLift;
+
+    namespace ex = pika::execution::experimental;
+
+    common::Pipeline<Communicator> chain(world);
```
albestro, line 58: Communicator `world` gets passed into Pipeline without being moved, and it gets used later for the signal message.

Copying a Communicator means copying the handle, so we end up using the same handle from different threads (in edge cases concurrently). It should be ok, since MPI calls should be thread-safe, moreover it is the `MPI_COMM_WORLD` communicator which is managed by MPI directly (so there are no additional caveats about lifetime management).

In case it is not, we may opt for cloning the Communicator on both ranks and using one for the `IRecv` communication and one for the signal part.

Let me know what you would prefer.

albestro:
> If you indeed had some idea for verifying this behaviour with a test I think that'd be great, even if it may be somewhat complicated...

@msimberg @teonnik

@rasolca came up with an idea for the test, that we briefly implemented (and quickly checked some edge-cases by altering the implementation of `transformMPI`).

test/unit/communication/test_transform_mpi.cpp
rasolca: No need to duplicate in this case as only two messages that have a unique match are exchanged.

rasolca (approved):

rasolca:
bors try

test/unit/communication/test_transform_mpi.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    // "ensure" (by waiting a reasonable amount of time) that IRecv doesn't complete
+    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+    EXPECT_FALSE(sent_guard);
+
+    // Note:
+    // At this point we checked if everything went as expected in the posting phase. Let's signal
+    // Rank 1 that it can send back the message that will unlock IRecv.
+    const int signal = 13;
+    MPI_Send(&signal, 1, MPI_INT, 1, 0, world);
```
teonnik, line 102: I'm not sure I get this part. The `MPI_Send` is in standard mode, so it's implementation dependent whether the call will complete before the matching `MPI_Recv` is issued. The call may complete as soon as the data is buffered (but not yet sent). Perhaps use `MPI_Rsend`? this will start the `send` only when the matching `recv` is posted.

rasolca: Not sure I understand the question. This MPI_Send is just triggering the other rank to send the message that match the receive posted with trasformMPI. When it returns doesn't influence the test.

The timeline in the following:
```
RANK 0:                                            |  Rank 1:
- Post MPI_Irecv via tranformMPI (l. 67)           | - MPI_Recv (l. 123)
- check that the comm pipeline has unlocked (92)   |    |
  (as the MPI_Irecv can be immediately posted)     |    |
- check that the continuation of the transformMPI  |    |
  hasn't been unlocked (l. 96)                     |    |
- MPI_Send (l. 102)        ------------------------|--> Ok
(Match with previous Irecv) <----------------------| - MPI_Send (l. 129)
- check that the continuation is now unlocked (111)
```
Note that all checks are performed with a delay, as certain operations might not execute immediately.

rasolca:
bors try

PR #574, opened 2022-06-19
Add color to ninja output

teonnik:


albestro (commented):
LGTM.

I don't want to overcomplicate this, at least for the moment, with a check about the availability of the option. But, as suggestion, I would consider to refactor

https://github.com/eth-cscs/DLA-Future/blob/5a23516e72a2cff20a8672e5a5bb8f71809ae092/cmake/DLAF_AddTargetWarnings.cmake#L18-L20

so that these variables are usable elsewhere in the project, so we can centralise the management of compilers.

Moreover, there are inconsistencies in the formatting:
- `if` without space before parenthesis
- both `add_compile_options` with extra space before the parenthesis

CMakeLists.txt
```diff
@@ -36,6 +36,15 @@ option(DLAF_BUILD_TESTING "Build tests" ON)
 option(DLAF_BUILD_DOC "Build documentation" OFF)
 option(DLAF_WITH_PRECOMPILED_HEADERS "Use precompiled headers." OFF)
 
+# Add color to ninja output
+if(CMAKE_GENERATOR MATCHES "Ninja")
```
msimberg, line 40: Is this using `MATCHES` instead of `STREQUAL` to support `Ninja Multi-Config`?

msimberg (approved):
Nothing to add over @albestro's comments.

Regarding formatting in CMake files, we use https://github.com/cheshirekow/cmake_format for that in pika and enforce it in CI (just like clang-format). Would that be interesting to enable for DLA-Future as well?

rasolca (approved):

PR #578, opened 2022-06-21
add cmake-format

rasolca:
Added cmake-format as of @msimberg suggestion:
https://github.com/eth-cscs/DLA-Future/pull/574#pullrequestreview-1013091560


.cmake-format.py
```diff
@@ -0,0 +1,230 @@
...
+
+  # If true, lines are indented using tab characters (utf-8 0x09) instead of
+  # <tab_size> space characters (utf-8 0x20). In cases where the layout would
+  # require a fractional tab character, the behavior of the  fractional
+  # indentation is governed by <fractional_tab_policy>
+  use_tabchars = False
+
+  # If an argument group contains more than this many sub-groups (parg or kwarg
+  # groups) then force it to a vertical layout.
+  max_subgroups_hwrap = 2
```
msimberg, line 73:
```suggestion
  max_subgroups_hwrap = 3 (or 4)
```
Maybe? This makes some more function calls fit on one line. Not a must as this is just my preference though...

msimberg (approved):
Thanks @rasolca for putting this so quickly together! Looks good to me. One minor suggestion for the configuration options.

albestro (approved):
Nice hint @msimberg and well done @rasolca! Thanks!

minor: remember to manage the license header in the `.cmake_format.py` configuration file 😉

.cmake-format.py
msimberg: I prefer 3 over 2. With 4 that output is a bit surprising and ugly.

rasolca:
Quick question:
I chose no space between `if/else/endif` and `(` as it was the most common case (especially for else and endif). However as if had many case with space as well I would like to hear your opinions.

Issue #568, opened 2022-06-12
Some unit tests are too slow


teonnik:
Anything over 5-10s is IMO too much. The waiting time may discourage developers from running unit tests frequently enough. On my laptop with an Intel i7-8550U (8) @ 4.000GHz, the following tests take too long to execute:

```
43/51 Test #43: test_reduction_to_band ...........   Passed   78.29 sec
44/51 Test #44: test_bt_reduction_to_band ........   Passed   55.41 sec
45/51 Test #45: test_gen_to_std ..................   Passed   17.93 sec
46/51 Test #46: test_cholesky ....................   Passed   15.64 sec
47/51 Test #47: test_compute_t_factor ............   Passed   28.92 sec
49/51 Test #49: test_multiplication_triangular ...   Passed  110.95 sec
51/51 Test #51: test_triangular ..................   Passed  349.46 sec
```

msimberg:
I agree with the tests taking very long to finish. There may be multiple reasons for it, but I wonder if one of them is simply oversubscription. I think most of those tests run with 6 ranks, and probably an unconstrained number of threads. Just for comparison, could you try running e.g. `test_triangular` with `--pika:threads=2 --pika:bind=none`? There would still be oversubscription, but not as much so I would maybe expect the test to finish faster.

teonnik:
Yes, indeed, that helped. The test speed up with more than twice. For example:

`test_compute_t_factor  ~ 10s`
`test_triangular ~ 150s`

msimberg:
Do we have a way to restrict a unit test to only use e.g. 4 ranks (for your case of 8 cores), or similar?

rasolca:
The problem with the triangular solver and multiplication are the 24 different cases that has to be tested.
(left/right, upper/lover, non/trans/conj non/diag) on different grid (Note: distributed triangular multiplication doesn't support transposed and conj yet, therefore is faster)

Implementing a cmake flag to reduce 6 rank tests to 4 ranks should be easy, but it remove the most important test: a non square-grid with non trivial communicators in both dimension.

This issue can be linked with #557.
My idea is to split some of the tests (blas/lapack/dlaf algotithms) in two parts:
- a unit test with few representative fast tests (few grids and selected tests)
- a more intensive test with more cases (all grids and larger matrices)

Other possible TODOs:
- separate the local and the distributed tests of the dlaf algorithms (currently the local tests are executed simultaneously on all ranks)
[...]

PR #582, opened 2022-06-22
CMake MPI preset: use --pika:threads and --pika:bind with `plain-mpi` preset

albestro:
I started working on one of the suggestions available here https://github.com/eth-cscs/DLA-Future/issues/568#issuecomment-1155175874

In particular this one
> Other possible TODOs:
> [...]
> improve the CMAKE logic that setups test, to add --pika:threads=? --pika:bind=none when slurm is not available.

Main changes:
- when `DLAF_MPI_PRESET=slurm`, check that `DLAF_TEST_THREAD_BINDING_ENABLED=TRUE` otherwise just emits a warning message (no value change)
- when `DLAF_MPI_PRESET=plain-mpi`:
  - `DLAF_TEST_THREAD_BINDING_ENABLED=FALSE` (forced, with warning message in case it was set differently)
  - for MPIPIKA tests, add `--pika:threads=x`, where `x` is the the maximum  between 2 and the number of cores per rank (test-dependent, function of number of ranks)

Moreover, there is a trivial **silent** PIKA_EXTRA_ARGS preprocessor, that "overwrites" user provided arguments in case dlaf internal logic wants to set the same parameter.

This needs to be tested and refined, looking forward to your comments.

### DIFF
--- a/cmake/DLAF_AddTest.cmake
+++ b/cmake/DLAF_AddTest.cmake
@@ -51,7 +51,7 @@ function(_set_element_to_fallback_value LIST_NAME ELEMENT_REGEX FALLBACK)
   set(_TMP_LIST ${${LIST_NAME}})
   list(FILTER _TMP_LIST INCLUDE REGEX ${ELEMENT_REGEX})
   list(LENGTH _TMP_LIST _NUM_TMP_LIST)
-  if (_NUM_TMP_LIST EQUAL 0)
+  if(_NUM_TMP_LIST EQUAL 0)
     list(APPEND ${LIST_NAME} ${FALLBACK})
     set(${LIST_NAME} ${${LIST_NAME}} PARENT_SCOPE)
   endif()
@@ -170,7 +170,9 @@ function(DLAF_addTest test_target_name)
         set(_DLAF_PIKA_THREADS 2)
       endif()
 
-      _set_element_to_fallback_value(_PIKA_EXTRA_ARGS_LIST "--pika:threads" "--pika:threads=${_DLAF_PIKA_THREADS}")
+      _set_element_to_fallback_value(
+        _PIKA_EXTRA_ARGS_LIST "--pika:threads" "--pika:threads=${_DLAF_PIKA_THREADS}"
+      )
     endif()
 
     list(APPEND _TEST_ARGUMENTS ${_PIKA_EXTRA_ARGS_LIST})
--- a/cmake/DLAF_AddTest.cmake
+++ b/cmake/DLAF_AddTest.cmake
@@ -144,8 +144,27 @@ function(DLAF_addTest test_target_name)
   if(IS_AN_PIKA_TEST)
     separate_arguments(_PIKA_EXTRA_ARGS_LIST UNIX_COMMAND ${DLAF_PIKATEST_EXTRA_ARGS})
 
+    # APPLE platform does not support thread binding
+    if(NOT APPLE)
+      # TODO pika:bind=none?
+      list(FILTER _PIKA_EXTRA_ARGS_LIST EXCLUDE REGEX "--pika:use-process-mask")
+      list(APPEND _PIKA_EXTRA_ARGS_LIST "--pika:use-process-mask")
+    endif()
+
     if(NOT DLAF_TEST_THREAD_BINDING_ENABLED)
-      list(APPEND _TEST_ARGUMENTS "--pika:bind=none")
+      list(FILTER _PIKA_EXTRA_ARGS_LIST EXCLUDE REGEX "--pika:bind")
+      list(APPEND _PIKA_EXTRA_ARGS_LIST "--pika:bind=none")
+    endif()
+
+    if(IS_AN_MPI_TEST AND DLAF_MPI_PRESET STREQUAL "plain-mpi")
+      if(DLAF_CORE_PER_RANK GREATER_EQUAL 2)
+        set(_DLAF_PIKA_THREADS ${DLAF_CORE_PER_RANK})
+      else()
+        set(_DLAF_PIKA_THREADS 2)
+      endif()
+
+      list(FILTER _PIKA_EXTRA_ARGS_LIST EXCLUDE REGEX "--pika:threads")
+      list(APPEND _PIKA_EXTRA_ARGS_LIST "--pika:threads=${_DLAF_PIKA_THREADS}")
     endif()
 
     list(APPEND _TEST_ARGUMENTS ${_PIKA_EXTRA_ARGS_LIST})
--- a/cmake/DLAF_AddTest.cmake
+++ b/cmake/DLAF_AddTest.cmake
@@ -146,9 +146,8 @@ function(DLAF_addTest test_target_name)
 
     # APPLE platform does not support thread binding
     if(NOT APPLE)
-      # TODO pika:bind=none?
-      list(FILTER _PIKA_EXTRA_ARGS_LIST EXCLUDE REGEX "--pika:use-process-mask")
       list(APPEND _PIKA_EXTRA_ARGS_LIST "--pika:use-process-mask")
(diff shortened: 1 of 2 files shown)


cmake/DLAF_AddTest.cmake
```diff
@@ -141,11 +141,25 @@ function(DLAF_addTest test_target_name)
 
     # APPLE platform does not support thread binding
     if (NOT APPLE)
-      list(APPEND _TEST_ARGUMENTS "--pika:use-process-mask")
+      # TODO pika:bind=none?
+      list(FILTER _PIKA_EXTRA_ARGS_LIST EXCLUDE REGEX "--pika:use-process-mask")
```
albestro, line 145: check this with @msimberg

test/CMakeLists.txt
```diff
@@ -25,8 +27,18 @@ if (NOT DLAF_MPI_PRESET STREQUAL _DLAF_MPI_PRESET)
...
+    if (DLAF_TEST_THREAD_BINDING_ENABLED)
+      message(WARNING "Disabling pika binding")
+      set(DLAF_TEST_THREAD_BINDING_ENABLED FALSE CACHE BOOL "" FORCE)
+    endif()
 
   elseif(DLAF_MPI_PRESET STREQUAL "slurm")
+    if(NOT DLAF_TEST_THREAD_BINDING_ENABLED)
+      message(WARNING
+        "When using DLAF_MPI_PRESET=slurm this option should be enabled. "
+        "You may incur in performance drop. Do it are your own risk")
```
msimberg, line 39:
```suggestion
        "You may incur in performance drop. Do it on your own risk.")
```

cmake/DLAF_AddTest.cmake
msimberg: I'd leave this as is, i.e. not use `pika:bind=none`. Pasting my summary from slack:

> No, they're largely orthogonal, though I admit the `bind=none` case and macos is in practice not well tested and may do surprising things.

> `--pika:bind` controls how _individual_ threads should be bound to hardware, if at all. They can be bound to a numa domain, core, pu, or not at all.

> `--pika:(ignore/use)-process-mask` controls what resources are available to start with to the whole runtime.

> In theory I think `--pika:bind=none --pika:use-process-mask` would choose the number of threads for the runtime (assuming `--pika:threads` isn't also set) based on the process mask, but then not bind the threads to any particular thread. In practice I'm not sure it works exactly like that because it's badly tested (and an unusual case).

About the actual change here, is the idea to move `--pika:use-process-mask` last in the list? If yes, why?

cmake/DLAF_AddTest.cmake
```diff
@@ -141,11 +141,25 @@ function(DLAF_addTest test_target_name)
...
     endif()
 
     if(NOT DLAF_TEST_THREAD_BINDING_ENABLED)
-      list(APPEND _TEST_ARGUMENTS "--pika:bind=none")
+      list(FILTER _PIKA_EXTRA_ARGS_LIST EXCLUDE REGEX "--pika:bind")
+      list(APPEND _PIKA_EXTRA_ARGS_LIST "--pika:bind=none")
+    endif()
+
+    if(IS_AN_MPI_TEST AND DLAF_MPI_PRESET STREQUAL "plain-mpi")
+      if(DLAF_CORE_PER_RANK GREATER_EQUAL 2)
```
msimberg, line 155: Is `DLAF_CORE_PER_RANK` guaranteed to be set? Or does the comparison gracefully handle it not being set...?

msimberg (changes requested):

cmake/DLAF_AddTest.cmake
albestro: > About the actual change here, is the idea to move `--pika:use-process-mask` last in the list? If yes, why?

It is just a leftover because I copy-pasted the same mechanism for all options. The idea is that I remove the option in case the user already specified it, in order to overwrite the value with the one chosen by this logic. In this case there is no need, since the option does not have a value and it will result, as you said, in just moving it at the end of the list.

cmake/DLAF_AddTest.cmake
albestro: I would expect it to be set whenever it is an MPI test or the user asked to run all test with MPI

https://github.com/eth-cscs/DLA-Future/blob/fadbecf2e427a896a1cb12a9d203ea12535c5eb9/cmake/DLAF_AddTest.cmake#L114-L118

msimberg: Ok. I'm not sure I understand all the logic behind the MPI options, but what about when `MPIEXEC_NUMCORE_FLAG` isn't set? I guess in that case you can't set the number of cores per rank anyway... but is this check: https://github.com/eth-cscs/DLA-Future/blob/fadbecf2e427a896a1cb12a9d203ea12535c5eb9/cmake/DLAF_AddTest.cmake#L107 required here as well?

albestro: Yeah, actually you're right. `DLAF_CORE_PER_RANK` has the fallback to 1 just in that branch when `MPIEXEC_NUMCORE_FLAG` exists.

The logic about option `--pika:threads` applies just for `plain-mpi` preset, which doesn't have neither `MPIEXEC_NUMCORE_FLAG` nor `MPIEXEC_NUMCORES`. This means that we would always end up not having `DLAF_CORE_PER_RANK` set, but moreover we end up not having information about how many cores there are on the system.

Should we just go for setting it to 2 without any clever logic inspecting number of cores available?

@rasolca @msimberg @teonnik

rasolca: What about `MPIEXEC_MAX_NUMPROCS`? Apparently it defaults to the number of cores.

albestro:
I did the last couple of things missing. Please let me know if there is anything else needed.

test/CMakeLists.txt
```diff
@@ -35,8 +37,18 @@ if(NOT DLAF_MPI_PRESET STREQUAL _DLAF_MPI_PRESET)
           FORCE
     )
     set(MPIEXEC_NUMCORES "" CACHE STRING "Number of cores available for each MPI rank." FORCE)
+    if(DLAF_TEST_THREAD_BINDING_ENABLED)
+      message(WARNING "Disabling pika binding")
+      set(DLAF_TEST_THREAD_BINDING_ENABLED FALSE CACHE BOOL "" FORCE)
+    endif()
```
msimberg, line 43: I'm sorry if this was already discussed somewhere: is the idea that if `preset == plain-mpi` then that essentially implies a debug environment where one never threads to be bound? If I'm on a machine with enough cores but still using `plain-mpi` I probably still want to bind threads because tests will likely still finish faster (even if I may not care about absolute performance).

So my question is: would it be appropriate to set the _default_ of `DLAF_TEST_THREAD_BINDING_ENABLED` to off if the preset is `plain-mpi` but still allow it to be changed back to on?

msimberg (approved):
Just one more comment/question, but it's not a blocker for this.

test/CMakeLists.txt
rasolca: This case is quite complicated as in general MPI(*) doesn't setup correct rank binding and therefore the different ranks (as our test spawn multiple rank per node) would anyway be bound to the same cores.
So the binding in that case would be beneficial only to 1 RANK tests.

(*) Even if MPI takes care of the rank binding, IIRC currently there is no way to set the thread/rank flag in the `plain-mpi` case and the binding would be wrong.

msimberg: > in general

is the key point in my opinion, because MPI can do this (sometimes). But my question is also not very important in the bigger scheme. I was just curious as to why it's forced rather than only "preferred", but I'm not expecting changes here!

rasolca (approved):

teonnik (approved):

rasolca:
bors try-

Issue #468, opened 2022-02-08
Replace futures/promises with `any_sender` or custom senders


msimberg:
The "straightforward" (i.e. possibly simpler, until we try it...) change would be to use `async_rw_mutex` (or a variation of it) to replace the current mechanism. It works similarly to the current mechanism. We should, however, see if there are alternative mechanisms that are worth considering while doing the refactoring.

msimberg:
The last explicit use of `shared_future` is for `taus` in some of the algorithms. Changing this to use senders will also remove the last use of `keepFuture`.

msimberg:
This is now completed. There are no more uses of `pika::future`, `pika::shared_future`, or `pika::promise`.

Issue #605, opened 2022-07-13
Update `selectGeneric`, `select`, `selectRead` to use sender access


msimberg:
They currently use `operator()` and `read` on `Matrix` instead of the sender equivalents. Updating this to use the sender versions means that one can't anymore query whether a sender is "ready" or not. Some tests rely on this functionality. It's convenient for testing to sanity check if tile senders/futures are ready or not, but it's not clear if there's are use cases outside of testing and the functionality should be kept in some form. Alternatively, can the tests be refactored not to rely on `is_ready`?

Part of #468. #564 is also related since it added helpers that look like `select` and `selectRead` but use the sender accessors: https://github.com/eth-cscs/DLA-Future/blob/9dd8678478581f3dd261fb1bb45bc6e20a22e8db/include/dlaf/util_matrix.h#L390-L410. When updating `select*` the `collect*` helpers should be removed.

albestro:
Trying to highlight main points of the problem blocking from switching to senders.

`test_matrix` and `test_matrix_view`, among other things, test if the tile "pipeline" mechanism works correctly by holding/releasing future<Tile> in a controlled way.

The main idea is that the test gets a specific sequence of `future<Tile>`, building up a DAG of dependencies, and step by step by releasing "one-by-one" the dependencies in the aforementioned sequence, it verifies that the DAG status is what is supposed to be (future is ready or not).

The tile mechanism is planned to be redesigned and rethought. However, if the general behaviour will still be valid, we can think about a possible solution already starting from now.

The more trivially and basic solution would be to implement something along the lines of what we did for `test_transfor_mpi`

https://github.com/eth-cscs/DLA-Future/blob/2258862a09350d0944add52eaeac626ec35b96ad/test/unit/communication/test_transform_mpi.cpp#L85-L92

It is not exactly the same behaviour, but probably is compatible with what we would like to test. We may try to think about something smarter, since in this case we would try to replicate the future mechanism with senders, requiring to keep the "status" of the sender in an atomic variable.

msimberg:
This was done in #766.

PR #606, opened 2022-07-13
Use `unique_any_sender` in `reduction_to_band`

msimberg:
Another tiny step towards #468.

### DIFF
--- a/include/dlaf/eigensolver/reduction_to_band/impl.h
+++ b/include/dlaf/eigensolver/reduction_to_band/impl.h
@@ -620,9 +620,9 @@ void updateTrailingPanelWithReflector(const bool has_head, comm::Communicator& c
   updateTrailingPanel(has_head, panel, j, w, tau);
 }
 
-template <class CommSender, class T>
+template <class TriggerSender, class CommSender, class T>
 pika::shared_future<common::internal::vector<T>> computePanelReflectors(
-    pika::future<void> trigger, comm::IndexT_MPI rank_v0, CommSender&& mpi_col_chain_panel,
+    TriggerSender&& trigger, comm::IndexT_MPI rank_v0, CommSender&& mpi_col_chain_panel,
     MatrixT<T>& mat_a, const common::IterableRange2D<SizeType, matrix::LocalTile_TAG> ai_panel_range,
     SizeType nrefls) {
   namespace ex = pika::execution::experimental;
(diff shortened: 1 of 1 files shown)


include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -1125,7 +1126,8 @@ common::internal::vector<pika::shared_future<common::internal::vector<T>>> Reduc
     // blocked waiting to do nothing on the next iteration (computePanel), while other ranks would be
     // stuck waiting for it for completing steps of the previous iteration, needed for the update of the
     // trailing matrix that will unlock the next iteration.
-    trigger_panel = pika::when_all(selectRead(x, x.iteratorLocal()), selectRead(xt, xt.iteratorLocal()));
+    trigger_panel = pika::future<void>(
```
albestro, line 1129: Just checking my understanding: could `drop_value` be used in place of `future<void>`?

Moreover in general I'm wondering if we could switch to something more "senderified", but probably we would need to use `selectRead` @teonnik variant (I remember of something he may have done in #436)

albestro (approved):
LGTM, just a minor comment.

include/dlaf/eigensolver/reduction_to_band/impl.h
msimberg: > Just checking my understanding: could `drop_value` be used in place of `future<void>`?

Actually, yes, that's a good point. The `select*` functions are the blocker for full senderification, but in fact I think this could be further senderified already now without that. I'll push some changes for that.
 
> Moreover in general I'm wondering if we could switch to something more "senderified", but probably we would need to use `selectRead` @teonnik variant (I remember of something he may have done in #436)

Yes, it should be updated. This issue tracks that: https://github.com/eth-cscs/DLA-Future/issues/605. The main blocker is that some tests rely on checking if a future is ready or not. Please comment over there if you have thoughts or ideas.

msimberg:
bors try

rasolca:
bors try

include/dlaf/eigensolver/reduction_to_band/impl.h
msimberg: In theory `pika::execution::experimental::{when_all,when_all_vector,drop_value}` could've been used here, but because of the usual problem of `Tile`s not being copyable from `shared_future`s without `keep_future` I've reverted the blind change. This is better updated as part of #605.

msimberg:
bors try-

msimberg:
bors try

PR #617, opened 2022-08-02
Tridiagonal Eigensolver on GPUs

teonnik:


include/dlaf/eigensolver/tridiag_solver/mc.h
```diff
@@ -203,53 +197,49 @@ void offloadDiagonal(Matrix<const T, Device::CPU>& mat_trd, Matrix<T, Device::CP
...
+
+  WorkSpaceHostMirror<T, device> ws_h{matrix::MatrixMirror<T, Device::CPU, device>(evals),     // evals
+                                      matrix::MatrixMirror<T, Device::CPU, device>(evecs),     // evecs
+                                      matrix::MatrixMirror<T, Device::CPU, device>(ws.mat1),   // mat1
+                                      matrix::MatrixMirror<T, Device::CPU, device>(ws.mat2)};  // mat2
+
   // Set `mat_ev` to `zero` (needed for Given's rotation to make sure no random values are picked up)
-  matrix::util::set0<Backend::MC, T, Device::CPU>(pika::execution::thread_priority::normal, mat_ev);
+  matrix::util::set0<backend, T, device>(pika::execution::thread_priority::normal, evecs);
+  ws_h.evecs.copySourceToTarget();  // copy from GPU to CPU if evecs is on GPU
```
msimberg, line 226: Minor optimization opportunity: `MatrixMirror` currently copies the contents from the source to the target on construction. If I understand this correctly, it would not be necessary in this particular situation because you are anyway operating on the source matrix first, and then copying the contents to the target.

Could you move the construction of `ws_h` to just below the `set0` call here? It looks like that would be sufficient for your use case and avoid the additional copy at the same time.

include/dlaf/eigensolver/tridiag_solver.h
```diff
@@ -64,23 +64,28 @@ void tridiagSolver(Matrix<T, device>& mat_trd, Matrix<T, device>& d, Matrix<T, d
...
+        for (auto el_idx : iterate_range2d(out.size())) {
+          out(el_idx) = std::complex<T>(in(el_idx), 0);
+        }
+      };
+
+      pika::execution::experimental::when_all(mat_real_ev_h.get().read_sender(tile_wrt_local),
+                                              mat_ev_h.get().readwrite_sender(tile_wrt_local)) |
+          dlaf::internal::transformDetach(dlaf::internal::Policy<Backend::MC>(),
+                                          std::move(convert_to_complex_fn));
+    }
```
msimberg, line 88: This looks like something that should be possible to do easily on the GPU. Am I missing something bigger that blocks having this on the GPU, or is it just something you haven't gotten to yet? In either case, not a blocker for this PR.

include/dlaf/eigensolver/tridiag_solver/mc.h
```diff
@@ -203,53 +197,49 @@ void offloadDiagonal(Matrix<const T, Device::CPU>& mat_trd, Matrix<T, Device::CP
...
-                  Matrix<SizeType, Device::CPU>(vec_size, vec_tile_size),
-                  Matrix<SizeType, Device::CPU>(vec_size, vec_tile_size),
-                  Matrix<ColType, Device::CPU>(vec_size, vec_tile_size)};
+  offloadDiagonal(tridiag_h.get(), ws_h.evals.get());
 
   // Each triad represents two subproblems to be merged
   for (auto [i_begin, i_split, i_end] : generateSubproblemIndices(distr.nrTiles().rows())) {
-    mergeSubproblems(i_begin, i_split, i_end, ws, offdiag_vals[to_sizet(i_split)], d, mat_ev);
+    mergeSubproblems<backend>(i_begin, i_split, i_end, offdiag_vals[to_sizet(i_split)], ws, ws_h, evals,
+                              evecs);
```
msimberg, line 241: Do `evals` and `evecs` not need to be copied back to the source before doing this? Possibly by scoping `ws_h` to go out of scope before reaching this for loop? Or is the synchronization done somewhere else implicitly?

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -124,6 +125,17 @@ struct WorkSpace {
   Matrix<ColType, Device::CPU> c;
 };
 
+template <class T, Device device>
+struct WorkSpaceHostMirror {
+  // Mirror to eigenvalues and eigenvectors on host memory
+  matrix::MatrixMirror<T, Device::CPU, device> evals;
+  matrix::MatrixMirror<T, Device::CPU, device> evecs;
+
+  // Mirrors to auxiliary matrices on the CPU
```
msimberg, line 134:
```suggestion
  // Mirror to eigenvalues and eigenvectors on host memory
  matrix::MatrixMirror<T, Device::CPU, device> evals;
  matrix::MatrixMirror<T, Device::CPU, device> evecs;

  // Mirrors to auxiliary matrices on host memory
```
or
```suggestion
  // Mirror to eigenvalues and eigenvectors on the CPU
  matrix::MatrixMirror<T, Device::CPU, device> evals;
  matrix::MatrixMirror<T, Device::CPU, device> evecs;

  // Mirrors to auxiliary matrices on the CPU
```
?

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -291,24 +303,15 @@ struct TileCollector {
...
-    for (SizeType i = 0; i < rows; ++i) {
-      TileElementIndex idx(i, 0);
-      out(idx) = in(idx);
-    }
-  };
   for (SizeType i = i_begin; i <= i_end; ++i) {
     GlobalTileIndex idx(i, 0);
     ex::when_all(in.read_sender(idx), out.readwrite_sender(idx)) |
-        di::transform(di::Policy<Backend::MC>(), copy_fn) | ex::start_detached();
+        di::transform(di::Policy<backend>(), matrix::internal::copy_o) | ex::start_detached();
```
msimberg, line 314:
```suggestion
        dlaf::tile::copy(di::Policy<CopyBackend_v<...>>()) | ex::start_detached();
```
? Or does that not work?

include/dlaf/eigensolver/tridiag_solver/mc.h
msimberg: Hopefully answering my own question after reading further: the synchronization is done inside `mergeSubproblems`.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -1004,8 +1012,8 @@ void mergeSubproblems(SizeType i_begin, SizeType i_split, SizeType i_end, WorkSp
   // - reorder columns in `mat_ev` using `i2` such that eigenvectors match eigenvalues
   //
   sortIndex(i_begin, i_end, k_fut, ws.dtmp, ws.i1, ws.i2);
-  applyIndex(i_begin, i_end, ws.i2, ws.dtmp, d);
-  dlaf::permutations::permute<Backend::MC, Device::CPU, T, Coord::Col>(i_begin, i_end, ws.i2, ws.mat1,
-                                                                       mat_ev);
+  applyIndex(i_begin, i_end, ws.i2, ws.dtmp, ws_h.evals.get());
+  dlaf::permutations::permute<backend, device, T, Coord::Col>(i_begin, i_end, ws.i2, ws.mat1, evecs);
```
msimberg, line 1016: Would it be useful to have `Matrix<T, Source> MatrixMirror::get_source()` so that you don't need to pass the original `evecs` into this function?

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -1004,8 +1012,8 @@ void mergeSubproblems(SizeType i_begin, SizeType i_split, SizeType i_end, WorkSp
...
+  ws_h.evecs.copySourceToTarget();  // copy from GPU to CPU if evecs in on GPU
```
msimberg, line 1017: Is this required because subsequent iterations may access target-side values that this iteration wrote into the source-side of `evecs`?

msimberg (commented):
Mostly some general questions at this point.

include/dlaf/eigensolver/tridiag_solver.h
teonnik: Yes, indeed, I just hadn't gotten to it yet. It should be in by the end of the week!

msimberg: Sounds good. Then I'm leaving this unresolved since it sounds like you'll be adding it to this PR.

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: Yes, I think it will be useful!

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: Yes, exactly. This is a temporary situation as more kernels are moved to the GPU.

include/dlaf/eigensolver/tridiag_solver/merge.h
msimberg: See https://github.com/eth-cscs/DLA-Future/pull/620. Does not have to be included in this PR.

include/dlaf/eigensolver/tridiag_solver.h
teonnik: This now done on the GPU with : https://github.com/eth-cscs/DLA-Future/pull/617/commits/b5bb0e5c2c5da3381d8959e4ad1b001da797a67a

include/dlaf/eigensolver/tridiag_solver/mc.h
teonnik: https://github.com/eth-cscs/DLA-Future/pull/617/commits/ff8ad699fd042f36dbc2834b29e12072fd8d8f24

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: https://github.com/eth-cscs/DLA-Future/pull/617/commits/ff8ad699fd042f36dbc2834b29e12072fd8d8f24

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -201,26 +196,52 @@ void assembleZVec(SizeType i_begin, SizeType i_split, SizeType i_end, pika::shar
...
+      if constexpr (D == Device::CPU) {
+        for (SizeType i = 0; i < tile.size().cols(); ++i) {
+          col(TileElementIndex(i, 0)) = sign * tile(TileElementIndex(row, i)) / T(std::sqrt(2));
+        }
+      }
+      else {
+        copyTileRowAndNormalizeOnDevice(sign, tile.size().cols(), tile.ld(),
+                                        tile.ptr(TileElementIndex(row, 0)), col.ptr(), ts...);
+      }
+    };
```
msimberg, line 220: Same comment here about dispatching. Even adding a `copyTileRowAndNormalizeOnDevice` overload without a stream argument would be sufficient, i.e. to allow something like:
```suggestion
    auto copy_fn = [top_tile](const auto& rho, const auto& tile, const auto& col,
                              [[maybe_unused]] auto&&... ts) {
      // Copy the bottom row of the top tile or the top row of the bottom tile
      SizeType row = (top_tile) ? col.size().rows() - 1 : 0;

      // Negate Q1's last row if rho < 0
      //
      // lapack 3.10.0, dlaed2.f, line 280 and 281
      int sign = (top_tile && rho < 0) ? -1 : 1;
      copyTileRowAndNormalizeOnDevice(sign, tile.size().cols(), tile.ld(),
                                      tile.ptr(TileElementIndex(row, 0)), col.ptr(), ts...);
      }
    };
```
I think just adding non-stream overloads to the other places with the same comment would also be an option.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -336,81 +349,101 @@ void sortIndex(SizeType i_begin, SizeType i_end, pika::shared_future<SizeType> k
...
+    if constexpr (D == Device::CPU) {
+      auto cmp = [v_ptr](SizeType i1, SizeType i2) { return v_ptr[i1] < v_ptr[i2]; };
+      pika::merge(pika::execution::par, begin_it, split_it, split_it, end_it, out_index_ptr,
+                  std::move(cmp));
+    }
+    else {
+#ifdef DLAF_WITH_GPU
+      mergeIndicesOnDevice(begin_it, split_it, end_it, out_index_ptr, v_ptr, ts...);
+#endif
+    }
```
msimberg, line 361: Same dispatching comment.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -336,81 +349,101 @@ void sortIndex(SizeType i_begin, SizeType i_end, pika::shared_future<SizeType> k
...
+      for (SizeType i = 0; i < n; ++i) {
+        out_ptr[i] = in_ptr[i_ptr[i]];
+      }
+    }
+    else {
+#ifdef DLAF_WITH_GPU
+      applyIndexOnDevice(n, i_ptr, in_ptr, out_ptr, ts...);
+#endif
     }
   };
```
msimberg, line 403: Dispatching.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -336,81 +349,101 @@ void sortIndex(SizeType i_begin, SizeType i_end, pika::shared_future<SizeType> k
...
-      out_ptr[in_ptr[i]] = i;
+    if constexpr (D == Device::CPU) {
+      for (SizeType i = 0; i < n; ++i) {
+        out_ptr[in_ptr[i]] = i;
+      }
+    }
+    else {
+      invertIndexOnDevice(n, in_ptr, out_ptr, ts...);
     }
   };
```
msimberg, line 434: Dispatching.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -458,28 +491,30 @@ inline pika::future<SizeType> stablePartitionIndexForDeflation(SizeType i_begin,
...
+          ct_tile(TileElementIndex(i, 0)) = val;
+        }
+      }
+      else {
+        setColTypeTile(val, ct_tile.size().rows(), ct_tile.ptr(), ts...);
+      }
+    };
+
+    ex::start_detached(
+        di::transform<di::TransformDispatchType::Plain>(di::Policy<DefaultBackend<D>::value>(),
```
msimberg, line 515: Side note, `TransformDispatchType::Plain` is the default so if you want you can leave it out here and in many other places (except if you need to disable unwrapping).

src/eigensolver/tridiag_solver/misc_gpu_kernels.cu
```diff
@@ -0,0 +1,241 @@
...
+
+// https://github.com/NVIDIA/thrust/issues/1515
+//
+template <class T>
+void mergeIndicesOnDevice(const SizeType* begin_ptr, const SizeType* split_ptr, const SizeType* end_ptr,
+                          SizeType* out_ptr, const T* v_ptr, cudaStream_t stream) {
+  auto cmp = [v_ptr] __device__(const SizeType& i1, const SizeType& i2) {
+    return v_ptr[i1] < v_ptr[i2];
+  };
+  // NOTE: The call may be synchronous, to avoid that either wrap in a __global__ function as shown in
```
msimberg, line 30: Is the idea with this that `thrust::merge` has a device-callable implementation as well, so that gives you control of the launch and synchronization instead?

I'd say we go with the current implementation even it may be blocking, open an issue about updating it, profile it and fix it as needed?

msimberg (commented):
This looks good to me on a high level.

Note that I'm not sure if my comments about dispatching are feasible/make sense. You may have requirements that make the current state a better option than what I'm suggesting.

include/dlaf/eigensolver/tridiag_solver/mc.h
```diff
@@ -122,28 +117,28 @@ void solveLeaf(Matrix<T, Device::CPU>& mat_trd, Matrix<T, Device::CPU>& mat_ev)
...
+    auto diag_fn = [](const auto& tridiag_tile, const auto& diag_tile, [[maybe_unused]] auto&&... ts) {
+      if constexpr (D == Device::CPU) {
+        for (SizeType i = 0; i < tridiag_tile.size().rows(); ++i) {
+          diag_tile(TileElementIndex(i, 0)) = tridiag_tile(TileElementIndex(i, 0));
+        }
+      }
+      else {
+        copyDiagTileFromTridiagTile(tridiag_tile.size().rows(), tridiag_tile.ptr(), diag_tile.ptr(),
+                                    ts...);
+      }
```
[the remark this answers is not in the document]
teonnik: > this copies the diagnoal from one tile to another?
Yes.

> Would it be worth adding a copyDiagFromTridiagTile helper for the CPU case as well and then use the various macros to create all the needed overloads for this?

Yes, it would be worth adding, not just for this one but also for other custom kernels in `misg_gpu_kernels.h`. I plan to do that after the performance benchmarking. Either in this PR or in a new PR.

include/dlaf/eigensolver/tridiag_solver/merge.h
msimberg: @teonnik note that #620 was merged so you should be able to use `getSource` now. However, I think the code has changed enough in the meantime that I can't quite tell if you need it anymore...

src/eigensolver/tridiag_solver/misc_gpu_kernels.cu
```diff
@@ -0,0 +1,613 @@
...
+DLAF_CUDA_CAST_TO_COMPLEX(, double);
+
+constexpr unsigned invert_index_kernel_sz = 256;
+
+__global__ void invertIndexOnDevice(SizeType len, const SizeType* in, SizeType* out) {
+  const SizeType i = blockIdx.x * invert_index_kernel_sz + threadIdx.x;
+  if (i >= len)
+    return;
+
+  out[in[i]] = i;
```
rasolca, line 140: Warning: non coalesced memory access.

teonnik:
bors try

teonnik:
bors try

teonnik:
bors try

teonnik:
bors try

include/dlaf/eigensolver/tridiag_solver/mc.h
```diff
@@ -203,53 +198,101 @@ void offloadDiagonal(Matrix<const T, Device::CPU>& mat_trd, Matrix<T, Device::CP
...
+      matrix::MatrixMirror<T, Device::CPU, device>(ws.mat1),       // mat1
+      matrix::MatrixMirror<T, Device::CPU, device>(ws.mat2),       // mat2
+      matrix::MatrixMirror<T, Device::CPU, device>(ws.dtmp),       // dtmp
+      matrix::MatrixMirror<T, Device::CPU, device>(ws.z),          // z
+      matrix::MatrixMirror<T, Device::CPU, device>(ws.ztmp),       // ztmp
+      matrix::MatrixMirror<SizeType, Device::CPU, device>(ws.i1),  // i1
+      matrix::MatrixMirror<SizeType, Device::CPU, device>(ws.i2),  // i2
+      matrix::MatrixMirror<SizeType, Device::CPU, device>(ws.i3),  // i3
+      matrix::MatrixMirror<ColType, Device::CPU, device>(ws.c)     // c
+  };
```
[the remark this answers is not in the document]
teonnik: Yes, you are right. However, I have now remove `MatrixMirror<>` from this part of the code and the initialization is different.

include/dlaf/eigensolver/tridiag_solver/mc.h
```diff
@@ -203,53 +198,101 @@ void offloadDiagonal(Matrix<const T, Device::CPU>& mat_trd, Matrix<T, Device::CP
...
+  TileElementSize vec_tile_size(distr.blockSize().rows(), 1);
+  WorkSpace<T, device> ws{Matrix<T, device>(distr),                           // mat1
+                          Matrix<T, device>(distr),                           // mat2
+                          Matrix<T, device>(vec_size, vec_tile_size),         // dtmp
+                          Matrix<T, device>(vec_size, vec_tile_size),         // z
+                          Matrix<T, device>(vec_size, vec_tile_size),         // ztmp
+                          Matrix<SizeType, device>(vec_size, vec_tile_size),  // i1
+                          Matrix<SizeType, device>(vec_size, vec_tile_size),  // i2
+                          Matrix<SizeType, device>(vec_size, vec_tile_size),  // i3
+                          Matrix<ColType, device>(vec_size, vec_tile_size)};  // c
```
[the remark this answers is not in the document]
teonnik: I find the code clearer with explicit constructors but if you prefer implicit constructors, I can change it.

include/dlaf/eigensolver/tridiag_solver/mc.h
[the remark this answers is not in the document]
teonnik: I created an issue about this here : https://github.com/eth-cscs/DLA-Future/issues/634 . I am not sure if it would be better addressed in a separate PR.

include/dlaf/eigensolver/tridiag_solver/mc.h
```diff
@@ -203,53 +198,101 @@ void offloadDiagonal(Matrix<const T, Device::CPU>& mat_trd, Matrix<T, Device::CP
...
+      namespace ex = pika::execution::experimental;
+      namespace di = dlaf::internal;
+      ex::start_detached(di::transform<di::TransformDispatchType::Plain>(
+          dlaf::internal::Policy<Backend::GPU>(),
+          [](const matrix::Tile<const T, Device::GPU>& in,
+             const matrix::Tile<std::complex<T>, Device::GPU>& out, cudaStream_t stream) {
+            castTileToComplex(in.size().rows(), in.size().cols(), in.ld(), in.ptr(), out.ptr(), stream);
+          },
+          std::move(sender)));
 #endif
```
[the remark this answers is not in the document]
teonnik: Relevant issue : https://github.com/eth-cscs/DLA-Future/issues/634

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -131,66 +147,45 @@ inline SizeType problemSize(SizeType i_begin, SizeType i_end, const matrix::Dist
...
+    auto init_fn = [tile_row](const auto& index_tile, [[maybe_unused]] auto&&... ts) {
+      if constexpr (D == Device::CPU) {
+        for (SizeType i = 0; i < index_tile.size().rows(); ++i) {
+          index_tile(TileElementIndex(i, 0)) = tile_row + i;
+        }
+      }
+      else {
+        initIndexTile(tile_row, index_tile.size().rows(), index_tile.ptr(), ts...);
+      }
+    };
```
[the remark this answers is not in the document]
teonnik: Relevant issue: https://github.com/eth-cscs/DLA-Future/issues/634

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: Relevant issue: https://github.com/eth-cscs/DLA-Future/issues/634

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: Relevant issue: https://github.com/eth-cscs/DLA-Future/issues/634

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: Relevant issue: https://github.com/eth-cscs/DLA-Future/issues/634

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: Relevant issue: https://github.com/eth-cscs/DLA-Future/issues/634

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: https://github.com/eth-cscs/DLA-Future/pull/617/commits/3a0c23bdae5064dd60f15e52074825dd14ef270b

include/dlaf/eigensolver/tridiag_solver/mc.h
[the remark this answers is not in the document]
msimberg: Separate PR is probably sanest, thanks!

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -115,82 +109,92 @@ struct WorkSpace {
...
+template <class T>
+Matrix<T, Device::CPU> initMirrorMatrix(Matrix<T, Device::GPU>& mat) {
+  return Matrix<T, Device::CPU>(mat.distribution());
+}
+
+template <class T>
+Matrix<T, Device::CPU>& initMirrorMatrix(Matrix<T, Device::CPU>& mat) {
+  return mat;
+}
+
```
msimberg, line 153: I'm just trying to understand this change. Did you stop using `MatrixMirror` because it didn't give you enough control over when copies are made or something else? Just asking in case we should consider further changes to `MatrixMirror` or if it's unnecessary since the mirroring here may be removed soon enough once everything is on the GPU?

include/dlaf/eigensolver/tridiag_solver/mc.h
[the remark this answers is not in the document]
msimberg: No, all good as it is. I was more concerned about the other case and this comment mostly tagged along for the ride. I was a bit on the fence about it anyway.

include/dlaf/eigensolver/tridiag_solver/mc.h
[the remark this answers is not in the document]
msimberg: I contined the `MatrixMirror` topic (by mistake) in another comment...

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: > Did you stop using MatrixMirror because it didn't give you enough control over when copies are made or something else? 
 
 Yes. Essentially, I only needed copies for a subset of tiles [i_begin, i_end] and I didn't need a copy upon destruction.

> Just asking in case we should consider further changes to MatrixMirror or if it's unnecessary since the mirroring here may be removed soon enough once everything is on the GPU?

I don't think we need to add new functionality to MatrixMirror at the moment.

src/eigensolver/tridiag_solver/misc_gpu_kernels.cu
msimberg: @teonnik provided more details, and more details here: https://github.com/NVIDIA/thrust/blob/main/examples/cuda/async_reduce.cu. This comment was only to increase my understanding of the potential workaround and does not block this PR.

teonnik: I opened an issue: https://github.com/eth-cscs/DLA-Future/issues/636.

The comment is inspired by an example in Thrust : thrust/examples/cuda/async_reduce.h .However, the approach of wrapping the kernel in a `__global__` function may now be outdated as it appears to rely on CUDA dynamic parallelism. This has to be investigated further.

include/dlaf/eigensolver/tridiag_solver.h
```diff
@@ -33,55 +33,34 @@ namespace eigensolver {
...
-void tridiagSolver(Matrix<T, device>& mat_trd, Matrix<T, device>& d, Matrix<T, device>& mat_ev) {
-  static_assert(std::is_same<T, float>::value || std::is_same<T, double>::value,
-                "tridagSolver accepts only real values (float, double)!");
+template <Backend backend, Device device, class T, class CT>
+void tridiagSolver(Matrix<T, device>& tridiag, Matrix<T, device>& evals, Matrix<CT, device>& evecs) {
+  static_assert(std::is_floating_point_v<T>,
+                "The tridiagonal matrix and the resulting eigenvalues must be real values (float, double)!");
+  static_assert(
+      std::is_same<T, CT>::value || std::is_same<std::complex<T>, CT>::value,
+      "Eigenvectors must either match the floating point type of the eigenvalues or be of the corresponding complex type (std::complex<>).");
```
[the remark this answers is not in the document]
teonnik: https://github.com/eth-cscs/DLA-Future/pull/617/commits/b0ec9ce67259cce4eb0d85898993b9dc8faa52ac

include/dlaf/eigensolver/tridiag_solver/mc.h
```diff
@@ -98,53 +71,100 @@ std::vector<pika::shared_future<T>> cuppensDecomposition(Matrix<T, Device::CPU>&
...
-                   tile::stedc(Policy<Backend::MC>(thread_priority::normal)));
-  }
-}
+    if constexpr (D == Device::CPU) {
+      whenAllLift(mat_trd.readwrite_sender(LocalTileIndex(i, 0)),
+                  mat_ev.readwrite_sender(LocalTileIndex(i, i))) |
+          tile::stedc(Policy<Backend::MC>(thread_priority::normal)) | start_detached();
+    }
+    else {
+#ifdef DLAF_WITH_CUDA
```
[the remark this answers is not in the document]
teonnik: Currently, no! I will try to have the HIP implementation in soon. Is this OK for now : https://github.com/eth-cscs/DLA-Future/pull/617/commits/4b2ffe6f4223f8cfcb360fab8fd8244f97fbffa9 ?

src/eigensolver/tridiag_solver/misc_gpu_kernels.cu
teonnik: I opened an issue about this here : https://github.com/eth-cscs/DLA-Future/issues/637 .

teonnik:
bors try

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -115,82 +109,92 @@ struct WorkSpace {
...
+template <class T>
+struct WorkSpaceHostMirror<T, Device::CPU> {
+  matrix::Matrix<T, Device::CPU>& evals;
+  matrix::Matrix<T, Device::CPU>& mat1;
+  matrix::Matrix<T, Device::CPU>& dtmp;
+  matrix::Matrix<T, Device::CPU>& z;
+  matrix::Matrix<T, Device::CPU>& ztmp;
+  matrix::Matrix<SizeType, Device::CPU>& i2;
+  matrix::Matrix<ColType, Device::CPU>& c;
```
albestro, line 141: Is this a duplication? Is this needed?

test/unit/eigensolver/test_tridiag_solver.cpp
```diff
@@ -241,18 +233,41 @@ void solveRandomTridiagMatrix(SizeType n, SizeType nb) {
...
 const std::vector<std::tuple<SizeType, SizeType>> tested_problems = {
     // n, nb
-    {0, 8}, {16, 8}, {16, 4}, {16, 5}, {100, 10}, {93, 7}};
+    {0, 8},
+    {16, 16},
+    {16, 8},
+    {16, 4},
+    {16, 5},
+    {100, 10},
+    //{93, 7},
```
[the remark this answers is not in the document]
teonnik: There is currently a problem with this setup that I am investigating. Occasionally it segfaults. I have't found the reason why yet.

src/permutations/general/perms.cu
```diff
@@ -0,0 +1,128 @@
...
+
+constexpr unsigned perms_kernel_sz = 32;
+
+template <Coord coord>
+constexpr SizeType getCoord(SizeType row, SizeType col) {
+  if constexpr (coord == Coord::Row)
+    return row;
+  else
+    return col;
+}
```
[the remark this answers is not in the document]
teonnik: Good catch. This was a leftover from a previous implementation!

src/eigensolver/tridiag_solver/gpu.cpp
```diff
@@ -0,0 +1,18 @@
...
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2022, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#include "dlaf/eigensolver/tridiag_solver/mc.h"
```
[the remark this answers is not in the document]
teonnik: https://github.com/eth-cscs/DLA-Future/pull/617/commits/cf67f39f356625d2c7c0efd05c4b12edd6ccf74c

test/unit/eigensolver/test_tridiag_solver.cpp
```diff
@@ -9,23 +9,29 @@
...
 #include "dlaf_test/util_types.h"
 
-template <typename Type>
-class TridiagEigensolverTest : public ::testing::Test {};
+using namespace dlaf;
+using namespace dlaf::test;
 
 template <typename Type>
 class CuppensTest : public ::testing::Test {};
+TYPED_TEST_SUITE(CuppensTest, RealMatrixElementTypes);
```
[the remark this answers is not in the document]
teonnik: https://github.com/eth-cscs/DLA-Future/pull/617/commits/a4d74bbb19cd8edbae2fcd927219b03eba84d61d

src/permutations/general/perms.cu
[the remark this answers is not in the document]
teonnik: https://github.com/eth-cscs/DLA-Future/pull/617/commits/ac21c2711261d2993b27fbd7cd17ae96045b5752

include/dlaf/permutations/general/impl.h
```diff
@@ -79,50 +80,48 @@ void applyPermutations(GlobalElementIndex out_begin, GlobalElementSize sz, SizeT
...
+        auto& tile_out = out_tiles[to_sizet(distr.globalTileLinearIndex(i_split_gl_out))];
 
-      if constexpr (D == Device::CPU) {
         dlaf::tile::lacpy<T>(region, i_subtile_in, tile_in, i_subtile_out, tile_out);
       }
-      else if constexpr (D == Device::GPU) {
+    });
+  }
+  else if constexpr (D == Device::GPU) {
 #if defined(DLAF_WITH_CUDA)
```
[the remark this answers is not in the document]
teonnik: Yes! Thanks for pointing this out. The same issue was also present in test_multiplication_general.cpp.

https://github.com/eth-cscs/DLA-Future/pull/617/commits/4a9c127fd1bb4bef2fa4168b72ceb3497c23c848

include/dlaf/eigensolver/tridiag_solver/merge.h
teonnik: When on CPU, these just serve as references to the original data. This is similar to how MatrixMirror works, except that I can use these to copy subranges of tiles for each matrix and there is no copy upon destruction. Note that no copy is performed if two references point to the same object as is the case on CPU. The copies only happen when executing on the GPU.

include/dlaf/eigensolver/tridiag_solver/mc.h
```diff
@@ -103,53 +76,99 @@ std::vector<pika::shared_future<T>> cuppensDecomposition(Matrix<T, Device::CPU>&
...
-
-template <class T>
-void copyDiagTile(const matrix::Tile<const T, Device::CPU>& tridiag_tile,
-                  const matrix::Tile<T, Device::CPU>& diag_tile) {
-  for (SizeType i = 0; i < tridiag_tile.size().rows(); ++i) {
-    diag_tile(TileElementIndex(i, 0)) = tridiag_tile(TileElementIndex(i, 0));
+    if constexpr (D == Device::CPU) {
+      whenAllLift(mat_trd.readwrite_sender(LocalTileIndex(i, 0)),
+                  mat_ev.readwrite_sender(LocalTileIndex(i, i))) |
+          tile::stedc(Policy<Backend::MC>(thread_priority::normal)) | start_detached();
```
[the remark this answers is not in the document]
teonnik: https://github.com/eth-cscs/DLA-Future/pull/617/commits/6b4e10da7c1cbbcf2707f6ac767d3752544addc9

msimberg:
Quick summary of the potential CUDA problems: https://github.com/pika-org/pika/pull/306 was added in 0.7.0 and makes CUDA kernel continuations run inline if the stream which was used to submit the previous kernel is already empty (i.e. the kernel finished). This seems to cause pathological behaviour and stack overflows in one of the test cases on this PR. I've yet to determine if this is "expected" behaviour and should be worked around in DLA-Future (or pika), or if it's a plain bug in pika where some continuations are run inline when they shouldn't.

@teonnik tested with pika 0.6.0 and the segfaults go away. If I can't find a reasonable fix until the 0.8.0 release I'll revert https://github.com/pika-org/pika/pull/306 for 0.8.0 and look into it after the release.

test/unit/eigensolver/test_tridiag_solver.cpp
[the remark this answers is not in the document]
teonnik: https://github.com/eth-cscs/DLA-Future/pull/617/commits/017472f6f561c594c3f44759ccb78a6cd119dfd7

include/dlaf/eigensolver/tridiag_solver/impl.h
```diff
@@ -76,37 +76,12 @@ std::vector<pika::shared_future<T>> cuppensDecomposition(Matrix<T, D>& mat_trd)
...
-      }
-    };
-
     auto sender =
         ex::when_all(mat_trd(LocalTileIndex(i_split, 0)), mat_trd(LocalTileIndex(i_split + 1, 0)));
 
     // Cuppen's tridiagonal decomposition
     offdiag_vals.push_back(
-        di::transform(di::Policy<DefaultBackend<D>::value>(), std::move(cuppens_fn), std::move(sender)) |
+        di::transform(di::Policy<DefaultBackend<D>::value>(), cuppensDecomp_o, std::move(sender)) |
```
msimberg, line 84: I just want to point out that this can be further reduced to `std::move(sender) | cuppensDecomp(di::Policy<DefaultBackend<D>::value>())` if you use `DLAF_MAKE_SENDER_ALGORITHM_OVERLOADS`. However, this change would be part of https://github.com/eth-cscs/DLA-Future/issues/634 (I'll link to this comment there).

teonnik: Yes, thanks for bringing this up here, I meant to talk about it in the weekly meeting. I actually intentionally left out the `DLAF_MAKE_SENDER_ALGORITHM_OVERLOADS` overloads because I wasn't sure if they add much value compared to just passing `_o` to `transform()`. Unless I am missing something, they seem to be mostly syntactic sugar which, unfortunately, adds an additional templating layer and more boilerplate. However, if there are good reasons to use them (be it just for the sake of convention) I wouldn't mind using them here.

msimberg: You're right about it being syntactic sugar, but it's not only that. Let me try to explain a few points.

1. The current set of macros definitely add bloat (and worse, increase compile times). This is bad, no doubt about it!
2. Your comment makes me think of a way to avoid some of the above. The current layering with `DLAF_MAKE_SENDER_ALGORITHM_OVERLOADS` going through a layer of `transform` is one way of implementing it but it need not be that way. The sender overloads can instead "inline" the body of `transform` directly. Since `transform` is anyway instantiated for each different function type and policy combination, this should not add bloat and hopefully should even reduce it.
3. There's a design reason for preferring `algo_name(...)` to be the entry point rather than `transform(..., algo_name)`. The former is much more straightforward to customize, and the latter is a bit sketchy to customize (and will not be possible to customize in some cases).

The two first points are relevant to https://github.com/pika-org/pika/issues/433 and I will try something out there to see if it helps.

The third point has the added dimension that there won't be the same concerns and constraints for all algorithms. For `tile::gemm` we eventually want to be able to use `std::blas::matrix_matrix_multiplication` and making this transition easier was the main intention for me to prefer the "non-transform" versions. However, we also have things like `cuppensDecomp` which will probably never be an algorithm in the standard library and where we control the implementation from top to bottom. There we're of course much more free to make various assumptions based on the implementation (such as removing a layer of function calls). If 2. can help remove the unnecessary bloat that is currently there I would always prefer using non-transform versions for consistency. If that doesn't help enough then we absolutely targeted changes in the knowledge that the changes are just that: bloat and compilation time optimizations.

Does that make sense?

albestro (approved):

include/dlaf/eigensolver/tridiag_solver/coltype.h
```diff
@@ -0,0 +1,40 @@
...
+    str << "Dense";
+  }
+  else if (ct == ColType::UpperHalf) {
+    str << "UpperHalf";
+  }
+  else {
+    str << "LowerHalf";
+  }
+  return str;
+}
```
msimberg, line 38: Related issue: https://github.com/eth-cscs/DLA-Future/issues/644. Not to be changed in this PR.

msimberg (approved):

include/dlaf/eigensolver/tridiag_solver/kernels.h
```diff
@@ -0,0 +1,292 @@
...
+
+#pragma once
+
+#include "dlaf/common/callable_object.h"
+#include "dlaf/eigensolver/tridiag_solver/coltype.h"
+#include "dlaf/matrix/tile.h"
+#include "dlaf/types.h"
+
+#ifdef DLAF_WITH_GPU
+#include <cusolverDn.h>
```
msimberg, line 19: I think you want to do this instead here:
```suggestion
#include <dlaf/gpu/lapack/api.h>
```

rasolca:
bors try

rasolca:
bors try

teonnik:
bors try

PR #629, opened 2022-08-19
Remove `enable_language(C)`

msimberg:
I'm sure I've missed some reason why it's required, but I'm naively assuming that this is ok and hoping that CI will tell me if it's not possible.

Has this been added for some `find_package(MPI)`/CMake nonsense?


msimberg:
bors try

teonnik (approved):

msimberg:
@rasolca as discussed on Monday, feel free to close this without merging if you think we might need it for MPI.

rasolca:
I'm not sure about it, but I would not like to risk.

PR #631, opened 2022-08-24
Find Umpire in `DLAFConfig.cmake`

msimberg:
Part of #628.

Interesting side observations from this:
- CMake did not complain during configuration because the Umpire target is only called `umpire` which becomes `-lumpire` if there's no target by that name
- Spack hides these kind of mistakes because it adds include directories by default
- This only came up when compiling the DLAF wrapper as a HIP file in CP2K, because the `clang++` that `enable_language(HIP)` uses is not the `clang++` spack wrapper and it did not know where to find Umpire includes

### DIFF
--- a/cmake/template/DLAFConfig.cmake.in
+++ b/cmake/template/DLAFConfig.cmake.in
@@ -44,6 +44,7 @@ endif()
 
 find_dependency(blaspp PATHS @blaspp_DIR@)
 find_dependency(lapackpp PATHS @lapackpp_DIR@)
+find_dependency(Umpire PATHS @Umpire_DIR@)
 
 # ----- pika
 find_dependency(pika PATHS @pika_DIR@)


msimberg:
> Is the mixed use of upper and lower case correct?
> 
> https://github.com/eth-cscs/DLA-Future/blob/2ff39be7f23c95572b300457cc1433e4cada5749/CMakeLists.txt#L128
> 
> https://github.com/eth-cscs/DLA-Future/blob/150628d126ca78571d878e00255b82d0d2b55cb4/src/CMakeLists.txt#L69-L76
> 
> https://github.com/eth-cscs/DLA-Future/blob/ba94985ac45b2e9a91ea90433db760f6a953029c/cmake/template/DLAFConfig.cmake.in#L47

Unfortunately yes... In theory `find_package(umpire)` also works since they use `kebab-case` for `umpire-config.cmake` (https://github.com/LLNL/Umpire/blob/develop/umpire-config.cmake.in), but the real Umpire project name is `Umpire` (https://github.com/LLNL/Umpire/blob/5dc5843c950c792f8a591ca5176c9a85fdc9e4c3/CMakeLists.txt#L16) and the target name is `umpire` (https://github.com/LLNL/Umpire/blob/5dc5843c950c792f8a591ca5176c9a85fdc9e4c3/src/umpire/CMakeLists.txt#L114).

albestro (approved):

rasolca:
bors try

PR #632, opened 2022-08-29
Fix empty cases for tridiagonal solver

### LINKED ISSUE #625: Tridiagonal solver problem with empty matrices
About empty matrices tests

https://github.com/eth-cscs/DLA-Future/blob/fa5ef85793db4de75329fa58c72a18a9fc04d8a6/test/unit/eigensolver/test_eigensolver.cpp#L52

https://github.com/eth-cscs/DLA-Future/blob/fa5ef85793db4de75329fa58c72a18a9fc04d8a6/test/unit/eigensolver/test_gen_eigensolver.cpp#L50

The problem is that here

https://github.com/eth-cscs/DLA-Future/blob/fa5ef85793db4de75329fa58c72a18a9fc04d8a6/include/dlaf/eigensolver/tridiag_solver/mc.h#L92-L98

`i_end` becomes negative when there are no tiles, i.e. with empty matrices, which is not valid for `reserve` (with DLAF_ASSERTs enabled the `to_sizet` correctly catches it).

It should be an easy fix, but probably it is even easier for you @teonnik to add the quick return in the right point of the code.

I can open an issue for that, since probably it would be better to add a test case for it in tridiagonal solver.

_Originally posted by @albestro in https://github.com/eth-cscs/DLA-Future/issues/547#issuecomment-1209186510_

albestro:
Fix #625

This is required by #547.

### DIFF
--- a/test/unit/eigensolver/test_tridiag_solver.cpp
+++ b/test/unit/eigensolver/test_tridiag_solver.cpp
@@ -99,6 +99,9 @@ void solveLaplace1D(SizeType n, SizeType nb) {
 
   eigensolver::tridiagSolver<Backend::MC>(tridiag, evals, evecs);
 
+  if (n == 0)
+    return;
+
   // Eigenvalues
   auto expected_evals_fn = [n](GlobalElementIndex i) {
     return RealParam(2 * (1 - std::cos(M_PI * (i.row() + 1) / (n + 1))));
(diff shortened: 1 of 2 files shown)


test/unit/eigensolver/test_tridiag_solver.cpp
```diff
@@ -158,7 +161,7 @@ void solveRandomTridiagMatrix(SizeType n, SizeType nb) {
   std::vector<RealParam> diag_arr(to_sizet(n));
   std::vector<RealParam> offdiag_arr(to_sizet(n));
   SizeType diag_seed = n;
-  SizeType offdiag_seed = n - 1;
+  SizeType offdiag_seed = n + 1;
```
teonnik, line 164: Hmm, why is this needed?

teonnik (approved):
LGTM overall!

test/unit/eigensolver/test_tridiag_solver.cpp
albestro: Because `n == 0` make `offdiag_seed = -1`, which will internally get casted to `size_t` (more generally an `unsigned int`) as it is the input seed for the random engine `std::mt19937_64`.

albestro:
bors try

Issue #607, opened 2022-07-19
Investigate clang build segfault with cuda 11.6 and 11.7


rasolca:
https://github.com/eth-cscs/DLA-Future/pull/594#issuecomment-1188954697

PR #640, opened 2022-09-08
Try to speed CI up

rasolca:
3 tricks are used to speed CI up:
1) Force using image if the given tag exists:
  - Pro: Unnecessary (and inexplicable) rebuilding of build image are avoided.
  - Cons: Changes which do not modify the tag do not trigger a rebuild. (Note: this should happen in no or extremely rare cases, and can be solved removing manually the previous image from jfrog)
  
2) Insert dependencies to reduce the amount of build task running concurrently.
  (I noticed that a full (i.e. the build image has been rebuilt from scratch) cpu release gcc11 build phase took only 60 min when run alone, while the same task without the full build image building took 1h30min when run together with other tasks. Therefore I suspect we overload the build system)

3) remove the CUDA + clang10 test:
  As this test use an old CUDA + a compiler which doesn't support C++20 (will be needed in the near future by pika), and generates some problems with the tridiagonal solver PR #617 we remove it.
  It will be reintroduced when we find a working combination of CUDA + clang (see #607)

### DIFF
--- a/ci/.gitlab-ci.yml
+++ b/ci/.gitlab-ci.yml
@@ -33,7 +33,7 @@ stages:
     - TAG_REPO=`find $SPACK_DLAF_REPO -type f -exec sha256sum {} \; | sha256sum - | head -c 16`
     - TAG_ENVIRONMENT=`sha256sum $SPACK_ENVIRONMENT | head -c 16`
     - TAG=${TAG_IMAGE}-${TAG_APTGET}-${TAG_COMPILER}-MKL${USE_MKL}-${TAG_DOCKERFILE}-${TAG_SPACK}-${TAG_REPO}-${TAG_ENVIRONMENT}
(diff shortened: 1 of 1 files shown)


rasolca:
bors try

rasolca:
bors try-

msimberg (approved):
Looks reasonable enough to me. Do I understand correctly that the rebuilds may have been happening if the local cache on the build node doesn't have the tagged image anymore, but by first pulling it you force it to be available locally (if it exists) to be reused? Or should remotely available layers be automatically be downloaded and reused locally as well (which is the inexplicable part)?

rasolca:
The cache works even if the content is not available locally.
docker downloads a manifest file (line 85 and 89 of the output below)
then each command of the docker file is compared with the cached one (e.g. line 98).
Docker then downloads the layers needed for the cache (lines 139-170) and, if some of the layers changed, builds them (none in the example)
https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/jobs/2978895135

Sometimes it happens that the first layer of the cache (the base image) is not used. Not sure why it happens as checking the output the base image is exactly the same (sha included). This is the inexplicable part.

The logic of the new way try to avoid the cache mechanism if possible:
First it tries to download the image. If the image exists it is tagged as `latest` as well and used,
otherwise it builds it using only `latest` as cache. (Note the `||` in https://github.com/eth-cscs/DLA-Future/blob/d1daf273312e66dfa4b848ff72ad38965c554af4/ci/.gitlab-ci.yml#L36 ).

rasolca:
bors try

rasolca:
bors try

Issue #602, opened 2022-07-11
Investigate problems of HIP builds


rasolca:
https://github.com/eth-cscs/DLA-Future/pull/459#issuecomment-1179232091

rasolca:
Update:
I rerun the test with the new image #592 (rocm5.1.3) and the results are:
a different error:
```
[ RUN      ] BacktransformationT2BTestGPU/0.CorrectnessLocal
/DLA-Future/test/unit/eigensolver/test_bt_band_to_tridiag.cpp:161: Failure
Failed
Error at index (1, 0): expected -0.571167 == -0.61407 (Relative diff: 0.0698675 > 2.86102e-06, Absolute diff: 0.0429035 > 2.86102e-06)

[  FAILED  ] BacktransformationT2BTestGPU/0.CorrectnessLocal, where TypeParam = float (11371 ms)
```
and again node killed by
`test_broadcast_panel`

rasolca:
During last attempt to run the CI image test on ault (with AMD GPUs) for #642 RANK_[1,2,4] test passed, however some RANK_6 tests drains the node (including `test_cholesky` that was passing in the past).
Not sure about the full list of the failing and node draining tests as I stopped testing after I drained 3 nodes.

rasolca:
Solved by #742.

PR #642, opened 2022-09-12
Replace hipblas with rocblas

aurianer:
Dependent on #657. Remove hipblas dependency and replace all hipblas usage by rocblas symbols.

Flyby:
- Require pika@0.9.0, since hipblas has been replaced in pika 0.9.0 and macro redefinition is allowed only if it has the same value.
- Update spack sha to have pika 0.9.0 available.


include/dlaf/blas/tile.h
```diff
@@ -81,6 +81,23 @@ inline void extendROCBlasWorkspace(cublasHandle_t handle,
   template <typename T>               \
   struct Name
 
+#ifdef DLAF_WITH_HIP
+    // rocblas functions needs to be lowercase
```
msimberg, line 85:
```suggestion
    // rocblas functions are lowercase
```
but I also think the comment doesn't add much value and could be removed?

include/dlaf/blas/tile.h
```diff
@@ -81,6 +81,23 @@ inline void extendROCBlasWorkspace(cublasHandle_t handle,
...
+  DLAF_DEFINE_GPUBLAS_OP(Name, double, d##f);              \
+  DLAF_DEFINE_GPUBLAS_OP(Name, std::complex<float>, c##f); \
+  DLAF_DEFINE_GPUBLAS_OP(Name, std::complex<double>, z##f)
+
+#define DLAF_MAKE_GPUBLAS_SYHE_OP(Name, f)                   \
+  DLAF_DECLARE_GPUBLAS_OP(Name);                             \
+  DLAF_DEFINE_GPUBLAS_OP(Name, float, Ssy##f);               \
+  DLAF_DEFINE_GPUBLAS_OP(Name, double, Dsy##f);              \
+  DLAF_DEFINE_GPUBLAS_OP(Name, std::complex<float>, Che##f); \
+  DLAF_DEFINE_GPUBLAS_OP(Name, std::complex<double>, Zhe##f)
```
msimberg, line 98:
```suggestion
  DLAF_DEFINE_GPUBLAS_OP(Name, float, ssy##f);               \
  DLAF_DEFINE_GPUBLAS_OP(Name, double, dsy##f);              \
  DLAF_DEFINE_GPUBLAS_OP(Name, std::complex<float>, che##f); \
  DLAF_DEFINE_GPUBLAS_OP(Name, std::complex<double>, zhe##f)
```
or is this what your last `TODO` commit is about...?

include/dlaf/gpu/blas/error.h
```diff
@@ -18,65 +18,26 @@
...
-#ifdef DLAF_WITH_HIP
-// Lifted from hipBLAS/library/src/hcc_detail/hipblas.cpp, with missing
-// enumeration values added. This should be removed when hipBLAS is no longer
-// used.
-inline hipblasStatus_t rocBLASStatusToHIPStatus(rocblas_status error) {
-  switch (error) {
-    case rocblas_status_continue:
-    case rocblas_status_size_unchanged:
-    case rocblas_status_size_increased:
-    case rocblas_status_success:
```
msimberg, line 53: Remember that these all need to be considered success with rocblas. They are returned when getting the workspace size.

msimberg (changes requested):

include/dlaf/gpu/blas/error.h
rasolca: @msimberg I'm wondering if separating the two thing would be better:
- when executing `rocblas_status_success` is the expected return code and is checked with `DLAF_GPUBLAS_CHECK_ERROR`
- when checking the rocblas workspace the expected return code are `rocblas_status_continue, rocblas_status_size_increased, rocblas_status_size_unchanged` which should be checked with a different method. (A possible name can be`DLAF_ROCBLAS_WORKSPACE_CHECK_ERROR`).

msimberg: Yeah, that sounds reasonable.

include/dlaf/blas/tile.h
msimberg: Unpushed?

include/dlaf/gpu/blas/error.h
```diff
@@ -18,67 +18,55 @@
...
     case rocblas_status_memory_error:
-      return HIPBLAS_STATUS_ALLOC_FAILED;
     case rocblas_status_internal_error:
-      return HIPBLAS_STATUS_INTERNAL_ERROR;
     case rocblas_status_perf_degraded:
     case rocblas_status_size_query_mismatch:
     case rocblas_status_check_numerics_fail:
-    default:
-      return HIPBLAS_STATUS_UNKNOWN;
+      return error;
```
msimberg, line 64: I can't add a diff here, but you can use `default` here for the "all-the-other-error-codes" case.

rasolca:
bors try-

spack/packages/dla-future/package.py
```diff
@@ -46,7 +46,7 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("umpire@6:")
 
     depends_on("pika +mpi")
-    depends_on("pika@0.8:")
+    depends_on("pika@0.9:")
```
msimberg, line 49: I'm slightly losing track of what versions are required for what. Is this bump necessary or just nice to have here? To be clear, I don't mind the bump here even it's not strictly necessary, just trying to keep track of the why.

include/dlaf/blas/tile.h
```diff
@@ -485,7 +502,7 @@ void trsm(cublasHandle_t handle, const blas::Side side, const blas::Uplo uplo, c
 #ifdef DLAF_WITH_CUDA
       blasToCublasCast(a.ptr());
 #elif defined(DLAF_WITH_HIP)
-      // The hipblas API requires a non-const argument
+      // The rocblas API requires a non-const argument
```
msimberg, line 505: Did you check by the way if the restriction is really there for rocblas as well, or if it was only a problem with the hipblas API?

include/dlaf/blas/tile.h
```diff
@@ -34,10 +34,10 @@
 #define DLAF_GET_ROCBLAS_WORKSPACE(f)                                                                 \
   [&]() {                                                                                             \
     std::size_t workspace_size;                                                                       \
-    DLAF_GPUBLAS_CHECK_ERROR(                                                                         \
+    DLAF_ROCBLAS_WORKSPACE_CHECK_ERROR(                                                               \
```
msimberg, line 37: I _think_ `rocblas_{start,stop}_device_memory_size_query` should not return any of the "special" statuses (i.e. `rocblas_status_size_increased`, `rocblas_status_size_unchanged`, or `rocblas_status_continue`) so you should be able to use the regular error checking macro here. Is that the case?

spack/packages/dla-future/package.py
rasolca, line 49: Does this PR depend on #657?

include/dlaf/matrix/print_gpu.h
```diff
@@ -51,7 +51,7 @@ void print(Format format, const Tile<const T, Device::GPU>& tile, whip::stream_t
 
 template <class Format, class T>
 void print(Format format, const Tile<const T, Device::GPU>& tile, std::ostream& os = std::cout) {
-  whip::stream_t stream;
+  whip::stream_t stream{};
```
rasolca, line 54: This has no effect... Is there a reason to change?

include/dlaf/blas/tile.h
rasolca: And if the restriction is still there: is it guaranteed that values of A are not changed?

spack/packages/dla-future/package.py
msimberg: I think if there's no need to restrict the version unless really required.

msimberg: > I mean in any case we have a fixed version for pika in the CMakeLists.txt, so I figured since there is no release for DLAF, it should be exactly 0.9 but I don't mind the column

I read this too quickly earlier. We _don't_ have a fixed version in `CMakeLists.txt` (there's no `EXACT`). However, see my other comment on https://github.com/eth-cscs/DLA-Future/pull/657.

msimberg: This comment: https://github.com/eth-cscs/DLA-Future/pull/657#discussion_r999371105.

msimberg: TL;DR: pika 0.9.0 _is_ required for these changes (for HIP, but we can require it for everything).

msimberg:
bors try

msimberg:
@aurianer I think this is still missing the fix that was made to have pika correctly depend on `whip`. Could you update the spack commit once more? Note that you can just take the commits from https://github.com/msimberg/spack/tree/umpire-misc-fixes-2 and create a branch for yourself that is rebased on a recent enough upstream `develop`. The branch/commit does not need to be on the upstream spack repo for CI to be able to download it.

rasolca:
The CI is building pika@main instead of pika@0.9.0 and fails.
https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/jobs/3200879815
I suspect that the reason is that SPACK_SHA was not updated.

@msimberg Might the old SPACK_SHA be the reason of pika@main not having the whip dependency? Otherwise I'll open an issue in pika.

msimberg:
> @msimberg Might the old SPACK_SHA be the reason of pika@main not having the whip dependency? Otherwise I'll open an issue in pika.

It's definitely the reason.

msimberg:
bors try

msimberg (approved):

ci/.gitlab-ci.yml
```diff
@@ -19,7 +19,7 @@ stages:
     - trying
   variables:
     GIT_SUBMODULE_STRATEGY: recursive
-    SPACK_SHA: 08efe13efb627793fbe048542890ee85668c8100
+    SPACK_SHA: e4f4b70ed153167c06aa63eefa41d975394af066
```
albestro, line 22: Should we add a comment here that specifies why we are using a custom version of spack?

albestro (approved):
Just a minor comment about spack version used.

teonnik (approved):
LGTM!

ci/.gitlab-ci.yml
albestro: FIY `pika@0.9.0` is available in official spack, so the only missing thing is the fix for `umpire`, specifically `CMAKE_HIP_ARCHITECTURES`, for which there is an open PR by @msimberg https://github.com/spack/spack/pull/32025

albestro: I would just change the comment, I don't want to keep this on hold 😄 thanks!

msimberg: Indeed, I wouldn't wait for that to be merged. I've sent some messages to try to get it in a bit faster but don't hold your breath.

ci/.gitlab-ci.yml
```diff
@@ -19,7 +19,8 @@ stages:
     - trying
   variables:
     GIT_SUBMODULE_STRATEGY: recursive
-    SPACK_SHA: 08efe13efb627793fbe048542890ee85668c8100
+    # Custom spack sha used to make a umpire fix + pika 0.9.0 available
```
msimberg, line 22: pika 0.9.0 is already available on develop without the extra patches. Maybe like this:
```suggestion
    # Custom spack sha used to make an umpire fix available
    # (https://github.com/aurianer/spack/commit/e4f4b70ed153167c06aa63eefa41d975394af066)
```

msimberg (approved):

rasolca:
bors try

rasolca:
Tested the rocm image on ault.
Result: RANK_[1,2,4] tests passed and problems with some RANK_6 tests (draining nodes). #602

Issue #623, opened 2022-08-09
Remove `mpi_invoke`?


msimberg:
It's currently unused. Unsure if it will be useful in the future, or if it can be safely removed.

msimberg:
@teonnik pointed out that https://github.com/eth-cscs/DLA-Future/pull/255 introduced `mpi_invoke` which was created for https://github.com/eth-cscs/DLA-Future/pull/228 which was never merged (old executors implementation). With that in mind I think we can remove this functionality (and easily reintroduce some notion of dealing with `MPI_THREAD_SERIALIZED`/`MPI_THREAD_MULTIPLE` at a later stage if needed).

PR #651, opened 2022-09-26
Check required MPI thread level in `dlaf::initialize`

msimberg:
To avoid unpleasant debugging when something other than DLAF initializes MPI with the wrong level of threading, I propose we check the level during initialization. This is the minimal required changed, but this is a draft because if I understand correctly we could remove the support for `MPI_THREAD_SERIALIZED`? Is that required for something? Related issue: #623.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -197,6 +197,17 @@ void initialize(pika::program_options::variables_map const& vm, configuration co
     std::exit(0);
   }
 
+  int mpi_initialized;
+  DLAF_MPI_CHECK_ERROR(MPI_Initialized(&mpi_initialized));
+  if (mpi_initialized) {
+    int provided;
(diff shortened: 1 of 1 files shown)


teonnik (approved):

msimberg:
Just for reference, #655 removes `mpi_invoke` etc. for #623.

albestro (approved):

msimberg:
bors try

msimberg:
Updated to guard the `MPI_Query_thread` check to only happen if MPI has been initialized, so that single-rank tests that don't use MPI, but still initialize DLA-Future, don't attempt to check for the thread level.

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

msimberg:
Some jobs seem to have failed to start the first time around. They were restarted and this looks ok from the CI perspective now.

PR #657, opened 2022-09-30
Porting to pika 0.9.0

aurianer:


msimberg:
What's the status of this PR?

CMakeLists.txt
```diff
@@ -116,7 +116,7 @@ else()
 endif()
 
 # ----- pika
-find_package(pika 0.8.0 REQUIRED)
+find_package(pika 0.9.0 REQUIRED)
```
msimberg, line 119:
```suggestion
find_package(pika 0.8.0 REQUIRED)
```

spack/packages/dla-future/package.py
```diff
@@ -46,7 +46,7 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("umpire@6:")
 
     depends_on("pika +mpi")
-    depends_on("pika@0.8:")
+    depends_on("pika@0.9")
```
msimberg, line 49:
```suggestion
    depends_on("pika@0.8:")
```

msimberg (changes requested):
I think we can keep the 0.8.0 version requirement as it is.

CMakeLists.txt
msimberg: In fact, for this PR specifically the requirement even seems to be 0.8.0 exactly because of HIP changes. pika 0.9.0 already uses rocblas exclusively so all the macro definitions related to cublas/rocblas/hipblas conflict between pika and DLA-Future. Conversely, once https://github.com/eth-cscs/DLA-Future/pull/642 is merged pika 0.8.0 can't be used because then the macros will again conflict but in the opposite direction (DLA-Future uses rocblas and pika uses hipblas).

So what I would suggest is that if https://github.com/eth-cscs/DLA-Future/pull/642 is close to ready we close this PR and only go with https://github.com/eth-cscs/DLA-Future/pull/642 (with a pika 0.9.0 version requirement). In theory 0.8.0 would be enough for everything non-HIP, but as before if we can avoid special cases I think we should unless there's a really good reason to accomodate the special case.

msimberg:
@aurianer I will close this since #642 is close to be merged and will bump the requirement to 0.9.0 anyway (and this can't go in on its own unless `+rocm` is restricted to `pika@0.8`).

PR #659, opened 2022-10-06
algo: Distributed sub-Gemm (trisolver use-case)

albestro:
This is a distributed gemm that covers a use-case needed for the distributed tridiagonal eigensolver. In particular, it allows to multiply two (distributed) square matrices selections of size `k`.

Changelog:
- bug-fix: `getSubMatrixMatrixMultiplication` was returning wrong values for multiplications not covering the full matrix;
- minor bug-fix: `std::polar` (and transitively `TypeUtilities::polar`) gets as first parameter `magnitude` value, which cannot be negative;
- extend test-cases for old test in `test_multiplication_general`, the partial multiplication was not tested.

TODO:
- [x] ~accept k as "future"~ (let's just skip this optimisation for the moment)
- [x] extend and check test for distributed case
- [x] add and verify test-cases for incomplete tiles
- [x] add and check assertions
- [x] enabling GPU (if it is straightforward, as expected)
- [ ] decide name and position for this algorithm

### DIFF
--- a/include/dlaf/multiplication/general.h
+++ b/include/dlaf/multiplication/general.h
@@ -14,6 +14,8 @@
 
 #include "dlaf/common/assert.h"
 #include "dlaf/communication/communicator_grid.h"
+#include "dlaf/matrix/distribution.h"
+#include "dlaf/matrix/index.h"
 #include "dlaf/matrix/matrix.h"
 #include "dlaf/multiplication/general/api.h"
 #include "dlaf/util_matrix.h"
@@ -47,14 +49,18 @@ void generalSubMatrix(const SizeType a, const SizeType b, const blas::Op opA, co
   DLAF_ASSERT(dlaf::matrix::square_blocksize(mat_b), mat_b);
   DLAF_ASSERT(dlaf::matrix::square_blocksize(mat_c), mat_c);
 
-  // TODO check assertions. these are superflous
-  DLAF_ASSERT(dlaf::matrix::square_size(mat_a), mat_a);
-  DLAF_ASSERT(dlaf::matrix::square_size(mat_b), mat_b);
-  DLAF_ASSERT(dlaf::matrix::square_size(mat_c), mat_c);
-
   DLAF_ASSERT(matrix::local_matrix(mat_a), mat_a);
   DLAF_ASSERT(matrix::local_matrix(mat_b), mat_b);
   DLAF_ASSERT(matrix::local_matrix(mat_c), mat_c);
+
+  // Note:
+  // This is an over-constraint, since the algorithm just cares about the sub-matrix size.
+  // It simplifies next check about [a,b] range validity, that otherwise would require it to be
+  // validated against every single sub-matrix in a, b and c that might have different element sizes.
+  //
+  // At the moment, we don't have this use-case, so let's keep it simple.
+  DLAF_ASSERT(equal_size(mat_a, mat_b), mat_a, mat_b);
+  DLAF_ASSERT(equal_size(mat_a, mat_c), mat_a, mat_c);
 
   const SizeType m = mat_a.nrTiles().rows();
   DLAF_ASSERT(a <= b, a, b);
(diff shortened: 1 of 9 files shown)


include/dlaf/multiplication/general.h
```diff
@@ -66,4 +67,32 @@ void generalSubMatrix(const SizeType a, const SizeType b, const blas::Op opA, co
   else
     DLAF_UNIMPLEMENTED(opA, opB);
 }
+
+template <Backend B, Device D, class T>
+void generalSubMatrixK(comm::CommunicatorGrid grid, const SizeType a, const SizeType b, const SizeType k,
```
msimberg, line 72:
```suggestion
void generalSubMatrixSquare(comm::CommunicatorGrid grid, const SizeType a, const SizeType b, const SizeType k,
```
, at least if I interpret the `K` as referring to the inputs being `k`×`k` matrices?

include/dlaf/multiplication/general/api.h
```diff
@@ -40,5 +40,28 @@ DLAF_MULTIPLICATION_GENERAL_ETI(extern, Backend::GPU, Device::GPU, double)
 DLAF_MULTIPLICATION_GENERAL_ETI(extern, Backend::GPU, Device::GPU, std::complex<float>)
 DLAF_MULTIPLICATION_GENERAL_ETI(extern, Backend::GPU, Device::GPU, std::complex<double>)
 #endif
+
+template <Backend B, Device D, class T>
+struct GeneralSubK {
```
msimberg, line 45: If the main algorithm is renamed, this should probably also be renamed to something consistent.

test/include/dlaf_test/matrix/util_generic_blas.h
```diff
@@ -61,22 +63,22 @@ auto getSubMatrixMatrixMultiplication(const SizeType a, const SizeType b, const
 
   auto elC = [a, b, k](const GlobalElementIndex ij) {
     if (ij.row() < a || ij.row() > b || ij.col() < a || ij.col() > b)
-      return TypeUtilities<T>::polar(-99, 99);
+      return -TypeUtilities<T>::polar(99, 99);
```
msimberg, line 66: Would it be worth adding an assertion to `TypeUtilities::polar` (possibly `DLAF_ASSERT_MODERATE` if it's expected to be called frequently somewhere). https://en.cppreference.com/w/cpp/numeric/complex/polar says "The behavior is undefined if r is negative or NaN, or if theta is infinite."

test/include/dlaf_test/matrix/util_generic_blas.h
```diff
@@ -61,22 +63,22 @@ auto getSubMatrixMatrixMultiplication(const SizeType a, const SizeType b, const
...
     const double j = ij.col();
 
     return TypeUtilities<T>::polar((i + k + 1) / (j + 5), i + j + k);
   };
 
-  auto elR = [a, b, k, alpha, beta](const GlobalElementIndex ij) {
+  auto elR = [a, b, sub, k, alpha, beta](const GlobalElementIndex ij) {
     if (ij.row() < a || ij.row() > b || ij.col() < a || ij.col() > b)
-      return TypeUtilities<T>::polar(13, -26);
+      return -TypeUtilities<T>::polar(99, 99);
```
msimberg, line 76: Were `13` and `-26` "important numbers" or is `99, 99` just as good (or better) in this situation? Asking because I can't judge if it's a copy-paste mistake or intentional.

test/include/dlaf_test/matrix/util_generic_blas.h
albestro: Yes, I'm with you, I thought about that too 😉 It might also reveal other cases under cover.

test/include/dlaf_test/matrix/util_generic_blas.h
albestro: It is the bug-fix I'm talking about in the first point of the changelog 😉

test/include/dlaf_test/matrix/util_generic_blas.h
albestro: Question: since this is in `DLAF_test` library, which does not depend on `dlaf`, should we just go for plain `assert`?

include/dlaf/multiplication/general.h
albestro: Just to give a bit more context, but you are already on the good track.

In the best case, this algorithm will accept three main parameters:
- idx_begin
- idx_last
- k

Given those, it will compute the gemm between the square matrices using A, B and C but just in the range of tiles `[idx_begin, idx_last]`, so it will be by construction a gemm over square matrices.

The `k` parameter, which we are not still sure it will be implemented in this first version, opens to an optimisation where just the first k rows/cols in these square sub-matrices are actually used for the gemm.

test/include/dlaf_test/matrix/util_generic_blas.h
msimberg: Good point. Yes, plain assert sounds good!

msimberg: Looks good, though I realize only now: why not `DLAF_ASSERT`? I actually thought you were referring to that with plain assert (as opposed to `DLAF_ASSERT_MODERATE` which is not "plain").

include/dlaf/multiplication/general.h
msimberg: After @albestro's answer and speaking to him offline I now understand why `generalSubMatrixSquare` doesn't make sense as a name, or at least doesn't make sense only for the `K` version: both `generalSubMatrix` and `generalSubMatrixK` operate on square submatrices by construction. An important distinction that I also didn't understand first is that `a` and `b` are tile indices, while `k` is an element index (or size). So I retract at least my original renaming suggestion. Perhaps some other name would still be more descriptive but I'm also ok with leaving the name as it is right now.

msimberg (approved):

rasolca (approved):

albestro:
bors try

albestro:
bors try

teonnik (approved):
LGTM!

Issue #612, opened 2022-07-29
Cholesky "external" interface C++ / C / Fortran


rasolca:
To simplify the usage in existing application a C/C++ Scalapack-like interface is needed that take care of resuming everything is needed to execute a DLAF algorithm execute it and suspend again DLAF.

Moreover Fortran bindings are also needed.

Issue #610, opened 2022-07-29
DLAF + pika initialization / resume / suspension / finalization


rasolca:
To be able to use the library in existing application we need:
- initialization: initialize pika and DLAF + suspend the runtime
- resume: will be called before the algorithm to resume pika runtime
- suspend: after the algorithm is completed suspend pika runtime
- finalization: release all DLAF resources and stop pika.

PR #668, opened 2022-10-11
Enforce single threaded parallel regions for blas/lapack calls

### LINKED ISSUE #611: Blas thread setting
When used in existing application the library should be able to control the number of BLAS thread.

In particular the following is needed:
- set BLAS number of threads to 1 before the execution of any algorithm
- reset the previous number of threads after the algorithm has completed.

Note: a RAII object can prevent forgetting to reset the number of threads.

msimberg:
Attempt to fix #611.

This adds an RAII class to set/reset the number of threads for OpenMP around blas/lapack calls. The current approach explicitly creates a `SingleThreadedOmpScope` object before blas/lapack calls. The downside of this is approach is that it requires explicitly creating the scope for every call and risks us forgetting to create it.

One alternative would be to add the creation into e.g. the callable object wrapper created by `DLAF_MAKE_CALLABLE_OBJECT` (or a similar macro) which would give us the creation "for free" for every blas/lapack call that goes through the callable objects. This, however, still doesn't cover specialized, direct, calls to blas/lapack like https://github.com/eth-cscs/DLA-Future/blob/a864d299e54bfa9a885cd3ceec479b49a0ff30b8/include/dlaf/eigensolver/band_to_tridiag/mc.h#L55-L59.

Finally, a short note on the use of `*set_num_threads*`:
- From what I can tell `omp_set_num_threads` only applies to the current thread. This is at least the case in practice in `gomp`. Reading the OpenMP spec I can't quite tell if this is actually guaranteed or implementation specific. There's wording that seems to indicate that it might be guaranteed, but it could also be the other way around...
- `mkl_set_num_threads` definitely applies to all threads according to the documentation (and this is consistent with the behaviour I've observed).
- `mkl_set_num_threads_local` definitely applies to the current thread only.

The above behaviour would also explain why I "couldn't" use `omp_set_num_threads` in CP2K.

I'm not sure we actually need to call both `omp_set_num_threads` and `mkl_set_num_threads_local` if MKL is enabled, but figured it doesn't hurt. If someone knows for sure it's not necessary I can change that. I'm also assuming that for standalone calls in tests we don't need to enforce single threaded blas because we control the environment where they run. In the worst case someone could of course configure DLA-Future with threaded blas and then everything would be terribly slow, but things should still work correctly. Is this a big enough issue to worry about?

To do/decide:
- [ ] namespace for helper (currently `dlaf::common::internal`)?
- [ ] wrap each individual call or put into some macro where possible?
- [ ] name of RAII helper/header?
- [ ] wrap blas/lapack calls in tests/miniapps?


msimberg:
bors try

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

rasolca:
> I'm not sure we actually need to call both `omp_set_num_threads` and `mkl_set_num_threads_local` if MKL is enabled, but figured it doesn't hurt. If someone knows for sure it's not necessary I can change that.

It is needed, as any mkl mechanism to set the number of threads has priority over openmp. (E.g. MKL_SET_NUM_THREADS env variable, or any previous mkl_set_num_threads invocation)

msimberg:
> > I'm not sure we actually need to call both `omp_set_num_threads` and `mkl_set_num_threads_local` if MKL is enabled, but figured it doesn't hurt. If someone knows for sure it's not necessary I can change that.
> 
> It is needed, as any mkl mechanism to set the number of threads has priority over openmp. (E.g. MKL_NUM_THREADS env variable, or any previous mkl_set_num_threads invocation)

What I meant by that is: I'm not sure we need to call `omp_set_num_threads` if MKL is enabled (since `mkl_set_num_threads_local` anyway should have priority over everything else). We could _only_ call `mkl_set_num_threads_local`.

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

msimberg:
bors try

CMakeLists.txt
```diff
@@ -26,7 +26,14 @@ endif()
 
 include(CMakeDependentOption)
 
+set(DLAF_WITH_OPENMP_DESCRIPTION "Enable OpenMP support (TODO: Explanation for why?)")
```
msimberg, line 29: Do you think it's necessary to explain here what this is for? If yes, I think the explanation would be similar to the warning message on line 53, which is printed if `DLAF_WITH_OPENMP` is disabled _and_ `DLAF_WITH_MKL` is disabled. However, if we print the warning anyway I'm not sure if there's a need to have a long explanation here. What do you think?

CMakeLists.txt
```diff
@@ -26,7 +26,14 @@ endif()
...
+set(DLAF_WITH_OPENMP_DEFAULT ON)
+if(APPLE)
```
msimberg, line 31: This is currently quite coarse. @albestro and @rasolca do you think this can be `if(APPLE AND "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")` (or is it `AppleClang`?) or is this ok?

CMakeLists.txt
```diff
@@ -93,6 +111,9 @@ endif()
 # ----- MPI
 find_package(MPI REQUIRED)
 
+# ----- OpenMP
+find_package(OpenMP REQUIRED)
+
```
rasolca, line 116: I think that the if statement is missing...

CMakeLists.txt
rasolca: I would limit it to Apple clang.

msimberg: Thanks. And what does it identify itself as (in terms of `CMAKE_CXX_COMPILER_ID`?)?

CMakeLists.txt
msimberg: Indeed, I've just pushed a fix for that. Please have another look.

msimberg:
The remaining to do items in the PR description I will do on request.

As for the name, one alternative could be `SingleThreadedBlas(Lapack)` since OpenMP is an implementation detail that is sometimes there and sometimes not.

msimberg:
bors try

msimberg:
bors try

msimberg:
I've pushed a couple of more changes here:

- A test to check that the thread setting is correctly applied
- The test is skipped for MKL if `mkl_set_num_threads` doesn't have an effect (this is a symptom of using a sequential variant of MKL)
- A spack variant to choose the MKL variant to be used. Currently this only allows choosing `seq`, `omp`, or `tbb`. Does this need to expose all the other options available as well?
- I've changed the release mode CI environment to have `mkl=omp` so that the thread setting is actually tested (the codecov environment uses openblas).

CMakeLists.txt
msimberg: I removed the TODO and marking this as resolved. However, feel free to reopen if you feel that something needs more explanation.

include/dlaf/common/single_threaded_omp.h
```diff
@@ -0,0 +1,32 @@
...
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+
+// TODO: internal?
+namespace dlaf::common::internal {
+class [[nodiscard]] SingleThreadedOmpScope {
```
rasolca, line 15: Not fully satisfied with the name as the main goal is to reduce blas threads.
What about `SingleThreadBlasScope`? Or `SingleThreadScope`?

msimberg: Indeed, I'm also not entirely happy with having `Omp` in the name, especially since it also sets MKL threads (which is not necessarily using OpenMP).

I like `SingleThread(ed)Blas(Scope)` as well. `SingleThreadScope` sounds slightly too generic for my taste, but it's also ok if you prefer that.

teonnik (approved):
LGTM!

I agree with the https://github.com/eth-cscs/DLA-Future/pull/668#discussion_r1012972086 about the naming of `SingleThreadedOmpScope`, `SingleThread(ed)Blas(Scope)` sounds good to me.

> I'm also assuming that for standalone calls in tests we don't need to enforce single threaded blas because we control the environment where they run. In the worst case someone could of course configure DLA-Future with threaded blas and then everything would be terribly slow, but things should still work correctly. Is this a big enough issue to worry about?

No, I don't think it is, because as you said, we almost always control the environment for the test cases and even if we don't, they can still run.

rasolca:
On hold, as it conflicts with #698.

msimberg:
Rebased, renamed the wrapper (and source files), and added wrappers to the direct blas/lapack calls in tests/miniapps. This needs another sanity check still when it's been rebased on master when it's time to be merged to check that all relevant calls have been wrapped.

include/dlaf/common/single_threaded_omp.h
msimberg: It's now called `SingleThreadedBlasScope`.

msimberg:
cscs-ci run

msimberg:
I went through all bare blas/lapack calls now and wrapped them all in the single threaded scope wrapper, even ones that don't necessarily need it. The last commit (https://github.com/eth-cscs/DLA-Future/commit/f8ae14ab3feae536ab9e3fe9b1e517f24a7caf75) for sure has wrappers that are not strictly needed. I decided it's easier this way since then I don't have to worry if I've missed some, but if you think it's better to leave it out in some situations I can still revert those.

msimberg:
cscs-ci run

msimberg:
There was one test failure that I find unlikely to be caused by these changes, but you never know... https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/3465819735#L1355

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
There are a couple more failures on the eigensolver tests: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/3477987284.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

ci/.gitlab-ci.yml
```diff
@@ -116,7 +116,7 @@ cpu release build clang13 cxx20:
     DEPLOY_DOCKER_FILE: ci/docker/deploy.Dockerfile
     BASE_IMAGE: docker.io/ubuntu:22.04
     DEPLOY_BASE_IMAGE: docker.io/ubuntu:22.04
-    EXTRA_APTGET: clang-13
+    EXTRA_APTGET: "clang-13 libomp-13-dev"
```
RMeli, line 119:
```suggestion
    EXTRA_APTGET: "clang-13"
```

ci/.gitlab-ci.yml
```diff
@@ -95,8 +95,8 @@ cpu release build clang12:
     DEPLOY_DOCKER_FILE: ci/docker/deploy.Dockerfile
     BASE_IMAGE: docker.io/ubuntu:22.04
     DEPLOY_BASE_IMAGE: docker.io/ubuntu:22.04
-    EXTRA_APTGET: clang-12
-    EXTRA_APTGET_DEPLOY: "glibc-tools"
+    EXTRA_APTGET: "clang-12 libomp-12-dev"
+    EXTRA_APTGET_DEPLOY: "glibc-tools strace"
```
RMeli, line 99:
```suggestion
    EXTRA_APTGET: "clang-12"
    EXTRA_APTGET_DEPLOY: "glibc-tools"
```

ci/docker/release-cpu.yaml
```diff
@@ -10,7 +10,7 @@
 
 spack:
   specs:
-    - dla-future@develop +miniapps +ci-test ^intel-mkl ^mpich
+    - dla-future@develop +miniapps +ci-test mkl=omp ^intel-mkl ^mpich
```
RMeli, line 13:
```suggestion
    - dla-future@develop +miniapps +ci-test ^intel-mkl threads=openmp ^mpich
```

ci/mpi-ctest
```diff
@@ -12,10 +12,10 @@ pushd /DLA-Future-build > /dev/null
 # Run the tests, only output on the first rank
 if [[ $SLURM_PROCID == "0" ]]; then
     TZ=CET date +"Run started at: %H:%M:%S %z"
-    ctest -V $@
+    strace -e clone3 ctest --repeat-until-fail 50 -V $@
```
RMeli, line 15:
```suggestion
    ctest -V $@
```

ci/mpi-ctest
```diff
@@ -12,10 +12,10 @@ pushd /DLA-Future-build > /dev/null
...
     TZ=CET date +"Run finished at: %H:%M:%S %z"
 else
-    ctest -Q $@
+    ctest -Q --repeat-until-fail 50 $@
```
RMeli, line 18:
```suggestion
    ctest -Q $@
```

ci/ctest_to_gitlab.sh
```diff
@@ -67,7 +67,7 @@ JOB_TEMPLATE="
   variables:
     SLURM_CPUS_PER_TASK: {{CPUS_PER_TASK}}
     SLURM_NTASKS: {{NTASKS}}
-    SLURM_TIMELIMIT: '15:00'
+    SLURM_TIMELIMIT: '30:00'
```
RMeli, line 70:
```suggestion
    SLURM_TIMELIMIT: '15:00'
```

src/common/single_threaded_blas.cpp
```diff
@@ -0,0 +1,39 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2022, ETH Zurich
```
RMeli, line 4:
```suggestion
// Copyright (c) 2018-2023, ETH Zurich
```

test/unit/common/test_single_threaded_blas.cpp
```diff
@@ -0,0 +1,52 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2022, ETH Zurich
```
RMeli, line 4:
```suggestion
// Copyright (c) 2018-2023, ETH Zurich
```

RMeli (changes requested):

ci/.gitlab-ci.yml
```diff
@@ -75,7 +75,7 @@ cpu release build gcc11:
     BASE_IMAGE: docker.io/ubuntu:22.04
     DEPLOY_BASE_IMAGE: docker.io/ubuntu:22.04
     EXTRA_APTGET: ""
-    EXTRA_APTGET_DEPLOY: "glibc-tools"
+    EXTRA_APTGET_DEPLOY: "glibc-tools strace"
```
RMeli, line 78:
```suggestion
    EXTRA_APTGET_DEPLOY: "glibc-tools"
```

spack/packages/dla-future/package.py
```diff
@@ -109,7 +109,12 @@ def cmake_args(self):
 
         # BLAS/LAPACK
         if "^mkl" in spec:
-            args.append(self.define("DLAF_WITH_MKL", True))
+            vmap = {"none": "seq", "openmp": "omp", "tbb": "tbb"} # Map MKL variants to LAPACK target name
+            # TODO: Generalise for intel-oneapi-mkl
+            args += [
+                self.define("DLAF_WITH_MKL", True),
+                self.define("MKL_LAPACK_TARGET", "mkl::mkl_intel_32bit_{0}_dyn".format(vmap[spec["intel-mkl"].variants["threads"].value])),
+            ]
```
msimberg, line 117: This can be left as an excercise for another PR, but I just want to note this here.

We should most likely be using something like this instead: https://spack-tutorial.readthedocs.io/en/sc20/tutorial_advanced_packaging.html#accessing-dependency-libraries. This would remove the need for `MKL_LAPACK_TARGET`. We do still need some of saying "we're using MKL" to be able to call `mkl_set_num_threads`. This could be automatically detected, but simpler is probably just to set `DLAF_WITH_MKL` as now since we know in the spack package if `intel-mkl` or `intel-oneapi-mkl` is being used.

test/include/dlaf_test/matrix/util_matrix.h
```diff
@@ -132,7 +132,7 @@ void check(ElementGetter&& expected, MatrixLocal<const T>& mat, ComparisonOp com
     if (!comp(expected(index), mat(index))) {
       ADD_FAILURE_AT(file, line) << "Error at index (" << index
                                  << "): " << err_message(expected(index), mat(index)) << std::endl;
-      return;
+      //return;
```
msimberg, line 135: These are debugging leftovers from me.

test/unit/multiplication/test_multiplication_general.cpp
```diff
@@ -212,6 +213,7 @@ TYPED_TEST(GeneralSubMultiplicationDistTestGPU, CorrectnessDistributed) {
       const TypeParam beta = TypeUtilities<TypeParam>::element(-2.6, .7);
       testGeneralSubMultiplication<TypeParam, Backend::GPU, Device::GPU>(comm_grid, a, b, alpha, beta, m,
                                                                          mb);
+      pika::threads::get_thread_manager().wait();
```
msimberg, line 216: These `wait`s should go away with a merge of master now that https://github.com/eth-cscs/DLA-Future/pull/844 is merged.

spack/packages/dla-future/package.py
RMeli: @msimberg should I address this here, or do you want to raise a separate issue?

msimberg: I'd be happy with a separate issue.

ci/.gitlab-ci.yml
msimberg, line 119: We might need to add `libomp` (or a similar package) for the ROCm environment further below to fix this: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/4196109579.

RMeli:
cscs-ci run

spack/packages/dla-future/package.py
RMeli: https://github.com/eth-cscs/DLA-Future/issues/851

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

test/unit/common/test_single_threaded_blas.cpp
albestro, line 1: Just an idea that might improve readability: you can factor out set/get, so that you have the logic in the single place.

miniapp/miniapp_cholesky.cpp
```diff
@@ -248,9 +249,11 @@ void setUpperToZeroForDiagonalTiles(Matrix<T, Device::CPU>& matrix) {
       continue;
 
     auto tile_set = [](typename Matrix<T, Device::CPU>::TileType&& tile) {
-      if (tile.size().rows() > 1)
+      if (tile.size().rows() > 1) {
```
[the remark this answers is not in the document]
msimberg: Presumably yes because the cholesky miniapp requires square blocks a bit further up?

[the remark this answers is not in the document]
albestro: Good point.

RMeli:
cscs-ci run

test/unit/common/test_single_threaded_blas.cpp
RMeli: Good point! I split it into two tests for better readability, I hope it addresses your comment.

RMeli:
cscs-ci run

RMeli:
cscs-ci run

ci/ctest_to_gitlab.sh
```diff
@@ -67,7 +67,7 @@ JOB_TEMPLATE="
   variables:
     SLURM_CPUS_PER_TASK: {{CPUS_PER_TASK}}
     SLURM_NTASKS: {{NTASKS}}
-    SLURM_TIMELIMIT: '15:00'
+    SLURM_TIMELIMIT: '35:00'
```
RMeli, line 70:
```suggestion
    SLURM_TIMELIMIT: '15:00'
```

ci/ctest_to_gitlab.sh
```diff
@@ -100,7 +100,7 @@ allocate:
   extends: .daint_alloc
   variables:
     PULL_IMAGE: 'YES'
-    SLURM_TIMELIMIT: '40:00'
+    SLURM_TIMELIMIT: '2:00:00'
```
RMeli, line 103:
```suggestion
    SLURM_TIMELIMIT: '40:00'
```

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
Thanks @albestro for the early comments. Switching from "draft" to PR since CI came back all green. I added `^intel-mkl threads=openmp` and `^openblas threads=openmp` to all the specs in `ci/docker`.

I tested locally as well, with the following environments:
```bash
- dla-future@master +miniapps +cuda cuda_arch=89 ^intel-mkl threads=openmp ^mpich device=ch3 netmod=tcp
- dla-future@master +miniapps +cuda cuda_arch=89 ^openblas threads=openmp ^mpich device=ch3 netmod=tcp
- dla-future@master +miniapps +cuda cuda_arch=89 ^openblas threads=none ^mpich device=ch3 netmod=tcp
```

All test pass and test timings look OK.

---

**Note**

Before https://github.com/eth-cscs/DLA-Future/pull/668/commits/de74b2a906bc37b73e8d2c6298d7388109638f94 everything seemed fine with `intel-mkl` but some `openblas` tests (related to the eigensolver) were failing randomly. The affected tests were actually failing consistently, but they were reported by `ctest` as "Passed" (I was not aware of this possible behavior):
```text
test 49
    Start 49: test_gen_eigensolver

49: Test command: /home/rmeli/spack/opt/spack/linux-ubuntu22.04-zen4/gcc-11.3.0/mpich-4.1.1-ykh74fw6bcgmzwkgszkbbwobvwqkhwhm/bin/mpiexec "-n" "6" "/home/rmeli/git/DLA-Future/build/test/unit/eigensolver/test_gen_eigensolver" "--pika:bind=none" "--pika:threads=2"
49: Working Directory: /home/rmeli/git/DLA-Future/build/test/unit/eigensolver
49: Test timeout computed to be: 1500
49: Running main() from gtest_mpipika_main.cpp
49: Running main() from gtest_mpipika_main.cpp
49: Running main() from gtest_mpipika_main.cpp
49: Running main() from gtest_mpipika_main.cpp
49: Running main() from gtest_mpipika_main.cpp
49: Running main() from gtest_mpipika_main.cpp
49: [==========] Running 16 tests from 8 test suites.
49: [----------] Global test environment set-up.
49: [----------] 2 tests from GenEigensolverTestMC/0, where TypeParam = float
49: [ RUN      ] GenEigensolverTestMC/0.CorrectnessLocal
2/2 Test #49: test_gen_eigensolver .............   Passed    1.00 sec
```

msimberg (commented):
I can't approve this myself, but looks good to me now. @RMeli thank you for taking this to the end!

albestro (approved):
LGTM.

Just a minor comment: is it wanted to have `SingleThreadedBlasScope` also in our DLAF test?

Am I wrong saying that it is something we introduced for compatibility when DLAF is used together with other libraries, but we could not care about it in our tests?

msimberg:
> LGTM.
> 
> Just a minor comment: is it wanted to have `SingleThreadedBlasScope` also in our DLAF test?

I would argue that we should use it also in tests. The only place where we could leave it out are `PLAIN` tests where we'd be ok with having BLAS use multiple threads. But even there I'd say we should use it just to exercise the same single-threaded paths that we'd end up using when running in the pika runtime. In `PIKA` tests we'd end up with BLAS threads contending with pika threads.
 
> Am I wrong saying that it is something we introduced for compatibility when DLAF is used together with other libraries, but we could not care about it in our tests?

I'd say in a roundabout way, yes, it's there for compatibility. When building and testing DLA-Future on its own we always have control over the BLAS implementation used. We always assumed that BLAS was single-threaded, even though multi-threaded BLAS was previously "supported". When using DLA-Future in other applications/libraries we can no longer assume that we have the power to choose a single-threaded BLAS, so we have to make sure we're still using a single thread for BLAS in those situations.

RMeli:
>  When building and testing DLA-Future on its own we always have control over the BLAS implementation used.

To add to this point, in this PR I changed the specs in `ci/docker/*.yaml` to either `^intel-mkl threads=openmp` or `^openblas threads=openmp`.

rasolca (approved):
LGTM.

However I'm convinced that at some point a blas call without single thread scope will end up in the code.
Can we have a way to really check at the end of each algorithm test that blas didn't initialized extra threads?

rasolca:
Note: did someone check that all blas calls are in a single thread scope?
Some calls might have been added in the meantime.

RMeli:
> This shouldn't happen, and it might be a wrong test implementation. We should investigate it as it can be source of problems in the future.

Agree. Just to clarify, the test failures were due to some LAPACK calls still missing `SingleThreadedBlasScope`. But the lack of failure from `ctest` is a bit concerning indeed.

> However I'm convinced that at some point a blas call without single thread scope will end up in the code.

Yes, this is probably bound to happen. I plan to have a look at all the open PRs and check that `SingleThreadedBlasScope` is used.

Yesterday I refactored/cleaned up the CP2K code quite a bit, and I think calling `SingleThreadedBlasScope` there isn't too bad after all (this is what @msimberg and Mathieu initially implemented). So a possible alternative to this PR would be to provide the `SingleThreadedBlasScope` for external users but not to use it internally, and eventually provide a ScaLAPACK-like interface (#612) which takes care of this (together with `pika` resume/suspend, #610).

The drawback of the latter approach, though, is that it would be trickier to build and test `dlaf` in a Spack environment concretised with multi-threaded BLAS implementation.

> Note: did someone check that all blas calls are in a single thread scope?

I checked with the latest `master`, but a second pair of eyes wouldn't hurt for sure. The most effective way I found is to use VSCode search, since it provides a list of hits that can be removed/hidden one at a time (once they have been checked).

I searched for `blas::` and `lapack::` calls, as well as `using namespace blas` and `using namespace lapack`.

msimberg:
> > However I'm convinced that at some point a blas call without single thread scope will end up in the code.
> 
> Yes, this is probably bound to happen. I plan to have a look at all the open PRs and check that `SingleThreadedBlasScope` is used.

I agree that we should keep our eyes out for this. However, even better would be an automatic check. I think it might be worth trying to set up a follow-up CI job on one of the multi-threaded configurations to run algorithms miniapps/unit tests through strace and check how many threads are spawned. I'd be happy to do this unless someone else wants to do it. Do you think this sounds reasonable?

> Yesterday I refactored/cleaned up the CP2K code quite a bit, and I think calling `SingleThreadedBlasScope` there isn't too bad after all (this is what @msimberg and Mathieu initially implemented). So a possible alternative to this PR would be to provide the `SingleThreadedBlasScope` for external users but not to use it internally, and eventually provide a ScaLAPACK-like interface (#612) which takes care of this (together with `pika` resume/suspend, #610).

Unfortunately this isn't enough, which is why this PR exists. It works for MKL, but not for other OpenMP-based BLAS-es. It worked ok so far in CP2K because we could just always use MKL. MKL has functions for setting the number of threads both globally and for the current thread (https://github.com/RMeli/cp2k/blob/0e982cab7f38a51db21ea224496cd1ad5e2ffc7c/src/fm/dlaf.cpp#L97 sets it globally). However, for OpenMP `omp_set_num_threads` only applies to the current thread.

PR #670, opened 2022-10-11
Make DLA-Future install/build verbose in CI

msimberg:
I'm doing this in the hope that build errors will be printed slightly more nicely. This targets specifically build errors from building DLA-Future itself. What I'm hoping to improve is e.g. this kind of formatting: https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/jobs/3157080069#L393. However, adding `--verbose` may not help and spack may still decide to wrap everything to 80 columns (or whatever it is there). If that's the case I will instead update this to print `spack-build-out.txt` on failure.

### DIFF
--- a/ci/ctest_to_gitlab.sh
+++ b/ci/ctest_to_gitlab.sh
@@ -41,7 +41,7 @@ allocate:
   extends: .daint_alloc
   variables:
     PULL_IMAGE: 'YES'
-    SLURM_TIMELIMIT: '40:00'
+    SLURM_TIMELIMIT: '1:00:00'
 
 {{JOBS}}
 
@@ -67,7 +67,7 @@ JOB_TEMPLATE="
   variables:
     SLURM_CPUS_PER_TASK: {{CPUS_PER_TASK}}
     SLURM_NTASKS: {{NTASKS}}
-    SLURM_TIMELIMIT: '10:00'
+    SLURM_TIMELIMIT: '15:00'
     PULL_IMAGE: 'NO'
     USE_MPI: 'YES'
     DISABLE_AFTER_SCRIPT: 'YES'
@@ -100,7 +100,7 @@ allocate:
   extends: .daint_alloc
   variables:
     PULL_IMAGE: 'YES'
-    SLURM_TIMELIMIT: '15:00'
+    SLURM_TIMELIMIT: '40:00'
 
 {{JOBS}}
 
@@ -116,7 +116,7 @@ JOB_TEMPLATE="
   variables:
     SLURM_CPUS_PER_TASK: {{CPUS_PER_TASK}}
     SLURM_NTASKS: {{NTASKS}}
-    SLURM_TIMELIMIT: '5:00'
+    SLURM_TIMELIMIT: '10:00'
     PULL_IMAGE: 'NO'
     USE_MPI: 'YES'
     DISABLE_AFTER_SCRIPT: 'YES'
--- a/ci/docker/codecov.Dockerfile
+++ b/ci/docker/codecov.Dockerfile
@@ -29,7 +29,7 @@ RUN spack repo rm --scope site dlaf && \
     spack -e ci concretize -f && \
     mkdir ${BUILD} && \
     ln -s ${BUILD} `spack -e ci location -b dla-future` && \
-    spack -e ci install --keep-stage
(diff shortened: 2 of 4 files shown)


msimberg:
In my opinion https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/jobs/3158440325#L518 is a slight improvement over https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/jobs/3158440325#L684 (the latter is still printed on failures).

ci/.gitlab-ci.yml
```diff
@@ -65,140 +65,140 @@ cpu release build gcc11:
...
-    CXXSTD: 17
-    USE_MKL: "OFF"
-    USE_ROCBLAS: "ON"
-    SPACK_ENVIRONMENT: ci/docker/amdgpu-release-rocm523.yaml
-    BUILD_IMAGE: $CSCS_REGISTRY_IMAGE/release-amdgpu-clang10/build
-    DEPLOY_IMAGE: $CSCS_REGISTRY_IMAGE/release-amdgpu-clang10/deploy:$CI_COMMIT_SHA
-    SLURM_CONSTRAINT: mc
-    THREADS_PER_NODE: 64
-    USE_CODECOV: "false"
+#cpu release build clang12:
```
msimberg, line 68: Revert before merging.

src/init.cpp
```diff
@@ -19,6 +19,8 @@
 #include <iostream>
 #include <memory>
 
+askjdhakshdaksh
```
msimberg, line 22: Revert before merging.

ci/.gitlab-ci.yml
msimberg: Reverted.

src/init.cpp
msimberg: Reverted.

msimberg:
As discussed with @rasolca I've also increased the time limits for slurm jobs (all doubled except for the full allocation time limit which I rounded up to a nice hour). This is done in the hope that some of the jobs that seem to be timing out in the middle of testing won't do so anymore (if e.g. pulling the image takes up a large part of the allocation, not leaving enough time for tests).

ci/ctest_to_gitlab.sh
```diff
@@ -67,7 +67,7 @@ JOB_TEMPLATE="
   variables:
     SLURM_CPUS_PER_TASK: {{CPUS_PER_TASK}}
     SLURM_NTASKS: {{NTASKS}}
-    SLURM_TIMELIMIT: '10:00'
+    SLURM_TIMELIMIT: '20:00'
```
rasolca, line 70: codecov runs got extra time because of report generation that was very slow (with lcov, now with fastcov is better).
Therefore, even keeping the extra 5 min for report generation it would be 15 min.
```suggestion
    SLURM_TIMELIMIT: '15:00'
```

rasolca (approved):

ci/ctest_to_gitlab.sh
```diff
@@ -41,7 +41,7 @@ allocate:
   extends: .daint_alloc
   variables:
     PULL_IMAGE: 'YES'
-    SLURM_TIMELIMIT: '40:00'
+    SLURM_TIMELIMIT: '1:00:00'
```
rasolca, line 44: This is ok if the values in line 54 and 70 are updated according to my comments.

ci/ctest_to_gitlab.sh
```diff
@@ -100,7 +100,7 @@ allocate:
   extends: .daint_alloc
   variables:
     PULL_IMAGE: 'YES'
-    SLURM_TIMELIMIT: '15:00'
+    SLURM_TIMELIMIT: '30:00'
```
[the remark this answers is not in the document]
msimberg: Updated.

ci/ctest_to_gitlab.sh
msimberg: Updated.

ci/ctest_to_gitlab.sh
msimberg: Updated the other ones.

msimberg:
Thanks for the comments @rasolca! I've updated this according to your suggestions.

msimberg:
Success on the first try! Hopefully that's not just a coincidence.

PR #671, opened 2022-10-12
Add LUMI system definitions for benchmarking

msimberg:
Note that the account is the one for CP2K development on LUMI. That may need to be changed depending on who is running the scripts.

I will separately add a script to bind the correct GPUs to ranks (if that's still needed after LUMI is rebooted again...).


rasolca (approved):

msimberg:
Merging without CI since this doesn't affect builds at all.

Issue #601, opened 2022-07-07
Implement distributed tridiagonal solver.


rasolca:
- [x] #659 @albestro
- [x] final sorting of eigenvectors (@teonnik)
- [x] eigenvector formation (@teonnik)
- [x] #675 givens rotation for deflation (@albestro)

rasolca:
Closed by #719

PR #674, opened 2022-10-13
Reduce debug bloat part 2

msimberg:
Not ready (please ignore commented out code), but I'd like to have early feedback from you.

This starts moving explicit template instantiations of some of the communication kernels into cpp files to reduce debug bloat (with `unique_any_sender`) and reduce compile times. I've changed some of the kernels here, and will continue if you are happy with the changes. The main downside of having `unique_any_sender` in the interfaces is that senders have to be explicitly changed to type erased senders at the call sites. If this is something that will happen everywhere we can also consider just directly type erasing e.g. the `Matrix::read(write)_sender` return types.

My test case has expanded from #672 to `time make test_reduction_to_band miniapp_cholesky && ls -lah miniapp/miniapp_cholesky src/libDLAF.a test/unit/eigensolver/test_reduction_to_band` (CPU only). With that the results on master are:

```
real    8m20.364s
user    7m51.750s
sys     0m26.652s
-rwxr-xr-x 1 mjs users 112M Oct 13 12:26 miniapp/miniapp_cholesky
-rw-r--r-- 1 mjs users 949M Oct 13 12:26 src/libDLAF.a
-rwxr-xr-x 1 mjs users 822M Oct 13 12:25 test/unit/eigensolver/test_reduction_to_band
```

and on https://github.com/eth-cscs/DLA-Future/commit/12ce5ce3529fc7a1857bba7050bd21d751fe5d2f:
```
real    7m9.256s
user    6m42.705s
sys     0m24.575s
-rwxr-xr-x 1 mjs users 105M Oct 13 12:16 miniapp/miniapp_cholesky
-rw-r--r-- 1 mjs users 828M Oct 13 12:16 src/libDLAF.a
-rwxr-xr-x 1 mjs users 743M Oct 13 12:16 test/unit/eigensolver/test_reduction_to_band
```
So the reduction in compile times and binary size is about 10-20%. I expect a bit more once all kernels have been changed.

Please comment on the changes if they are readable, understandable etc.

To do:
- [x] run benchmarks to see that there is no performance impact
- [x] change element type in `test_collective_async` to floating point type?

PR #675, opened 2022-10-13
algo: applyGivenRotations Distributed (MC)

albestro:
This is an implementation for one of the sub-tasks needed by #601.

At the time of writing is still in the early phases and I will refine it in the next days, but I hope that this premature publication might help in the integration with the rest of the code @teonnik.

Main points:
- `TileCollector` has been adapted to work also with distributed ranges
- `applyGivensRotationsToMatrixColumns` accepts a `Communicator` of the row, because it is up to the caller the correct management of multiple calls and ensure their independency (this is something we might want to discuss @teonnik)
- minor fix: naming of lambda function in communication kernel

TODO:
- [ ] ? Move implementation together with the other into `tridiag_solver_merge` (and the test too)
- [x] ~Evaluate how easy it is to extend to GPU~ (it will be topic for a separate PR)
- [x] Check if sub-tasking is implemented correctly
- [x] Add more test-cases
- [x] Check it works with incomplete tiles
- [x] Cleanup and refinement

### DIFF
--- a/include/dlaf/communication/kernels/p2p.h
+++ b/include/dlaf/communication/kernels/p2p.h
@@ -70,7 +70,7 @@ template <class CommSender, class Sender>
   using dlaf::internal::whenAllLift;
   using dlaf::internal::withTemporaryTile;
 
-  auto recv = [dest, tag, pcomm = std::forward<CommSender>(pcomm)](auto const& tile_comm) mutable {
+  auto send = [dest, tag, pcomm = std::forward<CommSender>(pcomm)](auto const& tile_comm) mutable {
     return whenAllLift(std::move(pcomm), dest, tag, std::cref(tile_comm)) | transformMPI(send_o);
   };
 
@@ -78,7 +78,7 @@ template <class CommSender, class Sender>
   constexpr Device comm_device_type = CommunicationDevice_v<in_device_type>;
 
   return withTemporaryTile<comm_device_type, CopyToDestination::Yes, CopyFromDestination::No,
-                           RequireContiguous::No>(std::forward<Sender>(tile), std::move(recv));
+                           RequireContiguous::No>(std::forward<Sender>(tile), std::move(send));
 }
 
 template <class CommSender, class Sender>
--- a/include/dlaf/eigensolver/tridiag_solver/merge.h
+++ b/include/dlaf/eigensolver/tridiag_solver/merge.h
@@ -14,10 +14,13 @@
 #include <pika/parallel/algorithms/for_each.hpp>
 #include <pika/unwrap.hpp>
 
+#include "dlaf/common/range2d.h"
 #include "dlaf/eigensolver/tridiag_solver/coltype.h"
 #include "dlaf/eigensolver/tridiag_solver/kernels.h"
 #include "dlaf/lapack/tile.h"
 #include "dlaf/matrix/copy_tile.h"
+#include "dlaf/matrix/distribution.h"
+#include "dlaf/matrix/index.h"
 #include "dlaf/matrix/matrix.h"
 #include "dlaf/multiplication/general.h"
 #include "dlaf/permutations/general.h"
@@ -223,30 +226,43 @@ pika::future<T> calcTolerance(SizeType i_begin, SizeType i_end, Matrix<const T, 
 }
 
 struct TileCollector {
-  SizeType i_begin;
-  SizeType i_end;
+  SizeType idx_begin;
+  SizeType idx_last;
 
 private:
-  template <class T, Device D>
-  std::pair<GlobalTileIndex, GlobalTileSize> getRange(Matrix<const T, D>& mat) {
-    SizeType ntiles = i_end - i_begin + 1;
-    bool is_col_matrix = mat.distribution().size().cols() == 1;
-    SizeType col_begin = (is_col_matrix) ? 0 : i_begin;
+  std::pair<GlobalTileIndex, GlobalTileSize> getRange(const matrix::Distribution& dist) const {
+    SizeType ntiles = idx_last - idx_begin + 1;
+    bool is_col_matrix = dist.size().cols() == 1;
+    SizeType col_begin = (is_col_matrix) ? 0 : idx_begin;
     SizeType col_sz = (is_col_matrix) ? 1 : ntiles;
-    return std::make_pair(GlobalTileIndex(i_begin, col_begin), GlobalTileSize(ntiles, col_sz));
+
+    return std::make_pair(GlobalTileIndex{idx_begin, col_begin}, GlobalTileSize{ntiles, col_sz});
   }
 
 public:
-  template <class T, Device D>
-  auto read(Matrix<const T, D>& mat) {
-    auto [begin, end] = getRange(mat);
-    return matrix::util::collectReadTiles(begin, end, mat);
+  auto iteratorLocal(const matrix::Distribution& dist) const {
+    auto [g_begin, g_size] = getRange(dist);
+    const LocalTileIndex begin{
+        dist.template nextLocalTileFromGlobalTile<Coord::Row>(g_begin.row()),
+        dist.template nextLocalTileFromGlobalTile<Coord::Col>(g_begin.col()),
+    };
+    const LocalTileSize size{
+        dist.template nextLocalTileFromGlobalTile<Coord::Row>(g_size.rows()),
+        dist.template nextLocalTileFromGlobalTile<Coord::Col>(g_size.cols()),
+    };
+    return std::make_pair(begin, size);
   }
 
   template <class T, Device D>
-  auto readwrite(Matrix<T, D>& mat) {
-    auto [begin, end] = getRange(mat);
-    return matrix::util::collectReadWriteTiles(begin, end, mat);
+  auto read(Matrix<const T, D>& mat) const {
+    auto [begin, size] = iteratorLocal(mat.distribution());
+    return matrix::util::collectReadTiles(begin, size, mat);
+  }
+
+  template <class T, Device D>
+  auto readwrite(Matrix<T, D>& mat) const {
+    auto [begin, size] = iteratorLocal(mat.distribution());
+    return matrix::util::collectReadWriteTiles(begin, size, mat);
   }
 };
 
--- a/include/dlaf/eigensolver/tridiag_solver/merge.h
+++ b/include/dlaf/eigensolver/tridiag_solver/merge.h
@@ -225,33 +225,37 @@ pika::future<T> calcTolerance(SizeType i_begin, SizeType i_end, Matrix<const T, 
          di::transform(di::Policy<Backend::MC>(), std::move(tol_fn)) | ex::make_future();
 }
 
-struct TileCollector {
+/// TileCollector is an helper that, given included boundaries of a range [idx_begin, idx_last],
+/// allows to obtain access to all tiles in that range for a matrix.
+///
+/// The range specified is generally considered a square, i.e. the range is applied to both rows and
+/// columns.
+/// Instead, if the matrix has a single column of tiles, the range is considered linear over rows.
+class TileCollector {
   SizeType idx_begin;
   SizeType idx_last;
 
-private:
-  std::pair<GlobalTileIndex, GlobalTileSize> getRange(const matrix::Distribution& dist) const {
-    SizeType ntiles = idx_last - idx_begin + 1;
-    bool is_col_matrix = dist.size().cols() == 1;
-    SizeType col_begin = (is_col_matrix) ? 0 : idx_begin;
-    SizeType col_sz = (is_col_matrix) ? 1 : ntiles;
-
-    return std::make_pair(GlobalTileIndex{idx_begin, col_begin}, GlobalTileSize{ntiles, col_sz});
-  }
-
-public:
   auto iteratorLocal(const matrix::Distribution& dist) const {
-    auto [g_begin, g_size] = getRange(dist);
+    const bool is_col_matrix = dist.nrTiles().cols() == 1;
+
+    const GlobalTileIndex g_begin(idx_begin, is_col_matrix ? 0 : idx_begin);
+    const GlobalTileIndex g_end(idx_last + 1, is_col_matrix ? 1 : idx_last + 1);
+
     const LocalTileIndex begin{
         dist.template nextLocalTileFromGlobalTile<Coord::Row>(g_begin.row()),
         dist.template nextLocalTileFromGlobalTile<Coord::Col>(g_begin.col()),
     };
-    const LocalTileSize size{
-        dist.template nextLocalTileFromGlobalTile<Coord::Row>(g_size.rows()),
-        dist.template nextLocalTileFromGlobalTile<Coord::Col>(g_size.cols()),
+    const LocalTileIndex end{
+        dist.template nextLocalTileFromGlobalTile<Coord::Row>(g_end.row()),
+        dist.template nextLocalTileFromGlobalTile<Coord::Col>(g_end.col()),
     };
+    const LocalTileSize size = end - begin;
+
     return std::make_pair(begin, size);
   }
+
+public:
+  TileCollector(SizeType i_begin, SizeType i_last) : idx_begin(i_begin), idx_last(i_last) {}
 
   template <class T, Device D>
   auto read(Matrix<const T, D>& mat) const {
--- a/include/dlaf/eigensolver/tridiag_solver/rot.h
+++ b/include/dlaf/eigensolver/tridiag_solver/rot.h
@@ -10,64 +10,132 @@
 #pragma once
 
 #include "dlaf/common/assert.h"
+#include "dlaf/common/range2d.h"
+#include "dlaf/common/round_robin.h"
 #include "dlaf/communication/communicator_grid.h"
+#include "dlaf/communication/kernels/p2p.h"
 #include "dlaf/eigensolver/tridiag_solver/merge.h"
+#include "dlaf/matrix/distribution.h"
+#include "dlaf/matrix/index.h"
+#include "dlaf/matrix/matrix.h"
+#include "dlaf/matrix/panel.h"
 #include "dlaf/sender/when_all_lift.h"
 
 namespace dlaf::eigensolver::internal {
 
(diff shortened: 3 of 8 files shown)


include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -223,30 +226,43 @@ pika::future<T> calcTolerance(SizeType i_begin, SizeType i_end, Matrix<const T,
...
+    bool is_col_matrix = dist.size().cols() == 1;
+    SizeType col_begin = (is_col_matrix) ? 0 : idx_begin;
     SizeType col_sz = (is_col_matrix) ? 1 : ntiles;
-    return std::make_pair(GlobalTileIndex(i_begin, col_begin), GlobalTileSize(ntiles, col_sz));
+
+    return std::make_pair(GlobalTileIndex{idx_begin, col_begin}, GlobalTileSize{ntiles, col_sz});
   }
 
 public:
+  auto iteratorLocal(const matrix::Distribution& dist) const {
```
msimberg, line 243: Perhaps we can consider a real type for the `tuple<tuple<GlobalTileIndex, GlobalTileIndex>, tuple<GlobalTileIndex, GlobalTileInde>>` here. However, this is absolutely not a blocker for this PR.

include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -0,0 +1,146 @@
...
+        cps.emplace_back(std::move(cp_recv));
+      }
+
+      // each one computes his own, but just stores either x or y
+      // (or both if are on the same rank)
+      T* x = hasX ? tile_x.ptr(idx_x) : tile_ws.ptr({0, 0});
+      T* y = hasY ? tile_y.ptr(idx_y) : tile_ws.ptr({0, 0});
+
+      tt::sync_wait(ex::when_all_vector(std::move(cps)) |
+                    ex::then([rot, n, x, y]() { blas::rot(n, x, 1, y, 1, rot.c, rot.s); }));
```
msimberg, line 137: Could you briefly explain the need for a `sync_wait` here (as we generally don't need it)?

msimberg (commented):
Just a couple of quick comments at this stage.

include/dlaf/eigensolver/tridiag_solver/rot.h
albestro: I'm not fully sure about this, so thanks for the question that gives me the chance to discuss about it.

Everything happens inside a task, which is given a list of rotations and more relevantly a list of tiles (the one inside the square sub-matrix selection).

Tiles lifetime is ensured till the end of the task. If we `start_detached` the sub-task above, afaik, it can be that the main task finishes releasing the tile, and the sub-task  might operate on a tile that no one is keeping alive anymore.

With `sync_wait`, we ensure that the task will wait for the conclusion of the sub-task. This has a "side-effect" of serialising the loop over rotations, which, afaict, it is what happens also in the current local implementation.

It must be said that it might also not be the best implementation ever (surely it isn't).

As discussed with @rasolca offline this morning, there are other options.
For instance, an option would be to schedule all communications and just later schedule computations (still waiting on completion of needed communications); this would require to allocate more workspaces (currently it is just one, and it represents another limiting factor to the parallelisation, but matches the one imposed by `sync_wait`). 
Another variant would be to avoid allocating one workspaces for each rotation and just allocate a bunch of them and group rotations together and execute them in chunks.
Somehow orthogonally to this, we can also opt to pack them to reduce the number of communications happening.

Anyway, even just considering this basic implementation, I'm really looking forward to your comments in order to understand if I'm missing something and it can be done differently and better. Thanks in advance!

include/dlaf/eigensolver/tridiag_solver/merge.h
albestro: I'm not super proud of this solution. IMO it is somehow overlapping with the existing `common::iterate_range2d`.

include/dlaf/eigensolver/tridiag_solver/rot.h
msimberg: I think this is probably worth an offline discussion so I'll just briefly respond: `let_value` would probably work as a drop-in replacement of the `transform` and you'd get the tiles to live long enough. Then you'd instead return the sender instead of `sync_wait`ing it as the last thing in the function. What might change the discussion is if you have other plans for the `transform` (that creates a task with `givens_rots_fn`), since at the moment it's not necessary. I'm guessing you're planning to implement a GPU version at some point and then the `transform` will bactually be required. But even then perhaps the `transform` can be called inside `givens_rots_fn`. Slightly too many unknowns still at this point to be able to say anything more.

test/unit/eigensolver/CMakeLists.txt
```diff
@@ -74,3 +74,11 @@ DLAF_addTest(
   USE_MAIN MPIPIKA
   MPIRANKS 6
 )
+
+DLAF_addTest(
+  test_tridiag_solver_rot
+  SOURCES test_tridiag_solver_rot.cpp
+  LIBRARIES dlaf.core dlaf.eigensolver dlaf.factorization dlaf.solver dlaf.tridiagonal_eigensolver dlaf.multiplication dlaf.permutations
```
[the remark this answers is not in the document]
albestro: Good point. I copied from the wrong template 😅 

At the moment I have this in a separate file, but in the end we might want to move it to `tridiag_solver_merge` together with its local counterpart (at least until we won't find a better arrangement for the others in there)

include/dlaf/communication/kernels/p2p.h
```diff
@@ -70,15 +70,15 @@ template <class CommSender, class Sender>
   using dlaf::internal::whenAllLift;
   using dlaf::internal::withTemporaryTile;
 
-  auto recv = [dest, tag, pcomm = std::forward<CommSender>(pcomm)](auto const& tile_comm) mutable {
+  auto send = [dest, tag, pcomm = std::forward<CommSender>(pcomm)](auto const& tile_comm) mutable {
```
msimberg, line 73: For your information, I've applied this change here as well: https://github.com/eth-cscs/DLA-Future/pull/674/commits/e879a337bad9b51fa7494e6a5c765f94d32bc097 (since this has moved to a source file the change would possibly be missed otherwise...).

include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -0,0 +1,175 @@
...
+
+        const comm::IndexT_MPI rank_partner = hasX ? rankColY : rankColX;
+
+        // TODO possible optimization, check if it is zero or not
+
+        const T* col_send = hasX ? col_x : col_y;
+        T* col_recv = hasX ? col_y : col_x;
+
+        cps.emplace_back(wrapper::scheduleSendCol<D, T>(comm_row, rank_partner, 0, col_send, m));
+        cps.emplace_back(wrapper::scheduleRecvCol<D, T>(comm_row, rank_partner, 0, col_recv, m));
```
teonnik, line 150: Here you can use `rot.i` and `rot.j` as tags depending on which one is sent and which one is received in the send and receive calls respectively.

include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -0,0 +1,175 @@
...
+      serializer =
+          di::whenAllLift(std::move(serializer), ex::when_all_vector(std::move(cps))) |
+          di::transform(di::Policy<DefaultBackend_v<D>>(), [rot, m, col_x, col_y](auto&&... ts) {
+            if constexpr (D == Device::CPU)
+              blas::rot(m, col_x, 1, col_y, 1, rot.c, rot.s);
+            else
+              // givensRotationOnDevice(m, x, y, rot.c, rot.s, ts...);
+              DLAF_STATIC_UNIMPLEMENTED(T);
+          });
+    }
```
teonnik, line 164: My understanding here is that this is done because you only have enough workspace to store a single vector at a time. That is why the Given's rotation has to be performed just after the communication is done and no other communication can be launched to prevent overriding the workspace. Is that correct? If so, perhaps you can include more details in your comment?

teonnik (changes requested):
Thanks for this @albestro, much appreciated!

An interesting test case that I would like to see before the PR is merged is when a rank sends multiple columns to another rank, that is, multiple vectors from one rank paired with multiple vectors on another rank. Perhaps with your current implementation that is not a problem because the communications are serialized but it may become one in the future if we decide to change this.

I see that alternative approaches to implementing this were already discussed here : https://github.com/eth-cscs/DLA-Future/pull/675#discussion_r994844374 . Indeed, if there are a lot of Givens rotations, we may have to revise the current serial approach in order to avoid the latency. That will indeed require more memory and it will also, probably, force us to pack the data to reduce the number of communications launched at the same time. However, as a first working version, I think your approach is good!

msimberg (commented):
Just putting a provisional "LGTM" here with the expectation that the open TODOs are either taken care of here or issues opened about them (including the potential use of `let_value` in `applyGivensRotationsToMatrixColumns`).

include/dlaf/eigensolver/tridiag_solver/rot.h
albestro: Exactly, since it is serialised it is not a problem...but! I should start using a tag because this can be called many times from the same caller which might pass the same communicator (i.e. without cloning it).

About this, in first instance I would add to the API a way to specify the tag to use for the call, leaving the responsibility to the caller to use independent tags for different calls to this.

I would prefer this solution because deciding tags independency from within it seems to me that we are embedding logic about its usage, which instead should be up to the caller.

What do you think? Would you prefer a different solution?

include/dlaf/eigensolver/tridiag_solver/rot.h
albestro: Yes, exactly. I will follow your advice and document a bit the serialisation 😉

include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -0,0 +1,243 @@
...
+        continue;
+
+      T* col_x = getColPtr(rot.i, hasX);
+      T* col_y = getColPtr(rot.j, hasY);
+
+      const bool hasBothXY = hasX && hasY;
+
+      std::vector<ex::unique_any_sender<>> cps;
+      if (!hasBothXY) {
+        // TODO compute TAG
```
albestro, line 180: As said, my proposal is to get it from the caller, which could pass a tag unique for each call, ensuring the inter-call separation, that will be re-used by all intra-call communications that will happen.

This should™️ work correctly given current implementation details.

But, if in the future we will switch to a smarter/more advanced implementation where intra-call communications run in parallel, then the actual tag used would have to be a function of the current tag used. At that point the caller should ensure that all generated tags starting from the "seed" tag will be independent both intra and inter call...

@teonnik @rasolca what do you suggest to do about this tag?

include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -0,0 +1,243 @@
...
+
+        const comm::IndexT_MPI rank_partner = hasX ? rankColY : rankColX;
+
+        // TODO possible optimization, check if it is zero or not
```
albestro, line 184: I would just go for dropping this TODO without creating an issue for tracking this chance. This algorithm implementation is a very basic one, and we might evaluate in the future if it is worth improving it, and in case how (there are many things that can be done in a smarter way).

include/dlaf/eigensolver/tridiag_solver/rot.h
albestro: @msimberg I added some documentation trying to explain why I think `sync_wait` is currently needed with the current implementation.

We can discuss about it offline if you want to have a support in getting the idea or if you want to share some suggestions. 😉

include/dlaf/eigensolver/tridiag_solver/merge.h
albestro: `TileCollector` is something we might want to look at during API refactoring, but I would not touch it more than what I need for this PR.

Just want to point out that, given two indices describing a bound included range, it actually encompasses two type of behaviours:
- if it is applied to a column matrix, the given range is considered linear, i.e. over rows of the single column
- otherwise, the given range is considered to be a 2D square

I don't like this very much, but I opted for keeping this functionality and just document it a bit in the code to make it more "transparent".

include/dlaf/eigensolver/tridiag_solver/rot.h
msimberg: Thanks! Yes, we can discuss details offline as I'm almost certain this can be rewritten using `let_value`.

msimberg (approved):

include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -0,0 +1,243 @@
...
+
+      if (!hasX && !hasY)
+        continue;
+
+      T* col_x = getColPtr(rot.i, hasX);
+      T* col_y = getColPtr(rot.j, hasY);
+
+      const bool hasBothXY = hasX && hasY;
+
+      std::vector<ex::unique_any_sender<>> cps;
```
msimberg, line 178: Sorry, pressed approve slightly too early. Could we have a slightly more descriptive name for this variable please?

msimberg (changes requested):

include/dlaf/eigensolver/tridiag_solver/rot.h
albestro: `comm_checkpoints`?

msimberg: Yeah, sure, sounds better to me.

include/dlaf/eigensolver/tridiag_solver/rot.h
teonnik: > Exactly, since it is serialised it is not a problem...but! I should start using a tag because this can be called many times from the same caller which might pass the same communicator (i.e. without cloning it).

This is a good point!

> About this, in first instance I would add to the API a way to specify the tag to use for the call, leaving the responsibility to the caller to use independent tags for different calls to this.

Yes, I think this is reasonable!

include/dlaf/eigensolver/tridiag_solver/rot.h
teonnik: Yes, I think the TODO can be dropped.

include/dlaf/eigensolver/tridiag_solver/rot.h
teonnik: > But, if in the future we will switch to a smarter/more advanced implementation where intra-call communications run in parallel, then the actual tag used would have to be a function of the current tag used. At that point the caller should ensure that all generated tags starting from the "seed" tag will be independent both intra and inter call...

Yes, indeed. We will need to discuss how to manage tags across algorithms (if they use the same communicator) and across kernels of the same algorithm in case there is more than one kernel using `MPI_Isend` and `MPI_Irecv`. However, I think both of these questions are beyond the scope of this PR. The solution to have tag as part of the API so that a user can pass it to the function is IMO good.

I opened an issue about this : https://github.com/eth-cscs/DLA-Future/issues/700

test/unit/eigensolver/test_tridiag_solver_rot.cpp
```diff
@@ -0,0 +1,112 @@
...
+    SizeType m;
+    SizeType mb;
+    SizeType i_begin;
+    SizeType i_last;
+    std::vector<di::GivensRotation<U>> rots;
+  };
+
+  const std::vector<config_t<T>> configs{
+      // range with one-sided margin
+      {9, 3, 1, 2, {di::GivensRotation<T>{3, 8, 0.5f, 0.5f}}},
```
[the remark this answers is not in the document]
albestro: Better? 😄 https://github.com/eth-cscs/DLA-Future/pull/675/commits/88366074ea47dc8f9f136ccc1af7e10744c2fd83

include/dlaf/eigensolver/tridiag_solver/rot.h
albestro: In the meanwhile, I did this https://github.com/eth-cscs/DLA-Future/pull/675/commits/a0d549ba75c2a9aac4e66a6ab4aaddfeb31f7a6d

msimberg (approved):

teonnik (approved):

rasolca (approved):

rasolca:
bors try

albestro:
Going to fix build problems with GCC11 that I didn't get because I was working with GPU and GCC9.

albestro:
@rasolca this is ready to be tested again, I should have fixed the warnings 🤞🏻

rasolca:
bors try

rasolca:
bors try-

albestro:
bors try

hopefully last fix for warnings

albestro:
> ## try
> Build failed:
> 
> * [ci/gitlab/full-pipeline](https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/pipelines/689932826)

It seems there are problems, apparently MC test in debug hangs...(all others work).
Tomorrow I'll try to see if I can reproduce that, and eventually I'll try to debug it...

albestro:
bors try

🤞

Bug fix for incomplete tiles https://github.com/eth-cscs/DLA-Future/pull/675/commits/e4c5de661284b178f830385232640e92d5103d6b

Added also a couple more of incomplete tile test-cases https://github.com/eth-cscs/DLA-Future/pull/675/commits/f704d42b549ab7abe2c2dcc7279ee63edfd19621

Additionally, fixed wrong usage of `whenAllLift` and `select` (see #711) https://github.com/eth-cscs/DLA-Future/pull/675/commits/85507f8493a020e282d57f7346c6143bb4788a4a

albestro:
> ## try
> Build failed:
> 
> * [ci/gitlab/full-pipeline](https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/pipelines/691425491)

I don’t know, I will have a look tomorrow.

albestro:
bors try

I erroneously reverted the fix, now it should be the right one 🤞

albestro:
bors try-

albestro:
bors try

in a last check of the code I found the same problem existed for the workspace allocation too, even if it was not as problematic as the first one.

PR #677, opened 2022-10-14
Update runners and actions in GitHub actions workflows

msimberg:
I'm hoping this will "just work", but let's see.

This is in reaction to https://github.com/actions/runner-images/issues/6002, i.e. the `ubuntu-18.04` runners being deprecated (and they will be removed at the end of the year).

### DIFF
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions/javascript-action@v1
 
       - name: Install tools
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install tools
         run: |
           sudo apt-get install --no-install-recommends clang-format-10 python3
-          pip3 install black==21.8b0
+          pip3 install black==22.10.0
           pip3 install cmakelang==0.6.13
 
       - name: Fetch master
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


msimberg:
The old `black` version was incompatible with the latest `click` (whatever that is), see https://github.com/psf/black/issues/2964. I've bumped the `black` version, but that leads to a (one-line) formatting change. Are you happy with that or would you prefer trying to downgrade `click` instead (I haven't tried if that would work as well)?

.github/workflows/check_format.yml
```diff
@@ -15,7 +15,7 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
rasolca, line 18: Please specify the version (I suppose latest means `22.04` now).
Otherwise we might unexpectedly end up with incompatible versions.

rasolca:
> The old `black` version was incompatible with the latest `click` (whatever that is), see [psf/black#2964](https://github.com/psf/black/issues/2964). I've bumped the `black` version, but that leads to a (one-line) formatting change. Are you happy with that or would you prefer trying to downgrade `click` instead (I haven't tried if that would work as well)?

I would update the source

.github/workflows/check_format.yml
msimberg: Ok, done. Are you ok with the `check_license_header.yaml` workflow uses `ubuntu-latest`? It doesn't install anything so there's a smaller chance of surprises there, but still.

msimberg:
Interestingly `clang-format-10` doesn't seem to be available on `ubuntu-22.04` but was there on `ubuntu-latest`. Are you also up for a `clang-format` version bump?

rasolca:
> Interestingly `clang-format-10` doesn't seem to be available on `ubuntu-22.04` but was there on `ubuntu-latest`. Are you also up for a `clang-format` version bump?

Not sure it is the right time to do it... IIRC 10 and 11 were very similar but 12 was producing quite different formats...
I suggest to keep ubuntu 20.04 until the end of the year.

msimberg:
All workflows now use `ubuntu-20.04` explicitly. The `black` version was updated (with a one-line formatting change). I also took the chance to update the `checkout` actions, since the old version is using a deprecated version of node. The `javascript-action` action is also still using a deprecated version of node, but there doesn't seem to be a newer version available yet so that is unchanged.

rasolca (approved):

PR #260, opened 2020-07-10
notify github properly

haampie:

PR #240, opened 2020-06-05
Add back warnings as errors for CI

albestro:
Add back warnings as errors, it has been missed and left out during migration from previous CI system.

I will address in this PR also minor changes needed to solve warnings, if present.

PR #678, opened 2022-10-18
Use ci-ext

finkandreas:
To do:
- [x] restore artifical dependencies between jobs? nope


finkandreas:
This is a working version of using CI with a ci-ext setup:
Changes:
- do not use JFrog anymore, but the gitlab registry (both should work, but yesterday JFrog was down, could be changed to JFrog registry again)
- use as k8s runner the rosa cluster, which does not offer docker but podman
- removed notification script, it's done automatically
- referenced the templates in `ci-ext` as external include, although it is nowhere used. It is rather meant as a "documentation", because there are podman templates how to build containers on this cluster. Can be removed.
- explicitly state images with the `docker.io` prefix, because podman has multiple default registries and needs a tty to choose the correct one
- removed the `only` clause, because branch restriction should be done on the ci-ext setup page

ci/.gitlab-ci.yml
```diff
@@ -33,12 +30,37 @@ stages:
...
+           --format docker --layers --isolation chroot
+           --build-arg BASE_IMAGE
+           --build-arg BUILDKIT_INLINE_CACHE=1
+           --build-arg SPACK_SHA
+           --build-arg EXTRA_APTGET
+           --build-arg COMPILER
+           --build-arg CXXSTD
+           --build-arg SPACK_ENVIRONMENT
+           --build-arg SPACK_DLAF_REPO --build-arg USE_MKL
+           --build-arg NUM_PROCS=$(echo "$(cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us) $(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us) $(nproc)" | awk ''{ print ($1/$2 > 0) ? int($1/$2+0.5) : $3; }'')
```
msimberg, line 47: As mentioned by @finkandreas, we should pass `NUM_PROCS` to `spack install` (the nodes have 256 cores (/hyperthreads?) but only 64 are allocated per pod. However, `nproc` reports 256 and the `/sys/fs/cgroup` stuff reports 64 (@finkandreas please correct me if I misunderstood).

I can push this change.

msimberg:
This is really nice, and based on the very statistically significant (not) first runs this is already much more reliable than the current CI.

As I understood it, the current caveats are:
- The `rosa` cluster may occasionally be down/reorganized if someone decides they have a better use for those nodes, but that should hopefully not be worse than the current stability issues on `fulen`. :shrug:
- Restarting jobs can only be done by gitlab users who are in @finkandreas's special group or "top-level" admins (which I think only @rasolca is of us). Assuming things are more stable this will hopefully not be an issue, but even then we at least have one person who can restart jobs. @finkandreas is thinking about ways to handle this (e.g. having a way to restart jobs through the custom CI interface (i.e. not through gitlab, which is an implementation detail)).

ci/.gitlab-ci.yml
finkandreas: You are absolutely right. NUM_PROCS works as is on both bare metal and inside a cgroup environment. `nproc` would give the physical amount of cores, independent what the k8s restrictions are. On bare metal `/sys/fs/cgroup/cpu/cpu.cfs_quota_us` is typically `-1`, i.e. we fallback to `nproc`.
This variable should of course be used inside the Dockerfile / spack environment file under the key `config:build_jobs`, or if you use the new Makefile approch then start installation of spack packages with `make -j $NUM_PROCS`.

msimberg:
bors try

finkandreas:
Yes, `rosa` should be a "stable" system because it is important for testing the next generation jupyter setup, however in the past it was shutdown for 3 days.

Yes, job retry would be only possible in the current implementation for @rasolca. Retrying jobs should not be necessary because in a perfect world everything just works™.

ci/.gitlab-ci.yml
rasolca: I have a couple of question as I'm afraid we can go easily out-of-memory...
I suppose that they are AMD nodes with two 64-cores sockets (x2 SMT).
How much memory do these nodes have?
64 cores are allocated per pod. Does this mean 32 real cores (x2 SMT) or 64 real cores? Or stated in a different way: are 2 or 4 pods per node allowed?

rasolca:
Looks good. A couple of questions:
@finkandreas 

I suppose that the last commit was not tested yet due to site-wide maintenance. Am I correct?

> * removed notification script, it's done automatically

Harmen added the notification stage because multi-stage pipeline were messing up the automatic notification (i.e. a failed sub-pipeline was reporting failed, but a later passing pipeline could overwrite the notification with passed). Is this problem solved now?

> * removed the only clause, because branch restriction should be done on the ci-ext setup page

How can we access the setup page?

I noticed that https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/3190759609 failed due to the 1h pod timeout. The rocm build image is really slow to build and need a longer timeout (I think that Harmen increased it to 3h or even 6h). Is it something has to be set globally or can be set per job or task?

I also noticed that you tried to add caching, but then removed it. What is the problem?
Otherwise, is it possible to use `--cache-from=$CI_REGISTRY_IMAGE:latest` to have something similar to the previous caching behaviour?

> Yes, job retry would be only possible in the current implementation for @rasolca. Retrying jobs should not be necessary because in a perfect world everything just works™.

A very optimistic view of the world... 😉

msimberg:
@finkandreas also tentatively removed the artifical job dependencies that were introduced to avoid too many concurrent jobs on the old setup. In theory this should now not be a problem, but we may need to observe the behaviour for a while to see if it's ok. Alternatively, we may just keep the dependencies anyway to avoid unnecessary builds of certain configurations if we already know that the "basic" configurations fail (which I think was a second reason for introducing the dependencies anyway). I've added a to do item in the PR description.

msimberg:
Even if everything "seems" to work ok at the moment, I would let this sit here unmerged for at least a week of CI retries to gather some data on how stable this is.

rasolca:
Note:



> To do:
> 
>     * [ ]  restore artifical dependencies between jobs?

This shouldn't be needed. It was introduced to limit the load and decrease the chances of errors with the previous setup.

msimberg:
> Note:
> 
> > To do:
> > ```
> > * [ ]  restore artifical dependencies between jobs?
> > ```
> 
> This shouldn't be needed. It was introduced to limit the load and decrease the chances of errors with the previous setup.

I added it with a question mark because while I think that was the primary reason, the secondary reason was to just avoid builds of other configurations if we know that the "easy" configurations fail to build. I think for throughput having them all run in parallel would be great, so I'm all for not having the dependencies, but I just wanted to mention this second reason there.

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

msimberg:
bors try

ci/.gitlab-ci.yml
rasolca: My concern the cores being too much, but memory that can be too few:
512GB means 2GB/thread and it is not much as we are also building in memory.
For images (which now are built in memory) I count ~5GB and when building DLAF with debug symbols each process needs more than 2GB. (once I was testing the dockerfile and make -j 4 was failing with 8GB of ram...).
So we can easily consume a lot of memory if many of our jobs are put on the same node with few cores (therefore less available memory) each.

finkandreas:
> The issue was with multi-stage pipelines, and we should make sure it has been fixed.
Some extra info: https://github.com/eth-cscs/DLA-Future/pull/260

Hm, I still don't understand what this has to do with multi-stage pipelines. I think your request is to not only get a notification on the SHA that bors creates but also additionally on the SHA of the latest PR commit. Right now, you'll only get a notification on the bors-created SHA. The parent (latest commit of PR) will not get a status notification.
Maybe for clarification what `bors-try` is doing: It creates a new branch named `trying` which is a the `master` branch with PR-branch merged. This of course is a new commit SHA, and this is the commit SHA that will be notified. Because bors also drops a comment with the commit sha, it magically appears in github on the PR website and you see the status of the build.
What is the reason why you also need the (same) status on the latest PR commit?

rasolca:
What you are describing is the nice to have feature that was implemented on top of the solution of the main problem.

### The main problem:

Unfortunately I don't remember all the details but I'll try to explain everything I remember (maybe @albestro remembers something more and can expand it):
Assume you have a multi-stage pipeline P which spawns SP1 and SP2, and SP1 fails while SP2 passes.
There was the possibility that the result of the full pipeline P was reported as passed by github but I don't remember the exact details anymore.

### The nice to have feature:

It was solving the problem that a commit was marked with a green tick even if the bors try on it failed, making very hard to understand what failed and what passed. E.g. in #240 
Screenshot 2022-10-19 at 12 54 50
Not sure if it is still relevant with ci-ext.

finkandreas:
>  Not sure if it is still relevant with ci-ext.

If you continue to use bors then yes, this is still relevant, and you would not get automatic CI notifications on the last PR commit, but only on the bors created SHA.

> Assume you have a multi-stage pipeline P which spawns SP1 and SP2, and SP1 fails while SP2 passes.
There was the possibility that the result of the full pipeline P was reported as passed by github but I don't remember the exact details anymore.

Hm, so you're saying that the status of the pipeline in gitlab was a different status than the notified status in github? I have not seen that yet, but I also have not experimented too much with multi-stage pipelines.
All I know for now is that the pipeline status of the gitlab pipeline was the same to what is reported to github.
All I see that is done in https://github.com/eth-cscs/DLA-Future/pull/260 is that also the parent commit is notified when the ref name is staging or trying.

rasolca:
> > Assume you have a multi-stage pipeline P which spawns SP1 and SP2, and SP1 fails while SP2 passes.
> > There was the possibility that the result of the full pipeline P was reported as passed by github but I don't remember the exact details anymore.
> 
> Hm, so you're saying that the status of the pipeline in gitlab was a different status than the notified status in github? I have not seen that yet, but I also have not experimented too much with multi-stage pipelines. All I know for now is that the pipeline status of the gitlab pipeline was the same to what is reported to github. All I see that is done in #260 is that also the parent commit is notified when the ref name is staging or trying.

I checked the first part of the changes of #260 and I think I now remember:
bors was adding the comment `try: build succeded` or `try: build failed` already when the first sub-pipeline was finished reporting the "partial" result as the "total" result.

As I hope to get rid of bors this shouldn't be a problem. Sorry for the noise...

@msimberg we should avoid to use bors in this PR and check how to configure `ci-ext` in the best way.

finkandreas:
This might help to understand when pipelines are triggered in `ci-ext`: https://gitlab.com/finkandreas/doc_cscsci/-/tree/main#understanding-when-ci-is-triggered

If you do not use bors at all and rely on automatic testing on each PR commit for whitelisted users / manual testing by comment for non-whitelisted users then there would not be this bors commit showing up in a PR, and the pipeline would really be running on the PR commit.

msimberg:
I think since CI is quite expensive we may not want automatic triggering, but possibly only manual testing by comment. On the other hand we (I) can also try to be more careful about not pushing potentially incomplete changes. On the other hand we also tend to open draft PRs for early comment which may not need testing on every push. This would also speak for manual triggering by comment. On the third hand, automatic testing would be really nice, but I think we have to do some work to speed up CI before we can do that. @rasolca do you agree with this summary?

@finkandreas is it possible to _only_ trigger by comment?

finkandreas:
> @finkandreas is it possible to only trigger by comment?

Sure, set the list of branches to `empty` and then a pipeline is only triggered by comment.

If you want automatic testing + full testing, I would suggest to have two pipelines, the first one being "cheap" and targeting the branch `master` (and as such also every PR targeting `master` would be tested automatically) and an expensive one, with an `empty` branch set, i.e. can only be triggered by comment.

msimberg:
Let's see if it's too soon...

cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

rasolca:
Limiting to make -j 16 the deploy image get a significant speedup. The build part has not be tested as the build image are all available.
How do we proceed? Do we set them as CI variable in gitlab interface, such that we can easily change both parameters without bloating the repo history?

msimberg:
> Limiting to make -j 16 the deploy image get a significant speedup. The build part has not be tested as the build image are all available. How do we proceed? Do we set them as CI variable in gitlab interface, such that we can easily change both parameters without bloating the repo history?

That sounds like a nice idea. The only possible concern I have is if the CI environment ever decides to allocate _less_ than 16 cores per CI job, do we then also want to use fewer than 16 jobs for builds? I'm thinking something like `NUM_PROCS=min(16, <current core count calculation>)`. This is a pretty minor concern though and if it's a CI variable we can easily change it if we find out there are fewer resources available for jobs. @finkandreas is this something we need to care about?

finkandreas:
I don't think that you will ever see less than 16 cores.
I find it difficult to compare the two runs, because the one with 64 cores was doing more work.

I'd be surprised if 16 is faster than 64 when both have to rebuild the base image. Probably 64 is not much faster than 16, but it should definitely not be slower when the same work is done.

msimberg:
> I don't think that you will ever see less than 16 cores. I find it difficult to compare the two runs, because the one with 64 cores was doing more work.
> 
> I'd be surprised if 16 is faster than 64 when both have to rebuild the base image. Probably 64 is not much faster than 16, but it should definitely not be slower when the same work is done.

Ok, good to know. As @rasolca already pointed out memory may in fact be an issue, so if it starts swapping (do the build nodes have swap...?) with 64 jobs then it'd for sure be slower than 16 jobs. @rasolca it sounds like we could just go ahead with your suggestion of making it a variable set through the GitLab interface (which we can only do now because we're special users...). Then we can play around with it a bit to see if 24 or 32 or something else would actually still give an improvement.

rasolca:
cscs-ci run

rasolca:
> I don't think that you will ever see less than 16 cores. I find it difficult to compare the two runs, because the one with 64 cores was doing more work.

Ups... I wrongly supposed that only `cuda codecov build gcc11` rebuilt the build image, and other tests only built the deploy image. 🙈 So in general it is fast...

Anyway something happened with `cuda codecov build gcc11` as it started with the other (the timestamp when timezone was installed is about the same (minute more or less)) but it took six times longer.

> I'd be surprised if 16 is faster than 64 when both have to rebuild the base image. Probably 64 is not much faster than 16, but it should definitely not be slower when the same work is done.

I made two variables and set 32 for the build (dependencies) and 16 for deploy (dlaf). This configuration should be a little bit more conservative on memory.

rasolca:
> @rasolca it sounds like we could just go ahead with your suggestion of making it a variable set through the GitLab interface (which we can only do now because we're special users...).

Mhm... Does this mean that a user which is the admin of a repo doesn't have access to the gitlab mirror of the repo?
Or there will be other mechanisms to prevent variables to be set?

rasolca:
cscs-ci run

I wrongly named the variable in gitlab 🙈, but doing so I discovered that the coverage (=debug) tests were not limited. 😱

rasolca:
@msimberg I suppose that cscs CI invocations are not set up to work as bors invocations with any message after them.
Is it possible to set it up?

rasolca:
cscs-ci run

rasolca:
cscs-ci run

rasolca:
cscs-ci run

Let try with a random comment

msimberg:
> > @rasolca it sounds like we could just go ahead with your suggestion of making it a variable set through the GitLab interface (which we can only do now because we're special users...).
> 
> Mhm... Does this mean that a user which is the admin of a repo doesn't have access to the gitlab mirror of the repo? Or there will be other mechanisms to prevent variables to be set?

In principle yes. @finkandreas can give you the longer answer, but at least in theory GitLab is an implementation detail and it should be possible to control everything through the custom CI interface that doesn't have anything to do with GitLab. Of course, at the moment we can assume that things are running through GitLab...

ci/docker/codecov.Dockerfile
```diff
@@ -22,14 +22,15 @@ SHELL ["/bin/bash", "-c"]
...
+ARG NUM_PROCS
 # Note: we force spack to build in ${BUILD} creating a link to it
 RUN spack repo rm --scope site dlaf && \
     spack repo add ${SOURCE}/spack && \
     spack -e ci develop --no-clone -p ${SOURCE} dla-future@develop && \
     spack -e ci concretize -f && \
     mkdir ${BUILD} && \
     ln -s ${BUILD} `spack -e ci location -b dla-future` && \
-    spack -e ci install --keep-stage --verbose
+    spack -e ci install --jobs ${NUM_PROC} --keep-stage --verbose
```
msimberg, line 33:
```suggestion
    spack -e ci install --jobs ${NUM_PROCS} --keep-stage --verbose
```

msimberg:
cscs-ci run

msimberg (approved):

PR #679, opened 2022-10-19
Add workaround to have transform handle senders that send nothing

msimberg:
@albestro had a use case where `transform` would take a sender that sends no types/values. This does not compile because `pika::unwrapping` can't handle nullary callables (`pika::unwrapping([]{})()` does not compile). Since we have no plans to fix `pika::unwrapping` I'm adding a temporary workaround in DLA-Future which simply skips unwrapping in the nullary callable case (since it wouldn't do anything anyway).

@albestro could I ask you to check if this works in your cases as well?

### DIFF
--- a/include/dlaf/sender/transform.h
+++ b/include/dlaf/sender/transform.h
@@ -26,6 +26,8 @@
 #include <pika/cuda.hpp>
 #endif
 
+#include <type_traits>
+
 namespace dlaf {
 namespace internal {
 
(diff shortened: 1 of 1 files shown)


albestro:
Thank you very much @msimberg! It works and the "workaround" with a fake parameter is not needed anymore.

Thanks for the super quick and neat fix!

msimberg:
After I did this change I also realized that we could possibly even remove `unwrapping` altogether since we only want one level of unwrapping. We could instead explicitly unwrap futures the way we do for `reference_wrapper`. I think this would get rid of the need for the custom `Unwrap` template parameter as well. It would maybe very slightly improve the debug bloat situation as well...

msimberg:
bors try

albestro (approved):

rasolca (approved):

msimberg:
bors try-

msimberg:
bors try

Issue #498, opened 2022-03-14
Towards an unified ETI usage


albestro:
With #172 Explicit Template Instantiation (ETI) has been introduced in our codebase.

Since then, we've adopted it for various parts of our codebase, and different point of views of its usage arose.

Main objective of this issue is to discuss and define our preferred solution, so that we can:
- unify all existing usages
- document a reference for the future implementations

From what I can say looking at the codebase, there are mainly two "methods" used:
- method A: public header include template definition
- method B: public header just provide symbol definition

Apart known ETI limitations already discussed in #169, both methods have their pros and cons.

I will try to quickly describe them in next paragraphs, but they are not meant to be exhaustive and they are there just for starting a discussion and collect all information and point of views, so we can go for the better solution in the end.

### Method A
Example: https://github.com/eth-cscs/DLA-Future/blob/master/include/dlaf/auxiliary/norm.h

<img width="400" alt="image" src="https://user-images.githubusercontent.com/9337627/158196235-e09bf3c9-6981-47b8-b69a-93525f843348.png">

- ✅ public functions are instantiated, not internal implementations
- ✅ by instantiating public functions, it is possible to have partial implementations, since just used internals gets instantiated
- ❌ template code is included in use-code (parsed by compiler)
- 🤔 user can instantiate algorithms for types different from what ETI already provides

### Method B
Example: https://github.com/eth-cscs/DLA-Future/blob/master/include/dlaf/factorization/cholesky.h

<img width="400" alt="image" src="https://user-images.githubusercontent.com/9337627/158198265-1673e5a0-1d6d-4db3-b238-102de55d6966.png">

- ✅ template code is not included in user-code (not parsed by compiler)
- ❌ instantiating internal structures makes more difficult to have "partial" implementations
 (e.g. CPU both local/distributed and GPU just local)
- 🤔 user can NOT instantiate algorithms for types different from what ETI already provides

PR #683, opened 2022-10-21
algo: Reduction to band Distributed GPU

albestro:
Adapt current implementation to work in GPU too.

Changes:
- Refactor and extend `ComputePanelHelper` to work with GPU (as for local, it will compute the panel on CPU)
- Algorithm trigger preventing deadlocks had to be adapted for the GPU implementation (notes and doc will come later)
- introduce `D` static attribute for both `Matrix` and `Panel`
- adapt test facility to GPU and add Distributed GPU test

TODO
- [x] Check necessity of `D` static attribute introduction
- [x] Document trigger problem (both for GPU and CPU)
- [x] Differentiate trigger configuration between implementations (due to different requirements)/

### DIFF
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -57,6 +57,7 @@ template <class T, Device D>
 class Matrix : public Matrix<const T, D> {
 public:
   static constexpr Device device = D;
+
   using ElementType = T;
   using TileType = Tile<ElementType, D>;
   using ConstTileType = Tile<const ElementType, D>;
@@ -150,6 +151,8 @@ private:
 template <class T, Device D>
 class Matrix<const T, D> : public internal::MatrixBase {
 public:
+  static constexpr Device device = D;
+
   using ElementType = T;
   using TileType = Tile<ElementType, D>;
   using ConstTileType = Tile<const ElementType, D>;
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -56,6 +56,7 @@ auto selectGeneric(Func&& f, common::IterableRange2D<SizeType, LocalTile_TAG> ra
 template <class T, Device D>
 class Matrix : public Matrix<const T, D> {
 public:
+  static constexpr Device device = D;
   using ElementType = T;
   using TileType = Tile<ElementType, D>;
   using ConstTileType = Tile<const ElementType, D>;
--- a/include/dlaf/eigensolver/reduction_to_band.h
+++ b/include/dlaf/eigensolver/reduction_to_band.h
@@ -7,6 +7,7 @@
 // Please, refer to the LICENSE file in the root directory.
 // SPDX-License-Identifier: BSD-3-Clause
 //
+
 #pragma once
 
 #include "dlaf/common/vector.h"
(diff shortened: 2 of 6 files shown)


include/dlaf/matrix/matrix.h
```diff
@@ -56,10 +56,12 @@ auto selectGeneric(Func&& f, common::IterableRange2D<SizeType, LocalTile_TAG> ra
 template <class T, Device device>
 class Matrix : public Matrix<const T, device> {
 public:
+  static constexpr Device D = device;
```
msimberg, line 59: I know I've made this same change earlier in `Tile`, but now I'm wondering if the names should be swapped like this:
```suggestion
template <class T, Device D>
class Matrix : public Matrix<const T, D> {
public:
  static constexpr Device device = D;
```
(and of course with the same change in `Tile`, and now `Panel`). I think more often than not we template as `template <class T, Device D>` (and it fits with template parameters being CapitalCase), and variables are usually snake_case (like `coord` in `Panel`).

Absolutely not a blocker, but wanted to mention it here since you have it as one of the points in the description.

albestro: If @rasolca agree on your proposal, I will quickly create a PR for that 😉

(I still haven't checked if I really need it in this PR, but I think it can be generally useful to be able to access this information.)

albestro:
8baa1ea18359b11e1741246750f0d964cc512501 here the trigger specialisation has been implemented and documented.

It is a quite specific workaround and it might still create some problem
https://github.com/eth-cscs/DLA-Future/blob/8baa1ea18359b11e1741246750f0d964cc512501/include/dlaf/eigensolver/reduction_to_band/impl.h#L1165

I will ask you reviewers @msimberg @teonnik @rasolca to read this doc and check if you are able to get a rough idea (I understand it is not so straightforward to fully get it).

For the still open problem with the TODO, I would propose to open an issue for that.

msimberg:
> I will ask you reviewers @msimberg @teonnik @rasolca to read this doc and check if you are able to get a rough idea (I understand it is not so straightforward to fully get it).
> 
> For the still open problem with the TODO, I would propose to open an issue for that.

I think the explanation makes sense at least on a high level. The TODO is really a problem though and long term we should try to fully move away from blocking communication, even if it might mean a performance penalty (but how much of a penalty we're willing to accept is another thing...).

include/dlaf/matrix/matrix.h
albestro: I opened #685 trying to address the minimal set that would be useful for this.

As soon as that one will be merged, I will rebase this one and eventually do the last changes here if needed (i.e. checking if adding `device` attribute to `Matrix` and `Panel` is actually needed)

msimberg (approved):

albestro:
bors try

albestro:
> 

I had to manually re-run a build task because of k8s but, apart that minor problem, all 🟢🟢🟢

teonnik (approved):
LGTM!

rasolca (changes requested):
ETI has not been updated.

albestro:
> ETI has not been updated.

Good catch! 🥇 Thanks!

albestro:
> ETI has not been updated.

🏆 Good catch!

In https://github.com/eth-cscs/DLA-Future/pull/683/commits/c177ee9a594f9bd004ad5acb55f98d3f7691fbfe I just simplified what was already there.

About improving the ETI, there is already an issue that keeps track of the issue #498 (just saying so that this strange ETI usage gets linked for future reference)

rasolca (approved):

rasolca:
bors try

rasolca:
bors try

rasolca:
bors try-

rasolca:
bors try

PR #685, opened 2022-10-25
Renaming D <-> Device (mainly for Matrix and similar/related)

albestro:
Following https://github.com/eth-cscs/DLA-Future/pull/683#discussion_r1003142943

I tried to keep it as "minimal" as possible, since I don't want to interfere much with others PRs, and moreover there is no need to actually fully address this issue right now.

Main points:
- Rename `device` to `D` as template parameter for Matrix, Tile, Panel and related
- Normalize Device D template in `util_blas`
- MatrixMemory ETI for GPU

Template naming will still not be fully consistent after this PR.

Just a couple of examples:
- not all algorithms follow the `Device D` template parameter naming
- `Backend backend` has not been yet addressed

Moreover, as a side discussion, it might be relevant to define (for a later PR), if we want to accomunate also `using T = Matrix::ElementType` to the snake_case convention (as we are apparently aiming for `constexpr Device device = Matrix::device`).

### DIFF
--- a/include/dlaf/communication/broadcast_panel.h
+++ b/include/dlaf/communication/broadcast_panel.h
@@ -49,8 +49,8 @@ std::pair<SizeType, comm::IndexT_MPI> transposedOwner(const matrix::Distribution
 ///                     on other ranks it is the destination panel
 /// @param serial_comm  where to pipeline the tasks for communications.
 /// @pre Communicator in @p serial_comm must be orthogonal to panel axis
-template <class T, Device device, Coord axis, class = std::enable_if_t<!std::is_const_v<T>>>
-void broadcast(comm::IndexT_MPI rank_root, matrix::Panel<axis, T, device>& panel,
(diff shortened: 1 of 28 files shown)


msimberg (approved):

msimberg:
> Moreover, as a side discussion, it might be relevant to define (for a later PR), if we want to accomunate also `using T = Matrix::ElementType` to the snake_case convention (as we are apparently aiming for `constexpr Device device = Matrix::device`).

I think this is still consistent with the current style guide. `constexpr Device device = Matrix::device` is a variable, whereas `using T = Matrix::ElementType` is a type (alias). I think template parameters are not explicitly mentioned in the style guide which means that things like `Backend B` or `Backend backend` are both ok (neither is ok!?).

albestro:
bors try

rasolca (approved):

teonnik (approved):
LGMT!

PR #686, opened 2022-10-25
Build static libraries of all the intermediate DLA-Future object libraries

msimberg:
This fixes #487.

This is down the rabbit hole of #674. In its current state (i.e. without this PR) #674 reduces the binary size of _some_ binaries, but increases it for many. The reason for this is that object libraries are unconditionally included in executables and libraries taking no regard to whether or not they're needed. Since communication kernels were added to `dlaf.core`, the size of `dlaf.core` increased noticeably, and as a result everything that links to `dlaf.core` but doesn't use the newly added instantiations of communication kernels will end up massively bloated.

When linking to static libraries, unlike object libraries, only required symbols are actually included by the linker. This means that e.g. a test that doesn't use communication kernels and links to a static library `dlaf.core` will not end up bloated by the communication kernels.

This PR fixes #487 and as a side-effect also significantly reduces the binary size of most executables.

Still to do:
- [x] remove unnecessary dependencies from tests (now that dependencies are transitively linked)
- [x] add CMake helper function for creating the `dlaf.X` libraries

For comparison, these are the sizes of tests and static libraries on current master:
```
21G     total
2.0G    src/libDLAF.a
1.7G    /apps/daint/UES/simbergm/src/DLA-Future/build/spack/test/unit/eigensolver/test_reduction_to_band
1.7G    /apps/daint/UES/simbergm/src/DLA-Future/build/spack/test/unit/eigensolver/test_gen_to_std
1.7G    /apps/daint/UES/simbergm/src/DLA-Future/build/spack/test/unit/eigensolver/test_gen_eigensolver
1.7G    /apps/daint/UES/simbergm/src/DLA-Future/build/spack/test/unit/eigensolver/test_eigensolver
1.7G    /apps/daint/UES/simbergm/src/DLA-Future/build/spack/test/unit/eigensolver/test_bt_reduction_to_band
1.7G    /apps/daint/UES/simbergm/src/DLA-Future/build/spack/test/unit/eigensolver/test_bt_band_to_tridiag
1.7G    /apps/daint/UES/simbergm/src/DLA-Future/build/spack/test/unit/eigensolver/test_band_to_tridiag
651M    /apps/daint/UES/simbergm/src/DLA-Future/build/spack/test/unit/eigensolver/test_tridiag_solver_merge
650M    /apps/daint/UES/simbergm/src/DLA-Future/build/spack/test/unit/eigensolver/test_tridiag_solver_full
[...]

PR #690, opened 2022-10-27
Reformat all `CMakeLists.txt` and fix `format.sh` script

msimberg:
This fixes the `format.sh` script to correctly check formatting in  `CMakeLists.txt` files. It also reformats the files according to `cmake-format`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -60,39 +60,37 @@ option(DLAF_ASSERT_HEAVY_ENABLE "Enable high impact assertions" ${DLAF_ASSERT_HE
 
 # Define DLAF's PUBLIC properties
 add_library(dlaf.prop INTERFACE)
-target_include_directories(dlaf.prop
-  INTERFACE
-    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
-    $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
-    $<INSTALL_INTERFACE:include>
+target_include_directories(
+  dlaf.prop INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+                      $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include> $<INSTALL_INTERFACE:include>
 )
-target_link_libraries(dlaf.prop
-  INTERFACE
-    MPI::MPI_CXX
-    DLAF::LAPACK
-    pika::pika
-    lapackpp
-    blaspp
-    umpire
-    $<$<BOOL:${DLAF_WITH_CUDA}>:CUDA::cublas>
-    $<$<BOOL:${DLAF_WITH_CUDA}>:CUDA::cudart>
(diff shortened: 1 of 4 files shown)


src/CMakeLists.txt
```diff
@@ -181,7 +190,11 @@ DLAF_addSublibrary(
           matrix_mirror.cpp
           memory/memory_view.cpp
           memory/memory_chunk.cpp
-  GPU_SOURCES cusolver/assert_info.cu cusolver/stedc.cu lapack/gpu/lacpy.cu lapack/gpu/laset.cu
+  GPU_SOURCES
+    cusolver/assert_info.cu
+    cusolver/stedc.cu
+    lapack/gpu/lacpy.cu
+    lapack/gpu/laset.cu
```
rasolca, line 197: The different handling of `SOURCES` and `GPU_SOURCES` is terrible. Any clue on why this is happening?

test/CMakeLists.txt
```diff
@@ -30,7 +30,11 @@ if(DLAF_CI_RUNNER_USES_MPIRUN)
 endif()
 
 # if a preset has been selected and it has been changed from previous configurations
-if(NOT DLAF_MPI_PRESET STREQUAL _DLAF_MPI_PRESET)
+if(NOT
+   DLAF_MPI_PRESET
+   STREQUAL
+   _DLAF_MPI_PRESET
+)
```
rasolca, line 37: IMO this is way more unreadable...

teonnik:
> Additionally, this applies the formatting change suggested in https://github.com/eth-cscs/DLA-Future/pull/686#discussion_r1005450353. I think the result is a net negative and I would drop the commit (https://github.com/eth-cscs/DLA-Future/commit/259ec75a2b9649d882fcee62d30f0871963f6579). The main problem seems to be that cmake-format seems to use max_pargs_hwrap to reformat regular positional arguments like this (this I don't like):

I agree that the change is a net negative. My https://github.com/eth-cscs/DLA-Future/pull/686#discussion_r1005450353 was based on the assumption that CMake formatting is done manually. To avoid clashes between manual formatting and `cmake-format`, I am OK with leaving things as they are.

src/CMakeLists.txt
msimberg: I didn't even notice how bad that was. It seems to be controlled by https://github.com/eth-cscs/DLA-Future/blob/c071856a6373b7b6a15b0a90d7e036ed109fb234/.cmake-format.py#L117. `GPU_SOURCES` is 11 characters long and forces the other layout.

msimberg:
Thanks for the comments all. @albestro I'm going to assume you also agree that the config change is not an improvement, but feel free to comment if that assumption is wrong.

I've reverted the config change. That only leaves the fix to `format.sh` and actually applying the formatting to files that weren't formatted. This is ready to go from my side.

msimberg:
Note, I'm not going to run the full CI since this is only a formatting change.

rasolca (approved):

msimberg:
bors try

msimberg:
CI is happy as well so this is ready to be merged.

PR #691, opened 2022-10-27
Fix band to tridiagonal backtransformation

### LINKED ISSUE #643: Numerical error in test for Backtransform Tridiagonal to Band
Test `test_bt_band_to_tridiag` fails randomly, it can happen after few tens of runs or few hundreds, with a numerical error.

I've seen it failing on GPU, never on MC, both Local and LocalSubBand.

I was able to reproduce it on:
- (current) master 0037c966aec1a2b6cbf99e9a68b8cbb1b572dd1c
- in CI environment with Sarus (manually) running `release-cuda-gcc11/deploy:0e8f3f4169ab0a32aa74951baef9ab4f835a7e07`

CI apparently never triggered it, probably due to the low probability of the event, but in #633 seems to be a bit more frequent (I think it is likely unrelated to the changes I did there).

msimberg:
The cause was two `keep_future`s on `pika::future`s which, as we've already seen in https://github.com/eth-cscs/DLA-Future/pull/409, leads to wrong behaviour. As a reminder, with `keep_future` the full `future` (rather than its wrapped value) gets passed to `unwrapping` and the wrapped value is consumed in the `unwrapping` call. The value (tile in this case) is then released without waiting for the GPU kernel to finish.

Fixes #643.

### DIFF
--- a/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
+++ b/include/dlaf/eigensolver/bt_band_to_tridiag/impl.h
@@ -504,13 +504,13 @@ struct HHManager<Backend::GPU, Device::GPU, T> {
                                  matrix::Tile<const T, D>(std::move(tile_w)));
         };
 
-    auto tup2 = ex::when_all(ex::keep_future(std::move(tile_v_h)), ex::keep_future(std::move(tile_t_h)),
-                             splitTile(mat_v(ij), helper.specHH()), splitTile(mat_t(ij_t), t_spec),
-                             splitTile(mat_w(ij), helper.specHH())) |
-                dlaf::internal::transform<
-                    dlaf::internal::TransformDispatchType::Blas>(dlaf::internal::Policy<Backend::GPU>(),
-                                                                 copyVTandComputeW) |
-                ex::make_future();
+    auto tup2 =
+        ex::when_all(std::move(tile_v_h), std::move(tile_t_h), splitTile(mat_v(ij), helper.specHH()),
+                     splitTile(mat_t(ij_t), t_spec), splitTile(mat_w(ij), helper.specHH())) |
+        dlaf::internal::transform<
+            dlaf::internal::TransformDispatchType::Blas>(dlaf::internal::Policy<Backend::GPU>(),
+                                                         copyVTandComputeW) |
+        ex::make_future();
 
     return pika::split_future(std::move(tup2));
   }


msimberg:
bors try

msimberg:
I ran this locally and was easily able to reproduce the failures before the change and could not reproduce them afterwards. This also ran 50 times in CI without failure: https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/pipelines/679060177.

I'm considering (for a follow-up PR) to slightly change the use of `keep_future`. We could:
1. Only use `keepIfSharedFuture` (which I see is nowadays unused again, so I should anyway do _something_ about it).
2. Add an alternative `dlaf` version, `keepFuture` which would have `keepFuture(pika::future<T>)` deleted to make it a compilation error.

In either of these cases the idea is that it's easier to scan for use of `pika::keep_future` and say _do not use `pika::keep_future` ever in DLA-Future_ instead of trying to figure out if a particular call to `keep_future` is ok or not. There are obviously exceptions in the internals of `Matrix` etc., but in the general case that should be an easier guide to follow. One step further would be to add an `inspect`-like check to check if `pika::keep_future` is used (with the option to override it for the few cases where it is actually needed). Do you think any of the above would be useful or overkill?

msimberg:
bors try

albestro:
I want to thank you a lot @msimberg for this! That's super and it was hard to discover! Sorry for that bug!

I agree with you that we should do something about `keep_future`, because at this point it is too easy to misuse with `future`s, too easy to forget about it and too hard to find a misuse.

At first sight (2) looks nice to me, but I'm open to discuss with others about what's best to do 😉

albestro (approved):
Super @msimberg! Thanks!

msimberg:
Tests failed to run because daint was down... Let's try again.

bors try

albestro:
> ## try
> Build failed:
> 
> * [ci/gitlab/full-pipeline](https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/pipelines/679974394)

It should be good to go now, I manually restarted a couple of tasks due to some launch problem (I would say related to unstable daint/services)

msimberg:
> I agree with you that we should do something about `keep_future`, because at this point it is too easy to misuse with `future`s, too easy to forget about it and too hard to find a misuse.
> 
> At first sight (2) looks nice to me, but I'm open to discuss with others about what's best to do wink

I slightly take back what I said earlier about `keep_future` sometimes being needed for `future`. It's _never_ needed, so "never use `keep_future` on `future`s" is a reasonable guideline. I'll make PR with the second option.

PR #692, opened 2022-10-31
Adapt tridiag eigensolver to hip

aurianer:
Based on #642 
Compiles but not tested yet


src/eigensolver/tridiag_solver/kernels.cu
```diff
@@ -173,7 +188,11 @@ T maxElementInColumnTile(const matrix::Tile<const T, Device::GPU>& tile, whip::s
   SizeType len = tile.size().rows();
   const T* arr = tile.ptr();
 
+#ifdef DLAF_WITH_CUDA
   auto d_max_ptr = thrust::max_element(thrust::cuda::par.on(stream), arr, arr + len);
+#elif defined(DLAF_WITH_HIP)
+  auto d_max_ptr = thrust::max_element(thrust::hip::par.on(stream), arr, arr + len);
+#endif
```
msimberg, line 195: Maybe you can do this instead?
```suggestion
#ifdef DLAF_WITH_CUDA
  constexpr auto par = thrust::cuda::par;
#elif defined(DLAF_WITH_HIP)
  constexpr auto par = thrust::hip::par;
#endif
  auto d_max_ptr = thrust::max_element(par.on(stream), arr, arr + len);
```

include/dlaf/gpu/cub/api.h
```diff
@@ -0,0 +1,151 @@
...
+ ******************************************************************************/
+
+#pragma once
+
+#if defined(DLAF_WITH_HIP)
+
+#include <whip.hpp>
+#include "rocprim/rocprim.hpp"
+
+namespace cub {
```
msimberg, line 46: Let's not put things into the `cub` namespace since it's not our namespace.

If you can reduce this to a less generic version only for the use cases required by DLA-Future that would be good (e.g. `DeviceSegmentedReduce` looks unused).

msimberg (changes requested):
Do you know what's happening in https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/jobs/3346506029#L405? Is `comgr` a new dependency pulled in by `rocprim` and/or `rocthrust` and it just happens to be broken currently?

include/dlaf/gpu/cub/api.h
```diff
@@ -0,0 +1,138 @@
...
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL NVIDIA CORPORATION BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
```
msimberg, line 36: I don't really know what we're allowed to do here. Are we supposed to keep the copyright? Probably yes. Are supposed to keep the license? I don't know, but probably yes? Are we supposed to keep our own license header?

Does anybody know what the right thing to do here is?

src/eigensolver/tridiag_solver/kernels.cu
```diff
@@ -16,14 +16,29 @@
...
-#include <cusolverDn.h>
+#include "dlaf/gpu/blas/api.h"
+#include "dlaf/gpu/cub/api.h"
+#include "dlaf/gpu/lapack/api.h"
+#include "dlaf/gpu/lapack/error.h"
+
+#ifdef DLAF_WITH_HIP
+// Silence rocthrust internal warnings
+#pragma clang diagnostic push
+//#pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
```
msimberg, line 27: Still needed? In general, are the thrust headers not included with `-isystem`?

src/CMakeLists.txt
```diff
@@ -207,14 +210,18 @@ DLAF_addSublibrary(
...
+          $<$<BOOL:${DLAF_WITH_GPU}>:eigensolver/tridiag_solver/gpu.cpp>
+  GPU_SOURCES eigensolver/tridiag_solver/kernels.cu
   LIBRARIES dlaf.factorization dlaf.multiplication dlaf.permutations dlaf.solver dlaf.core
 )
+# Tmp! hip adds --offload-arch compile flag globally for the CXX language in
+# hip-config. We would like therefore to only link to rocthrust when compiling
+# with the HIP language. However COMPILE_LANGUAGE generator expression is not
+# accepted in a target_link_libraries call. The LINK_LANGUAGE generator
+# expression is therefore used but it doesn't add the include directories to the
+# target it's linked to. This is a workaround for that.
```
msimberg, line 223: We should open an issue (in DLA-Future or rocthrust...) about this if this is meant to stay here for this PR.

include/dlaf/gpu/cub/api.h
msimberg: Indeed, that's why I assume we need their license and copyright. What about ours?

rasolca: It depends:
As stated in the license 
> Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.

therefore if we use any line of code (even if we modify it) we have to keep their copyright and license.
On the other hand if we make any change and/or we add code we should put our license.

msimberg: So based on that it sounds like we can just remove our own license and copyright. If we make additions and extensions they can be put in a separate file. @aurianer, the file is currently taken as is from cub with some parts removed, but nothing modified, right?

albestro:
bors try

albestro:
bors try

let's see... 🤞

include/dlaf/gpu/cub/api.h
rasolca: Based on this
> They were using internal macros that I changed to use the corresponding symbol directly, I don't know if that counts.. I also added the change of namespace `dlaf::gpucub` EDIT: I also add to change a few things to avoid warnings..

I would keep both license

rasolca (approved):
Note: Was this PR tested on AMD GPUs? (the CI simply tests that the image builds)

ci/docker/build.Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apt-get -yqq update && \
     apt-get -yqq install --no-install-recommends \
     software-properties-common \
     build-essential gfortran \
-    autoconf automake \
+    autoconf automake ninja-build pkg-config \
```
msimberg, line 18: Is this not already on master? :thinking: Could you rebase/merge with latest master?

albestro:
cherry-picking de36d58e9c700a93072e349d6a4eddd7e13b4a2b should be enough to dramatically reduce build times for the AMD CI part, since it opens to using more externally provided ROCm libraries.

ci/.gitlab-ci.yml
```diff
@@ -19,9 +19,9 @@ stages:
     - trying
   variables:
     GIT_SUBMODULE_STRATEGY: recursive
-    # Custom spack sha used to make an umpire fix available
-    # (https://github.com/msimberg/spack/commit/770d71dff6428876f2cc996f8a0551df36a59e4f)
-    SPACK_SHA: 770d71dff6428876f2cc996f8a0551df36a59e4f
+    # Custom spack sha used to make an umpire fix available + fix the comgr ci pb
+    # (https://github.com/aurianer/spack/commit/f9b6600356a47ab6f4ffd45cef10b3c18e72269c)
+    SPACK_SHA: f9b6600356a47ab6f4ffd45cef10b3c18e72269c
```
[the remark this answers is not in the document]
albestro: See https://github.com/eth-cscs/DLA-Future/pull/692#issuecomment-1330867621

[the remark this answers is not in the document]
albestro: Just for reference, currently `comgr` is still built.

With the "changes to external" https://github.com/eth-cscs/DLA-Future/pull/692#issuecomment-1330867621 it won't be involved at all (because it's a dep of `rocprim`, which will be provided externally)

In any case, I agree with you that the comment must be removed if it is not relevant anymore (@aurianer knows better the spack branch situation, so I'll let her decide what's correct)

albestro (commented):
I would ask you to check if there are other guards that can be easily generalized to GPU.

e.g.
https://github.com/eth-cscs/DLA-Future/blob/5ca8a1d575c412c3746c626178998ff837bda1fd/test/unit/eigensolver/test_tridiag_solver_rot.cpp#L133-L140

msimberg (approved):
Thanks @aurianer for all the work on this! Let's get it in even though there's some more work to be done.

Could I ask you to open issues/comment on already open issues with all the open problems that you're aware of at this point (two things off the top of my head: 1. using externally installed rocthrust/rocprim and 2. commenting about the segfaults on https://github.com/eth-cscs/DLA-Future/issues/602).

rasolca:
> I would ask you to check if there are other guards that can be easily generalized to GPU.
> 
> e.g.
> 
> https://github.com/eth-cscs/DLA-Future/blob/5ca8a1d575c412c3746c626178998ff837bda1fd/test/unit/eigensolver/test_tridiag_solver_rot.cpp#L133-L140

The goal of this PR is to make the local solver work. Distributed will follow in another PR.

PR #694, opened 2022-10-31
Add constrained version of `keep_future` for internal use in DLA-Future

msimberg:
This is based on the bug fix in #691 (specifically 2. in https://github.com/eth-cscs/DLA-Future/pull/691#issuecomment-1294513760).

This adds a constrained version of `pika::keep_future` called `dlaf::internal::keepFuture`. It only compiles with `pika::shared_future` and fails to compile with `pika::future`. It should be used unless there is a very good reason to use `pika::keep_future` directly.

There is exactly one place in DLA-Future which currently requires using `pika::keep_future` directly and I've expanded the comment there to explain why that is the case. Let me know if that comment can be clearer still.

I've renamed the header `keep_if_shared_future.h` to just `keep_future.h` as it only contains `keepFuture`.

I've changed `test_tile` to not use `keep_future` at all and instead just call plain `(shared_)future::get` instead of `sync_wait`. I originally added the `keep_future` to be able to use `sync_wait` to "future proof" it, but since the senders are still futures I'm simplifying it at this point to just use plain `get`. The `sync_wait` can be added back at some point when the senders are not futures.


msimberg:
I've also opened https://github.com/eth-cscs/DLA-Future/issues/696 to keep track of a potential simplification where `keepFuture` wouldn't need to be called in so many places.

test/unit/matrix/test_util_matrix.cpp
```diff
@@ -162,7 +162,8 @@ void check_is_hermitian(Matrix<const T, Device::CPU>& matrix, comm::Communicator
           return dlaf::conj(tile_original({index.col(), index.row()}));
         };
 
-        CHECK_TILE_NEAR(transposed_conj_tile, sync_wait(keep_future(tile_transposed)).get(),
+        CHECK_TILE_NEAR(transposed_conj_tile,
+                        sync_wait(dlaf::internal::keepFuture(tile_transposed)).get(),
```
albestro, line 166: Not sure about this line, I might be missing something. To me it seems a mixed solution, I would expect

```suggestion
                        tile_transposed.get(),
```

while there is both `sync_wait` treating it as a sender, and `.get()` which makes me think about it as a future.

albestro (approved):
Thanks, I think it is a really nice move.

See just a minor comment about going back to future usage in a test.

msimberg:
bors try

msimberg:
bors try-

test/unit/matrix/test_util_matrix.cpp
albestro: Thanks for the explanation offline!

I was missing the point that `sync_wait` waits on `keep_future`, which is a sender that sends a future. So, once we get the value sent with `sync_wait`, we actually get a future, on which we have to do the `.get()` in order to extract the value from it.

teonnik (approved):

rasolca (approved):

msimberg:
I will update this one next.

msimberg:
This is rebased and ready for borsing again. I haven't changed anything about the `keepFuture` wrapper. There were some new instances of `keep_future` that I replaced with `keepFuture`, but otherwise this is unchanged.

PR #697, opened 2022-11-02
Replace `pika::unwrapping` with a specialized unwrapper

### LINKED ISSUE #681: Remove `pika::unwrapping` from `transform` and do custom one-level unwrapping of futures instead
https://github.com/eth-cscs/DLA-Future/pull/679#issuecomment-1285083540

### LINKED ISSUE #693: `transform` tries to instantiate callables without a stream argument
#679 added a workaround to not use `unwrapping` with nullary callables. The workaround introduced another problem when the callable is nullary only for CPU and unary for GPU. The `is_invocable_v<F>` check (in https://github.com/eth-cscs/DLA-Future/blob/c1720b1d43f2beeb42dac4784d7c357e0e08fa3f/include/dlaf/sender/transform.h#L99) attempts to instantiate the callable with zero arguments also on the GPU path. If the callable is unconstrained this will lead to compilation errors. An example of a call that will currently fail:

```
transform(policy, [](auto&&... ts) {
    if constexpr (device == Device::CPU) {
        f_cpu();
    } else {
        f_gpu(ts...); // instantiated with ts as an empty pack by is_invocable_v<F>
    }
})
```
This should be fixed at the same time as #681, but if a workaround is needed one can write the lambda as something like this:
```
if constexpr (device == Device::CPU) {
    f_cpu();
} else if constexpr (sizeof...(ts) == 1) {
    f_gpu(ts...);
} // ignore the other cases
```

msimberg:
Don't pay much attention to this PR yet. I think I have at least one more bug to fix. I want to check with CI if it agrees or if there is more to fix...

Fixes #693 and fixes #679 in a better way. Fixes #681.

Removes the use of `pika::unwrapping` in favour of a custom unwrapper for specialized for DLA-Future. There is now:

- `dlaf::common::internal::Unwrapping`, a struct with a call operator similar to the type returned from `pika::unwrapping`
- `dlaf::common::internal::unwrap`, a function that unwraps a single value
- `dlaf::common::internal::Unwrapper`, a template struct that can be specialized for different types to define how something should be unwrapped. The default is to do nothing to a value.

`Unwrapper` currently has specializations for:
- `pika::future<T>`
- `pika::shared_future<T>`
- `std::reference_wrapper<T>`
- `PromiseGuard<T>` (all `PromiseGuards`, not only `PromiseGuard<Communicator>`)

These changes mean that e.g. `vector<future<T>>` is _not_ unwrapped anymore. Only the types explicitly specialized are unwrapped. This also removes the need for the `Unwrap` boolean template parameter to `transform` that was introduced some time ago in https://github.com/eth-cscs/DLA-Future/pull/564/commits/5135384398dfe64a7658ffc24fa1e2f559fba863.


include/dlaf/common/pipeline.h
```diff
@@ -94,3 +95,14 @@ class Pipeline {
       sender_;  ///< This contains always the "tail" of the queue of senders.
 };
 }
+
+namespace dlaf::common::internal {
+// TODO: Is unwrapping _all_ PromiseGuards overreaching? Should it be restricted to PromiseGuard<Communicator>?
```
msimberg, line 100: TODO

include/dlaf/common/unwrap.h
```diff
@@ -0,0 +1,75 @@
...
+
+/// Unwrapping is a callable object wrapper that calls unwrap on all arguments
+/// before calling the wrapped callable.
+template <typename F>
+struct Unwrapping {
+  std::decay_t<F> f;
+
+  template <typename... Ts>
+  auto operator()(Ts&&... ts)
+  //-> decltype(std::move(f)(Unwrapper<std::decay_t<Ts>>::unwrap(std::forward<Ts>(ts))...))
```
msimberg, line 67: TODO, this can be uncommented.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -450,14 +448,14 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
...
+    worker.startSweep(sweep, *a_ws);
   };
-  auto cont_sweep = [a_ws, b](SizeType nr_steps, PromiseGuard<SweepWorker<T>> worker,
+  auto cont_sweep = [a_ws, b](SizeType nr_steps, SweepWorker<T>& worker,
                               matrix::Tile<T, Device::CPU>&& tile_v, TileElementIndex index) {
     for (SizeType j = 0; j < nr_steps; ++j) {
-      worker.ref().compactCopyToTile(tile_v, index + TileElementSize(j * b, 0));
-      worker.ref().doStep(*a_ws);
+      worker.compactCopyToTile(tile_v, index + TileElementSize(j * b, 0));
+      worker.doStep(*a_ws);
```
msimberg, line 458: TODO, check with @albestro if this is really ok.

include/dlaf/sender/transform_mpi.h
```diff
@@ -9,39 +9,26 @@
...
-/// This helper "consumes" a PromiseGuard<Communicator> ensuring that after this call the one
-/// passed as argument gets destroyed.
+/// \overload consumePromiseGuardCommunicator
 template <typename T>
-void consumePromiseGuardCommunicator(T& t) {
-  if constexpr (IsPromiseGuardCommunicator<std::decay_t<T>>::value) {
-    [[maybe_unused]] auto t_local = std::move(t);
-  }
-}
+void consumePromiseGuardCommunicator(T&) {}
```
msimberg, line 31: Note, the only reason I separated this into two overloads is to avoid having to instantiate `IsPromiseGuardCommunicator`. I haven't measured the difference and I suspect the win is small, so if someone prefers the `if constexpr` version let me know and I'll happily revert this. This change has no impact on behaviour.

rasolca:
On hold as it conflicts with #698.

msimberg:
bors try

include/dlaf/common/unwrap.h
msimberg: Uncommented.

include/dlaf/eigensolver/band_to_tridiag/mc.h
msimberg: There's a slight semantic difference between the old and the new behaviour and whether or not it's a problem depends on what comes after `cont_sweep`. Though I see @rasolca has made the same change (to unwrap _all_ `PromiseGuards`) in https://github.com/eth-cscs/DLA-Future/pull/698/files, so that's reassuring.

The difference is that if `PromiseGuard<T>` is taken by value by the function it will be released at the end of the function scope. If the `PromiseGuard<T>` is unwrapped before being passed to the function it will live as long as the rvalue reference of it lives inside `then` (for CPU), which may be until the next `make_future`/`ensure_started`/`start_detached` (or other adaptor that releases previous operation states).

msimberg:
bors try

msimberg:
Now that #698 is merged I'm going to update this as the next thing. It turns out that this is useful for some work that @teonnik is doing so I'll try to get this rebased and cleaned up as soon as possible.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -264,15 +263,10 @@ void sortIndex(SizeType i_begin, SizeType i_end, pika::shared_future<SizeType> k
...
-      di::transform<di::TransformDispatchType::Plain, false>(di::Policy<DefaultBackend_v<D>>(),
-                                                             std::move(sort_fn), std::move(sender)));
+      di::transform(di::Policy<DefaultBackend_v<D>>(), std::move(sort_fn), std::move(sender)));
 }
 
 // Applies `index` to `in` to get `out`
-//
-// Note: `pika::unwrapping()` can't be used on this function because std::vector<Tile<const >> is
-// copied internally which requires that Tile<const > is copiable which it isn't. As a consequence the
-// API can't be simplified unless const is dropped.
```
msimberg, line 275: @teonnik do you remember if there's some simplification we can do here right away now that the unwrapping has been changed? Or is "the API can't be simplified" just referring to the `transform<..., false>` call?

include/dlaf/sender/with_temporary_tile.h
```diff
@@ -98,7 +99,7 @@ auto withTemporaryTile(InSender&& in_sender, F&& f) {
   return std::forward<InSender>(in_sender) |
          // Start a new asynchronous scope for keeping the input tile alive
          // until all asynchronous operations are done.
-         ex::let_value(pika::unwrapping([f = std::forward<F>(f)](auto& in) mutable {
+         ex::let_value(dlaf::common::internal::Unwrapping{[f = std::forward<F>(f)](auto& in) mutable {
```
msimberg, line 102: @teonnik with this change your use case of passing a `just(ref(tile))` into `withTemporaryTile` should work (though I may be missing something...).

include/dlaf/common/pipeline.h
msimberg: I've removed the TODO as this was anyway already being done on master now.

msimberg:
bors try

albestro (approved):
Neat! LGTM!

teonnik (approved):
LGTM!

rasolca (approved):

rasolca:
There are a couple of open conversations.
@msimberg can you confirm that they are resolved?

msimberg:
> There are a couple of open conversations. @msimberg can you confirm that they are resolved?

From my perspective they are resolved. They were mostly informational. https://github.com/eth-cscs/DLA-Future/pull/697#discussion_r1036391067 was more to check if there's some additional cleanup that could be done as a result of these changes, but resolving that is not a blocker for this and can be done afterwards (if there is in fact anything more to clean up there). I closed the other open conversations outright.

msimberg:
This also lead to a small debug bloat reduction. `libDLAF.so` was 1428553504 bytes before, now it's `1382810424` (CPU only). It's still ridiculous, but at least it's in the right direction...

PR #698, opened 2022-11-02
Distributed band to tridiagonal

rasolca:
It is finally working, but needs a cleanup before reviewing...


rasolca:
bors try

include/dlaf/eigensolver/band_to_tridiag.h
```diff
@@ -89,5 +89,32 @@ TridiagResult<T, Device::CPU> bandToTridiag(blas::Uplo uplo, SizeType band_size,
...
+template <Backend backend, Device device, class T>
+TridiagResult<T, Device::CPU> bandToTridiag(comm::CommunicatorGrid grid, blas::Uplo uplo,
+                                            SizeType band_size, Matrix<const T, device>& mat_a) {
+  DLAF_ASSERT(matrix::square_size(mat_a), mat_a);
+  DLAF_ASSERT(matrix::square_blocksize(mat_a), mat_a);
+  DLAF_ASSERT(matrix::equal_process_grid(mat_a, grid), mat_a, grid);
+
+  // If the grid contains only one rank force local implementation.
+  if (grid.size() == comm::Size2D(1, 1))
+    return bandToTridiag<backend, device, T>(uplo, band_size, mat_a);
```
msimberg, line 102: Nothing to do for this comment, but I'm just curious: is this a shortcut we can apply for other algorithms as well or is there something special about this one that allows it?

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -89,15 +94,92 @@ void applyHHRight(const SizeType m, const SizeType n, const T tau, const T* v, T
   blas::ger(ColMaj, m, n, -tau, w, 1, v, 1, a, lda);
 }
 
+// split versions of the previous operations
```
msimberg, line 97: I'm ignorant as usual about the algorithms, but "split" says nothing to me in this context. Is that something one knows if one knows how the algorithm works? If yes, this is probably fine. If it refers to something specifically in the way you've implemented the algorithm would you mind briefly explaining what split refers to?

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -89,15 +94,92 @@ void applyHHRight(const SizeType m, const SizeType n, const T tau, const T* v, T
...
+    DLAF_ASSERT(block_size_ % band_size_ == 0, block_size_, band_size_);
+    DLAF_ASSERT(0 <= id && id < ceilDiv(size_, block_size_), id, ceilDiv(size_, block_size_));
+  }
 
   T* ptr(SizeType offset, SizeType j) noexcept {
     DLAF_ASSERT_HEAVY(0 <= offset && offset < ld_ + 1, offset, ld_);
     DLAF_ASSERT_HEAVY(0 <= j && j < size_, j, size_);
-    return mem_(j * (ld_ + 1) + offset);
+
+    if (dist) {
```
msimberg, line 177:
```suggestion
    if constexpr (dist) {
```

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -269,12 +355,70 @@ struct BandBlock {
   }
 #endif
 
+  SizeType memoryIndex(SizeType j) {
+    if (dist) {
```
msimberg, line 359:
```suggestion
    if constexpr (dist) {
```

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -269,12 +355,70 @@ struct BandBlock {
...
+                                std::shared_ptr<BandBlock<T, true>> a_block, MPI_Request* req) {
+    DLAF_MPI_CHECK_ERROR(MPI_Isend(a_block->ptr(0, j), to_int(2 * b), dlaf::comm::mpi_datatype<T>::type,
+                                   dest, tag, comm, req));
+  };
+
+  auto comm_sender =
+      whenAllLift(std::forward<CommSender>(pcomm), a_block, std::forward<DepSender>(dep)) |
+      transformMPI(send);
+  // Ensure lifetime shared pointer.
+  return whenAllLift(std::move(a_block), std::move(comm_sender)) | ex::then([](auto) {});
```
msimberg, line 399: Is this last `whenAllLift(std::move(a_block), ...)` there just to make sure `a_block` lives long enough? If yes, I think you can rewrite it like this since `transformMPI` will ensure that the values live long enough:

```suggestion
  auto send = [dest, tag, b, j](const comm::Communicator& comm,
                                // The ref here is not a must, but avoids an extra copy of the shared_ptr
                                std::shared_ptr<BandBlock<T, true>>& a_block, MPI_Request* req) {
    DLAF_MPI_CHECK_ERROR(MPI_Isend(a_block->ptr(0, j), to_int(2 * b), dlaf::comm::mpi_datatype<T>::type,
                                   dest, tag, comm, req));
  };

  return whenAllLift(std::forward<CommSender>(pcomm), a_block, std::forward<DepSender>(dep)) |
      transformMPI(send);
```
Or is there another reason for the `whenAllLift`?

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -341,7 +485,7 @@ class SweepWorker {
   }
 
   T& tau() noexcept {
-    return *data_(0);
+    return *data_();
   }
   const T& tau() const noexcept {
     return *data_(0);
```
msimberg, line 491: Also `data_()` as above, or is this intentionally left different?

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -365,6 +509,126 @@ class SweepWorker {
...
+  auto recv = [sweep, step, src, tag](const comm::Communicator& comm, auto& worker, MPI_Request* req) {
+    worker.recv(sweep, step, comm, src, tag, req);
+  };
+
+  return whenAllLift(std::forward<CommSender>(pcomm), std::forward<PromiseSender>(worker)) |
+         transformMPI(recv);
+}
+
+template <class T>
+class vector2D {
```
msimberg, line 610: Am I missing something or is this unused? I guess you might (have) intend(ed) to use it to replace the `vector<vector<T>>` uses further down?

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -535,4 +798,548 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
   return {std::move(mat_trid), std::move(mat_v)};
 }
 
+std::ostream& operator<<(std::ostream& out, matrix::SubTileSpec spec) {
+  out << spec.origin << "+" << spec.size;
+  return out;
+}
```
msimberg, line 804: This should presumably be removed before merging.

include/dlaf/eigensolver/band_to_tridiag.h
rasolca: We can do it for all other algorithms. What is special about this one? The fact that the distributed is implemented assuming at least two ranks, so it will not work without it.

include/dlaf/eigensolver/band_to_tridiag/mc.h
rasolca: This is the kind of comment I'm looking for. 👍🏻

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -269,12 +355,70 @@ struct BandBlock {
...
+  using dlaf::internal::whenAllLift;
+
+  auto recv = [src, tag, b, j](const comm::Communicator& comm,
+                               std::shared_ptr<BandBlock<T, true>> a_block, MPI_Request* req) {
+    DLAF_MPI_CHECK_ERROR(MPI_Irecv(a_block->ptr(0, j), to_int(2 * b), dlaf::comm::mpi_datatype<T>::type,
+                                   src, tag, comm, req));
+  };
+
+  return whenAllLift(std::forward<CommSender>(pcomm), std::move(a_block), std::forward<DepSender>(dep)) |
+         transformMPI(recv);
```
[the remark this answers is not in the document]
rasolca: The reason it is not needed here is that if you receive a column, you are sure that a task which works on the same a_block depends on this mpi task and therefore it guarantees the lifetime.
Note: a_block doesn't use any auto dependency mechanism.

include/dlaf/eigensolver/band_to_tridiag/mc.h
rasolca: Yes, it is a remainder of debug printing...

include/dlaf/eigensolver/band_to_tridiag/mc.h
rasolca: It is a reminder of debugging... Fixed.

include/dlaf/eigensolver/band_to_tridiag/mc.h
rasolca: It was a first attempt, but then switched to `vector<vector<...>>` as the "internal" vector need to be resized.

include/dlaf/eigensolver/band_to_tridiag/mc.h
rasolca: Is it better now?

include/dlaf/sender/transform_mpi.h
```diff
@@ -73,20 +84,20 @@ struct MPICallHelper {
     // whereas we are only looking to guard the submission of the MPI operation.
     // Moreover, the callable requires a Communicator to make its job, and it should be agnostic of
     // it being wrapped in a PromiseGuard or not.
-    // We therefore use unwrapPromiseGuardCommunicator to pass the Communicator& into a transformMPI
+    // We therefore use unwrapPromiseGuard to pass the Communicator& into a transformMPI
     // callables, then after returning from the callable, the PromiseGuard<Communicator> gets explicitly
-    // released using the helper consumePromiseGuard.
+    // released using the helper consumePromiseGuardCommunicator.
```
albestro, line 89: If I get it correctly, this documentation has to be slightly revised. We have a "pre" and "post" phase.

In the "pre" phase, any `PromiseGuard`, no matter what type they are guarding, gets unwrapped and passed by reference to the function. Lifetime is kept by the `transformMPI`.

Inside `transformMPI`, after calling the function but before yielding for the result, there is the "post" phase.
Here ONLY `PromiseGuard<Communicator>` gets consumed manually in order to release it, because it is not needed anymore to keep the lifetime of the `Communicator` (required for posting, not for completing). But, other `PromiseGuard` have to be kept alive till the end of the function, so they don't get consumed.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -89,15 +94,94 @@ void applyHHRight(const SizeType m, const SizeType n, const T tau, const T* v, T
   blas::ger(ColMaj, m, n, -tau, w, 1, v, 1, a, lda);
 }
 
+// As the compact band matrix is stored in a rotating buffer it might happen that the operations
```
msimberg, line 97:
```suggestion
// As the compact band matrix is stored in a circular buffer it might happen that the operations
```
?

include/dlaf/eigensolver/band_to_tridiag/mc.h
msimberg: Yes, much better. Thank you!

include/dlaf/sender/transform_mpi.h
rasolca: I just revised the doc of what happens for Communicator PromiseGuard.
It can be extended, however as #697 is modifying it again, I would leave the doc udate for that PR.

albestro (commented):
I found difficult to understand the backbone of the algorithm, my little knowledge of it surely doesn't help. An idea could be to extract some lambda (at least the non-capturing ones) in the internal namespace.

Just some minor comments, since I don't have extended knowledge of the algorithm, even if it in the end I got a general idea.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -535,4 +773,610 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
...
+  SizeType nrtile = mat_a.nrTiles().cols();
+  SizeType nb = mat_a.blockSize().cols();
+  auto& dist_a = mat_a.distribution();
+
+  Matrix<BaseType<T>, Device::CPU> mat_trid({size, 2}, {nb, 2});
+  matrix::Distribution dist_v({size, size}, {nb, nb}, dist_a.commGridSize(), dist_a.rankIndex(),
+                              dist_a.sourceRankIndex());
+  Matrix<T, Device::CPU> mat_v(dist_v);
+
+  if (size == 0) {
```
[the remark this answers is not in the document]
rasolca: The quick return for `band_size == 1` is not so quick.
The tridiagonal matrix has to be rearranged and broadcasted.
I would rather go with band_size >=2.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -365,6 +507,103 @@ class SweepWorker {
...
+            MPI_Request* req) const noexcept {
+    DLAF_MPI_CHECK_ERROR(MPI_Isend(const_cast<T*>(data_()), to_int(band_size_ + 1),
+                                   comm::mpi_datatype<T>::type, dest, tag, comm, req));
+  }
+
+  void recv(SizeType sweep, SizeType step, const comm::Communicator& comm, comm::IndexT_MPI src,
+            comm::IndexT_MPI tag, MPI_Request* req) noexcept {
+    SweepWorker<T>::setId(sweep, step);
+    *data_(0) = T{9};
+    *data_(1) = T{9};
```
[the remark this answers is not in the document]
rasolca: 😓 something that was used during debugging... No need to set this values before receiving the correct one.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -535,4 +773,610 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
...
+  //   A3 B2 | C1 D0 | ** ..
+  //   -- --   -- --   --
+  //   A4 B3 | C2 D1 | E0 ..
+  //   .. .. | .. .. | .. ..
+  //
+  // the communication of a tile of a panel might be splitted in two parts (e.g. B2 B3 tile).
+  // If copyIsSplitted() is true the copy/communication has to happen in two parts Top and Bottom,
+  // otherwise only Top is set.
+
+  VAccessHelper(const comm::CommunicatorGrid& grid, const SizeType band, const SizeType sweeps,
```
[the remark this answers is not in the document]
rasolca: Everything is computed in the constructor. The only members that require the distributions are static members which do not benefit on storing the dists.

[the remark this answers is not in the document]
albestro: Sorry, I didn't see they were `static (and I would have expected they weren't).

Probably `static` function is something you opted for in order to not having to pass `VAccessHelper` around?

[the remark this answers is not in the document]
rasolca: I need to compute the panel index twice:
- when I generate the HH reflectors
- when I communicate/copy the panel to V
In the first case I need just the panel index and the other indices are not needed (not sure if they even make sense)
https://github.com/eth-cscs/DLA-Future/pull/698/files#diff-14958ca39c015bb93ece6c5aacfd53ef345d36af1687a04c9f1701ebb1c33436R1272
In the second case I need all the members to be available and therefore I instantiate the object (which call the static members)
https://github.com/eth-cscs/DLA-Future/pull/698/files#diff-14958ca39c015bb93ece6c5aacfd53ef345d36af1687a04c9f1701ebb1c33436R1303-R1304

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -535,4 +773,610 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
...
+  comm::IndexT_MPI rank_v_bottom_;
+  TileElementSize size_top_{0, 0};
+  TileElementSize size_bottom_{0, 0};
+};
+
+// Distributed implementation of bandToTridiag.
+template <Device D, class T>
+TridiagResult<T, Device::CPU> BandToTridiagDistr<Backend::MC, D, T>::call_L(
+    comm::CommunicatorGrid grid, const SizeType b, Matrix<const T, D>& mat_a) noexcept {
+  // Note on the algorithm, data distribution and dependency tracking:tiles_per_block
```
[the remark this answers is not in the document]
rasolca: 😶

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -365,6 +531,101 @@ class SweepWorker {
...
+      doStepSplit(a, n1);
+    }
+    else {
+      this->doStepFull(a);
+    }
+  }
+
+  void send(const comm::Communicator& comm, comm::IndexT_MPI dest, comm::IndexT_MPI tag,
+            MPI_Request* req) const noexcept {
+    DLAF_MPI_CHECK_ERROR(MPI_Isend(const_cast<T*>(data_()), to_int(band_size_ + 1),
```
msimberg, line 558: How sure are we that this `const_cast` can't be used wrongly? Why not
```suggestion
            MPI_Request* req) noexcept {
    DLAF_MPI_CHECK_ERROR(MPI_Isend(data_(), to_int(band_size_ + 1),
```
?

msimberg (approved):
Just one last question about the `const_cast`...

include/dlaf/eigensolver/band_to_tridiag/mc.h
rasolca: well spotted... the const cast should not be there (don't remember why I put it there during debugging)... the first parameter of Isend is a `const void*`.

rasolca: Fixing it in `include/dlaf/communication/kernels/p2p.h` as well..

teonnik (approved):
LGTM!

rasolca:
bors try

PR #714, opened 2022-11-10
Make sure split communicators in `CommunicatorGrid6RanksEnvironment` are freed by making them managed

### LINKED ISSUE #712: Too many communicators (0/4096 free on this process; ignore_id=0)
While I was investigating another problem, I stumbled upon this error message

```console
[MPI ERROR] /users/ialberto/workspace/dla-future.rot-dist/src/communication/communicator_grid.cpp:38 : dlaf::comm::CommunicatorGrid::CommunicatorGrid(dlaf::comm::Communicator, dlaf::comm::IndexT_MPI, dlaf::comm::IndexT_MPI, dlaf::common::Ordering)
Other MPI error, error stack:
PMPI_Comm_split(521)................: MPI_Comm_split(comm=0xc400b3d4, color=0, key=0, new_comm=0x15554570f1ec) failed
PMPI_Comm_split(503)................:
MPIR_Comm_split_impl(276)...........:
MPIR_Get_contextid_sparse_group(697): Too many communicators (0/4096 free on this process; ignore_id=0)
```

I got this not by using `ctest`, but manually running a distributed test on a single node using the `--gtest_repeat` command, which launches a single time an executable and repeat itself for the requested number of times.

```console
srun -u -n 6 -c 12 /users/ialberto/workspace/dla-future.rot-dist/build-mc-sysalloc/test/unit/communication/test_comm_reproducer --gtest_repeat=5000
```

I don't know if it is a problem correlated to gtest or it is a real problem (and in case it could be internal to DLAF, or even a problem in cray-mpich).

I was able to reproduce it with existing tests (the one I was working on, but also `test_cholesky` which is a "production" one), and also with a simple test I created

```cpp
// test_comm_reproducer.cpp
#include <gtest/gtest.h>

#include "dlaf_test/comm_grids/grids_6_ranks.h"

::testing::Environment* const comm_grids_env =
    ::testing::AddGlobalTestEnvironment(new dlaf::test::CommunicatorGrid6RanksEnvironment);

TEST(TestWithCommGrids, Minimal) {}
```

It seems deterministically reproducible. Indeed, I got for both my test `test_tridiag_solver_rot` (not yet on master) and `test_comm_reproducer` always a failure at iteration 4085, and for `test_cholesky` at iteration 4083 (run just one time because it is a long test).

msimberg:
Fixes #712.

The split communicator was previously unmanaged and thus not freed when the vector `comm_grids` was cleared. This makes it managed to make sure it's freed.

I checked for other uses of `MPI_Comm_split` and as far as I can tell they're all ok uses of unmanaged communicators (i.e. the communicators are explicitly freed). Do you see any reason to avoid a managed communicator in this instance and instead explicitly free the communicators that need freeing in the `TearDown` function?

As for adding a test, we could of course add @albestro's test from https://github.com/eth-cscs/DLA-Future/issues/712, but since that would only check one test environment I don't know if there's much value in that. So thinking about how we can avoid this situation in the API instead, I wonder if we could make managed communicators the default when constructing one with the regular `Communicator` constructor (with an exception for `MPI_COMM_WORLD`), and instead make the unmanaged case the verbose case. Currently managed communicators have to be constructed with the more verbose `make_managed_communicator`. What do you think?

### DIFF
--- a/test/include/dlaf_test/comm_grids/grids_6_ranks.h
+++ b/test/include/dlaf_test/comm_grids/grids_6_ranks.h
@@ -56,7 +56,7 @@ public:
       MPI_Comm split_comm;
       DLAF_MPI_CHECK_ERROR(MPI_Comm_split(world, color, world.rank(), &split_comm));
 
-      comm::Communicator comm(split_comm);
+      auto comm = comm::make_communicator_managed(split_comm);
       comm_grids.emplace_back(comm, rows, cols, common::Ordering::ColumnMajor);
     }
   }


albestro:
Thank you for the quick fix @msimberg! I just tested with my two tests (`test_tridiag_solver_rot` and `test_comm_reproducer`) and it works perfectly!

Before approving, I'd like to check a couple of things, that I look forward to do at latest next week (since it is a "minor" problem I marked this as low priority).

rasolca:
I would suggest to make explicit the choice between un/managed and provide the following.
```
comm::Communicator comm(comm, dlaf::comm:Managed());
comm::Communicator comm(comm, dlaf::comm::UnManaged());

auto comm = make_managed_communicator(comm)
auto comm = make_unmanaged_communicator(comm)
```
Note: for the constructor the distinction can be made at runtime with a enum class or at compilation using two overloads. I think the second approach is better.

rasolca (approved):

albestro (approved):

rasolca:
bors try

rasolca:
bors try-

Other PR with higher priority need to be tested.

rasolca:
bors try

PR #716, opened 2022-11-11
More build dependencies from apt-get

### LINKED ISSUE #699: Install more dependencies from Ubuntu repositories in CI
Packages like hwloc, ninja, libxml, pkgconf, etc. should have little to no impact on DLA-Future or pika behaviour and could be installed from Ubuntu's repositories to slightly speed up rebuilds of the CI images (only helps when rebuilding the base image).

rasolca:
Closes #699.

Note: as `hwloc` and `mpich` both depends on `libpciaccess` it make little difference to set `hwloc` external (only the compilation of `hwloc`) and I'm not confident about possible problems arising from multiple libpciaccess used. (@msimberg what do you think?)

### DIFF
--- a/ci/docker/amdgpu-release-rocm523.yaml
+++ b/ci/docker/amdgpu-release-rocm523.yaml
@@ -37,6 +37,10 @@ spack:
       - ~hwloc
       - device=ch3
       - netmod=tcp
+      - ~libxml2
+    hwloc:
+      variants:
+      - ~libxml2
     git:
       # Force git as non-buildable to allow deprecated versions in environments
       # https://github.com/spack/spack/pull/30040
--- a/ci/docker/build.Dockerfile
+++ b/ci/docker/build.Dockerfile
@@ -15,7 +15,7 @@ RUN apt-get -yqq update && \
     apt-get -yqq install --no-install-recommends \
     software-properties-common \
     build-essential gfortran \
-    autoconf automake \
+    autoconf automake ninja-build pkg-config \
     ${EXTRA_APTGET} \
     gawk \
     python3 python3-distutils \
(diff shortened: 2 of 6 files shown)


rasolca:
Note: based on CI outputs the speedup should be of ~5 min

msimberg (approved):

rasolca:
> we'd have to make `libpciaccess` also external.

I tried with `spack external find libpciaccess` without success...

msimberg:
> > we'd have to make `libpciaccess` also external.
> 
> I tried with `spack external find libpciaccess` without success...

I think the reason is that only packages that opt in to being findable with `spack external find`. Typically this will be packages with executables, and they opt-in by declaring something like this: https://github.com/spack/spack/blob/d49c992b230f544deeed13fb193d779a53f2f201/var/spack/repos/builtin/packages/ninja/package.py#L24 which spack then knows to search for.

I think one could add `libpciaccess` (and others) the same way all the ROCm packages are added, i.e. explicitly in the config: https://github.com/eth-cscs/DLA-Future/blob/306bc61ce74d6bcafa211b7081288bb9ad32c727/ci/docker/amdgpu-release-rocm523.yaml#L52-L76.

rasolca:
I would like to avoid specifying manually the packages as versions might change and everything can break without an apparent reason. rocm is an exception I wouldn't like to do again...

msimberg:
> I would like to avoid specifying manually the packages as versions might change and everything can break without an apparent reason. rocm is an exception I wouldn't like to do again...

Yep, fair enough! I think this already enough of an improvement on its own.

rasolca:
bors try

msimberg (approved):

PR #718, opened 2022-11-14
algo: applyGivenRotations Distributed (GPU)

### LINKED ISSUE #717: Porting application of Given Rotations to GPU
In #675 the algorithm has been implemented for MC, but for GPU it still requires some adaption (e.g. custom communication kernels have to be implemented)

albestro:
Close #717

Apart from enabling GPU tests, the main required change for the algorithm has been to port the custom communication kernels for P2P send/recv of a column to work also with GPU memory (via a CPU temporary buffer).

TODO:
- [x] ~#692 (see [comment](https://github.com/eth-cscs/DLA-Future/pull/718#issuecomment-1331823261))~
- [x] ~#719 (see [comment](https://github.com/eth-cscs/DLA-Future/pull/718#issuecomment-1326150675))~ after last changes in that PR, probably not really needed to wait for it anymore
- [x] ~Add dependency for temporary buffer allocation~
- [x] ~Refactor temporary buffer on CPU used for GPU memory communication~
- [x] Check if the workaround for `transform<GPU>` with a nullary callable is still required

### DIFF
--- a/include/dlaf/eigensolver/tridiag_solver/rot.h
+++ b/include/dlaf/eigensolver/tridiag_solver/rot.h
@@ -281,6 +281,11 @@ void applyGivensRotationsToMatrixColumns(comm::Communicator comm_row, comm::Inde
       // Moreover, even if rotations are independent, scheduling all communications all together
       // would require a tag ad hoc ensuring that communication between same ranks do not get mixed
       // (in addition to having a tag ensuring that other calls to this algorithm do not get mixed too).
+      //
+      // Current serialization of the algorithm ensures that communications cannot be scheduled all
+      // together beforehand, and with them, also temporary buffers for the GPU case. If this assumption
+      // will drop, it is relevant to highlight that there is nothing that would stop to schedule and,
+      // more importantly, allocate all of them together.
       tt::sync_wait(
           ex::when_all(ex::just(1),  // workaround for problem with unwrapping
                        ex::when_all_vector(std::move(comm_checkpoints))) |
--- a/include/dlaf/eigensolver/tridiag_solver/rot.h
+++ b/include/dlaf/eigensolver/tridiag_solver/rot.h
@@ -287,9 +287,8 @@ void applyGivensRotationsToMatrixColumns(comm::Communicator comm_row, comm::Inde
       // will drop, it is relevant to highlight that there is nothing that would stop to schedule and,
       // more importantly, allocate all of them together.
       tt::sync_wait(
-          ex::when_all(ex::just(1),  // workaround for problem with unwrapping
-                       ex::when_all_vector(std::move(comm_checkpoints))) |
-          di::transform(di::Policy<DefaultBackend_v<D>>(), [rot, m, col_x, col_y](int, auto&&... ts) {
+          ex::when_all_vector(std::move(comm_checkpoints)) |
+          di::transform(di::Policy<DefaultBackend_v<D>>(), [rot, m, col_x, col_y](auto&&... ts) {
             // Note:
             // each one computes his own, but just stores either x or y (or both if on the same rank)
             if constexpr (D == Device::CPU) {
--- a/include/dlaf/eigensolver/tridiag_solver/rot.h
+++ b/include/dlaf/eigensolver/tridiag_solver/rot.h
@@ -9,6 +9,10 @@
 //
 #pragma once
 
+#ifdef DLAF_WITH_GPU
+#include <whip.hpp>
+#endif
+
 #include "dlaf/common/assert.h"
 #include "dlaf/common/pipeline.h"
 #include "dlaf/common/range2d.h"
@@ -16,11 +20,16 @@
 #include "dlaf/communication/communicator.h"
 #include "dlaf/communication/communicator_grid.h"
 #include "dlaf/communication/kernels/p2p.h"
+#include "dlaf/eigensolver/tridiag_solver/kernels.h"
 #include "dlaf/eigensolver/tridiag_solver/tile_collector.h"
+#include "dlaf/matrix/copy_tile.h"
 #include "dlaf/matrix/distribution.h"
 #include "dlaf/matrix/index.h"
 #include "dlaf/matrix/matrix.h"
 #include "dlaf/matrix/panel.h"
+#include "dlaf/memory/memory_view.h"
+#include "dlaf/sender/policy.h"
+#include "dlaf/sender/transform.h"
 #include "dlaf/sender/transform_mpi.h"
 #include "dlaf/sender/when_all_lift.h"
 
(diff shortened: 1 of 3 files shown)


include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -211,8 +254,9 @@ void applyGivensRotationsToMatrixColumns(comm::Communicator comm_row, comm::Inde
       // would require a tag ad hoc ensuring that communication between same ranks do not get mixed
       // (in addition to having a tag ensuring that other calls to this algorithm do not get mixed too).
       tt::sync_wait(
-          di::whenAllLift(ex::when_all_vector(std::move(comm_checkpoints))) |
-          di::transform(di::Policy<DefaultBackend_v<D>>(), [rot, m, col_x, col_y](auto&&... ts) {
+          di::whenAllLift(ex::just(1),  // workaround for problem with unwrapping
```
albestro, line 257: @msimberg do you want to mention any specific PR/issues here for future reference?

msimberg: Just that https://github.com/eth-cscs/DLA-Future/pull/697 should fix this :)

albestro:
About the last open point in TOOD list
> Add dependency for temporary buffer allocation

Currently, the algorithm schedule an application of a rotation and `sync_wait`s on its completion before going on scheduling the next one. This fact ensures that communications cannot be scheduled all together beforehand, and with them, also temporary buffers for the GPU case. If this assumption will drop, it is relevant to highlight that there is nothing that would stop to schedule, and more importantly allocate, all of them together.

I think at the moment it might be too early to think about it considering current algorithm constraints, so I would just not care. Do you agree? @rasolca @msimberg @teonnik 

In case you agree, I think it is relevant to point that out for future reference.

msimberg (changes requested):

include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -31,43 +37,80 @@ namespace wrapper {
...
+  if constexpr (D == Device::CPU) {
+    return di::whenAllLift(std::forward<CommSender>(comm), dest, tag, col_data, n) |
+           dlaf::comm::internal::transformMPI(sendCol<D, T>);
+  }
+  else if constexpr (D == Device::GPU) {
+    namespace ex = pika::execution::experimental;
+    using dlaf::matrix::internal::CopyBackend_v;
+    using pika::execution::thread_priority;
+
+    return di::whenAllLift(memory::MemoryView<T, Device::CPU>{n}) |
```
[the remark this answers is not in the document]
albestro: I think it is more or less the same concern I stated in https://github.com/eth-cscs/DLA-Future/pull/718#issuecomment-1322268336. Is it?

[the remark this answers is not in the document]
msimberg: Indeed, it looks like it! Sorry for not seeing the connection.

I think this is fine as it is now, but a refactor would probably be a good idea. One potential way of "fixing" this is to have nested `let_value`s, with the outer one taking the communicator, and the inner taking the memory view. In that case the memory view is only going to be allocated when the pipeline allows communication to happen. However, by the time we get to that refactoring we might have refactored the communication functions to work more generically, not just with tiles and the problem might be fixed by that instead...

albestro:
With @teonnik we have found an API mismatch that has been easily solved (#721).

That change should™ not be in conflict with this PR.

include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -31,43 +37,80 @@ namespace wrapper {
...
+  else if constexpr (D == Device::GPU) {
+    namespace ex = pika::execution::experimental;
+    using dlaf::matrix::internal::CopyBackend_v;
+    using pika::execution::thread_priority;
+
+    return ex::just(memory::MemoryView<T, Device::CPU>{n}) |
+           ex::let_value([comm = std::forward<CommSender>(comm), dest, tag, col_data,
+                          n](memory::MemoryView<T, Device::CPU>& mem_view) mutable {
+             auto copy =
+                 di::whenAllLift(mem_view(), col_data, to_sizet(n) * sizeof(T),
```
msimberg, line 73: I think this is ok:
```suggestion
                 ex::just(mem_view(), col_data, to_sizet(n) * sizeof(T),
```
None of the arguments look like senders to me. This very slightly reduces template instantiations (`just` is cheaper than `when_all`).

include/dlaf/eigensolver/tridiag_solver/rot.h
msimberg, line 257: Conversely, I think all arguments are senders here:
```suggestion
          ex::when_all(ex::just(1),  // workaround for problem with unwrapping
```
Alternatively:
```suggestion
          di::whenAllLift(1,  // workaround for problem with unwrapping
```

msimberg (approved):
Two more minor style comments. Sorry for missing these earlier.

teonnik:
LGTM!

Note that the PR may conflict with changes to rot.h in #719 : [include/dlaf/eigensolver/tridiag_solver/rot.h](https://github.com/eth-cscs/DLA-Future/pull/719/files#diff-745638b97d24e04ee5c3094a75fac7810a334a4b4c790afcc395a7d90639dcc7L90-R124)

albestro:
> LGTM!
> 
> Note that the PR may conflict with changes to rot.h in #719 : [include/dlaf/eigensolver/tridiag_solver/rot.h](https://github.com/eth-cscs/DLA-Future/pull/719/files#diff-745638b97d24e04ee5c3094a75fac7810a334a4b4c790afcc395a7d90639dcc7L90-R124)

Yes, I will take care of that once your one gets merged.
Thanks for highlighting that 😉

msimberg (approved):

albestro:
> ## try
> Build failed:
> 
> * [ci/gitlab/full-pipeline](https://gitlab.com/cscs-ci/eth-cscs/DLA-Future/-/pipelines/708238331)

Everything is ok, apart ROCm build because of the `givensRotationOnDevice` kernel.

I would go for changing all guards to `DLAF_WITH_CUDA` because of this

https://github.com/eth-cscs/DLA-Future/blob/1332ce2e576772cd537bb7359d9a7dd3eed5d0fd/src/CMakeLists.txt#L208-L217

waiting for your feedback before proceeding.

@rasolca @teonnik @msimberg @aurianer

msimberg:
> Everything is ok, apart ROCm build because of the `givensRotationOnDevice` kernel.
> 
> I would go for changing all guards to `DLAF_WITH_CUDA` because of this
> 
> https://github.com/eth-cscs/DLA-Future/blob/1332ce2e576772cd537bb7359d9a7dd3eed5d0fd/src/CMakeLists.txt#L208-L217
> 
> waiting for your feedback before proceeding.

Slightly unsure since I don't know the extent of the missing HIP support in this PR, but would it make sense to make sure #692 is first merged, and then do whatever remaining adjustments here to make it work with HIP as well? Regular CUDA kernels should "just work" with HIP anyway so the question is whether this PR requires CUDA library calls that are not trivially translatable to HIP? Outside of the ones that have been updated in #692 of course.

albestro:
bors try

@rasolca as per agreement I unlock this from HIP.

@teonnik as soon as this gets merged, you can proceed with another PR adapting tridiag solver to use this one.

There is the open question about https://github.com/eth-cscs/DLA-Future/pull/718#issuecomment-1322268336 for which I ask all of you @msimberg @teonnik @rasolca if you prefer to have an issue tracking that or it is ok leaving it there like it is.

msimberg:
> bors try
> 
> @rasolca as per agreement I unlock this from HIP.
> 
> @teonnik as soon as this gets merged, you can proceed with another PR adapting tridiag solver to use this one.
> 
> There is the open question about [#718 (comment)](https://github.com/eth-cscs/DLA-Future/pull/718#issuecomment-1322268336) for which I ask all of you @msimberg @teonnik @rasolca if you prefer to have an issue tracking that or it is ok leaving it there like it is.

I guess an issue wouldn't hurt, unless you think it's trivial to remember to come back to it later...

albestro:
@rasolca this is ready for the last bors run after:
- adding note about https://github.com/eth-cscs/DLA-Future/pull/718#issuecomment-1322268336
- https://github.com/eth-cscs/DLA-Future/pull/718/commits/62e98054d3d91427cdfd5b9d021f5b7818ad5c02 revert of the workaround thanks to the new unwrapping (it required a rebase onto new master)

I'm going to mark this as RTMWTP and I give to you full control for calling the last bors 😉

PR #719, opened 2022-11-17
Distributed tridiagonal solver (MC)

teonnik:

PR #721, opened 2022-11-24
Fix requirement mismatch in applyGivenRotations (and improve doc)

albestro:
With @teonnik we realised that there was a mismatch between what tridiagonal solver is going to pass, i.e. columns indices relative to the sub-matrix range, compared to what applyGivenRotation function was expecting, i.e. global columns indices.

This PR fixes that (as tridiagonal solver expect it to use, i.e. with relative indices) and adapt relative test.

Moreover, I realised that documentation was not helping at all, so I tried to improve that. Have a look at that, once we converge on one, there is also the [local version one](https://github.com/eth-cscs/DLA-Future/blob/cd760bee66797830b0f48842994dde1eb50f7f3f/include/dlaf/eigensolver/tridiag_solver/merge.h#L560-L571) to be fixed accordingly.

### DIFF
--- a/include/dlaf/eigensolver/tridiag_solver/merge.h
+++ b/include/dlaf/eigensolver/tridiag_solver/merge.h
@@ -557,18 +557,23 @@ pika::future<std::vector<GivensRotation<T>>> applyDeflation(
                                                              std::move(deflate_fn), std::move(sender)));
 }
 
-// Assumption: the memory layout of the matrix from which the tiles are coming is column major.
-//
-// `tiles`: The tiles of the matrix between tile indices `(i_begin, i_begin)` and `(i_end, i_end)` that
-// are potentially affected by the Givens rotations. `n` : column size
-//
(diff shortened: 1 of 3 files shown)


teonnik (approved):

msimberg (approved):

rasolca (approved):

rasolca:
bors try

albestro:
bors try

some builds failed due to warnings, but apparently tests executed passed without problems. (FIY @teonnik)

albestro:
@rasolca @teonnik @msimberg

Are we done with the doc change?
Do you want me to fix the local one in this PR or in a separate one (in case @teonnik can you fix it in #719)?

rasolca:
> @rasolca @teonnik @msimberg
> 
> Are we done with the doc change? Do you want me to fix the local one in this PR or in a separate one (in case @teonnik can you fix it in #719)?

I would fix it here as well.

rasolca (approved):

rasolca:
bors try

PR #722, opened 2022-11-24
Prepend paths to `CMAKE_MODULE_PATH` to make sure DLA-Future's own `Find` modules take precedence

msimberg:
Instead of appending, prepend DLA-Future's paths to `CMAKE_MODULE_PATH`. This means that DLA-Future's own modules take precedence. This came up in CP2K where DLA-Future would first be configured using its own `FindMKL.cmake` module, but when DLA-Future is found by CP2K it would use CP2K's `FindMKL.cmake` module which doesn't define the same targets. Configuration would then fail.


rasolca (approved):

msimberg:
This should not affect anything else and should be mergeable as is. Holding off on the bors-ing if you want to trigger some other PR first.

PR #723, opened 2022-11-25
Fix MPI tag for band_to_tridiag

rasolca:
Fixes the bug the shows up sometimes in the CI

### DIFF
--- a/include/dlaf/eigensolver/band_to_tridiag/mc.h
+++ b/include/dlaf/eigensolver/band_to_tridiag/mc.h
@@ -1050,18 +1050,20 @@ TridiagResult<T, Device::CPU> BandToTridiagDistr<Backend::MC, D, T>::call_L(
   // The offset is set to the first unused tag by compute_v_tag.
   const comm::IndexT_MPI offset_col_tag = compute_v_tag(nrtile);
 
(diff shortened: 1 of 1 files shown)


rasolca:
tested the failing case (18,4,4) 5000 times without a failure (before ~1/5 was failing)

msimberg (approved):
Nice! I don't really understand what's happening here exactly, but it looks like a bug in some sort of index calculation? Are there any more general lessons to be learned from this? It looks very specific to this particular algorithm.

teonnik (approved):

rasolca:
> Nice! I don't really understand what's happening here exactly, but it looks like a bug in some sort of index calculation? Are there any more general lessons to be learned from this? It looks very specific to this particular algorithm.

The lesson learned: check any possible corner case when reusing MPI tags...
The problem is very specific to the algorithm:
Excluding the last column due to the dependencies two columns of the same block cannot be sent or received at the same and therefore the same tag for different sweeps can be used.
However, in some sweeps the last column is not touched and this can unlock two dependency at the same time: i.e. sending the last two columns, and therefore two different tags are needed.

rasolca:
bors try

PR #724, opened 2022-11-25
Panel StoreTransposed

albestro:
This is an extension of the `Panel` we are already used to, which turns out to be needed for #701.

A typical situation is this one

```c++
Panel<Col> panel;
Panel<Row> panelT

// broadcast transpose
broadcast(..., panel, panelT, ...)
```

where we have `panel` storing a (column) result and the algorithms needs to access its transposed variant `panelT` (row). Currently, we achieve that by mean of broadcast transpose which, given a tile at row `i`, in addition to share it with all ranks of the same row, sends it to all ranks of column (not row) `i` no matter the row they are on.

This is quite handy for various algorithms, like `cholesky`, `gen_to_std`, `reduction_to_band`, but as it is currently implemented presents a constraint.

*The key point is that we want to have a panel we can access with transposed coordinates, distributed according the transposed axis, but that will be stored as the original, i.e. not-tranposed* (computation takes care of data transposition).

Currently `panelT` is allocated as a correctly transposed panel, where also tile should be transposed, but thanks to the fact that we always use it in square matrices with square blocksizes, we can ignore that and skip the data transposition. So we are somehow "misusing" it, but it works because we are in a special case.

But, as soon as we switch to sub-block, the transposition of the tile shape result in the impossibility to "misuse" it, i.e. we cannot store anymore the original tile because of the tranposed shape.

Let's see an example. We have a square matrix `(m, m)` with blocksize `(mb, mb)`.

In this case `panel` will be `(m, mb)` and `panelT` will be `(mb, m)`. Each tile is always `(mb, mb)` in each panel, so it is a special case that works, since shape is not varying on transposition, we can "opt" for storing tiles as they were originally or tranposed (and we go for the former one).

Now we want to shrink the column panel by reducing the columns to `b`, leading to
- `panel` being `(m, b)` with tiles `(mb, b)`
- `panelT` being `(b, m)` with tile `(b, mb)`

as it can be seen, shape got tranposed, so we cannot store anymore the original tile.

What this extension do is to create a row panel distributed as the classic one, but where each tile is stored "tranposed" with respect to how it should be. So, if we have `panelT` that classically used to have `(b, mb)`, with the new extension it is possible to deal with tiles whose shape gets tranposed, resulting in tiles of `(mb, b)`, i.e. as they were in the original `panel`.

It's like having a `panelT` where each tile gets transposed individually a second time (in addition that goes from `panel` col to `panelT` row) to be stored as they were originally.

It starts as a draft PR so that we can opt for best naming of things. I will try to highlight main points in the comments.

TODO:
- [x] Adapt/extend test
- [x] Feedbacks on naming of `StoreTransposed`
- [x] Feedbacks on naming of `setWidth`/`setHeight`
- [x] Co-existance of both panel types (gen2std uses a panel in 2 ways, one of them requires the classic panel) or alternative implementation
- [x] Evaluate constraining `Panel<StoreTransposed>` to specific distributions or improve implementation (in case, this might be partially addressed here, and partially later tracking this with an issue)

### DIFF
--- a/include/dlaf/common/index2d.h
+++ b/include/dlaf/common/index2d.h
@@ -83,13 +83,13 @@ public:
 
   /// Return a copy of the row or the col index as specified by @p rc.
   template <Coord rc>
-  constexpr IndexT get() const noexcept {
-    if (rc == Coord::Row)
+  IndexT get() const noexcept {
+    if constexpr (rc == Coord::Row)
       return row_;
     return col_;
   }
 
-  constexpr IndexT get(const Coord rc) const noexcept {
+  IndexT get(const Coord rc) const noexcept {
     if (rc == Coord::Row)
       return row_;
     return col_;
--- a/include/dlaf/matrix/panel.h
+++ b/include/dlaf/matrix/panel.h
@@ -15,6 +15,7 @@
 #include "dlaf/common/assert.h"
 #include "dlaf/common/index2d.h"
 #include "dlaf/common/pipeline.h"
+#include "dlaf/common/range2d.h"
 #include "dlaf/common/vector.h"
 #include "dlaf/communication/communicator.h"
 #include "dlaf/communication/communicator_grid.h"
@@ -28,15 +29,17 @@
 namespace dlaf {
 namespace matrix {
 
+enum class StoreTransposed : bool { Yes = true, No = false };
+
 /// Panel (1D array of tiles)
 ///
 /// 1D array of tiles, i.e. a Row or Column panel strictly related to a given dlaf::Matrix (from the
 /// coords point of view)
-template <Coord axis, class T, Device D>
+template <Coord axis, class T, Device D, StoreTransposed isStoredTranposed = StoreTransposed::No>
 struct Panel;
 
 template <Coord axis, class T, Device D>
-struct Panel<axis, const T, D> {
+struct Panel<axis, const T, D, StoreTransposed::No> {
   // Note:
   // This specialization acts as base for the RW version of the panel,
   // moreover allows the casting between references (i.e. Panel<const T>& = Panel<T>)
@@ -62,11 +65,6 @@ struct Panel<axis, const T, D> {
   auto iteratorLocal() const noexcept {
     return common::iterate_range2d(LocalTileIndex(coord, rangeStartLocal()),
                                    LocalTileIndex(coord, rangeEndLocal(), 1));
-  }
-
-  /// Return the rank which this (local) panel belongs to
-  auto rankIndex() const noexcept {
-    return dist_matrix_.rankIndex();
   }
 
   /// Return Distribution used for construction
@@ -282,7 +280,7 @@ struct Panel<axis, const T, D> {
 
   /// Get the current height of the row panel.
   template <Coord A = axis, std::enable_if_t<A == axis && Coord::Row == axis, int> = 0>
-  SizeType getHeight() noexcept {
+  SizeType getHeight() const noexcept {
     return dim_ < 0 ? dist_matrix_.blockSize().template get<axis>() : dim_;
   }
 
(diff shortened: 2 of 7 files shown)


rasolca (commented):
LGTM

include/dlaf/matrix/panel.h
```diff
@@ -442,23 +438,139 @@ struct Panel<axis, const T, D> {
...
+  template <Coord A = axis, std::enable_if_t<A == axis && Coord::Row == axis, int> = 0>
+  SizeType getHeight() const noexcept {
+    return BaseT::getWidth();
+  }
+
+  // TODO fix setHeight (and setWidth accordingly)
+  template <Coord A = axis, std::enable_if_t<A == axis && Coord::Col == axis, int> = 0>
+  SizeType getWidth() const noexcept {
+    return BaseT::getHeight();
+  }
```
albestro, line 522: Width/Height are maintained as they were originally also for `StoreTransposed::Yes`, so the mutable size is:
- "width" in a Col panel
- "height" in a Row panel

To me it sounds reasonable, but before removing the TODO I wait for your feedbacks.

rasolca (approved):

rasolca:
cscs-ci run

include/dlaf/common/index2d.h
```diff
@@ -83,13 +83,13 @@ class basic_coords {
 
   /// Return a copy of the row or the col index as specified by @p rc.
   template <Coord rc>
-  constexpr IndexT get() const noexcept {
-    if (rc == Coord::Row)
+  IndexT get() const noexcept {
```
msimberg, line 86: I think you could keep the `constexpr` here (though I don't know if there's a use case for having it `constexpr`).

test/unit/communication/test_broadcast_panel.cpp
```diff
@@ -98,25 +94,37 @@ TYPED_TEST(PanelBcastTest, BroadcastRow) {
...
 std::vector<config_t> test_params_bcast_transpose{
     {{0, 0}, {1, 1}, {0, 0}},    // empty matrix
     {{10, 10}, {2, 2}, {5, 5}},  // empty panel (due to offset)
     {{20, 20}, {2, 2}, {9, 9}},  // just last tile (communicate without transpose)
     {{25, 25}, {5, 5}, {1, 1}},
 };
 
-template <class TypeParam, Coord PANEL_SRC_AXIS>
+template <class TypeParam, Coord AxisSrc, StoreTransposed storageT = StoreTransposed::No>
 void testBrodcastTranspose(const config_t& cfg, comm::CommunicatorGrid comm_grid) {
```
msimberg, line 117: Not from this PR, but:
```suggestion
void testBroadcastTranspose(const config_t& cfg, comm::CommunicatorGrid comm_grid) {
```

test/unit/communication/test_broadcast_panel.cpp
```diff
@@ -98,25 +94,37 @@ TYPED_TEST(PanelBcastTest, BroadcastRow) {
...
+
 std::vector<config_t> test_params_bcast_transpose{
     {{0, 0}, {1, 1}, {0, 0}},    // empty matrix
     {{10, 10}, {2, 2}, {5, 5}},  // empty panel (due to offset)
     {{20, 20}, {2, 2}, {9, 9}},  // just last tile (communicate without transpose)
     {{25, 25}, {5, 5}, {1, 1}},
 };
 
-template <class TypeParam, Coord PANEL_SRC_AXIS>
+template <class TypeParam, Coord AxisSrc, StoreTransposed storageT = StoreTransposed::No>
```
msimberg, line 116: Maybe in the test `storageT` can be explicit (i.e. not defaulted)?

msimberg (approved):
Only superficial comments from me, no blockers!

test/unit/communication/test_broadcast_panel.cpp
albestro: The rationale was: I don't want to touch the code, so I'd like to have tests that represents also this aspect.

I agree it is not so "transparent". Considering also #725, would you prefer to have the storage parameter explicit also in the code?

msimberg: So I think it's ok to have it be defaulted on the `Panel` class itself. To me it's clearer if it's explicit in the test because then I can see for sure that both cases are being tested, but it's really not a big issue. I'm happy the way it is as well.

include/dlaf/common/index2d.h
msimberg: As discussed separately, it's fine to remove the `constexpr`. It couldn't be used in a `constexpr` context anyway without further changes (the constructor would at least need to be `constexpr`).

albestro: I removed it because the returned parameter is not built as a constexpr. On the other side, I add it to the `if` because it works on a template parameter.

As discussed offline, it could be potentially be a constexpr function by making constexpr also the constructor. If you agree I would take this change here, and if in the future we introduce more constexpr we can fix this too.

rasolca:
cscs-ci run

PR #725, opened 2022-11-25
Panel StoreTranposed adaptations to algorithms

albestro:
This is a fork of #724 to isolate changes in the algorithms.

In the intentions, just the type of panels used transposed are changed, and it should not really affect anything, nor from the API point of view, nor from the allocation one (for current use-cases it should be allocated exactly as it was before).

Depending one the decisions in #724 this can change, but IMHO I hope not much because keeping the same API represents a big pro.


rasolca (approved):

msimberg (approved):

rasolca:
cscs-ci run

PR #739, opened 2022-12-05
Use more ROCm libs provided externally + minor spack change

### LINKED ISSUE #737: Install more HIP dependencies as externals
See https://github.com/eth-cscs/DLA-Future/pull/692#issuecomment-1330867621. 

@albestro do you want to rebase and push your commit from that comment as a PR or would you like me to do it?

albestro:
Close #737

@msimberg @rasolca 

Few days ago I tested that it was able to build both docker images (build + deploy) without errors (but there were compile warnings in dla-future...)

Do you want me to do/run something specific before merging this?


msimberg:
> Close #737
> 
> @msimberg @rasolca
> 
> Few days ago I tested that it was able to build both docker images (build + deploy) without errors (but there were compile warnings in dla-future...)
> 
> Do you want me to do/run something specific before merging this?

Do you have a link to the warnings? I assume they were new warnings that don't appear if these are not external?

msimberg (approved):

albestro:
> > Close #737
> > @msimberg @rasolca
> > Few days ago I tested that it was able to build both docker images (build + deploy) without errors (but there were compile warnings in dla-future...)
> > Do you want me to do/run something specific before merging this?
> 
> Do you have a link to the warnings? I assume they were new warnings that don't appear if these are not external?

I've seen them while building in my VM, but actually if you access the complete raw log of the CI, you can find them 😉 

This is the link to the [last](https://cdn.artifacts.gitlab-static.net/01/82/0182a4f5b6ec7369b61f9710697307dfe0e62b69b3c9e8aa12ec24305628843a/2022_12_05/3424339670/3741323197/job.log?response-content-type=text%2Fplain%3B%20charset%3Dutf-8&response-content-disposition=inline&Expires=1670317070&KeyName=gprd-artifacts-cdn&Signature=-K5n4t1EbMH1QlZf1dnwiA-lj0E=) one.

It seems that warning are enabled but they are not considered errors as for other builds.

rasolca:
> I've seen them while building in my VM, but actually if you access the complete raw log of the CI, you can find them 😉
> 
> It seems that warning are enabled but they are not considered errors as for other builds.

When we call cublas/solver rocblas/solver we should make sure that the integer type is correct (as e.g https://github.com/eth-cscs/DLA-Future/blob/master/include/dlaf/blas/tile.h#L392-L404) such that warnings should not be generated.

Moreover the variant `+ci-test` should enable `-Werror`. https://github.com/eth-cscs/DLA-Future/blob/master/spack/packages/dla-future/package.py#L128
Can you please check why it is not the case?

Do you want to fix these problems in this PR or in a new one?

msimberg:
@albestro my main question was whether the warnings are new for this PR or not, and it seems like the answer is that the warnings were probably introduced by #692 (but at least not from this PR).

So I think it would make sense to take care of the warnings separately, and I can have a go at doing so (unless you've already started @albestro...).

albestro:
> @albestro my main question was whether the warnings are new for this PR or not, and it seems like the answer is that the warnings were probably introduced by #692 (but at least not from this PR).

Yes, exactly. That's what I meant showing you the link to both this PR pipeline and the on currently on master. Unfortunately, I've never checked before merging of #692.

> So I think it would make sense to take care of the warnings separately, and I can have a go at doing so (unless you've already started @albestro...).

I agree with @msimberg . @rasolca I can take care of it in the next hours, but I would do it in a separate PR as suggested.

If you agree, I will open an issue about this 😉

rasolca (approved):

PR #743, opened 2022-12-07
Enable warnings as error for all languages

### LINKED ISSUE #740: AMD CI does not report warnings as errors
> I've seen them while building in my VM, but actually if you access the complete raw log of the CI, you can find them 😉
> 
> It seems that warning are enabled but they are not considered errors as for other builds.

When we call cublas/solver rocblas/solver we should make sure that the integer type is correct (as e.g https://github.com/eth-cscs/DLA-Future/blob/master/include/dlaf/blas/tile.h#L392-L404) such that warnings should not be generated.

Moreover the variant `+ci-test` should enable `-Werror`. https://github.com/eth-cscs/DLA-Future/blob/master/spack/packages/dla-future/package.py#L128
Can you please check why it is not the case?

Do you want to fix these problems in this PR or in a new one?

_Originally posted by @rasolca in https://github.com/eth-cscs/DLA-Future/issues/739#issuecomment-1339153660_

albestro:
Fix #740

With `+ci-test` variant, we were passing `-Werror` to CMake, but just for the C++ language, and not for GPU specific ones.

### DIFF
--- a/include/dlaf/gpu/cub/api.h
+++ b/include/dlaf/gpu/cub/api.h
@@ -104,19 +104,18 @@ struct Sum {
 
 class DeviceReduce {
 public:
-  template <typename InputIteratorT, typename OutputIteratorT, typename ReduceOpT, typename T,
-            typename NumItemsType>
+  template <typename InputIteratorT, typename OutputIteratorT, typename ReduceOpT, typename T>
   __host__ static whip::error_t Reduce(void* d_temp_storage, std::size_t& temp_storage_bytes,
-                                       InputIteratorT d_in, OutputIteratorT d_out,
-                                       NumItemsType num_items, ReduceOpT reduction_op, T init,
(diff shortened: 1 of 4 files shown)


albestro:
cscs-ci run

albestro:
This CI run should result in an error at least for the AMD build (not sure about CUDA).

If it fails, it is a success 😄

After that, I will fork this PR trying to fix existing warnings.

albestro:
CI now reports error also for AMD. No warnings were reported for NVIDIA, so I will just have to target the AMD build.

msimberg (approved):

albestro:
If no other change is required, this can be merged without being re-tested, since it has been tested the last commit merged from the forked PR (but super-powers are needed in that case).

teonnik (approved):

rasolca (approved):

rasolca:
cscs-ci run

PR #748, opened 2022-12-14
Don't include algorithm implementations in public headers

msimberg:
This updates a few more algorithms to only include the `api.h` header, instead of the full `impl.h` header. The difference is not dramatic, but I measured e.g. a recompilation time of `test_tridiag_solver_distributed` of 44 seconds with these changes vs 47 seconds before (after first fully compiling the test, then `touch`ing `include/dlaf/eigensolver/tridiag_solver.h`).

If I instead `touch` the `impl.h` file after fully building the build time is roughly the same if built in parallel, but the test file itself doesn't have to be rebuilt so the total amount of work is less.

I ask you to have a closer look especially at the ones that used to have ETI done separately with a `LOCAL` macro for GPUs. It's not 100% clear that I'm not adding instantiations that are actually not used.

### DIFF
--- a/include/dlaf/eigensolver/bt_reduction_to_band.h
+++ b/include/dlaf/eigensolver/bt_reduction_to_band.h
@@ -12,7 +12,7 @@
 #include <blas.hh>
 
 #include "dlaf/communication/communicator_grid.h"
-#include "dlaf/eigensolver/bt_reduction_to_band/impl.h"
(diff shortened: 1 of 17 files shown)


msimberg:
cscs-ci run

rasolca (approved):

albestro (approved):

PR #758, opened 2023-01-03
Fix deployment

albestro:
I started addressing the deployment problems.

~For the moment I started from an MC configuration, with MKL or OpenBLAS.~

Changelog:
- Fix LAPACK (non-MKL) dependency management
- Add `pika-algorithms` to dependencies
- Reorder `DLAFConfig.cmake` for an easier mapping with `CMakeLists.txt`
- Update spack package (remove unused variable + make GPU variable conditionally set)

PR #764, opened 2023-01-09
Make `miniapp` an independent project

albestro:
This is an attempt to make the `miniapp` folder an independent project. Apart the more clear separation between library and miniapps, this would also allow to test the deployment of the library.

The main requirement was that `miniapp` should be an independent project, so that it can be built separately specifying where DLAF library is located, but also together with the library seeming-lessly so that developer can easily work with them.

TODO
- [x] Requires #758
- [ ] Do we want to add a more specific check for `miniapp`? (see https://github.com/eth-cscs/DLA-Future/pull/764/commits/1eb8c749dac6640ed324b1cb59a407d2552444b0)
- [x] Open issue about https://github.com/eth-cscs/DLA-Future/pull/764/commits/6873c3268070c2872b952df39923a95bf49f34e9 https://github.com/eth-cscs/DLA-Future/issues/779

Changelog:
- For each EXPORTED targets an alias namespaced DLAF:: target is created
- Install also `DLAF_AddTargetWarnings` CMake script
- If `DLAF_BUILD_MINIAPPS=on`, create an empty `DLAFConfig.cmake` to use for embedded projects for which everything is already set correctly
- Make miniapp an independent project and migrate to using `DLAF::` namespaced targets
- Miniapp do not use anymore precompiled headers
- Remove miniapp dependency on `DLAF_test` (see issue #779)
- Exclude miniapp from check about "files with other extensions"

### DIFF
--- a/miniapp/CMakeLists.txt
+++ b/miniapp/CMakeLists.txt
@@ -14,12 +14,13 @@ project(DLAF-miniapps)
 list(PREPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 
 find_package(DLAF REQUIRED)
+find_package(pika 0.11.0 REQUIRED)
 
 include(DLAF_AddMiniapp)
 include(DLAF_AddTargetWarnings)
 
 add_library(DLAF_miniapp INTERFACE)
-target_link_libraries(DLAF_miniapp INTERFACE DLAF::DLAF)
+target_link_libraries(DLAF_miniapp INTERFACE DLAF::DLAF pika::pika)
 target_include_directories(DLAF_miniapp INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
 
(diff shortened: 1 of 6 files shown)


miniapp/CMakeLists.txt
```diff
@@ -8,17 +8,22 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
+cmake_minimum_required(VERSION 3.22)
+project(dlaf-miniapps)
```
msimberg, line 12:
```suggestion
project(DLAF-miniapps)
```
maybe? Just to keep the all-caps of `DLAF` for the main project.

msimberg (changes requested):
Looks good to me.

Do I understand correctly that you (or someone) is planning on separately adding the CI job that then tests that `find_package(DLAF)` works correctly?

rasolca (approved):

albestro:
cscs-ci run

msimberg (approved):

Issue #688, opened 2022-10-27
Reenable ignored warnings after pika 0.12.0 release


msimberg:
#674 disables some warnings that to the best of my knowledge are bogus. In pika 0.10.0 the warnings will be disabled locally in pika itself, but until then they have to be disabled in DLA-Future. This is the pika PR that disables them in pika itself: https://github.com/pika-org/pika/pull/489.

Edit: it seems like 0.10.0 is not enough and they will have to be silenced a different way. I've updated the title to reflect that.

msimberg:
This was fixed by #765, which locally ignores warnings rather than ignoring them through compiler flags.

PR #765, opened 2023-01-09
Ignore bogus pika warnings locally in source files

msimberg:
"Fixes" https://github.com/eth-cscs/DLA-Future/issues/688 by locally ignoring warnings in the source files instead of through compiler flags. I have not been able to silence these properly from pika so doing the next best thing and silencing them locally instead.

It seems like these bogus warnings may no longer be present in GCC 12, but I'm leaving the `#pragma GCC diagnostic ignored` for all versions because I've only tested with 12.2.0.

~To do: check if the `stringop-overflow` warning still needs to be silenced.~ It seems to not be required anymore.

### DIFF
--- a/include/dlaf/common/pipeline.h
+++ b/include/dlaf/common/pipeline.h
@@ -82,9 +82,16 @@ public:
     pika::lcos::local::promise<T> promise_next;
     sender_ = promise_next.get_future();
 
+#if defined(__GNUC__)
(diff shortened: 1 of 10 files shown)


msimberg:
cscs-ci run

rasolca (approved):

albestro (approved):

PR #782, opened 2023-01-17
Fix tiny typo in and add clarification to `--dlaf:eigensolver-min-band` command line option description

msimberg:
I'm also wondering if the description should say something like `When larger than the block size, the block size will be used instead.`?


rasolca:
> When larger than the block size, the block size will be used instead.

👍🏻

msimberg:
> > When larger than the block size, the block size will be used instead.
> 
> 👍🏻

See https://github.com/eth-cscs/DLA-Future/pull/782/commits/88057a9d4885e1075a17453a8fab9e66999a9e02.

rasolca (approved):

rasolca:
cscs-ci run

PR #785, opened 2023-01-20
Test about-to-be whip 0.1.0

msimberg:
This is just to check that the commit that will become whip 0.1.0 works fine in DLA-Future. There were only very minor changes, so I don't expect anything special here, but want to run it in any case.


msimberg:
cscs-ci run

PR #786, opened 2023-01-23
Update LUMI and hohgant system definitions

msimberg:
Also update plotting/benchmark scripts to allow fractional nodes.

### DIFF
--- a/scripts/gpu2ranks_slurm
+++ b/scripts/gpu2ranks_slurm
@@ -4,6 +4,11 @@
 #
 # This script is used to restrict visible GPUs when using multiple ranks per
 # node with slurm. It can be used with CUDA or HIP.
+#
+# Note that this script simply restricts the visible devices to SLURM_LOCALID.
+# If there is a different relationship between NUMA nodes and GPUs the remapping
+# must be done in systems.py by setting the --cpu-bind flag to srun. nvidia-smi
+# and rocm-smi can be used to find the affinity of GPUs to NUMA nodes.
 
 set -eu
 
--- a/scripts/gpu2ranks_slurm
+++ b/scripts/gpu2ranks_slurm
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Usage: `mpirun gpu2ranks_slurm <my_commands>`
+#
+# This script is used to restrict visible GPUs when using multiple ranks per
+# node with slurm. It can be used with CUDA or HIP.
+
+set -eu
+
+export CUDA_VISIBLE_DEVICES=$SLURM_LOCALID
+export ROCR_VISIBLE_DEVICES=$SLURM_LOCALID
+
+eval "$@"
--- a/scripts/systems.py
+++ b/scripts/systems.py
@@ -199,7 +199,7 @@ csc["lumi-cpu"] = {
 #SBATCH --job-name={run_name}_{nodes}
 #SBATCH --time={time_min}
 #SBATCH --nodes={nodes}
-#SBATCH --account=project_465000151
+#SBATCH --account=project_465000105
 #SBATCH --partition=standard
 #SBATCH --hint=multithread
 #SBATCH --output=output.txt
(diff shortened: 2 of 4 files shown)


scripts/systems.py
```diff
@@ -161,16 +224,17 @@
     "Allowed rpns": [8],
     "Multiple rpn in same job": True,
     "GPU": True,
-    "Run command": "srun -u -n {total_ranks} --cpu-bind=core -c {threads_per_rank}",
+    # Based on https://docs.lumi-supercomputer.eu/runjobs/scheduled-jobs/distribution-binding/#gpu-binding
+    "Run command": "srun -u -n {total_ranks} --cpu-bind=mask_cpu:ff000000000000,ff00000000000000,ff0000,ff000000,fe,ff00,ff00000000,ff0000000000 gpu2ranks_slurm",
```
rasolca, line 228: Why the 5th is `FE`?

scripts/gpu2ranks_slurm
```diff
@@ -0,0 +1,13 @@
...
+
+# Usage: `mpirun gpu2ranks_slurm <my_commands>`
+#
+# This script is used to restrict visible GPUs when using multiple ranks per
+# node with slurm. It can be used with CUDA or HIP.
+
+set -eu
+
+export CUDA_VISIBLE_DEVICES=$SLURM_LOCALID
+export ROCR_VISIBLE_DEVICES=$SLURM_LOCALID
```
rasolca, line 11: Might be dangerous. Is it not better to have system specific scripts that take care of GPU "position"?
(I see that for lumi you use a special CPU binding to take care of it what about hohgant?)

scripts/systems.py
msimberg: From the link in the comment:

```
The LUMI-G compute nodes have the low-noise mode activated. This mode reserve 1 core to the operating system. As a consequence only 63 cores are available to the jobs. Jobs requesting 64 cores/node will never run.
```

So one of the ranks will use one thread less because of that.

scripts/gpu2ranks_slurm
msimberg: There are two options for how to do the remapping: 1. Remap GPUs based on rank or 2. Remap "ranks" based on GPUs. For LUMI the scripts now do 2., i.e. available cores are remapped for each rank so that the rank-GPU mapping is the straightforward `{CUDA,ROCR}_VISIBLE_DEVICES=$SLURM_LOCALID`. If we do option 2. on all systems this will will work fine on those systems as well, but I definitely see your point about this being potentially dangerous.  

Speaking of which, I obviously should've done the same for hohgant... If I'm interpreting the output of `nvidia-smi` correctly the mapping is reversed (first numa domain has affinity to the last GPU etc.). I'll do some checks there before updating this again.

msimberg: I added masks for hohgant based on `rocm-smi`/`nvidia-smi` (see comments in `systems.py`). The hohgant topology on `amdgpu` is the same as on lumi. I ran a few tests with `miniapp_cholesky`/`miniapp_triangular_solver` with both the "naive"/"standard" bindings and the "affinity" bindings on `nvgpu` and noticed no difference, but I've left the mapping in place because it should not hurt at least.

msimberg: I have not done anything about the `gpu2ranks_slurm` script. If you prefer I can rename this into two `gpu2ranks_{hohgant,lumi}` scripts?

rasolca: No it is fine. However, please add a note in this script that GPU affinity has to be taken care in systems.py.

scripts/systems.py
rasolca: Ok,
Anyway, shouldn't it be:
`ff00000000000000ff000000000000,ff00000000000000ff00000000000000,ff00000000000000ff0000,ff00000000000000ff000000,fe00000000000000fe,ff00000000000000ff00,ff00000000000000ff00000000,ff00000000000000ff0000000000`?

scripts/gpu2ranks_slurm
msimberg: See https://github.com/eth-cscs/DLA-Future/pull/786/commits/7ffdcdebf62a52b7d7f3639c76145d266026f413. Is that ok?

scripts/systems.py
msimberg: It can be, but need not be since we're not using hyperthreads. Would you prefer the full mask?

rasolca: Yes, IIRC when we tested on daint (long time ago) it was giving better results.

msimberg: Done in https://github.com/eth-cscs/DLA-Future/pull/786/commits/92d50fe9e07ac7a1c42c7d7f7add9af7ab56d855.

rasolca (approved):

Issue #636, opened 2022-09-01
Investigate potential performance issues with `thrust::merge()` in the tridiagonal eigensolver


teonnik:
The kernel may be blocking the calling thread.

msimberg:
This requires Thrust 1.16.0 which comes with HIP 5.3.0. It's unclear which version of CUDA comes with 1.16.0.

msimberg:
I think this was fixed by #788.

PR #788, opened 2023-01-25
Make kernels in tridiagonal solver asynchronous

### LINKED ISSUE #784: Make memory copies actually asynchronous in `tridiag_solver/kernels.cu`
The asynchronous copies in https://github.com/eth-cscs/DLA-Future/blob/ee3a3223fb73d4b1f42d4fb5be88907b39b2c4ac/src/eigensolver/tridiag_solver/kernels.cu#L184-L187 and https://github.com/eth-cscs/DLA-Future/blob/ee3a3223fb73d4b1f42d4fb5be88907b39b2c4ac/src/eigensolver/tridiag_solver/kernels.cu#L106-L109 use a stack-local variable as the destination but has no synchronization to ensure that the destination lives until the copy is finished. This needs to be restructured to most likely pass in a reference to the destination, and have the destination lifetime managed outside of the functions.

msimberg:
Fixes #784 and sort of #636.

This changes all the calls to thrust in the tridiagonal solver to be asynchronous. It does not use `par_nosync` because:
1. it requires thrust 1.16 (which requires either a standalone thrust for CUDA or CUDA 12, and HIP 5.3.0)
2. it still doesn't make all kernels asynchronous (such as reductions which reduce to a host variable)
Instead this moves the calls to thrust into separate kernels and calls the device-side versions of the thrust functions (as suggested in https://github.com/eth-cscs/DLA-Future/blob/cbceab9ff876d98e30a093aef82acbc7ba6915cb/src/eigensolver/tridiag_solver/kernels.cu#L637-L638). It then manages the results manually using `MemoryChunk`s and asynchronous `memcpy`s.

Disappointingly this does not seem to change performance at all from a strong scaling test with 20k matrices.

On the positive side it seems to anecdotally "fix" the hangs I was seeing in `miniapp_tridiag_solver`. Where master would very frequently hang or segfault on the miniapp I was able to run a full strong scaling test with a few different matrix and block sizes on hohgant without a single failure. @albestro it'd be great if you could also at least do a quick check with your setup to see if the hangs still happen for you.

Finally, while performance doesn't seem to have improved, it did do something good. On a 10k matrix size with 1k block size test on a single rank, a single run of the old version showed ~16000 calls to `cudaEventRecord` (which is roughly the number of kernel launches). The new version has ~6000 calls. It's still a lot of (small) kernel launches which is probably still related to the performance issues, but at least it's less than before. Not quite a fair comparison, but `miniapp_triangular_solver` on a 10k/1k single iteration run does about 1000 calls to `cudaEventRecord`.

On the functions that I've updated I've done some variable renamings to add `_ptr` suffixes and explicit `host_` and `device_` prefixes. I may have made things less consistent with this, but please have a look and see if they seem like an improvement to you.

This also adds some more specialized code paths for GPU vs CPU which I'm not a big fan of, but refactoring those to be more generic is a task for the future.

To do:
- [x] Make this compile with HIP. Based on CI it seems like hipthrust may not have device-side versions of all the algorithms. I hope that's not the case...


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

include/dlaf/eigensolver/tridiag_solver/kernels.h
```diff
@@ -108,8 +110,22 @@ auto cuppensDecompAsync(TopTileSender&& top, BottomTileSender&& bottom) {
   namespace ex = pika::execution::experimental;
   namespace di = dlaf::internal;
 
-  auto sender = ex::when_all(std::forward<TopTileSender>(top), std::forward<BottomTileSender>(bottom));
-  return di::transform(di::Policy<DefaultBackend_v<D>>(), cuppensDecomp_o, std::move(sender));
+  using ElementType = dlaf::internal::SenderElementType<TopTileSender>;
+  constexpr auto backend = dlaf::DefaultBackend_v<D>;
+
+  if constexpr (backend == dlaf::Backend::GPU) {
```
msimberg, line 116: Note to self: be consistent about dispatching based on backend or device.

msimberg: This is now using `if constexpr (D == Device::CPU)` as elsewhere.

src/eigensolver/tridiag_solver/kernels.cu
```diff
@@ -171,20 +170,29 @@ DLAF_GPU_ASSEMBLE_RANK1_UPDATE_VECTOR_TILE_ETI(, float);
...
 #endif
-  auto d_max_ptr = thrust::max_element(par.on(stream), arr, arr + len);
-  T max_el;
-  // TODO: this is a performance pessimization, the value is on device
-  whip::memcpy_async(&max_el, d_max_ptr, sizeof(T), whip::memcpy_device_to_host, stream);
-  return max_el;
+
+  // NOTE: This could also use max_element. However, it fails to compile with
+  // HIP, so we use reduce as an alternative which works with both HIP and CUDA.
+  *device_max_el_ptr = thrust::reduce(par, begin_ptr, end_ptr, *begin_ptr, thrust::maximum<T>());
```
msimberg, line 183: This was previously using `max_element`. However, it doesn't compile with HIP in a device function (complains about internals trying to "call a host-device function from a host function"). For some reason `reduce` doesn't have this problem, so I'm using that instead.

src/eigensolver/tridiag_solver/kernels.cu
```diff
@@ -607,43 +615,59 @@ DLAF_GPU_COPY_1D_ETI(, float);
...
-  // The number of non-deflated values
+__global__ void stablePartitionIndexOnDevice(SizeType n, const ColType* c_ptr, const SizeType* in_ptr,
+                                             SizeType* out_ptr, SizeType* device_k_ptr) {
 #ifdef DLAF_WITH_CUDA
   constexpr auto par = thrust::cuda::par;
 #elif defined(DLAF_WITH_HIP)
   constexpr auto par = thrust::hip::par;
 #endif
-  SizeType k = n - thrust::count(par.on(stream), c_ptr, c_ptr + n, ColType::Deflated);
+  // Alternative: copy, then partition in place. Is it faster?
```
msimberg, line 634: Remove this comment. Look at it separately.

msimberg:
cscs-ci run

src/eigensolver/tridiag_solver/kernels.cu
```diff
@@ -171,20 +170,29 @@ DLAF_GPU_ASSEMBLE_RANK1_UPDATE_VECTOR_TILE_ETI(, float);
...
+  *device_max_el_ptr = thrust::reduce(par, begin_ptr, end_ptr, *begin_ptr, thrust::maximum<T>());
+}
+
+template <class T>
+void maxElementInColumnTile(const matrix::Tile<const T, Device::GPU>& tile, T* host_max_el_ptr,
+                            T* device_max_el_ptr, whip::stream_t stream) {
+  SizeType len = tile.size().rows();
+  const T* arr = tile.ptr();
+
+  DLAF_ASSERT(len > 0);
```
rasolca, line 192:
```suggestion
  DLAF_ASSERT(len > 0, len);
```

rasolca (approved):
It looks ok. Also the new varaible naming is better.

include/dlaf/eigensolver/tridiag_solver/kernels.h
```diff
@@ -229,8 +249,28 @@ DLAF_MAKE_CALLABLE_OBJECT(maxElementInColumnTile);
...
+    return std::forward<TileSender>(tile) |
+           di::transform(di::Policy<backend>(), maxElementInColumnTile_o);
+  }
+  else {
+#ifdef DLAF_WITH_GPU
+    return ex::when_all(std::forward<TileSender>(tile),
+                        ex::just(memory::MemoryChunk<ElementType, Device::CPU>{1},
+                                 memory::MemoryChunk<ElementType, Device::GPU>{1})) |
+           ex::let_value([](auto& tile, auto& host_max_el, auto& device_max_el) {
+             DLAF_ASSERT(tile.is_ready(), "");
```
albestro, line 267: debugging purpose?

albestro (approved):
Just a minor comment

msimberg:
cscs-ci run

msimberg:
cscs-ci run

include/dlaf/eigensolver/tridiag_solver/kernels.h
msimberg: Removed.

msimberg:
cscs-ci stop

src/eigensolver/tridiag_solver/kernels.cu
msimberg: Removed.

msimberg: https://github.com/eth-cscs/DLA-Future/issues/790

msimberg:
cscs-ci run

PR #789, opened 2023-01-26
Add tunable block size for band matrix in band_to_tridiag

rasolca:


rasolca:
cscs-ci run

test/unit/eigensolver/test_band_to_tridiag.cpp
```diff
@@ -157,24 +159,29 @@ void testBandToTridiag(CommunicatorGrid grid, blas::Uplo uplo, const SizeType ba
 TYPED_TEST(EigensolverBandToTridiagTest, CorrectnessLocalFromCPU) {
   const blas::Uplo uplo = blas::Uplo::Lower;
 
-  for (const auto& [m, mb, b] : sizes)
+  for (const auto& [m, mb, mb_1d, b] : sizes) {
+    getTuneParameters().band_to_tridiag_1d_block_size_base = mb_1d;
```
rasolca, line 163: For the local implementation it doesn't have any effect. I modified the parameter anyway, let me know if you prefer to not touch it.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -1040,14 +1041,14 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
   // The offset is set to the first unused tag by compute_copy_tag.
   const comm::IndexT_MPI offset_v_tag = compute_copy_tag(nrtile, false);
 
-  auto compute_v_tag = [offset_v_tag](SizeType i) {
+  auto compute_v_tag = [offset_v_tag](SizeType i, bool is_bottom) {
```
msimberg, line 1044: Maybe consider an enum to replace the `bool is_bottom`, but not a blocker.

test/unit/eigensolver/test_band_to_tridiag.cpp
```diff
@@ -131,6 +132,7 @@ void testBandToTridiag(const blas::Uplo uplo, const SizeType band_size, const Si
 template <Device D, class T>
 void testBandToTridiag(CommunicatorGrid grid, blas::Uplo uplo, const SizeType band_size,
                        const SizeType m, const SizeType mb) {
+  std::cout << m << std::endl;
```
msimberg, line 135: Debugging output?

test/unit/eigensolver/test_band_to_tridiag.cpp
msimberg: Fine by me. Would it ever be useful to make it have an effect for the local implementation?

msimberg (approved):
I can't comment on the algorithmic changes, only one high level suggestion.

test/unit/eigensolver/test_band_to_tridiag.cpp
rasolca: :face_with_peeking_eye:

include/dlaf/eigensolver/band_to_tridiag/mc.h
rasolca: I don't like much the idea to add enums for this kind of cases (lambda in a specific implemetation). Moreover this is a workaround to make the HH copy part working as it is. Anyway it still need to be improved.

test/unit/eigensolver/test_band_to_tridiag.cpp
rasolca: No, for local matrices the behaviour is similar as if the tuning value equal to the matrix size.

include/dlaf/eigensolver/band_to_tridiag/mc.h
msimberg: Ok, fair enough, I'm not insisting. For a more permanent solution I think it would aid readability with an enum (since the choice of `bool bottom` as the function parameters seems arbitrary and might as well be `bool top`), even if it's an enum only for use for this algorithm.

rasolca:
cscs-ci try

rasolca:
cscs-ci run

PR #797, opened 2023-02-13
ReductionToBand parallelise panel computation with bulk (Local, both MC and GPU)

albestro:
This PR increases parallelism with `bulk` for the computation of the panel in the ReductionToBand algorithm, which happens on Backend::MC for both Backend:MC and Backend::GPU versions of it.

- add tuneable parameter for number of threads to use for the computation
- add custom c'tor for `common::internal::vector`

TODO
- [ ] Finalize tune parameter (e.g. name, logic, default value for different backends)
- [ ] Track with an issue the fact that W reduction is done trivially
- [ ] Report about the blocking barrier requirement


include/dlaf/eigensolver/get_red2band_panel_nworkers.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "dlaf/common/assert.h"
+#include "dlaf/tune.h"
+
+namespace dlaf::eigensolver::internal {
+
+inline size_t getReductionToBandPanelNWorkers() noexcept {
+  const size_t nworkers = getTuneParameters().red2band_panel_nworkers;
+
+  DLAF_ASSERT(nworkers >= 1 && nworkers < pika::get_num_worker_threads(), nworkers);
```
albestro, line 22: @msimberg @rasolca not sure about the actual requirements. is it better to leave one for pika runtime so that other stuff can still be scheduled?

include/dlaf/tune.h
```diff
@@ -17,10 +17,13 @@ namespace dlaf {
 /// DLA-Future tuning parameters.
 ///
 /// Holds the value of the parameters that can be used to tune DLA-Future.
+/// - red2band_panel_nworkers: number of threads to use for computing the panel in the reduction to band algorithm.
 /// - eigensolver_min_band: The minimun value to start looking for a divisor of the block size.
 ///                         Set with --dlaf:eigensolver-min-band or env variable DLAF_EIGENSOLVER_MIN_BAND.
 /// Note to developers: Users can change these values, therefore consistency has to be ensured by algorithms.
 struct TuneParameters {
+  size_t red2band_panel_nworkers = std::max<size_t>(1, pika::get_num_worker_threads() / 2);
```
albestro, line 25: @rasolca I think we might want to have different defaults for `Backend::MC` and `Backend::GPU` (e.g. half resources on MC and "all" CPU resources on GPU)

include/dlaf/eigensolver/get_red2band_panel_nworkers.h
msimberg: Algorithmically I don't know, but with the yielding pika barrier I don't see a reason to leave threads unused. However, I guess this should actually be `pika::resource::get_pool("default").get_os_thread_count()`, the MPI pool thread is included in `pika::get_num_worker_threads()`?

albestro: > Algorithmically I don't know, but with the yielding pika barrier I don't see a reason to leave threads unused. 

Ok.

> I guess this should actually be pika::resource::get_pool("default").get_os_thread_count()

It's a good point. Both here and in the default value setup we want to know the amount of available threads, not considering MPI dedicated thread.

> the MPI pool thread is included in pika::get_num_worker_threads()?

I don't know, I have to check.

msimberg: > I don't know, I have to check.

Sorry @albestro, the parentheses were not a question but a statement. It does include the MPI pool (it's essentially the same as what you pass to `--pika:threads`, or what's inferred for it internally, which includes all pools).

include/dlaf/tune.h
```diff
@@ -17,10 +17,13 @@ namespace dlaf {
 /// DLA-Future tuning parameters.
 ///
 /// Holds the value of the parameters that can be used to tune DLA-Future.
+/// - red2band_panel_nworkers: number of threads to use for computing the panel in the reduction to band algorithm.
```
albestro, line 20: note-to-self: enable env variable input in `init.cpp`

include/dlaf/tune.h
```diff
@@ -17,10 +17,13 @@ namespace dlaf {
...
 /// - eigensolver_min_band: The minimun value to start looking for a divisor of the block size.
 ///                         Set with --dlaf:eigensolver-min-band or env variable DLAF_EIGENSOLVER_MIN_BAND.
 /// Note to developers: Users can change these values, therefore consistency has to be ensured by algorithms.
 struct TuneParameters {
+  size_t red2band_panel_nworkers = std::max<size_t>(1, pika::get_num_worker_threads() / 2);
+
   SizeType eigensolver_min_band = 100;
   SizeType band_to_tridiag_1d_block_size_base = 8192;
```
albestro, line 28: note-to-self: add documentation for this parameter too

albestro: @rasolca is going to take care of this in another PR

include/dlaf/eigensolver/get_red2band_panel_nworkers.h
msimberg: > is it better to leave one for pika runtime so that other stuff can still be scheduled?

I'm coming back to this again, and indeed it might be a good idea to do `max(num_threads - 1, 1)`. Whether it's needed for sure will probably depend on the algorithm and need profiling, but it's most likely quite a safe choice to leave one thread out.

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -36,15 +39,28 @@
...
 
 #include "dlaf/eigensolver/reduction_to_band/api.h"
 #include "dlaf/factorization/qr.h"
 
 namespace dlaf::eigensolver::internal {
 
+// Given a vector of vectors, reduce all vectors in the first one using sum operation
+template <class T>
+void reduceColumnVectors(std::vector<common::internal::vector<T>>& battery) {
+  for (size_t i = 1; i < battery.size(); ++i) {
```
msimberg, line 57: Here and everywhere else in this PR:
```suggestion
  for (std::size_t i = 1; i < battery.size(); ++i) {
```
Also, include `cstddef`.

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -287,10 +292,54 @@ auto computePanelReflectors(MatrixLike& mat_a, const matrix::SubPanelView& panel
     panel_tiles.emplace_back(matrix::splitTile(mat_a(i), spec));
   }
 
-  return ex::when_all_vector(std::move(panel_tiles)) |
-         dlaf::internal::transform(dlaf::internal::Policy<Backend::MC>(
-                                       pika::execution::thread_priority::high),
-                                   std::move(panel_task)) |
+  const size_t nthreads = getReductionToBandPanelNWorkers();
+  return ex::when_all(ex::just(std::make_shared<pika::barrier<>>(nthreads)),
+                      ex::just(std::vector<common::internal::vector<T>>{}),  // w (interally required)
```
msimberg, line 297:
```suggestion
                      ex::just(std::vector<common::internal::vector<T>>{}),  // w (internally required)
```

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -287,10 +292,54 @@ auto computePanelReflectors(MatrixLike& mat_a, const matrix::SubPanelView& panel
...
+                      ex::just(common::internal::vector<T>{}),               // taus
+                      ex::when_all_vector(std::move(panel_tiles))) |
+         ex::let_value([nthreads, nrefls, cols = panel_view.cols()](auto& barrier_ptr, auto& w,
```
msimberg, line 300: Just for scoping `cols = panel_view.cols()` can move to the second lambda, right?

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -287,10 +292,54 @@ auto computePanelReflectors(MatrixLike& mat_a, const matrix::SubPanelView& panel
...
+                                                                    auto& taus, auto& tiles) {
+           return ex::schedule(dlaf::internal::getBackendScheduler<Backend::MC>(
+                      pika::execution::thread_priority::high)) |
+                  ex::bulk(nthreads,
+                           [=, &barrier_ptr, &taus, &w, &tiles](const size_t index) {
```
msimberg, line 305: This is a note for a potential simplification which can be done in a separate PR as well. I'm realizing now that the outer `let_value` may not be necessary because `bulk` has a similar capability.

Instead of
```
when_all(...) |
let_value([](...) {
    return schedule() | bulk([](...) { ... });
});
```
I think this can be written as
```
when_all(...) |
transfer() |
bulk([](int i, t1& x1, t2& x2, ...) { ... }) |
then(drop_whatever_you_dont_need_from_when_all); // bulk will send x1, x2 etc. by value
```
See https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2300r6.html#example-async-inclusive-scan for a similar example.

Changing this is not a blocker for this PR.

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -36,15 +39,28 @@
...
 #include "dlaf/util_matrix.h"
 
 #include "dlaf/eigensolver/reduction_to_band/api.h"
 #include "dlaf/factorization/qr.h"
 
 namespace dlaf::eigensolver::internal {
 
+// Given a vector of vectors, reduce all vectors in the first one using sum operation
+template <class T>
+void reduceColumnVectors(std::vector<common::internal::vector<T>>& battery) {
```
msimberg, line 56: What does `battery` mean in this context?

msimberg (changes requested):
Looks good to me generally, just a few minor comments.

include/dlaf/tune.h
rasolca: #803

include/dlaf/eigensolver/reduction_to_band/impl.h
msimberg: So how about just `columns`? Or `columnVectors`? Or `vectors`? Or more something more specific with a suffix of `columns` or `vectors` if it makes sense for the algorithm. The function is called `reduceColumnVectors` after all.

msimberg: Regarding using `Matrix`: is it a ragged vector of vectors, i.e. are the vectors different lengths

msimberg: I guess it's not a ragged vector if they're reduced to the first vector... This is side-tracking, but what keeps you from using a `Tile` or a `Matrix` here?

albestro: IIRC it was about when the memory was allocated.

With a `just(vector<vector<>>)` I can decide at runtime how much but more importantly to when to actually allocate, i.e. just at runtime and not at schedule-time.

Currently, with `Matrix`/`Tile` we can just pre-allocate at schedule-time, and it might consume a lot of memory even when just a single buffer can be re-used thanks to memory managers (and not manually, like with Panel).

I would just go for renaming, but in case @msimberg or @rasolca you have a better idea, I'm open to do it now or in a next PR 😉

Any preference for the name? `columnVectors`?

msimberg: I'd have to look closer to be able to comment about the `vector` vs `Tile` issue, but as you said that can be done separately.

I'd be happy with `columnVectors`. I think it describes what it is better than `battery`.

albestro:
cscs-ci run

albestro:
Currently, in `pika` we have a "yielding" barrier. This is not optimal since lightweight tasks of the bulk might get rescheduled and this add overheads.

It would be mitigated with a "blocking" barrier: in pika is not yet implemented, but a feasibility test has been done with a `pthread_barrier` (really dangerous, since it can lead to deadlocks, since it blocks the os-thread).

rasolca (approved):

msimberg (approved):

PR #798, opened 2023-02-14
computeTFactor parallelise computation with bulk (MC Local and Distributed)

albestro:
This PR improves computeTFactor computation by parallelising the computation that previously was serialised because of the reduction over a single tile.

The algorithm implementation has been split in two variants:
- `Backend::MC` now has the `bulk`-based implementation
- `Backend::GPU` still use the previous version of the code
This required a couple of changes to the interface structure of the algorithm, and I took the chance to fix also some problem with the documentation.

Note: Look at first three commits to see the main implementation change in a more "isolated" way from the rest.

TODO
- [x] Split Distributed MC/GPU implementations too
- [x] Make `nthreads` used for bulk tunable
- [x] Complete documentation fixes
- [x] Code cleanup

### DIFF
--- a/include/dlaf/factorization/qr/t_factor_impl.h
+++ b/include/dlaf/factorization/qr/t_factor_impl.h
@@ -98,7 +98,7 @@ struct Helpers<Backend::MC, Device::CPU, T> {
       blas::gemv(blas::Layout::ColMajor, blas::Op::ConjTrans, va_size.rows(), va_size.cols(), -tau,
                  tile_v.ptr(va_start), tile_v.ld(), tile_v.ptr(vb_start), 1, 1, tile_t.ptr(t_start), 1);
     }
-    return std::move(tile_t);
+    return tile_t;
   };
 
   template <class TSender>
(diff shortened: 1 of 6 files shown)


include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -234,20 +245,117 @@ struct Helpers<Backend::GPU, Device::GPU, T> {
...
+                   ex::when_all_vector(std::move(panel_tiles)), std::move(taus), std::move(t)) |
+      ex::let_value([=](auto& barrier_ptr, auto&& panel, const common::internal::vector<T>& taus,
+                        matrix::Tile<T, D>& t) {
+        matrix::Matrix<T, D> t_all({t.size().rows() * to_SizeType(nthreads - 1), t.size().cols()},
+                                   t.size());
+        return ex::when_all_vector(
+                   select(t_all, common::iterate_range2d(t_all.distribution().localNrTiles()))) |
+               ex::transfer(
+                   dlaf::internal::getBackendScheduler<B>(pika::execution::thread_priority::high)) |
+               ex::bulk(nthreads, [=, &barrier_ptr, &t, &taus,
```
albestro, line 296: @msimberg differently from the other PR where you suggested to use `bulk` functionality of returning all senders instead of `let_value`, in this case `let_value` body allows to allocate memory just when dependencies are ready and not at scheduling time.

Do you see any other way for improving this part?

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -59,74 +68,78 @@ struct Helpers<Backend::MC, Device::CPU, T> {
...
+  static auto trmv_func(matrix::Tile<T, Device::CPU>&& tile_t) {
+    for (SizeType j = 0; j < tile_t.size().cols(); ++j) {
+      const TileElementIndex t_start{0, j};
+      const TileElementSize t_size{j, 1};
+
+      blas::trmv(blas::Layout::ColMajor, blas::Uplo::Upper, blas::Op::NoTrans, blas::Diag::NonUnit,
+                 t_size.rows(), tile_t.ptr(), tile_t.ld(), tile_t.ptr(t_start), 1);
+    }
+    // TODO: Why return if the tile is unused?
+    return std::move(tile_t);
```
rasolca, line 133: Can the TODO be solved?

rasolca (approved):

include/dlaf/factorization/qr/t_factor_impl.h
albestro: As discussed offline, the code might be ok like this for the moment.

What @msimberg suggested is the right point: we need a different object for this use-case, with synchronous access. I would add that, from previous experiences in similar circumstances, such object should also provide a way to not allocate the memory at schedule time. Indeed, if we would switch this code right now to use `MatrixLocal`, we would easily end up allocating space at schedule-time.

This is something we should think about during the refactoring 😉

rasolca:
cscs-ci run

rasolca:
The eigensolver test is deadlocking 😱

msimberg (approved):

albestro:
cscs-ci run

checking if the workaorund works also in the CI.

In case, I will document more the case and use it to investigate deeper the problem.

PR #766, opened 2023-01-09
Use pika's `async_rw_mutex` to replace tile dependency mechanism

msimberg:
And remove the use of most futures/promises while doing so.

~Don't pay much attention to this yet. I'm opening this mainly to get broader CI feedback. I will expand this description once it's closer to being cleaned up. Requires pika main with changes not yet merged (https://github.com/pika-org/pika/pull/548 and https://github.com/pika-org/pika/pull/549).~

To do:
- [x] renamings (new matrix sender access function names are placeholders, they will replace the future functions before merging; fixes #340)
- [x] re-enable or remove disabled tests
- [x] check if all temporary "workarounds" are real workarounds or can be solved differently
- [x] benchmarks
- [x] change `async_rw_mutex.hpp` include (needs new pika release)

PR #802, opened 2023-02-17
ExtraBuffers: revamping of the idea

albestro:
This is a revamping of a concept we already tested in #386. Due to the important changes happened to the codebase in the meanwhile, together with the fact that it was just a feasibility test, we opted for the easier path of just re-implementing the idea.

ExtraBuffers is a utility that comes handy for the "reduce" pattern, i.e. multiple tasks wants to write on the same memory. ExtraBuffers gives you a set of "buffer" tiles that tasks can use to write in parallel, and at the end the reduction is done on request.

In this PR there is also a first usage example, in W2 computation inside ReductionToBand.


include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -455,20 +456,24 @@ void gemmComputeW2(matrix::Matrix<T, D>& w2, matrix::Panel<Coord::Col, const T,
...
+  for (const auto& index_tile : w.iteratorLocal()) {
     ex::start_detached(dlaf::internal::whenAllLift(Op::ConjTrans, Op::NoTrans, T(1),
                                                    w.read_sender(index_tile), x.read_sender(index_tile),
-                                                   T(1), w2.readwrite_sender(LocalTileIndex(0, 0))) |
+                                                   T(1), buffers.readwrite_sender(index_tile.row())) |
                        tile::gemm(dlaf::internal::Policy<B>(thread_priority::high)));
+  }
+
+  ex::start_detached(tile::set0(dlaf::internal::Policy<B>(), w2.readwrite_sender(LocalTileIndex(0, 0))));
+  ex::start_detached(buffers.reduce(w2.readwrite_sender(LocalTileIndex(0, 0))));
```
albestro, line 476: Probably better to create a continuation with a single task?

include/dlaf/matrix/extra_buffers.h
```diff
@@ -0,0 +1,79 @@
...
+
+  auto readwrite_sender(SizeType index) {
+    return Matrix<T, D>::readwrite_sender(internalIndex(index));
+  }
+
+  template <class TileSender>
+  [[nodiscard]] auto reduce(TileSender tile) {
+    namespace ex = pika::execution::experimental;
+
+    std::vector<ex::any_sender<pika::shared_future<matrix::Tile<const T, D>>>> buffers;
```
albestro, line 44: The type is so cumbersome, probably @msimberg is aware of this and is going to say that it is something we will manage when we will make the tile sender mechanism pika::future free.

I don't know if there is any workaround available, but AFAIK currently our `Unwrapping` facilities does not unwrap `vector`, so we just get it sent as it is even using our `dlaf::internal::transform`.

include/dlaf/matrix/extra_buffers.h
```diff
@@ -0,0 +1,79 @@
...
+                                         else if constexpr (D == Device::GPU) {
+                                           dlaf::tile::internal::add(T(1), buffer.get(), tile, ts...);
+                                         }
+#endif
+                                         else {
+                                           DLAF_STATIC_UNIMPLEMENTED(T);
+                                         }
+                                       }
+                                     });
+  }
```
albestro, line 70: This is really not nice...15 lines for saying 

```
for (const auto& buffer : buffers)
  dlaf::tile::internal::add(T(1), buffer.get(), <optional whip::stream_t>);
```

I know we can go with something like

```
dlaf::tile::internal::add(T(1), buffer.get(), ts...);
```

but we will loose a bit of static checking...

do we have any better alternative?

include/dlaf/matrix/extra_buffers.h
msimberg: I see your complaint, but `typedef` exists :wink: I see that complex eventually being a helper typedef inside e.g. `Matrix`, or at least in the `dlaf::matrix` namespace and then you'll be able to write `std::vector<ReadOnlyTileSender>` or something like that. Would that partially or completely take care of your concerns here? This is what I'm planning on doing in #766 (because currently it's also dealing with long unwieldy types).

PR #813, opened 2023-03-17
Permutations: use ReadOnly access when possible

albestro:
Use readonly access instead of the full readwrite one, when it is not strictly required. The former, by not requiring exclusive access, might allow more tasks working in parallel.

With the current algorithm it might not change much, but with the algorithmic upgrades we are developing/testing, these additional superfluous constraints might represent an hindering factor for parallelisation (and, consequentially, for performance).

### DIFF
--- a/include/dlaf/permutations/general/impl.h
+++ b/include/dlaf/permutations/general/impl.h
@@ -120,8 +120,8 @@ void applyPermutations(GlobalElementIndex out_begin, GlobalElementSize sz, SizeT
   }
   else if constexpr (D == Device::GPU) {
 #if defined(DLAF_WITH_GPU)
-    applyPermutationsOnDevice<T, coord>(out_begin, sz, in_offset, distr, perm_arr, in_tiles, out_tiles,
-                                        args...);
+    applyPermutationsOnDevice<T, coord>(out_begin, sz, in_offset, distr, perm_arr, in_tiles_fut,
+                                        out_tiles, args...);
 #endif
   }
 }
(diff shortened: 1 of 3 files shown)


include/dlaf/permutations/general/impl.h
```diff
@@ -31,6 +31,7 @@
 #include "dlaf/util_matrix.h"
 
 #include <mpi.h>
+#include <pika/futures/future.hpp>
```
msimberg, line 34:
```suggestion
#include <pika/future.hpp>
```

msimberg (approved):
Looks good to me, just one include needs to be changed.

include/dlaf/permutations/general/impl.h
```diff
@@ -231,7 +232,7 @@ void all2allData(common::Pipeline<comm::Communicator>& sub_task_chain, int nrank
         std::transform(recv_counts.cbegin(), recv_counts.cend(), recv_counts.begin(), mul_const);
 
         // Note: that was guaranteed to be contiguous on allocation
-        T* send_ptr = send_tiles[0].ptr();
+        const T* send_ptr = send_tiles_fut[0].get().ptr();
```
albestro, line 235: @msimberg @rasolca just highlighting this, where I get a raw pointer from a tile in a shared_future.

AFAICT this should not be a problem since the shared_future is kept alive by `transfroMPI`.

Do you see any issue?

msimberg: This looks ok to me.

msimberg (approved):

rasolca (approved):

rasolca:
cscs-ci run

albestro:
cscs-ci stop

rasolca:
cscs-ci run

PR #814, opened 2023-03-22
Add CI configuration with GCC and stdexec

aurianer:
I think it would be a good idea to rename the pika variant from p2300 to stdexec, if @msimberg agrees, I'll update that

- [x] Add build CI with stdexec

### DIFF
--- a/ci/docker/release-cpu-stdexec.yaml
+++ b/ci/docker/release-cpu-stdexec.yaml
@@ -1,0 +1,26 @@
+#
+# Distributed Linear Algebra with Future (DLAF)
+#
+# Copyright (c) 2018-2023, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+spack:
+  include:
+  - common.yaml
+
+  view: false
+  concretizer:
+    unify:
+      true
+
+  specs:
+    - dla-future@develop +miniapps +ci-test +ci-check-threads ^intel-mkl threads=openmp ^mpich ^pika+stdexec ^stdexec@git.7a47a4aa411c1ca9adfcb152c28cc3dd7b156b4d=main
+
+  packages:
+    all:
+      variants:
+        - 'build_type=Release'
--- a/include/dlaf/sender/traits.h
+++ b/include/dlaf/sender/traits.h
@@ -61,7 +61,7 @@ struct SenderSingleValueTypeImpl<
 };
 
 // The type sent by Sender, if Sender sends exactly one type.
-#if defined(PIKA_HAVE_P2300_REFERENCE_IMPLEMENTATION)
+#if defined(PIKA_HAVE_STDEXEC)
 template <typename Sender>
 using SenderSingleValueType =
     typename SenderSingleValueTypeImpl<pika::execution::experimental::value_types_of_t<
--- a/include/dlaf/sender/traits.h
+++ b/include/dlaf/sender/traits.h
@@ -9,6 +9,9 @@
 //
 #pragma once
 
+#include <functional>
+#include <type_traits>
+
 #include <pika/async_rw_mutex.hpp>
 #include <pika/execution.hpp>
 #include <pika/future.hpp>
@@ -27,7 +30,7 @@ struct SenderSingleValueTypeImpl<TypeList<TypeList<T>>> {
   using type = T;
 };
 
-struct empty_env {};
+struct EmptyEnv {};
 
 // We are only interested in the types wrapped by future and shared_future since
 // we will internally unwrap them.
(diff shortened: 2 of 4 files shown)


msimberg:
cscs-ci run

ci/docker/release-cpu-stdexec.yaml
```diff
@@ -0,0 +1,26 @@
...
+  include:
+  - common.yaml
+
+  view: false
+  concretizer:
+    unify:
+      true
+
+  specs:
+    - dla-future@develop +miniapps +ci-test +ci-check-threads ^intel-mkl threads=openmp ^mpich ^pika+stdexec@git.7a47a4aa411c1ca9adfcb152c28cc3dd7b156b4d=main
```
msimberg, line 21:
```suggestion
    - dla-future@develop +miniapps +ci-test +ci-check-threads ^intel-mkl threads=openmp ^mpich ^pika+stdexec ^stdexec@git.7a47a4aa411c1ca9adfcb152c28cc3dd7b156b4d=main
```

include/dlaf/sender/traits.h
```diff
@@ -9,12 +9,17 @@
...
+#include <type_traits>
+
 #include <pika/async_rw_mutex.hpp>
 #include <pika/execution.hpp>
 #include <pika/future.hpp>
 
 #include <dlaf/types.h>
 
+#include "dlaf/types.h"
+
```
msimberg, line 22:
```suggestion
```

include/dlaf/sender/traits.h
```diff
@@ -27,6 +32,54 @@ struct SenderSingleValueTypeImpl<TypeList<TypeList<T>>> {
...
 
+// Utilities to use unique_t
+template <typename T>
+struct is_false : std::integral_constant<bool, !(bool) T::value> {};
+
+template <typename T>
+inline constexpr bool is_false_v = is_false<T>::value;
+
+template <typename... Ts>
+struct always_false : std::false_type {};
```
msimberg, line 43:
```suggestion
// Utilities to use unique_t
template <typename T>
struct IsFalse : std::integral_constant<bool, !(bool) T::value> {};

template <typename T>
inline constexpr bool IsFalse_v = IsFalse<T>::value;

template <typename... Ts>
struct AlwaysFalse : std::false_type {};
```
etc.

include/dlaf/sender/traits.h
```diff
@@ -60,12 +113,15 @@ struct SenderSingleValueTypeImpl<
   using type = RType;
 };
 
+template <typename... Ts>
+using decayed_typelist = TypeList<std::decay_t<Ts>...>;
+
```
msimberg, line 118:
```suggestion
template <typename... Ts>
using DecayedTypeList = TypeList<std::decay_t<Ts>...>;

```

msimberg (changes requested):
Thanks @aurianer for all the work on this!

I'm 99% sure some tests will fail (or more likely hang). If that's the case, let's make this PR only add the build stage with `+stdexec` so that we can get that tested on master as soon as possible, and separately fix the tests. If everything passes, let's do build and test here!

include/dlaf/sender/traits.h
```diff
@@ -60,12 +111,15 @@ struct SenderSingleValueTypeImpl<
   using type = RType;
 };
 
+template <typename... Ts>
+using DecayedTypeList = TypeList<std::decay_t<Ts>...>;
```
msimberg, line 115: Maybe you can group this together with `TypeList`?

include/dlaf/sender/traits.h
```diff
@@ -27,7 +30,55 @@ struct SenderSingleValueTypeImpl<TypeList<TypeList<T>>> {
   using type = T;
 };
 
-struct empty_env {};
+// Utilities to use unique_t
```
msimberg, line 33: Leave out or update to `UniqueType`? The utilities may be useful on their own at some point so my preference would be to leave it out.

msimberg (approved):
This looks good to me, thanks @aurianer! Just a couple of minor optional cosmetic comments.

include/dlaf/sender/traits.h
```diff
@@ -19,6 +22,58 @@ namespace dlaf::internal {
...
+struct Unique;
+
+template <template <typename...> class Pack, typename... Ts>
+struct Unique<Pack<Ts...>> : UniqueHelper<Pack<>, Pack<Ts...>> {};
+
+/// Remove duplicate types in the given pack.
+template <typename Pack>
+using UniqueType = typename Unique<Pack>::type;
+
+struct EmptyEnv {};
```
rasolca, line 75: Just for curiosity:
Why are these struct needed for stdexec?

rasolca (approved):
LGTM

include/dlaf/sender/traits.h
msimberg: It's technically also needed when stdexec is disabled in pika _in the general case_. However, practically we don't need the decaying in the use cases that currently exist in DLA-Future.

The reason is that some stdexec sender adaptors send completion signatures with `T&&` where pika would send `T` and we need to make sure we treat the two the same (as `T`).

PR #817, opened 2023-03-23
tridiagonal eigensolver -> improve perf on AMD gpus

rasolca:
- the leaf solver is moved on CPU
- the Cuppen decomposition is done on CPU as well
-> the tridiagonal matrix can now be kept on CPU and unnecessary copies can be avoided

This solves the huge performance bottleneck on AMD GPUs. For P100 there is a marginal improvement, while for A100 the change is almost irrelevant.

See https://confluence.cscs.ch/display/SCISWDEV/2023-03-14+Optimisation+Round+2


msimberg (approved):
No comments on the algorithmic changes, looks good to me otherwise.

PR #818, opened 2023-03-27
Pre-commit git hook helper script

biddisco:


biddisco:
I have modified the primary .github/format.sh script to allow git precommit use. 
Instead of reformatting the user's files, when used as a hook, the script builds a list of files that are incorrectly formatted and outputs the list along with a command line that can be invoked to fix the files. (Same as pika pre-commit hook).

PR #820, opened 2023-03-28
Add eigensolver for pre-allocated eigenvalues and eigenvector matrices

RMeli:
Only tested on CPU (locally).

### DIFF
--- a/include/dlaf/eigensolver/eigensolver.h
+++ b/include/dlaf/eigensolver/eigensolver.h
@@ -123,7 +123,6 @@ void eigensolver(comm::CommunicatorGrid grid, blas::Uplo uplo, Matrix<T, D>& mat
 template <Backend B, Device D, class T>
 EigensolverResult<T, D> eigensolver(comm::CommunicatorGrid grid, blas::Uplo uplo, Matrix<T, D>& mat) {
   const SizeType size = mat.size().rows();
-
   matrix::Matrix<BaseType<T>, D> eigenvalues(LocalElementSize(size, 1),
                                              TileElementSize(mat.blockSize().rows(), 1));
   matrix::Matrix<T, D> eigenvectors(GlobalElementSize(size, size), mat.blockSize(), grid);
--- a/include/dlaf/eigensolver/eigensolver.h
+++ b/include/dlaf/eigensolver/eigensolver.h
@@ -28,6 +28,29 @@ namespace dlaf::eigensolver {
 ///
 /// Implementation on local memory.
 ///
+/// @param uplo specifies if upper or lower triangular part of @p mat will be referenced
+/// @param mat contains the Hermitian matrix A
+/// @param eigenvectors matrix of eigenvectors
+/// @param vector of eigenvalues
+template <Backend B, Device D, class T>
+void eigensolver(blas::Uplo uplo, Matrix<T, D>& mat, Matrix<T, D>& eigenvectors,
+                 Matrix<BaseType<T>, D>& eigenvalues) {
+  DLAF_ASSERT(matrix::local_matrix(mat), mat);
+  DLAF_ASSERT(square_size(mat), mat);
+  DLAF_ASSERT(square_blocksize(mat), mat);
+
+  internal::Eigensolver<B, D, T>::call(uplo, mat, eigenvectors, eigenvalues);
+}
+
+/// Standard Eigensolver.
+///
+/// It solves the standard eigenvalue problem A * x = lambda * x.
+///
+/// On exit, the lower triangle or the upper triangle (depending on @p uplo) of @p mat_a,
+/// including the diagonal, is destroyed.
+///
+/// Implementation on local memory.
+///
 /// @return struct ReturnEigensolverType with eigenvalues, as a vector<T>, and eigenvectors as a Matrix
 /// @param uplo specifies if upper or lower triangular part of @p mat will be referenced
 /// @param mat contains the Hermitian matrix A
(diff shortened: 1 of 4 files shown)


test/unit/eigensolver/test_eigensolver.cpp
```diff
@@ -114,7 +115,7 @@ void testEigensolverCorrectness(const blas::Uplo uplo, Matrix<const T, Device::C
   CHECK_MATRIX_NEAR(res, workspace, 2 * m * TypeUtilities<T>::error, 2 * m * TypeUtilities<T>::error);
 }
 
-template <class T, Backend B, Device D>
+template <class T, Backend B, Device D, bool preallocated = false>
```
msimberg, line 118: Since this is a test I would just remove the default and always explicitly say whether or not the result should be preallocated. Possibly with tag types/enums so that one doesn't have to wonder at the callsite if the boolean is `preallocated` or `allocated` (or whatever one would call the opposite).

test/unit/eigensolver/test_eigensolver.cpp
```diff
@@ -128,18 +129,37 @@ void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb)
...
 
-  eigensolver::EigensolverResult<T, D> ret = [&]() {
-    MatrixMirror<T, D, Device::CPU> mat_a(mat_a_h);
-    return eigensolver::eigensolver<B>(uplo, mat_a.get());
-  }();
+  if constexpr (!preallocated) {
+    eigensolver::EigensolverResult<T, D> ret = [&]() {
+      MatrixMirror<T, D, Device::CPU> mat_a(mat_a_h);
+      return eigensolver::eigensolver<B>(uplo, mat_a.get());
+    }();
```
rasolca, line 136: You can avoid some code duplication if you move the constexpr branch inside the lambda
```suggestion
  eigensolver::EigensolverResult<T, D> ret = [&]() {
    MatrixMirror<T, D, Device::CPU> mat_a(mat_a_h);
    if constexpr (preallocated) {
      const SizeType size = mat_a_h.size().rows();
      Matrix<BaseType<T>, D> eigenvalues(LocalElementSize(size, 1),
                                     TileElementSize(mat_a_h.blockSize().rows(), 1));
      Matrix<T, D> eigenvectors(LocalElementSize(size, size), mat_a_h.blockSize());
      eigensolver::eigensolver<B>(uplo, mat_a.get(), eigenvectors, eigenvalues);
      return eigensolver::EigensolverResult<T, D>(std::move(eigenvalues), std::move(eigenvectors));
    else {
      return eigensolver::eigensolver<B>(uplo, mat_a.get());
    }
  }();
```

include/dlaf/eigensolver/eigensolver.h
```diff
@@ -40,6 +63,30 @@ EigensolverResult<T, D> eigensolver(blas::Uplo uplo, Matrix<T, D>& mat) {
...
+/// @param mat contains the Hermitian matrix A
+/// @param eigenvectors matrix of eigenvectors
+/// @param eigenvalues vector of eigenvalues
+template <Backend B, Device D, class T>
+void eigensolver(comm::CommunicatorGrid grid, blas::Uplo uplo, Matrix<T, D>& mat,
+                 Matrix<T, D>& eigenvectors, Matrix<BaseType<T>, D>& eigenvalues) {
+  DLAF_ASSERT(matrix::equal_process_grid(mat, grid), mat);
+  DLAF_ASSERT(square_size(mat), mat);
+  DLAF_ASSERT(square_blocksize(mat), mat);
+
```
msimberg, line 86: Could this assert that `eigenvectors` and `eigenvalues` are appropriate sizes? Are there any requirements for the layout and/or distribution?

include/dlaf/eigensolver/eigensolver.h
```diff
@@ -19,6 +19,29 @@
...
+///
+/// Implementation on local memory.
+///
+/// @param uplo specifies if upper or lower triangular part of @p mat will be referenced
+/// @param mat contains the Hermitian matrix A
+/// @param eigenvectors matrix of eigenvectors
+/// @param eigenvalues vector of eigenvalues
+template <Backend B, Device D, class T>
+void eigensolver(blas::Uplo uplo, Matrix<T, D>& mat, Matrix<T, D>& eigenvectors,
+                 Matrix<BaseType<T>, D>& eigenvalues) {
```
msimberg, line 37: Is this more consistent with the other overload? I'm not quite sure myself if this is actually better...
```suggestion
void eigensolver(blas::Uplo uplo, Matrix<T, D>& mat, EigensolverResult<T, D>& result) {
```
If anything the `eigenvalues` and `eigenvectors` should possibly be passed in the same order as they're declared in `EigensolverResult`?

test/unit/eigensolver/test_eigensolver.cpp
```diff
@@ -128,18 +129,37 @@ void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb)
...
+
+    if (mat_a_h.size().isEmpty())
+      return;
+
+    testEigensolverCorrectness(uplo, reference, eigenvectors, eigenvalues);
+  }
 }
 
-template <class T, Backend B, Device D>
+template <class T, Backend B, Device D, bool preallocated = false>
```
rasolca, line 162: In this case I would avoid the default template value.

include/dlaf/eigensolver/eigensolver/impl.h
```diff
@@ -73,14 +75,22 @@ EigensolverResult<T, D> Eigensolver<B, D, T>::call(comm::CommunicatorGrid grid,
...
+}
+
+template <Backend B, Device D, class T>
+EigensolverResult<T, D> Eigensolver<B, D, T>::call(comm::CommunicatorGrid grid, blas::Uplo uplo,
+                                                   Matrix<T, D>& mat_a) {
+  const SizeType size = mat_a.size().rows();
+
+  matrix::Matrix<BaseType<T>, D> evals(LocalElementSize(size, 1),
+                                       TileElementSize(mat_a.blockSize().rows(), 1));
+  matrix::Matrix<T, D> mat_e(GlobalElementSize(size, size), mat_a.blockSize(), grid);
```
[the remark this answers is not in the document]
msimberg: Yeah, that seems reasonable. That should get rid of two overloads?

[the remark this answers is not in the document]
rasolca: exactly, it gets rid of the two `call` overloads without preallocated matrices.

test/unit/eigensolver/test_eigensolver.cpp
RMeli: I used `enum class` (much cleaner, thanks!) but then I wanted to check the `else` statement (since it's no longer a boolean). I added a `static_assert` in the `else` branch, but it's not clean at all... Should I just drop it, since the `enum class` should only contain valid values?

msimberg: Yep, this looks reasonable enough to me. It's quite unlikely that enum is going to be extended anyway (but you never know). No need to change anything from my point of view, but the `Allocation` template parameter need not be a template parameter either, and then you could just do a plain old `switch-case`. But as I said, I'm happy with the current version as well.

msimberg (approved):
Nice @RMeli!

rasolca (approved):

albestro (approved):
LGTM too

rasolca:
cscs-ci run

PR #821, opened 2023-03-29
Add generalised eigensolver for pre-allocated eigenvalues and eigenvector matrices

### LINKED ISSUE #804: Eigensolver API
As discussed with @msimberg, in order to avoid unnecessary allocations and copies it might be useful to have an additional `eigensolver` which takes pre-allocated `eigenvalues` and `eigenvectors` matrices as input.

RMeli:
Follow-up to #820. Only tested on CPU (locally).

Fix #804


msimberg (approved):

albestro (approved):
LGTM.

Just a bit of code duplication in the test...but it's not a big deal.

Thanks!

rasolca (commented):
I'm thinking if the same approach used in the standard case would be better.
(i.e. (a) implement the case with the inputs, (b) allocate, call (a) and return, for the returning case.)
It will avoid some code duplication.

albestro:
About the local vs distributed, recently we opted for merging them with something like this

https://github.com/eth-cscs/DLA-Future/blob/f5aa38428bfee864f79b49d4bf34c4e1b644a4e4/test/include/dlaf_test/eigensolver/test_eigensolver_correctness.h#L30-L33

RMeli:
@rasolca your comments should be addressed.

@albestro I collapsed `testGenEigensolver` into one. If this is better (and similar to what you had in mind), I can apply the same to the `testEigensolver` to have better consistency.

rasolca (approved):

rasolca:
cscs-ci run

RMeli:
cscs-ci run

PR #827, opened 2023-03-31
Fix small typos in `test_matrix`

msimberg:
The test was checking the wrong vector of futures.

Internally `checkFuturesStep` was expecting the futures to return `false` from `is_ready`. The futures in `fut3` are in principle ready, but since they have already been checked, the futures are empty, and when a future is empty (no shared state) `is_ready` still returns `false` so the test was passing anyway.


rasolca (approved):

rasolca:
cscs-ci run

PR #834, opened 2023-04-14
Keep `-Werror` flags in CI

msimberg:
~I'm speculating, but I think warnings stopped being errors on CI some time ago. I suspect it's this PR that introduced it: https://github.com/spack/spack/pull/30882. This PR attempts to enable warnings as errors again in CI.~

~I have not tried to fix warnings yet in this PR, I first want to see that this works.~

This does seem to reenable warnings as errors in CI again. I'll work on fixing the warnings.

### DIFF
--- a/ci/docker/codecov.Dockerfile
+++ b/ci/docker/codecov.Dockerfile
@@ -30,7 +30,7 @@ RUN spack repo rm --scope site dlaf && \
     spack -e ci concretize -f && \
     mkdir ${BUILD} && \
     ln -s ${BUILD} `spack -e ci location -b dla-future` && \
-    spack -e ci install --jobs ${NUM_PROCS} --keep-stage --verbose
+    spack -e ci --config "config:flags:keep_werror:all" install --jobs ${NUM_PROCS} --keep-stage --verbose
 
 # Prune and bundle binaries
 RUN mkdir ${BUILD}-tmp && cd ${BUILD} && \
--- a/test/unit/communication/test_communicator_grid.cpp
+++ b/test/unit/communication/test_communicator_grid.cpp
@@ -29,7 +29,7 @@ void test_grid_communication(CommunicatorGrid& grid) {
   if (MPI_COMM_NULL == grid.rowCommunicator() || MPI_COMM_NULL == grid.colCommunicator())
     return;
 
-  const int buffer_send = 1;
+  int buffer_send = 1;
   int buffer, buffer_recv;
 
   // All Communication
--- a/test/unit/communication/test_transform_mpi.cpp
+++ b/test/unit/communication/test_transform_mpi.cpp
@@ -98,7 +98,7 @@ TEST_F(TransformMPITest, PromiseGuardManagement) {
     // Note:
     // At this point we checked if everything went as expected in the posting phase. Let's signal
     // Rank 1 that it can send back the message that will unlock IRecv.
-    const int signal = 13;
+    int signal = 13;
     MPI_Send(&signal, 1, MPI_INT, 1, 0, world);
 
     // Note:
--- a/ci/docker/deploy.Dockerfile
+++ b/ci/docker/deploy.Dockerfile
@@ -27,7 +27,7 @@ RUN spack repo rm --scope site dlaf && \
     spack -e ci concretize -f && \
     mkdir ${BUILD} && \
     ln -s ${BUILD} `spack -e ci location -b dla-future` && \
-    spack -e ci install --jobs ${NUM_PROCS} --keep-stage --verbose
+    spack -e ci --config "config:flags:keep_werror:all" install --jobs ${NUM_PROCS} --keep-stage --verbose
 
 # Test deployment with miniapps as independent project
 RUN pushd ${SOURCE}/miniapp && \
--- a/include/dlaf/lapack/tile.h
+++ b/include/dlaf/lapack/tile.h
@@ -353,7 +353,8 @@ void hegst(const int itype, const blas::Uplo uplo, const Tile<T, Device::CPU>& a
   DLAF_ASSERT(itype >= 1 && itype <= 3, itype);
 
   common::internal::SingleThreadedBlasScope single;
-  auto info = lapack::hegst(itype, uplo, a.size().cols(), a.ptr(), a.ld(), b.ptr(), b.ld());
+  [[maybe_unused]] auto info =
+      lapack::hegst(itype, uplo, a.size().cols(), a.ptr(), a.ld(), b.ptr(), b.ld());
 
   DLAF_ASSERT(info == 0, info);
 }
@@ -371,7 +372,7 @@ long long potrfInfo(const blas::Uplo uplo, const Tile<T, Device::CPU>& a) {
 
 template <class T>
 void potrf(const blas::Uplo uplo, const Tile<T, Device::CPU>& a) noexcept {
-  auto info = potrfInfo(uplo, a);
+  [[maybe_unused]] auto info = potrfInfo(uplo, a);
 
   DLAF_ASSERT(info == 0, info);
 }
--- a/include/dlaf/multiplication/general.h
+++ b/include/dlaf/multiplication/general.h
@@ -63,7 +63,7 @@ void generalSubMatrix(const SizeType a, const SizeType b, const blas::Op opA, co
   DLAF_ASSERT(equal_size(mat_a, mat_b), mat_a, mat_b);
   DLAF_ASSERT(equal_size(mat_a, mat_c), mat_a, mat_c);
 
-  const SizeType m = mat_a.nrTiles().rows();
+  [[maybe_unused]] const SizeType m = mat_a.nrTiles().rows();
   DLAF_ASSERT(a <= b, a, b);
   DLAF_ASSERT(a >= 0 && a < m, a, m);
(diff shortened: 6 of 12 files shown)


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
This seems to have silenced the remaining errors. I've simply `const_cast`ed away some of the `-Wtype-safety` errors, the same as we'd done in other MPI calls (presumably while we were still getting warnings as errors).

An open question is if we want to add a CI build in release mode (or more specifically, with all assertions disabled)? The unused variable warnings that I've fixed here are a result of disabling assertions but those warnings do not currently show up in CI because we don't disable assertions in any configuration.

msimberg:
cscs-ci run

test/unit/communication/test_transform_mpi.cpp
```diff
@@ -99,7 +99,7 @@ TEST_F(TransformMPITest, PromiseGuardManagement) {
     // At this point we checked if everything went as expected in the posting phase. Let's signal
     // Rank 1 that it can send back the message that will unlock IRecv.
     const int signal = 13;
-    MPI_Send(&signal, 1, MPI_INT, 1, 0, world);
+    MPI_Send(const_cast<int*>(&signal), 1, MPI_INT, 1, 0, world);
```
rasolca, line 102: As `signal` is used only here:
```suggestion
    int signal = 13;
    MPI_Send(&signal, 1, MPI_INT, 1, 0, world);
```

msimberg: Updated.

msimberg:
@albestro we discussed today adding a CI configuration that would completely disable assertions to catch the otherwise unused variables. Currently adding special CMake variables would require a special variant per CMake flag. An option that came up is having something like https://github.com/spack/spack/pull/28519. That PR seems to have been mostly forgotten, but do you think it would be a good idea to revive it? Alternatively, we could simply add a `cmakeflags` variant to DLA-Future even before anything is done with the spack PR. It would also allow removing the special `ci-test` variant and replacing it with `cmakeflags`. Do you have thoughts about this? Can you think of other ways to do it?

msimberg:
cscs-ci run

msimberg:
This seems to have failed due to something completely unrelated to our tests or this PR: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/4168406537#L283.

albestro:
> @albestro we discussed today adding a CI configuration that would completely disable assertions to catch the otherwise unused variables. Currently adding special CMake variables would require a special variant per CMake flag. An option that came up is having something like [spack/spack#28519](https://github.com/spack/spack/pull/28519). That PR seems to have been mostly forgotten, but do you think it would be a good idea to revive it?

I can check if it is easy to revive it and do some progress there 😉 

> Alternatively, we could simply add a `cmakeflags` variant to DLA-Future even before anything is done with the spack PR. It would also allow removing the special `ci-test` variant and replacing it with `cmakeflags`. Do you have thoughts about this?

Since I don't think it will merged soon enough (or at all), I would opt for alternatives like you already suggested. In case something happens in the official spack, we can eventually change accordingly later.

> Can you think of other ways to do it?

Whatever I think of at the moment looks a bit ugly. I was thinking of some kind of environment variable guided spack package, e.g. it might accept `DLAF_CMAKE_FLAGS` as env variable which gets injected (somehow as a replacement for the `cmakeflags`).


Just as reminder, there is also this thing for Codecov builds that gravitates around the same kind of problem

https://github.com/eth-cscs/DLA-Future/blob/bbc403cfe95872fc699e593ae3149b67dad0ed35/ci/docker/codecov.Dockerfile#L22-L23

ci/docker/codecov.Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN spack repo rm --scope site dlaf && \
     spack -e ci concretize -f && \
     mkdir ${BUILD} && \
     ln -s ${BUILD} `spack -e ci location -b dla-future` && \
-    spack -e ci install --jobs ${NUM_PROCS} --keep-stage --verbose
+    spack -e ci --config "config:flags:keep_werror:all" install --jobs ${NUM_PROCS} --keep-stage --verbose
```
albestro, line 33: It might be an overshoot, but I think it might be worth highlight it anyway.

We could have a folder with a common set of configurations, that we can include in all spack environments.

See https://spack.readthedocs.io/en/latest/environments.html#included-configurations

test/unit/communication/test_transform_mpi.cpp
```diff
@@ -98,7 +98,7 @@ TEST_F(TransformMPITest, PromiseGuardManagement) {
     // Note:
     // At this point we checked if everything went as expected in the posting phase. Let's signal
     // Rank 1 that it can send back the message that will unlock IRecv.
-    const int signal = 13;
+    int signal = 13;
```
albestro, line 101: 🤔 

```c
int MPI_Send(const void* buf, int count, MPI_Datatype datatype, int dest, int tag, MPI_Comm comm) 
```

(source :https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node47.htm#Node47)

test/unit/communication/test_communicator_grid.cpp
```diff
@@ -29,7 +29,7 @@ void test_grid_communication(CommunicatorGrid& grid) {
   if (MPI_COMM_NULL == grid.rowCommunicator() || MPI_COMM_NULL == grid.colCommunicator())
     return;
 
-  const int buffer_send = 1;
+  int buffer_send = 1;
```
albestro, line 32: 🤔

```c
int MPI_Allreduce(const void* sendbuf, void* recvbuf, int count, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm) 
```

(source: https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node117.htm#Node117)

ci/docker/codecov.Dockerfile
msimberg: Hmm, interesting. I'll see how easy it would be to update that in this PR still.

test/unit/communication/test_communicator_grid.cpp
msimberg: @albestro what you're posting is correct. What clang is saying is dumb. We're basically silencing a bogus warning from clang.

ci/docker/codecov.Dockerfile
albestro: Just saying, in case you go for it, we might also move other things that are already duplicated.

https://github.com/eth-cscs/DLA-Future/blob/bbc403cfe95872fc699e593ae3149b67dad0ed35/ci/docker/release-cpu.yaml#L19-L21

https://github.com/eth-cscs/DLA-Future/blob/bbc403cfe95872fc699e593ae3149b67dad0ed35/ci/docker/debug-cpu.yaml#L32-L44

test/unit/communication/test_communicator_grid.cpp
rasolca: Does it happens with newer clang versions as well?

msimberg: I was apparently typing at the same time as @rasolca so my response did not take this into account

> Does it happens with newer clang versions as well?

I don't know. But I can try if I attempt to add some new CI configurations.

ci/docker/codecov.Dockerfile
msimberg: See https://github.com/eth-cscs/DLA-Future/pull/845 for the separate config file.

I'm marking this resolved based on the assumption that it's not worth creating a config file only for the installation of DLA-Future which needs `keep_werror:all`. Please unresolve if you disagree.

test/unit/communication/test_communicator_grid.cpp
msimberg: I'll have a go at https://github.com/eth-cscs/DLA-Future/issues/847 once this and #845 are merged.

msimberg:
Adding these here for reference: https://github.com/llvm/llvm-project/issues/62512, https://github.com/pmodels/mpich/issues/6496. Do you want to wait for more detailed responses or would you like to see another approach before we get a response there? Even in the case that we do get a response I suppose we wouldn't have an upstream fix very soon anyway (next releases of mpich and/or clang earliest).

rasolca:
I would disable the clang checks for the specific lines of code and not mess with the constness of the arguments (until the check is fixed).

msimberg:
cscs-ci run

msimberg:
> I would disable the clang checks for the specific lines of code and not mess with the constness of the arguments (until the check is fixed). I would not wait for a fix.

See https://github.com/eth-cscs/DLA-Future/pull/834/commits/c0ae7306e801c3791fd5e42cd955ddcf1349a2ee.

msimberg:
cscs-ci run

rasolca (approved):

albestro (approved):

msimberg:
cscs-ci run

PR #835, opened 2023-04-14
Permute Distributed: migrate from AllToAll to Send/Recv

albestro:
Optimizations:
- Split unpacking of permutations from "all-at-once" to "local + others".
"locals" do not depend anymore on communication and can happen in parallel with the communication of "others".
- Migrate from MPI_AllToAllv to MPI_Send/MPI_Recv
  - Might reduce amount of data to be communicated (local are not communicated anymore)
  - Do not require anymore contiguousness of the MPI datatype
  For permute-row, this means not having anymore to transpose data, which implied a copy on a support memory (alternatively mat_in and mat_out...).
  - Not having to use `mat_in` as support memory, it was possible to fix const-correctness of the API

Additional Changelog:
- Test has been improved: now it also checks out-of-the-range invariability + there is a test-config for checking an edge case of the new implementation (related to local unpacking)
- Fix const-correctness also in `test_permutations_local`
- minor: Fix warning in MPI call inside gtest_listener
- some renamings to make some parts of the code more self-explanatory

_Note: I'm not super-happy about the code duplication introduced by `applyPermutationsFiltered` (more or less a copy of `applyPermutations`). I don't want to spend to much now on it, but in the near future I might have to do some other change on that part and your suggestion might be helpful._

### DIFF
--- a/include/dlaf/permutations/general/impl.h
+++ b/include/dlaf/permutations/general/impl.h
@@ -145,7 +145,7 @@ void applyPermutationsFiltered(
 
   // Parallelized over the number of permutations
   pika::for_loop(pika::execution::par, to_sizet(0), to_sizet(nperms), [&](SizeType i_perm) {
-    if (filter(perm_arr[i_perm]))
+    if (!filter(perm_arr[i_perm]))
       return;
 
     for (std::size_t i_split = 0; i_split < splits.size() - 1; ++i_split) {
@@ -584,7 +584,7 @@ void permuteOnCPU(common::Pipeline<comm::Communicator>& sub_task_chain, SizeType
     // [a, b)
     applyPermutationsFiltered<T, D, C>({0, 0}, sz, 0, subm_dist, perm_offseted.data(), mat_in_tiles,
                                        mat_out_tiles,
-                                       [a, b](SizeType i_perm) { return i_perm < a || i_perm >= b; });
+                                       [a, b](SizeType i_perm) { return i_perm >= a && i_perm < b; });
   };
 
   ex::when_all(send_counts_sender, recv_counts_sender, whenAllReadOnlyTilesArray(unpacking_index),
@@ -609,11 +609,11 @@ void permuteOnCPU(common::Pipeline<comm::Communicator>& sub_task_chain, SizeType
 
     // [0, a)
     applyPermutationsFiltered<T, D, C>({0, 0}, sz, 0, subm_dist, perm_arr, mat_in_tiles, mat_out_tiles,
-                                       [a](SizeType i_perm) { return i_perm >= a; });
+                                       [a](SizeType i_perm) { return i_perm < a; });
 
     // [b, end)
     applyPermutationsFiltered<T, D, C>({0, 0}, sz, 0, subm_dist, perm_arr, mat_in_tiles, mat_out_tiles,
-                                       [b](SizeType i_perm) { return i_perm < b; });
+                                       [b](SizeType i_perm) { return i_perm >= b; });
   };
 
   ex::when_all(recv_counts_sender, whenAllReadOnlyTilesArray(unpacking_index),
--- a/include/dlaf/permutations/general/impl.h
+++ b/include/dlaf/permutations/general/impl.h
@@ -212,20 +212,21 @@ void all2allEmptyData(common::Pipeline<comm::Communicator>& sub_task_chain, int 
 template <class T, Device D, class SendCountsSender, class RecvCountsSender>
 void all2allDataCol(common::Pipeline<comm::Communicator>& sub_task_chain, int nranks,
                     LocalElementSize sz_loc, SendCountsSender&& send_counts_sender,
-                    Matrix<T, D>& send_mat, RecvCountsSender&& recv_counts_sender,
+                    Matrix<const T, D>& send_mat, RecvCountsSender&& recv_counts_sender,
                     Matrix<T, D>& recv_mat) {
   namespace ex = pika::execution::experimental;
 
-  auto sendrecv_f = [vec_size = sz_loc.rows()](comm::Communicator& comm, std::vector<int> send_counts,
-                                               std::vector<int> send_displs,
-                                               const std::vector<matrix::Tile<T, D>>& send_tiles,
-                                               std::vector<int> recv_counts,
-                                               std::vector<int> recv_displs,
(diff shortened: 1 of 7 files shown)


include/dlaf/permutations/general/impl.h
```diff
@@ -191,119 +236,135 @@ auto whenAllReadOnlyTilesArray(Matrix<const T, D>& matrix) {
...
+                ex::just() | dlaf::comm::internal::transformMPI([=](MPI_Request* req) {
+                  const SizeType nperms = send_counts[rank_partner_index];
+                  auto message = dlaf::comm::make_message(
+                      DataDescriptor<const T>(send_ptr +
+                                                  send_displs[rank_partner_index] * send_perm_stride,
+                                              C == Coord::Col ? nperms : vec_size,
+                                              C == Coord::Col ? vec_size : nperms, send_ld));
+
+                  DLAF_MPI_CHECK_ERROR(MPI_Isend(message.data(), message.count(), message.mpi_type(),
+                                                 rank_partner, 0, comm, req));
```
albestro, line 288: note-to-self: use a better TAG, in order to not rely on other constraints for avoiding mismatched communication.

@rasolca suggests to use `index_split` (wrt tridiagonal solver)

test/src/gtest_mpi_listener.cpp
```diff
@@ -160,7 +160,8 @@ void MPIListener::OnTestEndAllRanks(const ::testing::TestInfo& test_info) const
 
 namespace internal {
 void mpi_send_string(const std::string& message, int to_rank) {
-  MPI_Send(message.c_str(), static_cast<int>(message.size()) + 1, MPI_CHAR, to_rank, 0, MPI_COMM_WORLD);
+  MPI_Send(const_cast<char*>(message.c_str()), static_cast<int>(message.size()) + 1, MPI_CHAR, to_rank,
+           0, MPI_COMM_WORLD);
```
msimberg, line 164: FYI, also changed in #834.

include/dlaf/permutations/general/impl.h
```diff
@@ -126,9 +129,51 @@ void applyPermutations(GlobalElementIndex out_begin, GlobalElementSize sz, SizeT
   }
 }
 
+template <class T, Device D, Coord C>
+void applyPermutationsFiltered(
+    GlobalElementIndex out_begin, GlobalElementSize sz, SizeType in_offset,
+    const matrix::Distribution& subm_dist, const SizeType* perm_arr,
+    const std::vector<pika::shared_future<matrix::Tile<const T, D>>>& in_tiles_fut,
+    const std::vector<matrix::Tile<T, D>>& out_tiles, std::function<bool(SizeType)> filter) {
```
msimberg, line 137: Regarding the duplication here, you could do:
```suggestion
    const std::vector<matrix::Tile<T, D>>& out_tiles, std::function<bool(SizeType)> filter = [](SizeType){ return false; }) {
```
If you also template the callable the compiler might completely optimize it away. The downside is another template parameter, and if you want to check that the callable is a `bool(SizeType)` the signature becomes more verbose.

msimberg, line 137: I realize the direction is arbitrary, so I don't feel too strongly about this, except that it should written in the documentation explicitly whether the predicate filters in or out.

My suggestion would be to rename `filter` to `pred` and have it filter items out when the predicate returns `false`. This would be consistent with https://en.cppreference.com/w/cpp/ranges/filter_view. However, I can see why the current option might be more intuitive as well, which is why the first part of this comment is the important one to avoid confusion.

include/dlaf/permutations/general/impl.h
```diff
@@ -126,9 +129,51 @@ void applyPermutations(GlobalElementIndex out_begin, GlobalElementSize sz, SizeT
...
+  constexpr auto OC = orthogonal(C);
+  std::vector<SizeType> splits =
+      dlaf::util::interleaveSplits(sz.get<OC>(), subm_dist.blockSize().get<OC>(),
+                                   subm_dist.distanceToAdjacentTile<OC>(in_offset),
+                                   subm_dist.distanceToAdjacentTile<OC>(out_begin.get<OC>()));
+
+  const SizeType nperms = subm_dist.size().get<C>();
+
+  // Parallelized over the number of permutations
+  pika::for_loop(pika::execution::par, to_sizet(0), to_sizet(nperms), [&](SizeType i_perm) {
```
msimberg, line 147: This is for a follow-up, i.e. no change requested here, but I want to comment to spread awareness: calls to pika's parallel algorithms should be replaced with the asynchronous versions. In this particular case we could simply use `bulk`.

include/dlaf/permutations/general/impl.h
albestro: It's a super good point.

I thought it was a good idea to check what other public APIs do...I looked at other functional approaches (e.g. [haskell](https://hackage.haskell.org/package/base-4.18.0.0/docs/Prelude.html#v:filter), [python](https://docs.python.org/3/library/functions.html?highlight=filter#filter)) and then I forgot to do apply the change 😛 

I fully agree with you to filter out when false. I'll do it.

test/unit/permutations/test_permutations_distributed.cpp
```diff
@@ -33,86 +41,136 @@ struct PermutationsDistTestMC : public TestWithCommGrids {};
...
-  Matrix<T, Device::CPU> mat_out_h(dist);
+  const Distribution dist(size, block_size, grid.size(), grid.rank(), src_rank_index);
 
-  SizeType index_start = dist.globalElementFromGlobalTileAndTileElement<C>(i_begin, 0);
-  SizeType index_finish = dist.globalElementFromGlobalTileAndTileElement<C>(i_end, 0) +
-                          dist.tileSize(GlobalTileIndex(i_end, i_end)).get<C>();
-  dlaf::matrix::util::set(perms_h, [index_start, index_finish](GlobalElementIndex i) {
-    if (index_start > i.row() || i.row() >= index_finish)
-      return SizeType(0);
+  const auto [index_start, index_end] = tileToElementRange(n, nb, i_begin, i_last + 1);
```
albestro, line 107: note: clang@12 complains because structured binding does not create variables but just "names", and it conflicts with what can be captured in lambdas (just variables).

rasolca:
cscs-ci run

msimberg (approved):

PR #840, opened 2023-04-17
scripts: create scripts for all miniapps {strong, weak} x {gpu, mc} + include local benchmarks

albestro:
Along `gen_dlaf_{strong,weak}`, this PR introduces a new pair of generator scripts in order to have a "template" for both `{mc,gpu}` (specifically it is PizDaint defaults).

⚠️ This operation in git history is not clean, and diffs between old and new files are a bit messy. Might be worth during the review to have a look at the file ignoring the diff proposed.

Main changes:
- now we have 4 generator scripts for all miniapps `gen_dlaf_{strong,weak}-{mc,gpu}`
- the `strong` version of the generator scripts includes an additional "local" benchmark, which runs a set of benchmarks with `--local` flag; these are useful as reference for distributed single-rank implementations
- all scripts have the main parameters in the beginning, so it is easier to customise the benchmark runs


albestro:
@rasolca I would take the chance of this April benchmark session for testing these new scripts, so that we can test/refine them as needed 😉

msimberg (commented):
Looks good!

However, why don't we add `createAndSubmitRun` as a helper into `miniapps.py` (or some other file if `miniapps.py` is too "official")?

I'm also wondering if it would make sense to have an API like this for `add`:
```
run.add([mp.chol, mp.evp, mp.trsm], "dlaf", dlafpath, {"rpn": rpn, ...})
```
The upside is that this more or less removes the need for a wrapper function (though a wrapper function can still be useful to combine `add`ing and `submit`ting). The downside is that we end up passing "illegal" keyword arguments to some algorithms and those would simply have to be ignored (or the `run` function would be bit more complex to only pass allowed arguments, but I'm not sure I like that). Too much magic?

scripts/gen_dlaf_weak-gpu.py
```diff
@@ -18,17 +18,27 @@
 import miniapps as mp
 import systems
 
-system = systems.cscs["daint-mc"]
+system = systems.cscs["daint-gpu"]
 
-dlafpath = "<path_to_dlaf>"
+dlafpath = "<path_to_dlaf_build_dir>"
 
 run_dir = f"~/ws/runs"
```
rasolca, line 25: having the same default directory for strong and weak is dangerous.

scripts/gen_dlaf_strong-gpu.py
```diff
@@ -0,0 +1,134 @@
...
+
+system = systems.cscs["daint-gpu"]
+
+dlafpath = "<path_to_dlaf_build_dir>"
+
+run_dir = f"~/ws/runs"
+
+time = 400  # minutes
+nruns = 5
+nodes_arr = [1, 2, 4, 8, 16]
```
rasolca, line 29: please add `extra_flags = "--dlaf:bt-band-to-tridiag-hh-apply-group-size=128"` (64 for MC)

scripts/gen_dlaf_strong-gpu.py
```diff
@@ -0,0 +1,134 @@
...
+# actual benchmark
+createAndSubmitRun(
+        run_dir,
+        nodes_arr)
+
+# additional benchmark collecting "local" implementation results in <run_dir>/local sub-directory
+createAndSubmitRun(
+        run_dir + "/local",
+        [1 / rpn],
+        extra_flags="--local")
```
rasolca, line 134:
```suggestion
        extra_flags="--local " + extra_flags)
```

scripts/gen_dlaf_strong-gpu.py
```diff
@@ -0,0 +1,134 @@
...
+        nruns, **kwargs,
+    )
+
+    run.submit(run_dir, debug=debug)
+
+
+# actual benchmark
+createAndSubmitRun(
+        run_dir,
+        nodes_arr)
```
rasolca, line 128:
```suggestion
        nodes_arr,
        extra_flags=extra_flags)
```

rasolca (commented):
@msimberg
- `createAndSubmitRun` is different for strong and weak benchmarks, moreover I prefer to have it in the script such that it is easier to quickly adapt the script to a subset of benchmarks without affecting other scripts.
- It might be a good idea to be able to add multiple routine in one step, however I would keep it simple for this round of benchmarks.

scripts/gen_dlaf_strong-gpu.py
```diff
@@ -0,0 +1,134 @@
...
+
+# additional benchmark collecting "local" implementation results in <run_dir>/local sub-directory
+createAndSubmitRun(
+        run_dir + "/local",
```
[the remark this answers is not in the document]
albestro: See https://github.com/eth-cscs/DLA-Future/pull/840/commits/4e371648b2dab7ac17b87cec8cfa393e544e818b

scripts/gen_dlaf_strong-gpu.py
```diff
@@ -0,0 +1,134 @@
...
+        {"rpn": rpn, "m_sz": m_szs, "mb_sz": mb_szs, "min_band": None},
+        nruns, **kwargs,
+    )
+    run.add(
+        mp.gevp,
+        "dlaf",
+        dlafpath,
+        {"rpn": rpn, "m_sz": m_szs, "mb_sz": mb_szs, "min_band": None},
+        nruns, **kwargs,
+    )
```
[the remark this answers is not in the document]
albestro: See https://github.com/eth-cscs/DLA-Future/pull/840/commits/3d7efe39f56cb43891c78cda9dbed18f12972f2a

scripts/gen_dlaf_strong-gpu.py
albestro: See https://github.com/eth-cscs/DLA-Future/pull/840/commits/b573badc5bef61b3759346caf935f0f98bb192e2

scripts/gen_dlaf_weak-gpu.py
albestro: See https://github.com/eth-cscs/DLA-Future/pull/840/commits/3300d59a66caa3a932078adb831100d72d78cc29

scripts/gen_dlaf_strong-gpu.py
```diff
@@ -0,0 +1,139 @@
...
+    run.submit(run_dir, debug=debug)
+
+
+# actual benchmark
+createAndSubmitRun(
+        run_dir,
+        nodes_arr,
+        extra_flags=extra_flags)
+
+# additional benchmark collecting "local" implementation results in <run_dir>/local sub-directory
```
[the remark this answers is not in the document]
albestro: See https://github.com/eth-cscs/DLA-Future/pull/840/commits/956737f858d26992757dccc7aabde5e973f67aab

rasolca (approved):

scripts/gen_dlaf_strong-mc.py
```diff
@@ -0,0 +1,143 @@
...
+
+import argparse
+import miniapps as mp
+import systems
+
+system = systems.cscs["daint-mc"]
+
+dlafpath = "<path_to_dlaf_build_dir>"
+
+run_dir = f"~/ws/runs/strong"
```
msimberg, line 25: I guess we can leave out the `f`:
```suggestion
run_dir = "~/ws/runs/strong"
```
?

PR #841, opened 2023-04-20
Remove unnecessary `barrier::arrive_and_wait` from `reduction_to_band`

msimberg:
~The end of the `bulk` region is anyway synchronized before the `then` continuation is called, so there's no need for the extra `arrive_and_wait`.~ The synchronization is probably still needed between iterations of the for-loop, but not for the last iteration.

This improves performance by 0 to a few percent. A couple of plots comparing this PR (`dlaf_4b965...`) to [master](https://github.com/eth-cscs/DLA-Future/commit/c1b515f911d4670fed46b57944c49ba166824497) (`dlaf_c1b51...`):
red2band_strong_time_10240_512_128
red2band_strong_time_20480_512_128
The difference seems to be smaller with smaller blocksizes. There's still quite a bit of noise, but I think it looks like there's some signal there as well to say that it makes a small difference.

### DIFF
--- a/include/dlaf/eigensolver/reduction_to_band/impl.h
+++ b/include/dlaf/eigensolver/reduction_to_band/impl.h
@@ -338,7 +338,9 @@ auto computePanelReflectors(MatrixLike& mat_a, const matrix::SubPanelView& panel
 
                       // STEP3: update trailing panel (multi-threaded)
                       updateTrailingPanel(has_head, tiles, j, w[0], taus.back(), begin, end);
-                      barrier_ptr->arrive_and_wait();
+                      if (j + 1 < nrefls) {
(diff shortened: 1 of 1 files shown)


msimberg:
Hmm, ignore this for now. This doesn't seem safe as it is. Let's see with some tests.

cscs-ci run

msimberg:
Sorry for the noise here, with the updated version of removing the `arrive_and_wait` only from the last iteration of the loop the benefit is negligible, if not just noise. The only reason to keep this change would be for the semantic correctness but it adds noise in the code. I'll close this and if someone feels like this is still a worthwhile change, please comment anyway. I suppose `nrefls` being just 1 or a small number is not a very common case, or at least not common enough to make a difference?

PR #845, opened 2023-04-27
Factor out common CI spack configurations into separate file

msimberg:
Based on the discussion in https://github.com/eth-cscs/DLA-Future/pull/834#discussion_r1177808829. Untested.

### DIFF
--- a/ci/docker/debug-cpu.yaml
+++ b/ci/docker/debug-cpu.yaml
@@ -9,36 +9,17 @@
 #
 
 spack:
+  include:
+  - common.yaml
+
   specs:
     - dla-future@develop build_type=Debug +miniapps +ci-test ^openblas ^mpich
-  view: false
-  concretizer:
-    unify:
-      true
 
   packages:
     all:
-      target: [x86_64]
       variants:
         - 'build_type=Release'
-    blaspp:
-      variants:
-        - '~cuda'
-        - '~openmp'
     pika:
       variants:
         - 'build_type=Debug'
         - 'malloc=system'
-    mpich:
-      version:
-        - 3.4.2
-      variants:
-        - '~fortran'
-        - '~libxml2'
-    hwloc:
-      variants:
-        - '~libxml2'
-    git:
-      # Force git as non-buildable to allow deprecated versions in environments
-      # https://github.com/spack/spack/pull/30040
-      buildable: false
--- a/ci/docker/common.yaml
+++ b/ci/docker/common.yaml
@@ -1,0 +1,37 @@
+#
+# Distributed Linear Algebra with Future (DLAF)
+#
+# Copyright (c) 2018-2023, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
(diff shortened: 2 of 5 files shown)


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

rasolca (changes requested):
TAG_ENVIRONMENT has to include the common environment as well,
otherwise changes in common.yaml will not trigger a rebuild.

msimberg:
> TAG_ENVIRONMENT has to include the common environment as well, otherwise changes in common.yaml will not trigger a rebuild.

Good catch. Update in https://github.com/eth-cscs/DLA-Future/pull/845/commits/358853655a32b839c4d07c37124f94746d8b7e6c.

msimberg:
cscs-ci run

ci/docker/build.Dockerfile
```diff
@@ -94,13 +94,18 @@ RUN spack repo add --scope site /user_repo
...
+ARG COMMON_SPACK_ENVIRONMENT
 # Build dependencies
 # 1. Create a spack environment named `ci` from the input spack.yaml file
 COPY $SPACK_ENVIRONMENT /spack_environment/spack.yaml
 RUN spack env create --without-view ci /spack_environment/spack.yaml
-# 2. Set the C++ standard
+# 2. Copy the common environment configuration into the environment directory
+# Note that the destination path is hardcoded because we cannot dynamically get
+# the destination path. This will fail if the environment path ever changes.
+COPY $COMMON_SPACK_ENVIRONMENT /opt/spack/var/spack/environments/ci
```
msimberg, line 105: This is a bit ugly with the hardcoded path, but my Dockerfile knowledge is not enough to say if there are better alternatives to this. We can get the path with `spack location -e ci` but we can't use that in the `COPY` directive directly. I don't know if there are some roundabout ways of using it there.

At least it seems like spack will complain loudly if it can't include `common.yaml` so if this ever changes we would notice, but I'm open to better suggestions here.

ci/docker/debug-cpu.yaml
```diff
@@ -9,36 +9,22 @@
...
 spack:
-  specs:
-    - dla-future@develop build_type=Debug +miniapps +ci-test ^openblas ^mpich
+  include:
+  - common.yaml
+
   view: false
   concretizer:
     unify:
       true
```
msimberg, line 18: Note that this is common to all the environments, but it seems like spack allow having these in the common config file, so I've left this duplicated.

rasolca: You mean **doesn't** allow?

rasolca (approved):

ci/docker/debug-cpu.yaml
msimberg: > Note that this is common to all the environments, but it seems like spack (edit:) *does not* allow having these in the common config file, so I've left this duplicated.

albestro (approved):

PR #849, opened 2023-04-28
Add `cmakeflags` variant to spack package

msimberg:
This is definitely not ready, and I'm not convinced this is a good idea after giving this a try. The main problem I see with this is that it's not (as far as I know) possible to combine `cmakeflags=X` and `cmakeflags=Y` from different sources (e.g. defaults in `common.yaml`, HIP/CUDA-specific flags in environment, coverage in Dockerfile). Based on #834 and #845 so only look at the last commit.

What do you think?


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
I'm going to close this since it doesn't seem to be a promising direction, and it's low priority. We can always come back to it.

PR #853, opened 2023-05-03
Add GCC 12 CI configuration

### LINKED ISSUE #839: Add CI gcc configuration with C++20
To prepare for stdexec, and to fix warnings about capturing `this` in lambdas.

msimberg:
Fixes #839.

### DIFF
--- a/ci/.gitlab-ci.yml
+++ b/ci/.gitlab-ci.yml
@@ -163,7 +163,7 @@ cpu release build clang16 cxx20:
     DEPLOY_DOCKER_FILE: ci/docker/deploy.Dockerfile
     BASE_IMAGE: docker.io/ubuntu:23.04
     DEPLOY_BASE_IMAGE: docker.io/ubuntu:23.04
-    EXTRA_APTGET: clang-16
+    EXTRA_APTGET: "clang-16 libomp-16-dev"
     EXTRA_APTGET_DEPLOY: "glibc-tools"
     # glibc-tools is needed for libSegFault on ubuntu:23.04
     COMPILER: clang@16.0.0
--- a/ci/.gitlab-ci.yml
+++ b/ci/.gitlab-ci.yml
@@ -134,6 +134,28 @@ cpu release build clang13 cxx20:
     THREADS_PER_NODE: 72
     USE_CODECOV: "false"
 
+cpu release build clang16 cxx20:
+  extends: .build_spack_common
+  variables:
+    BUILD_DOCKER_FILE: ci/docker/build.Dockerfile
+    DEPLOY_DOCKER_FILE: ci/docker/deploy.Dockerfile
+    BASE_IMAGE: docker.io/ubuntu:23.04
+    DEPLOY_BASE_IMAGE: docker.io/ubuntu:23.04
+    EXTRA_APTGET: clang-16
+    EXTRA_APTGET_DEPLOY: "glibc-tools"
+    # glibc-tools is needed for libSegFault on ubuntu:23.04
+    COMPILER: clang@16.0.0
+    CXXSTD: 20
+    USE_MKL: "ON"
+    USE_ROCBLAS: "OFF"
+    SPACK_ENVIRONMENT: ci/docker/release-cpu.yaml
+    COMMON_SPACK_ENVIRONMENT: ci/docker/common.yaml
+    BUILD_IMAGE: $CSCS_REGISTRY_PATH/release-cpu-clang16-20/build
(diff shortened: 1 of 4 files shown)


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
https://github.com/eth-cscs/DLA-Future/pull/853/commits/b173a1f83e4b630011a920aa4223e76ddfc2f8db works around a bug in GCC that seems to appear in GCC 12... and GCC 13 prerelease available in Ubuntu 23.04. See here for a minimal example: https://godbolt.org/z/KP4W689Pb. Clang seems to have no trouble with the example with any version. `gcc-13` in Ubuntu 23.04 reports as its version
```
gcc-13 (Ubuntu 13-20230320-1ubuntu1) 13.0.1 20230320 (experimental) [master r13-6759-g5194ad1958c]
```
so I'm guessing the fix was maybe not there yet in that version (GCC 13 on compiler explorer has no problems).

msimberg:
cscs-ci run

msimberg:
cscs-ci run

ci/.gitlab-ci.yml
```diff
@@ -134,6 +156,28 @@ cpu release build clang13 cxx20:
...
     USE_CODECOV: "false"
 
+cpu release build clang16 cxx20:
+  extends: .build_spack_common
+  variables:
+    BUILD_DOCKER_FILE: ci/docker/build.Dockerfile
+    DEPLOY_DOCKER_FILE: ci/docker/deploy.Dockerfile
+    BASE_IMAGE: docker.io/ubuntu:23.04
+    DEPLOY_BASE_IMAGE: docker.io/ubuntu:23.04
+    EXTRA_APTGET: clang-16
```
RMeli, line 166:
```suggestion
    EXTRA_APTGET: "clang-16 libomp-16-dev"
```
We can do this here or in #668

msimberg: Good point. I think we can have #668 in first. I'll update this here already now.

msimberg: https://github.com/eth-cscs/DLA-Future/pull/853/commits/3ee241fc7295b4b2143e9996b47ff118cfa25efc

msimberg:
Waiting for #668 to be merged before this.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
It seems like the problem of `blaspp` not configuring with clang 16 (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/4265672150#L1368) was due to the linker not finding `libomp` (even though `libomp-16-dev` was installed). From exploring a bit it seems like for clang 13 and `libomp-13-dev` the linker explicitly adds a path to an llvm directory which contains `libomp.so`. With clang 14, 15, and 16 that path is not set by the linker so it doesn't find `libomp.so`. Instead of trying to patch this by setting linker paths manually I tried the LLVM repository and there everything seems to be correctly configured. So I switched the base image back to 22.04 for clang and I'm now using the LLVM repo instead for the clang 16 configuration.

Because I changed the clang 16 configuration back to 22.04 I also changed the GCC 13 configuration to GCC 12 on top of 22.04. Now all configurations are again on the LTS release. Additionally, the GCC 13 in 23.04 was anyway a pre-release so it's maybe better to stick with 12 just because of that as well.

Once I've figured out what I messed up in the configuration file syntax this should be ready to go again...

msimberg:
cscs-ci run

msimberg:
I need more investigation on the clang behaviour so I've removed the clang 16 configuration from this PR. The state before removing the clang 16 configuration is still here: https://github.com/eth-cscs/DLA-Future/compare/master...msimberg:DLA-Future:clang-16-ci-config.

My previous investigation was sloppy: with the packages from Ubuntu's repositories I couldn't even compile with `clang++ -fopenmp file.cpp`, much less `clang++ -lomp file.cpp`. With clang from LLVM's repositories I tested that `clang++ -fopenmp file.cpp` worked. However, `clang++ -lomp file.cpp` did not work which is why CI was still failing. Because of https://github.com/spack/spack/blob/bcb7af6eb371b8958824e2f0551c4baa88d754a3/lib/spack/spack/build_systems/intel.py#L704 mkl is adding `-lomp` to `blas_libs`, not `-fopenmp` (or `-fopenmp=libomp`). I don't know if that's correct or if that should be fixed in spack. I'm going to see how a clang-built clang 16 behaves separately from this PR.

msimberg:
cscs-ci run

RMeli (approved):

rasolca (approved):

PR #843, opened 2023-04-25
tridiagonal eigensolver: uniformize usage of end index

rasolca:
Closes #708

Refactored tridiagonal eigensolver to use open ranges.

- [x] #854
- [x] remove some duplications (Waiting for #766 before moving some functions)
- [x] clang-format (only at the end to simplify diffs)

PR #858, opened 2023-05-06
Remove not needed column permutations

rasolca:
Based on #843

PR #860, opened 2023-05-09
TridiagSolver (local): "bulkerify" rank1 problem solution

albestro:
This PR aims at reducing micro-tasking for rank1 problem solution (local).

TODO:
- [x] #843 
- [x] Is re-use of `z` vector workspace a good thing? (it allows to allocate just one vector<vector> workspace for partial results, see next point)
- [x] Workspaces for multi-threading in bulk are allocated as vector<vector> instead of re-using a Matrix in a workspace. Is it ok?
- [x] ~Apply also row-permutations in bulk? (@rasolca)~
- [x] Evaluate if something can be done for set0
- [x] ~Do we want to embed `setUnitDiag`?~
- [x] How many `nthreads` for bulk?

### DIFF
--- a/include/dlaf/eigensolver/get_tridiag_rank1_nworkers.h
+++ b/include/dlaf/eigensolver/get_tridiag_rank1_nworkers.h
@@ -19,13 +19,13 @@
 namespace dlaf::eigensolver::internal {
 
 inline std::size_t getTridiagRank1NWorkers() noexcept {
+  // Note: precautionarily we leave at least 1 thread "free" to do other stuff (if possible)
+  const std::size_t available_workers = pika::resource::get_thread_pool("default").get_os_thread_count();
+  const std::size_t min_workers = 1;
+  const auto max_workers = std::max(min_workers, available_workers - 1);
+
   const std::size_t nworkers = getTuneParameters().tridiag_rank1_nworkers;
-
-  // Note: precautionarily we leave at least 1 thread "free" to do other stuff
-  const std::size_t max_workers = pika::resource::get_thread_pool("default").get_os_thread_count() - 1;
-
-  // 1 <= number of workers < max_workers
-  return std::max<std::size_t>(1, std::min<std::size_t>(max_workers, nworkers));
+  return std::clamp(nworkers, min_workers, max_workers);
 }
 
 }
--- a/include/dlaf/eigensolver/get_tridiag_rank1_nworkers.h
+++ b/include/dlaf/eigensolver/get_tridiag_rank1_nworkers.h
@@ -1,0 +1,30 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2023, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+#pragma once
+
+#include <cstdint>
+
+#include <pika/runtime.hpp>
+
+#include "dlaf/tune.h"
+
+namespace dlaf::eigensolver::internal {
+
+inline std::size_t getTridiagRank1NWorkers() noexcept {
+  const std::size_t nworkers = getTuneParameters().tridiag_rank1_nworkers;
+
+  // Note: precautionarily we leave at least 1 thread "free" to do other stuff
+  const std::size_t max_workers = pika::resource::get_thread_pool("default").get_os_thread_count() - 1;
+
+  // 1 <= number of workers < max_workers
+  return std::max<std::size_t>(1, std::min<std::size_t>(max_workers, nworkers));
+}
+
+}
--- a/include/dlaf/eigensolver/get_tridiag_rank1_nworkers.h
+++ b/include/dlaf/eigensolver/get_tridiag_rank1_nworkers.h
@@ -9,6 +9,7 @@
 //
 #pragma once
 
+#include <algorithm>
 #include <cstdint>
 
 #include <pika/runtime.hpp>
--- a/include/dlaf/eigensolver/tridiag_solver/merge.h
+++ b/include/dlaf/eigensolver/tridiag_solver/merge.h
@@ -776,6 +776,9 @@ void mergeSubproblems(const SizeType i_begin, const SizeType i_split, const Size
   //
   invertIndex(i_begin, i_end, ws_h.i3, ws_hm.i2);
 
+  // Note:
+  // This is neeeded to set to zero elements of e2 outside of the k by k top-left part.
+  // The input is not required to be zero for solveRank1Problem.
   matrix::util::set0<Backend::MC>(pika::execution::thread_priority::normal, idx_loc_begin, sz_loc_tiles,
                                   ws_hm.e2);
   solveRank1Problem(i_begin, i_end, k, scaled_rho, ws_hm.d1, ws_hm.z1, ws_h.d0, ws_hm.e2);
--- a/include/dlaf/eigensolver/tridiag_solver/merge.h
+++ b/include/dlaf/eigensolver/tridiag_solver/merge.h
@@ -9,6 +9,8 @@
 //
 #pragma once
 
+#include <algorithm>
+
 #include <pika/algorithm.hpp>
 #include <pika/barrier.hpp>
 #include <pika/execution.hpp>
@@ -28,6 +30,7 @@
 #include "dlaf/matrix/distribution.h"
 #include "dlaf/matrix/index.h"
 #include "dlaf/matrix/matrix.h"
+#include "dlaf/memory/memory_view.h"
 #include "dlaf/multiplication/general.h"
 #include "dlaf/permutations/general.h"
 #include "dlaf/permutations/general/impl.h"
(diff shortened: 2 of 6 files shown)


rasolca:
>     * [ ]  Is re-use of `z` vector workspace a good thing? (it allows to allocate just one vector workspace for partial results, see next point)
`z` (`z0` in #858) is needed only in the deflation step. Then it is permuted in `ztmp` (`z1`) and not needed anymore until the parent merge. As the next `z` can be assembled only when the child eigenvectors are computed, overlap is not possible in any case and therefore it is safe to reuse it.

>     * [ ]  Workspaces for multi-threading in bulk are allocated as vector instead of re-using a Matrix in a workspace. Is it ok?
Ok for allocating in the bulk, but vectors bypass umpire. Is it ok?

>     * [ ]  Apply also row-permutations in bulk? (@rasolca)
It would be good to already embed it.

>     * [ ]  Evaluate if something can be done for set0
Not sure what you are talking about.

>     * [ ]  Do we want to embed `setUnitDiag`?
It is not needed, however if we opt for the row permutation in bulk approach, `setUnitDiag` needs to be rewritten anyway.
Embedding it means avoiding to release all the tiles of the evec matrix and reacquire them in a when_all.

>     * [ ]  How many `nthreads` for bulk?
Depends on the current merge size. For sure it needs a tuning option, but we should think about which is the best way.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -439,44 +440,124 @@ auto applyDeflation(const SizeType i_begin, const SizeType i_end, RhoSender&& rh
...
 
-  auto sender =
-      ex::when_all(std::forward<KSender>(k), std::forward<RhoSender>(rho),
-                   ex::when_all_vector(tc.read(d)), ex::when_all_vector(tc.read(z)),
-                   ex::when_all_vector(tc.readwrite(evals)), ex::when_all_vector(tc.readwrite(evecs)));
+        // - compute productorial (thread-local)
+        for (auto j = to_SizeType(begin); j < to_SizeType(end); ++j) {
+          for (auto i = 0; i < k; ++i) {
+            if (i == j)
+              continue;
```
rasolca, line 514: `if` statements in inner loops should be avoided (performance problem).
It might have a negligible effect on the full algorithm, but maybe the two loops option should be considered.

If you want to keep this approach please consider to embed ll 500-508 in the `i==j` part (improved code readability)

albestro:
cscs-ci run

rasolca (approved):

include/dlaf/eigensolver/get_tridiag_rank1_nworkers.h
```diff
@@ -0,0 +1,30 @@
...
+namespace dlaf::eigensolver::internal {
+
+inline std::size_t getTridiagRank1NWorkers() noexcept {
+  const std::size_t nworkers = getTuneParameters().tridiag_rank1_nworkers;
+
+  // Note: precautionarily we leave at least 1 thread "free" to do other stuff
+  const std::size_t max_workers = pika::resource::get_thread_pool("default").get_os_thread_count() - 1;
+
+  // 1 <= number of workers < max_workers
+  return std::max<std::size_t>(1, std::min<std::size_t>(max_workers, nworkers));
```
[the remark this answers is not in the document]
msimberg: `std::min/max` also need `#include <algorithm>`.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -9,12 +9,16 @@
 //
 #pragma once
 
+#include <algorithm>
+
 #include <pika/algorithm.hpp>
```
msimberg, line 14:
```suggestion
```

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -445,47 +450,140 @@ auto applyDeflation(const SizeType i_begin, const SizeType i_end, RhoSender&& rh
...
+      ex::when_all(ex::just(std::make_shared<pika::barrier<>>(nthreads)), std::forward<KSender>(k),
+                   std::forward<RhoSender>(rho), ex::when_all_vector(tc.read(d)),
+                   ex::when_all_vector(tc.readwrite(z)), ex::when_all_vector(tc.readwrite(evals)),
+                   ex::when_all_vector(tc.readwrite(evecs)),
+                   ex::just(std::vector<memory::MemoryView<T, Device::CPU>>())) |
+      ex::transfer(di::getBackendScheduler<Backend::MC>(pika::execution::thread_priority::high)) |
+      ex::bulk(nthreads, [nthreads, n, nb](std::size_t thread_idx, auto& barrier_ptr, auto& k, auto& rho,
+                                           auto& d_tiles_futs, auto& z_tiles, auto& eval_tiles,
+                                           auto& evec_tiles, auto& ws_vecs) {
+        common::internal::SingleThreadedBlasScope single;
```
msimberg, line 477: I realize now that the scope of this has to be narrower. There's a risk that the barrier yields and the pika task is resumed on another threads, in which case the destructor of `SingleThreadedBlasScope` may reset the number of threads on the wrong worker thread.

I will check if this also needs to be changed in the other bulk algorithms. We can also have a debug check to make sure the same worker thread sets and restores the number of threads.

include/dlaf/tune.h
```diff
@@ -32,8 +35,10 @@ namespace dlaf {
 ///     DLAF_BT_BAND_TO_TRIDIAG_HH_APPLY_GROUP_SIZE.
 /// Note to developers: Users can change these values, therefore consistency has to be ensured by algorithms.
 struct TuneParameters {
-  size_t red2band_panel_nworkers =
-      std::max<size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count() / 2);
+  std::size_t red2band_panel_nworkers =
+      std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count() / 2);
+  std::size_t tridiag_rank1_nworkers =
+      std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count() / 2);
```
msimberg, line 41: This needs a comment above in the docstring.

src/init.cpp
```diff
@@ -206,6 +209,9 @@ pika::program_options::options_description getOptionsDescription() {
   desc.add_options()(
       "dlaf:band-to-tridiag-1d-block-size-base", pika::program_options::value<SizeType>(),
       "The 1D block size for band_to_tridiagonal is computed as 1d_block_size_base / nb * nb. (The input matrix is distributed with a {nb x nb} block size.)");
+  desc.add_options()(
+      "dlaf:tridiag-rank1-nworkers", pika::program_options::value<std::size_t>(),
+      "Maximum number of threads to use for computing rank1 problem solution in tridiagonal solver algorithm.");
```
msimberg, line 214: Only for consistency with other descriptions: 
```suggestion
      "The maximum number of threads to use for computing rank1 problem solution in tridiagonal solver algorithm.");
```

msimberg (changes requested):

include/dlaf/eigensolver/get_tridiag_rank1_nworkers.h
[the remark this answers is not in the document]
albestro: Actually the comment above the return is wrong in many ways, but the most important wrong bit is that number of workers is not ensured to be inside that range.

With `max_workers = 0`, that we get when there is a single thread inside the default pool, current implementation does `max(1, min(0, x)) = 1` (given `x >= 0`).

The first trivial `clamp` replacement would look like

```cpp
std::clamp<std::size_t>(nworkers, 1, max_workers);
```

which specifically ends up in the corner case with following values

```cpp
std::clamp<std::size_t>(nworkers, 1, 0); // corner case
```

But from `clamp` doc

> The behavior is undefined if the value of lo is greater than hi.

so we should manage differently this corner case. The solution might be

```cpp
  // `max_workers` should be at least 1
  const std::size_t max_workers = std::max(1, pika::resource::get_thread_pool("default").get_os_thread_count() - 1);
  return std::clamp<std::size_t>(nworkers, 1, max_workers);
```

Anyhow, whatever option we will take, a probably worth discussion to have is if it is a safety requirement to leave 1 rank free from bulk operation or it is a strong requirement (and we require at least two threads in default pool).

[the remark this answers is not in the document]
msimberg: Good point about the undefined behaviour. In that case the `min`/`max` version that you already had there is possibly the clearest. I'm happy with either version.

Regarding number of threads, I don't think it should be a strong requirement that the default pool has at least two worker threads. If it is, we should make sure to lift that requirement (since it implies we're doing something blocking).

Independent of `clamp` vs. `min`/`max`, couldn't you leave out at least some of the `std::size_t` template parameters:
```suggestion
  // Note: precautionarily we leave at least 1 thread "free" to do other stuff
  const std::size_t max_workers = pika::resource::get_thread_pool("default").get_os_thread_count() - 1;

  // 1 <= number of workers < max_workers
  return std::max(std::size_t(1), std::min(max_workers, nworkers));
```
?

include/dlaf/eigensolver/tridiag_solver/merge.h
albestro: If I get it correctly, we should not end up having `SingleThreadedBlasScope` crossing the boundary of a barrier yield.

If that's correct, I will fix it accordingly in this bulk, where we will need two of them, one for `laed4` and one for `nrm2`.

msimberg: Correct.

More generally: anything that may yield should not have a `SingleThreadedBlasScope` around it (the number of places where this can happen "silently" will slightly reduce with pika 0.16.0 because of https://github.com/pika-org/pika/pull/672).

include/dlaf/eigensolver/get_tridiag_rank1_nworkers.h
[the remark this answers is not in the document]
albestro: what about https://github.com/eth-cscs/DLA-Future/pull/860/commits/8eab3d1ed9b860911810d55ea7039b49fda06acc?

include/dlaf/tune.h
albestro: See https://github.com/eth-cscs/DLA-Future/pull/860/commits/72bdd5ad2951f9947367d429ad10b197d44c2e84

msimberg: Thanks, looks good.

include/dlaf/eigensolver/tridiag_solver/merge.h
albestro: Fixed in https://github.com/eth-cscs/DLA-Future/pull/860/commits/c31139be7001630d8cc9d0af6548252fc0971198

msimberg: Perfect, thanks.

msimberg (approved):

PR #862, opened 2023-05-11
Check single-threaded BLAS in CI

### LINKED ISSUE #861: Check that single-threaded BLAS is correctly enforced in CI
As a follow-up to #668.

msimberg:
Fixes #861. In MC configurations in CI, checks (using strace) that tests only spawn pika threads if they spawn any threads.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
This is not ready, but it's time for an intermediate update because at least the basic principle is working. See here for an example of the test output: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/4286400096. In this run I've disabled `SingleThreadedBlasScope` in `trmm3` which means that some miniapps spawn more threads than they should and some spawn the right amount. The whole job fails because of this.

The idea behind this is to:
1. Run a test with `--pika:exit --pika:print-bind` to get the number of threads that we expect pika to spawn for its thread pools (e.g. 4)
2. The number of threads we expect to actually be spawned is one more than the number of threads in the thread pools (e.g. 4 + 1 = 5; one thread is used for shutdown by the pika runtime)
3. Run a test under `strace -f -e clone3 executable` to print all calls to `clone3`, which is what gets called when pthreads are created.
4. Compare the number of expected threads to the number of `clone3` syscalls, fail if they don't match because in that case OpenMP or MKL threads were spawned (unless something went even more badly wrong).

Additional notes:
- The `strace` call uses `-f`, i.e. `--follow-forks`. This means that threads spawned from child threads will also be logged (e.g. OpenMP threads spawned from a pika worker thread). Without this option everything will look correct, but it may not actually be the case.
- In the runner the syscall which creates new threads is `clone3` so we check for that with `-e clone3`. This is apparently a newer version of the `clone` syscall (https://www.man7.org/linux/man-pages/man2/clone.2.html). E.g. on my own laptop some threads are created through `clone` and others through `clone3`. This is just to say that on other systems you may need to check for `clone` instead of `clone3`.
- I currently run the test through `bash -c $test` because simply running `$test` expands to a single "command" with arguments included and the "executable" is not found (e.g. `/path/to/miniapp_cholesky --grid-rows=2 --grid-cols=2`). `bash -c` separates the arguments again, but if someone knows a more idiomatic way of doing this let me know.

Cleanup that should be done next (especially things that I'm realizing now that I'm summarizing how this works).
- [ ] Call test executables only once. At the moment I'm calling them first once with `--pika:exit`  to get the number of threads pika will spawn, and then once for the actual test. However, this can be combined into one run of the test.
- [ ] Put `strace` output directly into a file with the `-o` flag. This avoids redirecting all output into a file, and allows us to keep the "normal" test output while also using `strace`. This would also allow us to check the number of threads in all configurations without downsides (possibly?).
- [ ] Because of the above: separate running miniapps as tests into a new PR. It's not necessary for this PR.
- [ ] Check what non-pika tests do with `--pika:print-bind`. Do we have to inject `--pika:print-bind` from cmake to `MPIPIKA` and `PIKA` tests, or can the flag be added unconditionally to all tests in the `mpi-ctest` script?
- [ ] add comment about `clone` vs `clone3`
- [ ] Use long option `--follow-forks` and add a comment about why it's used.

Let me know if you disagree with some of the above cleanup points, or if you see some other direction that could be cleaner. I'll start cleaning this up later in the week.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

ci/.gitlab-ci.yml
```diff
@@ -186,7 +186,7 @@ cuda release build gcc11:
     BASE_IMAGE: docker.io/nvidia/cuda:11.7.0-devel-ubuntu22.04
     DEPLOY_BASE_IMAGE: docker.io/ubuntu:22.04
     EXTRA_APTGET: ""
-    EXTRA_APTGET_DEPLOY: "glibc-tools"
+    EXTRA_APTGET_DEPLOY: "glibc-tools jq strace"
```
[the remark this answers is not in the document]
msimberg: Yep, no longer needed. I'll remove this.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

spack/packages/dla-future/package.py
```diff
@@ -149,6 +149,9 @@ def cmake_args(self):
             args.append(self.define("BUILD_TESTING", True))
             args.append(self.define("DLAF_BUILD_TESTING", True))
             args.append(self.define("DLAF_CI_RUNNER_USES_MPIRUN", True))
+
+            if "~cuda" in self.spec and "~rocm" in self.spec:
+                args.append(self.define("DLAF_TEST_PREFLAGS", "check-threads"))
```
msimberg, line 154: I added a `DLAF_TEST_PREFLAGS` CMake variable which is included in between MPI flags (if any) and the test executable. I use this to add the `check-threads` wrapper script. For symmetry I also added the option for `DLAF_TEST_POSTFLAGS`, but this is unused.

I currently have the logic for choosing to use `check-threads` into the spack package itself. There are (at least) two other options:

1. Inject `DLAF_TEST_PREFLAGS` the same way the coverage flag is injected, i.e. by rewriting the package file in CI.
2. Add a variant to enable this (`variant("ci-check-threads", ...)`?)
3. ???

Would you prefer any of these over the current implementation?

ci/docker/deploy.Dockerfile
```diff
@@ -38,14 +38,14 @@ RUN pushd ${SOURCE}/miniapp && \
 
 # Prune and bundle binaries
 RUN mkdir ${BUILD}-tmp && cd ${BUILD} && \
-    export TEST_BINARIES=`ctest --show-only=json-v1 | jq '.tests | map(.command[0]) | .[]' | tr -d \"` && \
+    export TEST_BINARIES=`PATH=${SOURCE}/ci:$PATH ctest --show-only=json-v1 | jq '.tests | map(.command | .[] | select(contains("check-threads") | not)) | .[]' | tr -d \"` && \
```
msimberg, line 41: This change is related to the change from `${test_target_name}` to `$<TARGET_FILE:${test_target_name}>` in `DLAF_AddTest.cmake.

CMake/CTest assumes that the first element in the `COMMAND` for a test is an executable name. If CTest can't find the location of the executable the json output will _simply not have a `command` entry_. Therefore we have to make sure `check-threads` is found in `PATH` when we populate `TEST_BINARIES`.

Additionally, when we inject `check-threads` as a wrapper the test itself is now the first _argument_ to the executable. This means that the json output will _not_ expand a test target name in that position. This is why I now changed all uses of `${test_target_name}` in the `COMMAND` to use `$<TARGET_FILE:${test_target_name}>` so that the path is expanded already by CMake.

Finally, because `check-threads` is now in the `command` array of the tests I filter it out for `TEST_BINARIES`.

Let me know if and how you'd like to see this commented in the CMake config and/or Dockerfile.

msimberg:
cscs-ci run

msimberg:
There are a couple of open cleanup questions, but the suggestion to separate out the thread checking into a separate script cleaned this up a lot. Please have another look.

msimberg:
cscs-ci run

spack/packages/dla-future/package.py
rasolca: I'm starting to think that we should add these "developer" variants in the spack package in our repository, and only remove them when we upload it to the spack repo.
In this way the full set of variants is available, if we want to do a quick manual test on any system, and we avoid package rewriting workaround in the CI. These variants will not be exposed to the users as they will mainly use the package from the spack repo.

ci/docker/deploy.Dockerfile
rasolca: Not sure. That line was already obscure before and not commented.
I easily figured out the reason of the changes without the comment (the same changes appear in the codecov dockerfile which displays before this file in the PR...)
Does anyone think a comment is needed?

rasolca (commented):
Looks good.

spack/packages/dla-future/package.py
msimberg: I'd be in favour of this. We're not going to run out of variants, so no need to stuff all the options behind `+ci-test`.

ci/.gitlab-ci.yml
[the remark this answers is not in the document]
msimberg: This was removed.

ci/ctest_to_gitlab.sh
```diff
@@ -109,6 +109,12 @@ deallocate:
   extends: .daint_dealloc
 "
 
+if [ "$SLURM_CONSTRAINT" = "mc" ]; then
```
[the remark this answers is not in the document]
msimberg: This was removed. Related discussion continues in https://github.com/eth-cscs/DLA-Future/pull/862#discussion_r1203873349.

ci/mpi-ctest-check-threads
```diff
@@ -0,0 +1,91 @@
...
+    REPORT_NAME=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
+
+    mkdir -p "$SHARED_REPORTS"
+fi
+
+pushd /DLA-Future-build >/dev/null
+
+# Get commands of all tests, since we will run the commands directly under
+# strace.
+IFS=$'\n' all_tests=($(ctest $@ --show-only=json-v1 | jq --raw-output '.tests | map(.command) | .[] | join(" ")'))
```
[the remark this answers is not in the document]
msimberg: It's using `ctest` again, closing.

msimberg:
cscs-ci run

rasolca (approved):

albestro (approved):
LGTM

PR #863, opened 2023-05-11
Print binary sizes in CI

msimberg:
Poor man's solution to keep track of binary sizes. This won't warn us about it, but at least we'd have the numbers in CI to compare without having to manually build different configurations.

### DIFF
--- a/ci/docker/codecov.Dockerfile
+++ b/ci/docker/codecov.Dockerfile
@@ -35,6 +35,9 @@ RUN spack repo rm --scope site dlaf && \
 # Prune and bundle binaries
 RUN mkdir ${BUILD}-tmp && cd ${BUILD} && \
     export TEST_BINARIES=`ctest --show-only=json-v1 | jq '.tests | map(.command[0]) | .[]' | tr -d \"` && \
+    echo "Binary sizes:" && \
+    ls -lh ${TEST_BINARIES} && \
+    ls -lh src/lib* && \
     libtree -d ${DEPLOY} ${TEST_BINARIES} && \
     rm -rf ${DEPLOY}/usr/bin && \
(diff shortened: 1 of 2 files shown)


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
The builds failed because the the tests failed to get an allocation. Otherwise this is ready to go from my point of view. This basically just adds output like https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/4269236035#L1696 to the builds.

albestro (approved):
Good idea!

Minor comment: It's easy enough to spot in a log, but probably it might helpful to have a tag/hook in the output so that in case we will have to look at more files, it might be easier to just do CTRL+F"<TAG>" (if we go for an easy to remember "TAG" 😄)

msimberg:
> Minor comment: It's easy enough to spot in a log, but probably it might helpful to have a tag/hook in the output so that in case we will have to look at more files, it might be easier to just do CTRL+F"" (if we go for an easy to remember "TAG" smile)

Also good idea! :smile: In the roughly 1-day lifetime of this PR, I was using `ls -lh` as the "tag" to search for but maybe something a bit more descripting would be useful. Any suggestions? `echo "Binary sizes:"`?

msimberg:
Tentatively until others have a chance to comment: https://github.com/eth-cscs/DLA-Future/commit/6e5b812fe48920e75fe1208ecb311e13ce049a10#diff-56687a3c6a70d6474d1713c07b9c8f0793c019444352b3685c98fd477f73134dR38.

RMeli (approved):

albestro (approved):

PR #864, opened 2023-05-12
Use busy waiting barrier in reduction to band

### LINKED ISSUE #833: Need of a "semi" blocking barrier for bulk
Currently, in `pika` we have a "yielding" barrier. This is not optimal since lightweight tasks of the bulk might get rescheduled and this add overheads.

It would be mitigated with a "blocking" barrier: in pika is not yet implemented, but a feasibility test has been done with a `pthread_barrier` (really dangerous, since it can lead to deadlocks, since it blocks the os-thread).

_Originally posted by @albestro in https://github.com/eth-cscs/DLA-Future/issues/797#issuecomment-1507284800_

msimberg:
Fixes #833, using the busy wait parameter to `barrier::wait` in https://github.com/pika-org/pika/pull/685. I've added configuration option for the busy wait time (in microseconds) specifically for reduction to band. https://github.com/eth-cscs/DLA-Future/pull/860 will probably also need to use the busy waiting barrier. We can see if we want to merge the options into a single `BARRIER_BUSY_WAIT_US` option, or if we need separate `TRIDIAG_SOLVER_BARRIER_BUSY_WAIT_US`, `RED2BAND_BARRIER_BUSY_WAIT_US` etc. The default value for the busy wait is 1 ms which seemed to show the best performance across daint and eiger, while still being relatively low (about the min task size that we anyway recommend).

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -174,6 +174,9 @@ void updateConfiguration(const pika::program_options::variables_map& vm, configu
   updateConfigurationValue(vm, param.tridiag_rank1_nworkers, "TRIDIAG_RANK1_NWORKERS",
                            "tridiag-rank1-nworkers");
 
+  updateConfigurationValue(vm, param.red2band_barrier_busy_wait_us, "TRIDIAG_RANK1_BARRIER_BUSY_WAIT_US",
+                           "tridiag-rank1-barrier-busy-wait-us");
+
   updateConfigurationValue(vm, param.bt_band_to_tridiag_hh_apply_group_size,
                            "DLAF_BT_BAND_TO_TRIDIAG_HH_APPLY_GROUP_SIZE",
                            "bt-band-to-tridiag-hh-apply-group-size");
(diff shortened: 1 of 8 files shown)


msimberg:
cscs-ci run

msimberg:
This is high priority for performance in reduction to band, but on hold until the next pika release.

src/init.cpp
```diff
@@ -199,7 +202,9 @@ pika::program_options::options_description getOptionsDescription() {
   // Tune parameters command line options
   desc.add_options()(
       "dlaf:red2band-panel-nworkers", pika::program_options::value<std::size_t>(),
-      "Maximum number of threads to use for computing the panel in the reduction to band algorithm.");
+      "The maximum number of threads to use for computing the panel in the reduction to band algorithm");
```
rasolca, line 205: Why did you remove the `.`?

msimberg: No reason I think. This looks like a mistake!

msimberg, line 205:
```suggestion
      "The maximum number of threads to use for computing the panel in the reduction to band algorithm.");
```

src/init.cpp
```diff
@@ -199,7 +202,9 @@ pika::program_options::options_description getOptionsDescription() {
   // Tune parameters command line options
   desc.add_options()(
       "dlaf:red2band-panel-nworkers", pika::program_options::value<std::size_t>(),
-      "Maximum number of threads to use for computing the panel in the reduction to band algorithm.");
+      "The maximum number of threads to use for computing the panel in the reduction to band algorithm.");
```
msimberg, line 205: The added `The` at the beginning is only for consistency with the description in `dlaf/tune.h` and the other options, but it doesn't really belong in this PR. Happy to move it out if you prefer that.

msimberg:
cscs-ci run

msimberg:
I've added a config option to busy wait in the tridiagonal solver barriers as well but I have _not_ yet tested if it has the same effect as in the reduction to band algorithm.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
I rebased this and made the default busy wait time 0 for the rank 1 problem barrier.

rasolca (approved):

PR #874, opened 2023-05-17
Only ignore `build*` in root of repository

msimberg:
For the longest time I was wondering why files like `build.Dockerfile` didn't show up in file searches for me. It turns out it was because it's using `fd` which uses `.gitignore` by default and it ignores `build*` in _any_ directory. I've updated the `.gitignore` pattern to be `/build*` which will ignore `build*` only in the root of the repository. I'm guessing this was the original intent of this pattern. Is it? Do you have use cases for ignoring `build*` also in subdirectories?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 .DS_Store
 *~
 *.swp
-build*
+/build*
 .ptp-sync-folder
 .vscode/
 !build.sh


albestro:
I don't have a use-case, but just as alternative, we can use `build*/` to ignore any build DIRECTORY instead of both FILES+DIR.

rasolca (approved):

msimberg:
> I don't have a use-case, but just as alternative, we can use `build*/` to ignore any build DIRECTORY instead of both FILES+DIR.

Good point, I wasn't aware of that option. I'm happy with any of `/build*`, `/build*/`, or `build*/`.

RMeli (approved):
I don't mind any of the suggested options.

albestro (approved):
It's ok for me too. In case of any problem we can always change later 😉

PR #410, opened 2021-08-17
algo: reduction to band (local)

albestro:

PR #877, opened 2023-05-17
Update coding style

### LINKED ISSUE #42: clang-format headers order
Currently clang-format format headers according to the coding style:
- if the main header is in a separated block (no way to fix it, but already mentioned in the coding style)
- if C and C++ standard headers are included using `<>`
- if third party libraries and DLAF headers are included with `""`.

The coding style do not mention any restriction on `<>` or `""`

Question:
Should we restrict it? At what level?

IMHO we should restrict C and C++ to `<>` and DLAF to `""`, but I'm not sure about third party libs.

### LINKED ISSUE #576: Consider changing naming style
TL;DR: It'd be nice to follow the standard library (and pika, etc.) naming style of `snake_case` for almost everything except template parameters (which are `CapitalCase`).

Trigger: With senders and sender adaptors the pattern of "callable objects" (CPO, niebloid, whatever you want to call them) is more common. With our current naming scheme a plain function is:
```
T myFunction(...);
```
whereas a callable object would be
```
struct MyFunction { T operator()(...); } my_function{};
```
which means at a callsite the user now sees inconsistent casing depending on the _implementation_ of a function, which they don't really need to care about. (Semi-related: I'd like to start getting rid of the `_o` suffix for the auto-generated callable objects as that is also an implementation detail).

Other reasons:
- personal preference
- consistency with pika and the standard library

I know that this is ultimately a subjective matter, but like #340 we shouldn't be afraid to make this change sooner rather than later while we can still do it without breaking users. Of course, if there's no consensus I suppose the status quo wins.

### LINKED ISSUE #871: Revise coding style
- [ ] revise naming conventions #576 
- [ ] introduce a guideline for indices naming
- [ ] bump clang format version (version 10 is quite outdated)

rasolca:
- [x] revise naming conventions #576 
- [x] introduce a guideline for indices naming
- [x] bump clang format version (version 10 is quite outdated)

Warning: this PR cannot be squash-merged, otherwise `.git-blame-ignore-revs` will not work correctly
- [x] check that ignored commit are still correct before merging (rebasing change them)


misc/coding.md
```diff
@@ -555,3 +438,29 @@ The post-increment and post-decrement operators create an unnecessary copy, that
...
+- `i` and `m` in general refers to rows (exceptions square sizes and diagonal),
+- `j` and `n` in general refers to columns (exceptions square sizes and diagonal).
+
+Note: if the use of the row or column index/size is determined by a template or a function parameter `i`, `m` should be used for the "parallel" coordinate and `j`, `n` for the orthogonal one.
+
+2D Indices:
+- Use the two indices to name the 2D object. E.g. `ij`. Note: In this way extracting a single index with `SizeType i = ij.row();` is intuitive.
+- The use of `id`, `idx` and similar names is discouraged.
+2D Size:
+- Name the object to identify what it refers to. E.g. `size_a` represent the number of global tiles of the matrix A.
```
rasolca, line 455: In this case I decided to not use combined names as they are hard to read. E.g. I find `mn` hard to interpret...

misc/coding.md
```diff
@@ -283,6 +297,8 @@ Do not use scopes for formatting reason.
 No trailing whitespaces should be added to any line.
 Use no space before a comma (`,`) and a semicolon (`;`) and add a space after them if they are not at the end of a line.
 
+For C++ files the clang-format style included in the repository provides the correct settings.
+
 ### Parenthesis
 
 Parenthesis should have no internal padding.
```
msimberg, line 304:
```suggestion
### Parentheses

Parentheses should have no internal padding.
```

misc/coding.md
```diff
@@ -555,3 +438,29 @@ The post-increment and post-decrement operators create an unnecessary copy, that
 ### Alternative Operator Representations
 
 Use the primary token for operators (E.g. `&&`, `||`, and `!`), alternative representations are not allowed (`and`, `or`, and `not`).
+
+## Index naming convention
+
+To have an uniform usage of names for indices and sizes in the library, please follow the following conventions:
+- Use `m`, `n` and `k` for sizes and `i`, `j` and `l`,
```
msimberg, line 445:
```suggestion
- Use `m`, `n` and `k` for sizes and `i`, `j` and `l` for indices,
```
?

misc/coding.md
```diff
@@ -29,10 +29,6 @@ They can only be included in specific parts of the code.
 
 In general the definition of template classes, functions, ... should be included in the header files.
 
-All header files should be self-contained.
-This means that each header can be included without defining anything special.
-In particular each header file should include all the headers needed and have `#pragma once` to prevent multiple inclusion.
-
 ### Includes
 Header files should be included with the full path based on the `include` directory.
 For example, the file `<libname>/include/<libname>/foo/bar.hpp` should be included as
```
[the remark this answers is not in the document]
rasolca: It should already match, but there is at least one case that has to be handled manually.
E.g. `dlaf/matrix.h` should be included first in `test_matrix.cpp`. clang-format-10 didn't support it.
This is the reason why we require a blank line after the related header.

Before updating this section I wanted to decide which version of clang-format to use and see if the limitation disappeared.

misc/coding.md
```diff
@@ -52,6 +48,8 @@ For readability it is suggested to use the standard order of includes:
 
 In each section the included files should be sorted in alphabetical order.
 
+Note: An empty line can be used to separate the different groups.
+
```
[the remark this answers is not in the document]
rasolca: I added it explicitly as we require a blank line after the related header, therefore it can be interpreted that no other blank lines are allowed.

misc/coding.md
```diff
@@ -29,10 +29,6 @@ They can only be included in specific parts of the code.
```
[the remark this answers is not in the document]
rasolca: section removed

misc/coding.md
```diff
@@ -65,31 +57,34 @@ functionalities. E.g trait helpers ends with `_t` and `_v`.
...
 ### Function Names
 
-Function names should start with a lowercase character and have a capital letter for each new word.
-Underscores (`_`) are not allowed.
-
-The only exception to this rule are getter and setter methods for class data members.
-They follow the naming convention `set_variable` and `get_variable`, respectively (see [Class Format](#class-format)).
+Variable names should contain only lowercase characters and underscores (`_`).
+The underscore is used to separate words and is not allowed as first or last character.
+Two consecutive underscores (`__`) are not allowed.
```
msimberg, line 87: This actually has an effect on the release: the algorithm function names would be changed by this. Should we sneak this in before 0.1.0 or do it afterwards? Not the end of the world to do it after 0.1.0 since, as far as I can tell, CP2K only needs `cholesky` and `eigensolver` (not `genEigensolver`) whose names would not change by this guideline.

rasolca: My idea was to make this task part of the refactoring and not part of the release.
The update of the API will therefore happen in 0.2.0.

misc/coding.md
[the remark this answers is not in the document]
msimberg: Ok, it's at least not currently taking into account C vs C++ headers but that can easily be updated. The `^<ext/.*\.h>` rule can probably be removed as well.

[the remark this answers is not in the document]
rasolca: #42 is still open...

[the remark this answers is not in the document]
rasolca: Currently it partially support it with the `^<.*\.h>` (C) and `^<.*` (C++) rules.
"Partially" as it rely on the fact that C and c++ uses `<>` and the rest `""`
A definitive solution to this point depends on #42 as well.

rasolca:
cscs-ci run

misc/coding.md
```diff
@@ -16,47 +16,70 @@ Tests should be placed according to their type in `test/<type>/`.
...
+
+In each group the included files should be sorted in alphabetical order.
+In this case Do not put empty lines between groups to allow clang-format to sort the headers in the correct order.
+
+**Note**
+The `.clang-format` style for unit tests is different from the style provided for the rest of the code.
+In tests, to allow the user to freely choose the tested header, include sorting is preserving the block structure.
+Therefore,
+* insert a blank line between the tested header and the other files such that clang-format keeps them on top of the list,
+* do not insert any other blank line to allow clang-format to sort correctly the rest of the headers.
```
rasolca, line 75: See main comment.

rasolca (commented):
I had to use two styles and two different clang format styles to automate as much as possible sorting the include files.

For include, source and miniapp files I use the option `IncludeBlocks: Regroup`, which means that include statements are combined (i.e. empty lines are ignored), sorted and re-grouped. In this case sorting is fully automatic.

For tests it is important to include the tested header in the first position, to check if the header is self-contained. Clang format has a cumbersome mechanism (you can specify only a regex for the suffix) to identify if the current source file is the main source of the  header. I considered renaming all test_*.cpp to *_test.cpp, but unfortunately it doesn't work either, as we use the full include path (`dlaf/...`).
I found two workarounds to this problem:
- include the related header with the quote syntax and introduce a special group with the regex `^"dlaf/` with priority -1.
- use `IncludeBlocks: Preserve` for tests and rely on the empty line to keep the related header in the first position.

I decided for the second option, however I'm open to switch to the first option if you find it better. (or feel free to propose a better solution...)

@msimberg this answers the question in #882.

rasolca:
cscs-ci run

.clang-format
```diff
@@ -38,56 +60,83 @@ BraceWrapping:
...
   - BOOST_FOREACH
-IncludeBlocks:   Preserve
+IfMacros:
+  - KJ_IF_MAYBE
+IncludeBlocks:   Regroup
 IncludeCategories:
-  - Regex:           'dlaf.*/'
+  - Regex:           '^<dlaf_test'
+    Priority:        7
+    SortPriority:    0
```
msimberg, line 100: Do I understand correctly that `SortPriority` is for sorting of _groups_ of includes, rather than individual includes? And setting them all to `0` means that all includes are going to be regrouped into a single group?

.clang-format
```diff
@@ -38,56 +60,83 @@ BraceWrapping:
...
   - foreach
   - Q_FOREACH
   - BOOST_FOREACH
-IncludeBlocks:   Preserve
+IfMacros:
+  - KJ_IF_MAYBE
+IncludeBlocks:   Regroup
 IncludeCategories:
-  - Regex:           'dlaf.*/'
+  - Regex:           '^<dlaf_test'
```
msimberg, line 98: May I ask you to sort the categories by priority (in the same order they would appear in source files, so priority 1 first?)?

.git-blame-ignore-revs
```diff
@@ -0,0 +1,8 @@
+# Reformat with clang-format-15
+423a4667c203a60f5f6eb2c82dba0f829129b137
```
msimberg, line 2: I guess you'll not squash merge this PR? This commit already seems to be out of date.

msimberg (changes requested):
Only minor comments. Thanks for doing this @rasolca, looks really nice!

.clang-format
rasolca: Not possible, the order is the evaluation order and is some cases changing the order produce different groups

.clang-format
rasolca: Different priority means different group, while `SortPriority` is used to specify the order in the given group.
`Priority: 0` is the special group clang-format uses for the corresponding header

e.g.
* header C: P 1 SP 0
* header B: P 0 SP 0
* header A: P 1 SP 2
will be ordered as
```
#include <B>

#include <C>
#include <A>
```

.git-blame-ignore-revs
rasolca: Please read the first comment of the PR...

.clang-format
rasolca: E.g. 
```
  - Regex:           '^<.*>'
    Priority:        2
```
will match almost all include headers.

.git-blame-ignore-revs
msimberg: Sorry about that.

I'm still leaving this comment unresolved though until rebasing and the hash has been updated.

.clang-format
msimberg: Makes sense, thanks!

msimberg (approved):

include/dlaf/common/callable_object.h
```diff
@@ -21,12 +21,12 @@
...
-    auto operator()(Ts&&... ts) const noexcept(noexcept((fname)(std::forward<Ts>(ts)...))) \
-        -> decltype((fname)(std::forward<Ts>(ts)...)) {                                    \
-      return (fname)(std::forward<Ts>(ts)...);                                             \
-    }                                                                                      \
-  } fname##_o {                                                                            \
+#define DLAF_MAKE_CALLABLE_OBJECT(fname)                                                    \
+  constexpr struct fname##_t {                                                              \
+    template <typename... Ts>                                                               \
+    auto operator()(Ts&&... ts) const noexcept(noexcept((fname) (std::forward<Ts>(ts)...))) \
+        -> decltype((fname) (std::forward<Ts>(ts)...)) {                                    \
```
albestro, line 28: Probably you might want to check this
```suggestion
        -> decltype((fname)(std::forward<Ts>(ts)...)) {                                    \
```

misc/coding.md
```diff
@@ -555,3 +442,30 @@ The post-increment and post-decrement operators create an unnecessary copy, that
...
+- in general, `j` and `n` refers to columns (exceptions square sizes and diagonal).
+
+**Note**
+if the use of the row or column index/size is determined by a template or a function parameter `i`, `m` should be used for the "parallel" coordinate and `j`, `n` for the orthogonal one.
+
+2D Indices:
+- Use the two indices to name the 2D object. E.g. `ij`. In this way extracting a single index with `SizeType i = ij.row();` is intuitive.
+- The use of `id`, `idx` and similar names is discouraged.
+2D Size:
+- Name the object to identify what it refers to. E.g. `size_a` represent the number of global tiles of the matrix A.
```
albestro, line 460:
```suggestion
- Name the object to identify what it refers to. E.g. `size_a` represents the number of global tiles of the matrix A.
```

test/unit/matrix/test_matrix_output.cpp
albestro, line 1: Probably this is not following the coding style...

test/unit/eigensolver/test_gen_eigensolver.cpp
```diff
@@ -134,7 +134,7 @@ void testGenEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType
 
   const TileElementSize block_size(mb, mb);
 
-  auto create_reference = [&]() -> auto {
+  auto create_reference = [&]() -> auto{
```
albestro, line 137: not sure if you can control this

```suggestion
  auto create_reference = [&]() -> auto {
```

(in case, there is a similar thing in `test_eigensolver`)

test/include/dlaf_test/matrix/util_matrix_local.h
```diff
@@ -13,21 +13,22 @@
 /// @file
 
 #include <functional>
-#include <lapack/util.hh>
 #include <type_traits>
 
-#include <gtest/gtest.h>
+#include <lapack/util.hh>
```
albestro, line 18: From a quick check, we can probably remove this include.

Apparently, it was introduced for `MatrixType` in https://github.com/eth-cscs/DLA-Future/pull/410, but then it became useless with the refactoring in https://github.com/eth-cscs/DLA-Future/pull/484

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -10,34 +10,33 @@
...
 
-#include <pika/future.hpp>
-
 #include <blas.hh>
 
+#include <pika/future.hpp>
+
 #ifdef DLAF_WITH_GPU
 #include <whip.hpp>
 #endif
```
albestro, line 19: If I'm not wrong, this should be moved at the very end in the "conditionally included" section.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -10,34 +10,34 @@
...
 
-#include "dlaf/eigensolver/band_to_tridiag/api.h"
-
 #include <pika/future.hpp>
 
+#include <dlaf/eigensolver/band_to_tridiag/api.h>
+
 #ifdef DLAF_WITH_GPU
 #include <whip.hpp>
 #endif
```
albestro, line 19: Shouldn't this be moved at the end? See also the other similar comment

rasolca:
cscs-ci run

rasolca:
cscs-ci run

PR #880, opened 2023-05-19
Introduce HDF5

albestro:
This PR aims at introducing HDF5 support in order to be able to dump/load matrices from disk.

The first practical target of this PR is allowing to save on disk data feeded to the tridiagonal eigensolver embedded in the "full" eigensolver pipeline, so that we can feed the miniapp with same data and analyse the run in an easier way. This might be helpful for understanding the performance problem we saw in the last benchmark campaigns. We speculate it is a data-dependent problem, and we want to try understanding better in order to not miss any potential problem we have in the implementation.

Currently, I see it as a development/debug feature, so I'm not sure about its positioning and exposure inside the library.

TODO
- [x] #913 
- [x] ~wait for all pika tasks before HDF5 operation that implies MPI collective? (e.g. probably createDataset?)~
- [x] check open/create flags
- [x] doc for HDF5 helpers
- [x] cleanup code
- [x] ~remove temporary dump in eigensolver~ dump of trisolver input/output data conditional on DLAF_DEBUG_DUMP_TRISOLVER_DATA
- [x] add basic test


include/dlaf/matrix/hdf5.h
```diff
@@ -0,0 +1,274 @@
...
+                has_mpio_);
+
+    const hsize_t dims_file[3] = {
+        dlaf::to_sizet(matrix_host.size().cols()),
+        dlaf::to_sizet(matrix_host.size().rows()),
+        internal::hdf5_datatype<T>::dims,
+    };
+    H5::DataSpace dataspace_file(3, dims_file);
+
+    // TODO it might be needed to wait all pika tasks
```
albestro, line 187: I've found this reference https://support.hdfgroup.org/HDF5/doc/RM/CollectiveCalls.html

AFAIU the points we should care about because they use MPI collectives internally are:
- file creation
- dataset creation

include/dlaf/matrix/hdf5.h
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  /// Create a file that, besides read, supports writing in parallel from different ranks.
+  ///
+  /// @p comm Communicator grouping all ranks that will be able to write in parallel to the file
+  /// @p filepath filepath where the file will be created
+  ///
+  /// @pre @p filepath should not exist
+  /// @post file created will support parallel-write
+  FileHDF5(comm::Communicator comm, const std::string& filepath) {
```
albestro, line 157: `filepath` always first?

include/dlaf/eigensolver/eigensolver/impl.h
```diff
@@ -62,8 +64,19 @@ void Eigensolver<B, D, T>::call(comm::CommunicatorGrid grid, blas::Uplo uplo, Ma
   auto taus = reductionToBand<B>(grid, mat_a, band_size);
   auto ret = bandToTridiag<Backend::MC>(grid, uplo, band_size, mat_a);
 
+#ifdef DLAF_WITH_HDF5
+  matrix::FileHDF5 file(grid.fullCommunicator(), "trid-ref.h5");
+
+  file.write(ret.tridiagonal, "/tridiag");
+#endif
```
rasolca, line 71: What about keeping it, and enabling it conditionally on an env variable?
(e.g. DLAF_EIGENSOLVER_WRITE_TRIDIAGONAL)

include/dlaf/matrix/hdf5.h
rasolca: In general we use grid as first argument.

include/dlaf/eigensolver/eigensolver/impl.h
albestro: It's not a bad idea, but it might require to slightly change some policy.

Each time a solver runs it will store in the same file, in the same dataset. As it is now implemented, it will error on the second run because in order to write it expects that the file does not exist.

I opted for this behaviour with the aim of not overwriting/truncating existing files involuntarily. But we can change this if we prefer to.

albestro:
cscs-ci run

include/dlaf/eigensolver/eigensolver/impl.h
msimberg: > on an env variable

If we do this it should probably also be a configuration option and command line option (perhaps this was implied)?

> not overwriting/truncating existing files involuntarily

What about defaulting the filename to some "unique" filename (with a timestamp or similar)? And perhaps including an option to force overwrite an existing file?

These are extra features though and need not be included in this PR.

include/dlaf/matrix/hdf5.h
```diff
@@ -0,0 +1,278 @@
...
+template <class T>
+const H5::PredType& hdf5_datatype<std::complex<T>>::type = hdf5_datatype<T>::type;
+
+template <class T>
+struct hdf5_datatype<const T> : public hdf5_datatype<T> {};
+
+// Helper function that for each local tile index in @p dist, gets a sender of a tile with
+// @p get_tile and sends it to a function that takes care of the mapping between file and memory.
+// Then, this function, passes all required arguments to @p dataset_op which should be either
+// dataset.read or dataste.write, so it has the following signature
```
msimberg, line 60:
```suggestion
// dataset.read or dataset.write, so it has the following signature
```

include/dlaf/matrix/hdf5.h
```diff
@@ -0,0 +1,278 @@
...
+}
+
+class FileHDF5 final {
+public:
+  /// File access modes:
+  /// - READONLY, the file will be opened (if should already exist)
+  /// - READWRITE, the file will be created (if should not exist)
+  enum class FileMode {
+    READONLY,
+    READWRITE,
```
msimberg, line 137:
```suggestion
    readonly,
    readwrite,
```
?

spack/packages/dla-future/package.py
```diff
@@ -136,6 +140,10 @@ def cmake_args(self):
                 arch_str = ";".join(archs)
                 args.append(self.define("CMAKE_CUDA_ARCHITECTURES", arch_str))
 
+        # HDF5 support
+        if "+hdf5" in spec:
+            args.append(self.define("DLAF_WITH_HDF5", True))
```
msimberg, line 145:
```suggestion
        args.append(self.define_from_variant("DLAF_WITH_HDF5", "hdf5"))
```

msimberg (changes requested):

include/dlaf/matrix/hdf5.h
albestro: See [633616e](https://github.com/eth-cscs/DLA-Future/pull/880/commits/633616ecd5c72ab9f5e7a39e0b8756748d57e743)

rasolca (approved):

include/dlaf/matrix/hdf5.h
albestro: @msimberg for the moment I didn't "guarded" above calls with the pika call to wait all tasks.

Do you think it can be problematic? Those calls happen at schedule time at the moment.

msimberg: I'm quite certain they can be problematic, for the same reasons that all other blocking calls can be problematic. I assume you haven't had problems with this in practice which is why you've left it out now? We can always see how far we get without the `wait` and add it in later if it seems ok for now.

include/dlaf/tune.h
```diff
@@ -21,6 +21,11 @@ namespace dlaf {
 /// DLA-Future tuning parameters.
 ///
 /// Holds the value of the parameters that can be used to tune DLA-Future.
+/// - debug_dump_trisolver_data:
+///     Enable dump of trisolver input/output data to "trid-ref.h5" file that will be created in the
+///     working folder (it should not exist before the execution).
+///     WARNING: just a single execution can be dumped on disk, and any subsequent call fails.
+///     Set with environment variable DLAF_DEBUG_DUMP_TRISOLVER_DATA.
```
msimberg, line 28: Maybe it makes sense to put this somewhere else given that it's:
1. not a tuning parameter
2. a debugging parameter
? Not a blocker for me, but it would avoid having to put a warning public documentation.

src/CMakeLists.txt
```diff
@@ -88,6 +88,7 @@ target_link_libraries(
     $<$<BOOL:${DLAF_WITH_HIP}>:roc::rocprim>
     $<$<LINK_LANGUAGE:HIP>:roc::rocthrust> # hip adds offload-arch globally for the CXX language in hip-config
     $<$<BOOL:${DLAF_WITH_GPU}>:whip::whip>
+    $<$<BOOL:${DLAF_WITH_HDF5}>:hdf5_cpp-shared>
```
msimberg, line 91: Is `-shared` a requirement or a preference? What happens if `hdf5` is built with `~shared`?

src/init.cpp
```diff
@@ -122,6 +122,13 @@ struct parseFromString<SizeType> {
   };
 };
 
+template <>
+struct parseFromString<bool> {
+  static std::size_t call(const std::string& var) {
+    return !var.empty();
```
msimberg, line 128: `env DEBUG_DUMP_TRISOLVER_DATA=OFF` will set `debug_dump_trisolver_data` to `true`, correct?

albestro: Yeah, I thought a bit about that. Would you prefer something like this?

```cpp
template <>
struct parseFromString<bool> {
  static std::size_t call(std::string var) {
      constexpr const char* true_values[] = {"ON", "TRUE", "YES"};
      constexpr std::size_t n = sizeof(true_values)/sizeof(const char*);
      std::transform(var.cbegin(), var.cend(), var.begin(), [](unsigned char c){ return std::toupper(c); });
      bool is_true = (true_values + n) != std::find(true_values, true_values + n, str);
      return is_true;
  }
};
```

msimberg: I think so, yes. I would maybe add a `1` in there as well. For that version you could maybe instead this for the array
```
constexpr const std::array true_values{"ON", "TRUE", "YES", "1"};
```

However, the alternative to the above would be to treat `DEBUG_DUMP_TRISOLVER_DATA` being set as true, which is maybe what you were going for? If yes, that'd also be fine by me, but that requires some other changes outside of `parseFromString` (or just always returning `true` for `bool` with the assumption that `parseFromString` is only called if there is an environment variable defined; I don't like this assumption though...).

msimberg: We had a quick chat about this and I just want to document some of the options that we discussed:

| option             | a     | b     | c     | d     |
|--------------------|-------|-------|-------|-------|
| unset DEBUG        | false | false | false | false |
| export DEBUG=      | true  | false | false | true  |
| export DEBUG=TRUE  | true  | true  | true  | true  |
| export DEBUG=FALSE | true  | false | true  | false |

The current implementation on this branch is `c`. The suggestion by @albestro is `b`. `a` is "if the environment variable is set the option is true", and `d` is a mix: "if the environment variable is set to empty, or a true-ish non-empty value the option is true".

include/dlaf/tune.h
albestro: It is a good point.

@rasolca what do you think? probably not for this PR, but we can track this idea with an issue if you agree with @msimberg

src/init.cpp
albestro: In order to raise a point, I would suggest a small change to the first row of the above table:

| option             | a     | b     | c     | d     |
|--------------------|-------|-------|-------|-------|
| unset DEBUG        | default | default | default | default |
| export DEBUG       | true  | false | false | true  |
| export DEBUG=TRUE  | true  | true  | true  | true  |
| export DEBUG=FALSE | true  | false | true  | false |

One point of giving meaning to the value of the variable is that user will have the chance to set the option as preferred and not just `{default, true}`, which in the case `default=true` it means that the variable is meaningless.

For this reason I would drop A. Moreover, I would drop also C because it looks counter-intuitive at first sight (DEBUG=FALSE -> TRUE).

Finally, I would go for B or D. Actually D looks the more reasonable one to me too (probably it was also what @msimberg was supporting).

@rasolca do you agree? do you have any preference?

rasolca: Either b or d.

But I would suggest `e`:
`true_values{"ON", "TRUE", "YES", "1"};` `false_values{"OFF", "FALSE", "NO", "0"};` and the rest default.

albestro: @rasolca proposes E

| option |  b | d | e |
| -- |  -- | -- | --  |
| unset DEBUG |  default  | default | default |
| export DEBUG |  false | true | default |
| export DEBUG=TRUE |  true | true | true |
| export DEBUG=FALSE  | false | false | false |
| export DEBUG=*  | ? | ? | default | 

that can be described as "the variable should be interpreted as true-ish or false-ish, otherwise default"

include/dlaf/matrix/hdf5.h
```diff
@@ -0,0 +1,274 @@
...
+  };
+
+  /// Create/open a local file.
+  ///
+  /// @p filepath filepath where the file to be opened is or where it will be created
+  /// @p mode file access mode (see FileMode for more details)
+  ///
+  /// @post if mode == READWRITE, being a local file it will not support parallel-write
+  FileHDF5(const std::string& filepath, const FileMode& mode) {
+    file_ = H5::H5File(filepath, mode2flags(mode));
```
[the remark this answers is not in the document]
albestro: Let me take the chance also to show you the error messages, since in this PR we don't "filter" them and we let HDF5 output its own error messages.

This is a run on RO filesystem
```bash
ialberto@daint106:/project/csstaff/ialberto> DLAF_DEBUG_DUMP_TRISOLVER_DATA=1 srun -n 1 -c72 ~/workspace/dla-future.hdf5/build-mc/miniapp/miniapp_eigensolver --matrix-size 10240 --block-size 512 --nwarmups=0 --nruns 1
[0]
HDF5-DIAG: Error detected in HDF5 (1.14.0) MPI-process 0:
  #000: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5F.c line 836 in H5Fopen(): unable to synchronously open file
    major: File accessibility
    minor: Unable to open file
  #001: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5F.c line 796 in H5F__open_api_common(): unable to open file
    major: File accessibility
    minor: Unable to open file
  #002: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5VLcallback.c line 3863 in H5VL_file_open(): open failed
    major: Virtual Object Layer
    minor: Can't open object
  #003: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5VLcallback.c line 3675 in H5VL__file_open(): open failed
    major: Virtual Object Layer
    minor: Can't open object
  #004: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5VLnative_file.c line 128 in H5VL__native_file_open(): unable to open file
    major: File accessibility
    minor: Unable to open file
  #005: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5Fint.c line 1809 in H5F_open(): unable to open file: name = 'trid-ref.h5', tent_flags = 11
    major: File accessibility
    minor: Unable to open file
  #006: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5FD.c line 769 in H5FD_open(): open failed
    major: Virtual File Layer
    minor: Unable to initialize object
  #007: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5FDmpio.c line 907 in H5FD__mpio_open(): MPI_File_open failed: MPI error string is 'Read-only file or filesystem name, error stack:
ADIOI_CRAY_OPEN(111): Read access is required to this file'
    major: Internal error (too specific to document in detail)
    minor: Some MPI function failed
terminate called after throwing an instance of 'pika::detail::exception_with_info<pika::exception>'
  what():  attempted to call set_error on empty any_receiver: pika(bad_function_call)
srun: error: nid00008: task 0: Exited with exit code 255
srun: launch/slurm: _step_signal: Terminating StepId=47422047.0
```

And this on a RW filesyem, that I did just for double checking that it actually works as expected.
```bash
ialberto@daint106:/scratch/snx3000/ialberto> DLAF_DEBUG_DUMP_TRISOLVER_DATA=1 srun -n 1 -c72 ~/workspace/dla-future.hdf5/build-mc/miniapp/miniapp_eigensolver --matrix-size 10240 --block-size 512 --nwarmups=0 --nruns 1
[0]
[0] 62.6493s dL (10240, 10240) (512, 512) 128 (1, 1) 36 MC
```

If then you re-run it, since the file already exists, it aborts

```bash
ialberto@daint106:/scratch/snx3000/ialberto> DLAF_DEBUG_DUMP_TRISOLVER_DATA=1 srun -n 1 -c72 ~/workspace/dla-future.hdf5/build-mc/miniapp/miniapp_eigensolver --matrix-size 10240 --block-size 512 --nwarmups=0 --nruns 1
[0]
HDF5-DIAG: Error detected in HDF5 (1.14.0) MPI-process 0:
  #000: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5D.c line 186 in H5Dcreate2(): unable to synchronously create dataset
    major: Dataset
    minor: Unable to create file
  #001: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5D.c line 134 in H5D__create_api_common(): unable to create dataset
    major: Dataset
    minor: Unable to create file
  #002: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5VLcallback.c line 1876 in H5VL_dataset_create(): dataset create failed
    major: Virtual Object Layer
    minor: Unable to create file
  #003: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5VLcallback.c line 1841 in H5VL__dataset_create(): dataset create failed
    major: Virtual Object Layer
    minor: Unable to create file
  #004: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5VLnative_dataset.c line 269 in H5VL__native_dataset_create(): unable to create dataset
    major: Dataset
    minor: Unable to initialize object
  #005: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5Dint.c line 349 in H5D__create_named(): unable to create and link to dataset
    major: Dataset
    minor: Unable to initialize object
  #006: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5Lint.c line 516 in H5L_link_object(): unable to create new link to object
    major: Links
    minor: Unable to initialize object
  #007: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5Lint.c line 759 in H5L__create_real(): can't insert link
    major: Links
    minor: Unable to insert object
  #008: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5Gtraverse.c line 836 in H5G_traverse(): internal path traversal failed
    major: Symbol table
    minor: Object not found
  #009: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5Gtraverse.c line 613 in H5G__traverse_real(): traversal operator failed
    major: Symbol table
    minor: Callback failed
  #010: /tmp/ialberto/spack-stage/spack-stage-hdf5-1.14.0-6nwv5jrxawmg6fqv7z35kbhocugdu47y/spack-src/src/H5Lint.c line 551 in H5L__link_cb(): name already exists
    major: Links
    minor: Object already exists
srun: error: nid00008: task 0: Segmentation fault (core dumped)
srun: launch/slurm: _step_signal: Terminating StepId=47422047.2
```

src/CMakeLists.txt
albestro: Actually, if we want to go for using `hdf5` shared libraries we should add it as requirement in `dla-future` spack package, because currently it is not enforced (by default in spack it is `hdf5+shared`).

About preference, not really, I just thought that not having them statically linked might have been a good idea. Do you have any suggestion about it?

src/init.cpp
albestro: Just opened #913. Let's set the discussion there, merge it and then we can use it here 😉

include/dlaf/tune.h
albestro: Just created #918

include/dlaf/matrix/hdf5.h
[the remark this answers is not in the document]
msimberg: Awesome, thanks! Those are not perfect, but better than I expected. I've seen some other applications only report:
```
srun: error: nid00008: task 0: Segmentation fault (core dumped)
```
in similar cases which is why I asked. This is far better than that.

src/CMakeLists.txt
msimberg: Not really. I'd expect both to work the same, but supporting both may need some additional CMake logic (to detect if the static or shared or both libraries are available)? How about we enforce `depends_on("hdf +shared", when="...")` for now with a comment since I assume the HDF5 option is not going to upstream spack (yet...)?

albestro: Fully agree. See https://github.com/eth-cscs/DLA-Future/pull/880/commits/b57d53a1a5f726e1d906af0a3b63af8328dd15c1

include/dlaf/eigensolver/eigensolver/impl.h
msimberg: This is now implemented, right? (In a more elaborate form since you both the environment variable and the command line option.)

msimberg (approved):

msimberg:
cscs-ci run

rasolca (approved):

include/dlaf/eigensolver/eigensolver/impl.h
msimberg: Right, I'm confusing myself. All good.

PR #882, opened 2023-05-23
Disable extension check for hidden files

rasolca:
Needed for #877 to have a different .clang-format for unit tests.

Extra: re-enable extension check in the miniapp directory (disabled by mistake in #764).


msimberg (approved):
Looks good to me. However, what's the use case for having a separate `.clang-format` file for tests?

Issue #865, opened 2023-05-16
Documentation for release 0.1


rasolca:
Review and publish the documentation for release 0.1

Part of #630

PR #884, opened 2023-05-25
Automatically upload master doxygen documentation to github pages.

rasolca:
The previous `docs` action now runs only on pull request.
For pushes in the `master` branch a new action that also pushes the documentation in the `docs`  branch (which is used by github pages).

See https://eth-cscs.github.io/DLA-Future/master/.

Part of #865.

### DIFF
--- a/.github/workflows/doc-push.yml
+++ b/.github/workflows/doc-push.yml
@@ -1,9 +1,6 @@
 name: Docs
 
 on:
-  pull_request:
-    branches:
-      - master
   push:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


msimberg (approved):
I'm happy with this version. A potential refactoring would be to combine the workflows into one and make the fetching and pushing steps conditional on the branch (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif).

RMeli (approved):

albestro (approved):

PR #888, opened 2023-05-30
Test that headers are self-contained

### LINKED ISSUE #883: add separate compilation-only tests to check for header self-containment
As per suggestion in #877, we should introduce an automatic check for header self-containment.

msimberg:
Fixes #883.

Adds a `DLAF_BUILD_TESTING_HEADER` CMake option (off by default) which adds object library targets which build a cpp file for each header file in `include`. This option is enabled under the `ci-test` spack variant. This also adds a top-level pseudotarget `test_header` (should this be `header_test` or something else?) which depends on all the individual header tests.

Note, this uses `file(GLOB)` which means that headers that are added to the `include` directory will not be picked up without an explicit CMake reconfiguration. For CI this is not a problem, but may be annoying for local development. The alternative is to explicitly list all header files though, which I find more error prone, especially since all headers will still be checked in CI.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

albestro (approved):

rasolca (approved):

PR #890, opened 2023-05-30
Ignore Spack builds

RMeli:


RMeli:
Used `develop` instead of `master`...

PR #894, opened 2023-06-05
Remove spack.cloud cache key and mirror

msimberg:
Since the mirror isn't update and doesn't give any cache hits I'm removing it from the dockerfile.


msimberg:
cscs-ci run

rasolca (approved):

Issue #868, opened 2023-05-16
Create submatrix view (reference and pipelined)


rasolca:
The following is needed:

[ ] tile sub-pipeline
[ ] create distribution for submatrix
[ ] tests.

more details: https://confluence.cscs.ch/x/hYBZHw

msimberg:
@albestro identified a potentially high impact use case for the sub-pipelined sub-matrix in in the tridiagonal eigensolver GEMM optimization work. He currently gets a sub-pipelined full matrix for tasks and then in many cases uses only a small part of the full matrix. On one hand the unused tiles in the sub-pipeline get released as soon as possible and may only have a small impact of scheduling. On the other hand there may be a lot of unused sub-pipelines which all add dynamic allocations.

rasolca:
There will be no impact after the clean-up, as only the relevant submatrix will be passed to `merge`.

PR #901, opened 2023-06-08
Allow incomplete first tile/block in `Distribution`

msimberg:
Part of #868.

This adds support for having incomplete first tiles in `Distribution`. This is introduced through a global offset in terms of number of elements. The size of the distribution is relative to the offset.

The offset is allowed to be arbitrarily large (i.e. larger than the block size). If the offset is larger than the block size I normalize it to be smaller than the block size and instead shift the source rank to account for the larger-than-block-size offset.

At the moment I've simply added an optional last `offset` parameter to all the `Distribution` constructors to not change all the callsites drastically. I'll change/add constructors when I know what we want them to look like. Do we want to:
- [x] Have the offset somewhere else in the parameter list? (I think last is ok, especially since it allows it to be defaulted; most cases will not use a non-zero offset) _Leave as is._
- [x] Make the offset compulsory? (I think not, see above) _Keep it optional._
- [x] Add other constructors: specify offset in terms of number of tiles? _May be useful._
- [x] Specify offset in terms of number of tiles and elements within the first tile? _May be useful._
- [x] Specify offset in terms of blocks? (this is the same as shifting the source rank but slightly more convenient since the caller doesn't have to wrap the source rank; I don't know if there's a use case for this though) _Not necessary for now._
- [x] Specify offset in terms of blocks and elements within the first tile? _Not necessary for now._
- [x] Specify offset in terms of `GlobalElementIndex` or `GlobalElementSize`? (I chose the former since it seems more appropriate to say "the matrix starts at this element index" rather than "the matrix has an empty whole in the beginning with size offset")
- [x] Add `LocalElementIndex` for the local constructor? A `LocalElementIndex` typedef is conspicuously missing from https://github.com/eth-cscs/DLA-Future/blob/0d88dfb4e85856a7f03651740cb742bffd7e2865/include/dlaf/matrix/index.h#L30. _Do not add, start migrating away from "local" indices/sizes_.
- [x] Add a "copy constructor" that takes an existing `Distribution` and a "sub matrix spec" to produce a new `Distribution`? (we'll need this in some form, the question is if it should be a free function, constructor, or member function) _Not needed in this PR, will discuss separately._

Some TODOs are still in the code. I will deal with those once the general picture looks ok.

This does not add tests for `block_size != tile_size`, but some should probably be added to `test_distribution`. `test_retiled_matrix` covers some of it, but it would be good to have more thorough tests explicitly on `Distribution` itself.

This does not and will not change `Matrix` or other functionality outside of `Distribution`.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
I added constructors that take an element-wise _and_ tile-wise offset (and some tests to go with it). I also cleaned up some of the TODOs. This is ready for reviews.

msimberg:
cscs-ci run

include/dlaf/matrix/distribution.h
```diff
@@ -85,6 +130,10 @@ class Distribution {
     return !operator==(rhs);
   }
 
+  const GlobalElementIndex& offset() const noexcept {
```
rasolca, line 133:
```suggestion
  const TileElementIndex& offset() const noexcept {
```
If I remember correctly our discussion, from any combination of offsets we set the source rank index such that the offset is in a single tile.

include/dlaf/matrix/distribution.h
```diff
@@ -388,22 +452,15 @@ class Distribution {
...
-    SizeType l = local_size_.get<rc>();
-    SizeType nb = tile_size_.get<rc>();
-    SizeType nbr = std::min(nb, l - i_loc_last * nb);  // size of last local tile along `rc`
-    return (i_loc_last - i_loc_begin) * nb + nbr;
+    SizeType i_loc_end = nextLocalTileFromGlobalTile<rc>(i_end);
+    return localElementDistanceFromLocalTile<rc>(i_loc_begin, i_loc_end);
   }
 
-  /// \overlap localElementDistanceFromGlobalTile
+  /// \overload localElementDistanceFromGlobalTile
```
rasolca, line 461: overload simply group the function under a single doc (the one without overload)

Please check the output.

include/dlaf/matrix/distribution.h
msimberg: This may be what you meant, but just to avoid confusion I want to be precise. Currently the offset is normalized to be an offset within the first _block_ which as you know is not necessarily the same as tile offset. I do this because the source rank index and distribution in general only concerns itself with blocks, not tiles. So technically, it should be
```suggestion
  const BlockElementIndex& offset() const noexcept {
```
Do we want to add an alias like that? Or should we treat `TileElementIndex` as something that can be used to index both into tiles and blocks? Alternatively, if `offset` is to return a `TileElementIndex` it should be renamed to `tileElementOffset` and we need a separate `(Global?)TileIndex tileOffset` member.

include/dlaf/matrix/distribution.h
```diff
@@ -31,30 +31,74 @@ class Distribution {
...
+  ///            of the matrix, used in addition to @p tile_offset,
+  /// @pre size.isValid(),
+  /// @pre !tile_size.isEmpty(),
+  /// @pre !grid_size.isEmpty(),
+  /// @pre rank_index.isIn(grid_size),
+  /// @pre source_rank_index.isIn(grid_size).
+  Distribution(const GlobalElementSize& size, const TileElementSize& block_size,
+               const comm::Size2D& grid_size, const comm::Index2D& rank_index,
+               const comm::Index2D& source_rank_index, const GlobalTileIndex& tile_offset,
+               const GlobalElementIndex& element_offset = {0, 0});
```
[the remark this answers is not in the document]
msimberg: Maybe, see https://github.com/eth-cscs/DLA-Future/pull/901#discussion_r1251791171. The offset here is also allowed to be larger than a tile/block offset and it's trivially handled, but I don't know if it's a useful thing to allow.

include/dlaf/matrix/distribution.h
msimberg: Indeed, doing
```
/// \overload ...
LocalElementSize localElementDistanceFromGlobalTile(...)
```
gives an entry in brief summary at the top but no detailed entry further down. The function name is not a link.

Doing 
```
/// some text
LocalElementSize localElementDistanceFromGlobalTile(...)
```
gives an entry in the summary at the top with `some text` as the description in the summary, but no entry further down. The function name is a link that goes to the brief summary.

Doing 
```
/// some text
///
/// some more text
LocalElementSize localElementDistanceFromGlobalTile(...)
```
gives an entry in the summary at the top with no description, but an entry further down with `some text` and `some more text` in the description. The function name is a link that goes to the detailed description further down.

I think we want option 2?

rasolca: Option 2 or 3 is ok. `\overload` should be used only to share the same documentation.

include/dlaf/matrix/distribution.h
albestro: From my naive point of view, I would suggest adding an alias even if it is just for a few places. IMO there is no reason for not adding an alias and ending up re-using a similar one that might create semantic confusion (not a suggestion for now, but for the next future).

albestro (approved):
After a quick chat with @rasolca I'm not fully sure that all Distribution c'tors should be public, but I'll leave this up for a future discussion 😉

include/dlaf/matrix/distribution.h
msimberg: See https://github.com/eth-cscs/DLA-Future/pull/901/commits/9479d377f028fa385e8f77c4f2015e7a0ebd2b52. I've gone with option 2.

msimberg:
> After a quick chat with @rasolca I'm not fully sure that all Distribution c'tors should be public, but I'll leave this up for a future discussion wink

No objections, but if that's the case which ones should be private and why?

albestro:
> > After a quick chat with @rasolca I'm not fully sure that all Distribution c'tors should be public, but I'll leave this up for a future discussion wink
> 
> No objections, but if that's the case which ones should be private and why?

I'm still not fully sure as as I said, but I see it strange that I can specify both a `tile_offset` and a `source_rank_index`, because to me they look like they should be strictly related.

msimberg:
cscs-ci run

rasolca (approved):

Issue #905, opened 2023-06-12
Change `RetiledMatrix` to `Matrix`


msimberg:
`RetiledMatrix` looks the same as `Matrix`, with the only difference being a `done` method on `RetiledMatrix`, and _should_ be possible to rename to `Matrix` (accessed through `Matrix::subPipeline(?)`?).

- Is it worth keeping a separate type for `done`?
- Can `Matrix` have a `done` method?

PR #908, opened 2023-06-16
Replace use of `shared_future` for taus by `Matrix`

msimberg:
Part of #468.

This replaces the use of `vector`s of `shared_future`s of `vector`s for taus in `{backtransformation,}ReductionToBand`. The 

Still to do:
- [x] check GPU compilation
- [x] check if taus need to be in column vector (instead of current row vector)
- [x] clean up generation of taus to be directly placed into `Matrix` instead of going via intermediate `vector`
- [x] check and add more assertions
- [x] lots of random cleanup (TODOs in code)
- [x] rerun benchmarks as a sanity check (`backTransformationreductionToBand` is unchanged, `reductionToBand` is slightly faster on this branch!)

Do we want an overload of `reductionToBand` that takes an existing `Matrix` for taus like #820? I'd leave it out if there's no immediate need for it, but if you already know it will be needed I can add it right away.


include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -1109,13 +1106,23 @@ common::internal::vector<pika::shared_future<common::internal::vector<T>>> Reduc
   // Reflector of size = 1 is not considered whatever T is (i.e. neither real nor complex)
   const SizeType nrefls = std::max<SizeType>(0, dist.size().rows() - band_size - 1);
 
-  common::internal::vector<pika::shared_future<common::internal::vector<T>>> taus;
+  // Row-vector that is distributed over columns, but exists locally on all rows of the grid
+  // TODO: transpose into column vector?
```
msimberg, line 1110: To do: change to column vector. Row vector has `ld = 64` between elements.

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -146,26 +143,28 @@ struct Helpers<Backend::GPU, Device::GPU, T> {
...
         whip::stream_t stream;
         DLAF_GPUBLAS_CHECK_ERROR(cublasGetStream(handle, &stream));
 
-        whip::memcpy_2d_async(tile_t.ptr(), to_sizet(tile_t.ld() + 1) * sizeof(T), taus.data(),
-                              sizeof(T), sizeof(T), to_sizet(k), whip::memcpy_default, stream);
+        // Assume taus is a column vector with column major layout, i.e. elements are contiguous
+        // TODO: Can this be asserted more directly?
+        DLAF_ASSERT(taus.size().cols() == 1, taus);
+        whip::memcpy_2d_async(tile_t.ptr(), to_sizet(tile_t.ld() + 1) * sizeof(T), taus(), sizeof(T),
+                              sizeof(T), to_sizet(k), whip::memcpy_default, stream);
```
msimberg, line 163: To do: the memory access needs to be checked.

include/dlaf/matrix/matrix.tpp
```diff
@@ -41,7 +41,7 @@ Matrix<T, D>::Matrix(Distribution distribution, const LayoutInfo& layout) noexce
   DLAF_ASSERT(this->distribution().localSize() == layout.size(),
               "Size of distribution does not match layout size!", distribution.localSize(),
               layout.size());
-  DLAF_ASSERT(this->distribution().blockSize() == layout.blockSize(), distribution.blockSize(),
+  DLAF_ASSERT(this->distribution().baseTileSize() == layout.blockSize(), distribution.baseTileSize(),
```
msimberg, line 44: Revert and do a separate PR.

test/unit/eigensolver/test_reduction_to_band.cpp
```diff
@@ -188,74 +191,66 @@ void splitReflectorsAndBand(MatrixLocal<const T>& mat_v, MatrixLocal<T>& mat_b,
...
+      for (SizeType k = 0; k < chunk_size; ++k) {
+        taus.push_back(tile_local(TileElementIndex(0, k)));
+      }
     }
 
     // copy each chunk contiguously
-    std::copy(chunk_data.begin(), chunk_data.end(), std::back_inserter(taus));
   }
 
+  // TODO: is the early exit required?
```
msimberg, line 244: @albestro, @rasolca there was an early exit here (https://github.com/eth-cscs/DLA-Future/blob/34934c60c5cc6f0c1b71fc5e2148e2c89b6903d1/test/unit/eigensolver/test_reduction_to_band.cpp#L225-L232 on master). The comment says "This is just and early exit" but it's not quite clear to me what the purpose of the early exit is. Is it optional? If the condition isn't true it seems like the whole test should fail. Should it be an `ASSERT*` or `TEST*` instead?

msimberg:
cscs-ci run

include/dlaf/eigensolver/bt_reduction_to_band/impl.h
```diff
@@ -317,12 +315,11 @@ void BackTransformationReductionToBand<B, D, T>::call(
         }
       }
 
+      const GlobalTileIndex taus_index{Coord::Col, k};
```
msimberg, line 318: @albestro, @aurianer, @rasolca, @RMeli, I've used the `Coord::X` constructor version in some places for `taus` indices, but I haven't been consistent with them yet (other places use e.g. `*Index{0, k}`. Do you prefer the `Coord` version generally? Eventually we still need some sort of `Vector` to abstract away this use case, but I will not add that in this PR.

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -146,26 +143,28 @@ struct Helpers<Backend::GPU, Device::GPU, T> {
...
+      DLAF_ASSERT(taus.size().rows() == k, taus.size().rows(), k);
 
       if (first_row_tile == 0) {
         whip::stream_t stream;
         DLAF_GPUBLAS_CHECK_ERROR(cublasGetStream(handle, &stream));
 
-        whip::memcpy_2d_async(tile_t.ptr(), to_sizet(tile_t.ld() + 1) * sizeof(T), taus.data(),
-                              sizeof(T), sizeof(T), to_sizet(k), whip::memcpy_default, stream);
+        // Assume taus is a column vector with column major layout, i.e. elements are contiguous
+        // TODO: Can this be asserted more directly?
```
msimberg, line 160: Am I missing some handy way to assert that a tile is column-major and has only one column? As far as I can tell we implicitly assume that all matrices are column major.

include/dlaf/factorization/qr/t_factor_impl.h
msimberg: This has been updated (taus is in a column vector now).

include/dlaf/eigensolver/reduction_to_band/impl.h
msimberg: Changed to a column vector.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

include/dlaf/matrix/matrix.tpp
msimberg: This has been reverted, but leaving open for now as a reminder to open a new PR with the changes.

miniapp/miniapp_reduction_to_band.cpp
```diff
@@ -111,6 +111,7 @@ struct reductionToBandMiniapp {
         matrix_host.waitLocalTiles();
         DLAF_MPI_CHECK_ERROR(MPI_Barrier(world));
 
+        // TODO: hangs?
```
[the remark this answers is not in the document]
msimberg: This is hanging "just" for the regular problem of doing synchronous communication on the default pool (at least I think so, this hangs also on master, with 2 threads per rank).

msimberg:
There are a few open TODOs in comments/code, but I'm opening this up for reviews as it's functionally done and passing tests.

I changed the vector in the end to be a column vector to make it contiguous. It's not 100% required, but it does simplify some of the copying to/from the vector (tiles), so it's probably better that way. The downside is that the direction of the vector is a bit unintuitive. However, if we eventually refactor things so that we have a dedicated 1D distributed vector the direction won't matter so much anymore (except for the distribution of course).

In the tests I left some of the generation/checking of taus to copy from/to `vector`s since it looked like the code was simpler that way. If someone who understands the code better thinks that we can easily and cleanly skip the use of `vector` please let me know and I can look into refactoring that.

include/dlaf/matrix/matrix.tpp
msimberg: https://github.com/eth-cscs/DLA-Future/pull/909

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -1109,13 +1104,23 @@ common::internal::vector<pika::shared_future<common::internal::vector<T>>> Reduc
...
+                                                       comm::Index2D(0, mat_a.sourceRankIndex().col())));
 
   if (nrefls == 0)
-    return taus;
+    return mat_taus;
 
-  const SizeType nblocks = (nrefls - 1) / band_size + 1;
-  taus.reserve(nblocks);
+  matrix::RetiledMatrix<T, Device::CPU> mat_taus_retiled(
+      mat_taus, LocalTileSize(1, mat_a.blockSize().cols() / band_size));
```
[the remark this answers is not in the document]
msimberg: This could become easier once #905 is done. It would allow us to use a `Matrix<...>&` to either the original unretiled matrix or to the retiled matrix since the types would be the same.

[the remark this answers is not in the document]
msimberg: #910

test/unit/eigensolver/test_reduction_to_band.cpp
rasolca: Not sure. @albestro?
Anyway, is this routine still needed? Can you replace it with one of the matrix allgather methods available in `dlaf_test` or it doesn't work as the distribution is "hand made"?

include/dlaf/factorization/qr/t_factor_impl.h
rasolca: Tiles are column major. In this case you only need to check that the size is (k x 1).

include/dlaf/eigensolver/bt_reduction_to_band.h
```diff
@@ -65,15 +65,14 @@ void backTransformationReductionToBand(
 /// @param mat_c contains the (m x n) matrix C (blocksize (mb x nb)), while on exit it contains Q C.
 /// @param mat_v is (m x m) matrix with blocksize (mb x mb), which contains the Householder reflectors.
 /// The j-th HH reflector is v_j = (1, V(mb + j : n, j)).
-/// @param taus is a (blocked) vector of size m (blocksize mb). The j-th element is the scaling factor
+/// @param mat_taus is a column vector of size m (blocksize mb). The j-th element is the scaling factor
```
rasolca, line 68: As above.

include/dlaf/eigensolver/reduction_to_band.h
```diff
@@ -72,14 +28,16 @@ auto groupTausFromBandsToTiles(
 /// band-diagonal result together with the elementary reflectors. Just the tiles of the lower
 /// triangular part will be used.
 /// @param band_size size of the band of the resulting matrix (main diagonal + band_size sub-diagonals)
+/// @return a column vector of size max(0, mat_a.size().rows() - band_size - 1)
+/// (block size mat_a.blockSize.rows()) containing taus
```
rasolca, line 32: similarly
```suggestion
/// @return the tau vector as needed by backtransformationReductionToBand.
```

include/dlaf/eigensolver/reduction_to_band.h
```diff
@@ -72,14 +28,16 @@ auto groupTausFromBandsToTiles(
...
 ///
 /// @pre mat_a has a square size
 /// @pre mat_a has a square block size
 /// @pre mat_a is a local matrix
 /// @pre mat_a.blockSize().rows() % band_size == 0
 template <Backend B, Device D, class T>
-common::internal::vector<pika::shared_future<common::internal::vector<T>>> reductionToBand(
-    Matrix<T, D>& mat_a, const SizeType band_size) {
+// TODO: Pass in reference to taus matrix?
```
rasolca, line 39: Not sure what is the advantage.

On the other hand a couple of disadvantage:
- the (advanced) user has to allocate tau with the correct distribution and size
- we have to check that tau has correct shape and distribution
- the documentation can be simple (especially with the current distribution workarounds)

include/dlaf/eigensolver/reduction_to_band.h
```diff
@@ -138,23 +95,23 @@ v v v v * *
 /// band-diagonal result together with the elementary reflectors as described above. Just the tiles of
 /// the lower triangular part will be used.
 /// @param band_size size of the band of the resulting matrix (main diagonal + band_size sub-diagonals)
+/// @return a column vector of size max(0, mat_a.size().rows() - band_size - 1)
+/// (block size mat_a.blockSize.rows()) containing taus
```
rasolca, line 99: As above.

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -292,55 +293,53 @@ auto computePanelReflectors(MatrixLike& mat_a, const matrix::SubPanelView& panel
...
-                    }
-                  }) |
-         ex::then([](auto barrier_ptr, auto w, auto taus, auto tiles) {
-           di::silenceUnusedWarningFor(barrier_ptr, w, tiles);
-           return taus;
-         }) |
-         ex::make_future();
+  auto s =
+      ex::when_all(ex::just(std::make_unique<pika::barrier<>>(nthreads),
+                            std::vector<common::internal::vector<T>>{}),  // w (internally required)
```
rasolca, line 298: `when_all(just(x), just(y), ...)` vs `when_all(just(x, y), ...)`:
Is just a matter of style, or is more efficient?

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -939,10 +931,9 @@ struct ComputePanelHelper<Backend::GPU, Device::GPU, T> {
 }
 
 /// Local implementation of reduction to band
-/// @return a vector of shared futures of vectors, where each inner vector contains a block of taus
+/// @return a column vector of taus with block size mat_a.blockSize().cols()
```
rasolca, line 934: Doxygen documentation should not be here.

At most I would keep the comment
`// Local implementation of reduction to band`

include/dlaf/factorization/qr/api.h
```diff
@@ -43,15 +42,15 @@ struct QR_Tfactor {
   /// @param k the number of elementary reflectors to use (from the beginning of the tile)
   /// @param v where the elementary reflectors are stored
   /// @param v_start tile in @p v where the column of reflectors starts
-  /// @param taus array of taus, associated with the related elementary reflector
+  /// @param taus column vector of taus, associated with the related elementary reflector
```
rasolca, line 45:
```suggestion
  /// @param taus tile of the taus vector, associated with the related elementary reflector
```

test/unit/communication/test_comm_sender.cpp
```diff
@@ -64,9 +63,8 @@ TEST(Bcast, Polling) {
   double val = (comm.rank() == root_rank) ? 4.2 : 1.2;
   std::vector<double> buf(static_cast<std::size_t>(size), val);
 
-  sync_wait(when_all(just(buf.data()), pika::make_ready_future<int>(size), just(dtype, root_rank, comm),
-                     pika::make_ready_future<void>()) |
-            transformMPI(MPI_Ibcast) | then([](int e) { DLAF_MPI_CHECK_ERROR(e); }));
+  sync_wait(when_all(just(buf.data(), size, dtype, root_rank, comm)) | transformMPI(MPI_Ibcast) |
```
rasolca, line 66: Is the `when_all` needed?

rasolca:
> In the tests I left some of the generation/checking of taus to copy from/to `vector`s since it looked like the code was simpler that way. If someone who understands the code better thinks that we can easily and cleanly skip the use of `vector` please let me know and I can look into refactoring that.

For now I would keep it as it is. We can simplify it later with the Vector type.

albestro (approved):
The only "bad" of this PR is the fact that not having the 1D Distributed `Vector`, you had to spread all over the code this "counter-intuitive" transposed usage of the Matrix with the ad-hoc distribution.

I'm confident that we will be able to "fix" these parts when we will have the new feature. Probably worth keeping track of this PR, so that it will be easier to spot where it can be introduced.

include/dlaf/eigensolver/bt_reduction_to_band.h
```diff
@@ -31,14 +31,14 @@ namespace eigensolver {
...
```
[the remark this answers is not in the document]
msimberg: https://github.com/eth-cscs/DLA-Future/pull/908/commits/5eb06b27420fee8e6aea5e076c129ff814a6bbf6

include/dlaf/eigensolver/reduction_to_band.h
msimberg: https://github.com/eth-cscs/DLA-Future/pull/908/commits/5eb06b27420fee8e6aea5e076c129ff814a6bbf6

include/dlaf/eigensolver/bt_reduction_to_band.h
msimberg: https://github.com/eth-cscs/DLA-Future/pull/908/commits/5eb06b27420fee8e6aea5e076c129ff814a6bbf6

include/dlaf/eigensolver/reduction_to_band.h
msimberg: https://github.com/eth-cscs/DLA-Future/pull/908/commits/5eb06b27420fee8e6aea5e076c129ff814a6bbf6

include/dlaf/eigensolver/reduction_to_band.h
msimberg: I don't have a use case for it. I was wondering if there is a use case for it since it's similar to the eigensolvers having overloads taking the eigenvalues/vectors and didn't want to forget to ask about it. I'm happy not to add it and I've removed the TODO in https://github.com/eth-cscs/DLA-Future/pull/908/commits/676ce6d0528d83e4a2df45010e62a0e1a9900df0.

include/dlaf/eigensolver/reduction_to_band/impl.h
msimberg: Mostly style. I wouldn't object if you prefer the former stylistically. There's in theory a small difference in instantiations, function calls, and operation states created when using separate `just`s, but I wouldn't expect it to have a measurable impact at least at runtime. I have not measured it.

test/unit/communication/test_comm_sender.cpp
msimberg: Nope, good catch. Removed in https://github.com/eth-cscs/DLA-Future/pull/908/commits/bda63508ee2127713df21ce0888553f054f414e6.

include/dlaf/factorization/qr/api.h
msimberg: https://github.com/eth-cscs/DLA-Future/pull/908/commits/5a3c9331068ddcc5e1916dddb722a1e6aebda1d1

include/dlaf/factorization/qr/api.h
```diff
@@ -71,7 +70,7 @@ struct QR_Tfactor {
...
```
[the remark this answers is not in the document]
msimberg: https://github.com/eth-cscs/DLA-Future/pull/908/commits/5a3c9331068ddcc5e1916dddb722a1e6aebda1d1

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -1078,14 +1073,14 @@ common::internal::vector<pika::shared_future<common::internal::vector<T>>> Reduc
     v.reset();
   }
 
-  return taus;
+  return mat_taus;
 }
 
 /// Distributed implementation of reduction to band
-/// @return a vector of shared futures of vectors, where each inner vector contains a block of taus
+/// @return a column vector of taus with block size mat_a.blockSize().cols()
```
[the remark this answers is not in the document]
msimberg: https://github.com/eth-cscs/DLA-Future/pull/908/commits/4322a278222a47da74af0cd9737df40c32bde8b4

include/dlaf/eigensolver/reduction_to_band/impl.h
msimberg: https://github.com/eth-cscs/DLA-Future/pull/908/commits/4322a278222a47da74af0cd9737df40c32bde8b4

include/dlaf/factorization/qr/t_factor_impl.h
msimberg: Updated in https://github.com/eth-cscs/DLA-Future/pull/908/commits/61e43b0cb08a04ac7be968a6f3653c3aa0be6881. I left the comment and moved the assertion up to the top of the function along with the other assertions.

test/unit/eigensolver/test_reduction_to_band.cpp
msimberg: I hadn't thought about using `allGather`. In the distributed case it fails on https://github.com/eth-cscs/DLA-Future/blob/2634d5bde89efc80b18b76b4afaeefb65618a1e6/test/include/dlaf_test/matrix/util_matrix_local.h#L104, understandably. Am I missing some other helper that you think could help? If not, I would probably leave this as it is given that the plan is to clean this up once we have a distributed vector. What do you think?

msimberg:
cscs-ci run

include/dlaf/eigensolver/reduction_to_band/impl.h
rasolca: No strong preference on my side.

test/unit/eigensolver/test_reduction_to_band.cpp
rasolca: Back to the original question:
You can use either `DLAF_ASSERT` or `EXPECT_EQ`.
`ASSERT_EQ` cannot be used in non-void functions.

rasolca (approved):
Note: CI failed due to a config problem #927.
No need to rerun it.

test/unit/eigensolver/test_reduction_to_band.cpp
msimberg: I added back a check (as `EXPECT_EQ`) in https://github.com/eth-cscs/DLA-Future/pull/908/commits/f856457f4dba1ec0046fb704f703096eadc9248f for the distributed case. Not sure it adds much, but probably doesn't hurt for now. We can always remove it if seems superfluous or annoying in the future.

msimberg:
cscs-ci run

msimberg:
I only added an extra test check in the last commit. I guess this is ready to go unless there are further comments.

msimberg:
cscs-ci run

rasolca (approved):

test/unit/eigensolver/test_reduction_to_band.cpp
albestro: I admit my comment is not helpful at all.

Not sure why this check is there...after a quick view I was not able to fully reconstruct the history of this change. It might be either an additional check I added during development or a leftover. It is something marginal, so I would not bother too much. Anyway, your adaptation looks good 😉

albestro (approved):

PR #909, opened 2023-06-21
Support different tile and block size in `Matrix`

msimberg:
Moving over the changes from https://github.com/eth-cscs/DLA-Future/pull/908#discussion_r1233952427 to a separate PR.

~Do we want to rename `LayoutInfo::blockSize` to `LayoutInfo::tileSize`? `tileSize` can be equal to `blockSize` but more generally `LayoutInfo` is concerned with the tile size.~ Edit: I haven't done any renamings like this because in some cases we call things like `square_blocksize` and for that we want everything to have a `blockSize` method. I'm punting this down the road.

### DIFF
--- a/include/dlaf/matrix/distribution.h
+++ b/include/dlaf/matrix/distribution.h
@@ -44,7 +44,7 @@ public:
   /// @param[in] source_rank_index is the rank of the process which contains the top left tile of the matrix,
   /// @param[in] element_offset is the element-wise offset of the top left tile of the matrix ,
   /// @pre size.isValid(),
-  /// @pre !tile_size.isEmpty(),
+  /// @pre !block_size.isEmpty(),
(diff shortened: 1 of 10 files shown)


msimberg:
cscs-ci run

msimberg:
Waiting for #898 before I continue here since these will both touch `test_matrix.cpp`.

msimberg:
cscs-ci run

msimberg:
This so far makes some minimal relaxations to allow constructing matrices with different tile size and block size. I've added some tests, but coverage could likely be better. I'm currently only exposing this functionality through the `Matrix(Distribution)` constructor, i.e. I haven't added any "custom" constructors that allow passing a tile and block size directly (the existing constructors which take a block size are still there).

rasolca (approved):

PR #912, opened 2023-06-23
TridiagSolver: number of bulk workers for rank1 solver

albestro:
TridiagSolver computes the solution by working on sub-problems that varies a lot in terms of sizes. For this reason, rank1 problem solution now adopts, for both local and distributed implementation, an "auto-limiting" policy in terms of workers used per problem.

Smaller problems will have less workers, allowing to solve more of them concurrently, and bigger problems will be assigned more workers, so that the computation will be carried out faster.

In addition, this PR also changes the default number of workers used: all threads in default pool are now considered for being used in the rank1 bulk. In particular, no thread is left out as previously done for others, when for "staying on the safe side" we opted for leaving out a thread.

- [x] #904

### DIFF
--- a/include/dlaf/tune.h
+++ b/include/dlaf/tune.h
@@ -60,8 +60,7 @@ struct TuneParameters {
   std::size_t red2band_panel_nworkers =
       std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count() / 2);
   std::size_t red2band_barrier_busy_wait_us = 1000;
-  std::size_t tridiag_rank1_nworkers =
-      std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count());
+  std::size_t tridiag_rank1_nworkers = pika::resource::get_thread_pool("default").get_os_thread_count();
   std::size_t tridiag_rank1_barrier_busy_wait_us = 0;
 
   SizeType eigensolver_min_band = 100;
--- a/include/dlaf/tune.h
+++ b/include/dlaf/tune.h
@@ -61,7 +61,7 @@ struct TuneParameters {
       std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count() / 2);
   std::size_t red2band_barrier_busy_wait_us = 1000;
   std::size_t tridiag_rank1_nworkers =
-      std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count() / 2);
+      std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count());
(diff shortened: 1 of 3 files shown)


include/dlaf/tune.h
```diff
@@ -53,7 +53,7 @@ struct TuneParameters {
       std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count() / 2);
   std::size_t red2band_barrier_busy_wait_us = 1000;
   std::size_t tridiag_rank1_nworkers =
-      std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count() / 2);
+      std::max<std::size_t>(1, pika::resource::get_thread_pool("default").get_os_thread_count());
```
msimberg, line 56: The `std::max` shouldn't be needed anymore, right? It was just there in case `num_threads` is `1` and `num_threads / 2` is `0`.

msimberg:
I have nothing against these changes but I'm wondering if you can expand on why you've changed these now? "It doesn't make a difference for performance and this is simpler" is perfectly acceptable :smile: Even better if these defaults perform better. Also, where does the 2 columns per thread come from?

albestro:
> I have nothing against these changes but I'm wondering if you can expand on why you've changed these now? "It doesn't make a difference for performance and this is simpler" is perfectly acceptable 😄 Even better if these defaults perform better. Also, where does the 2 columns per thread come from?

This is just a slightly more thought default than the previous no-brainer one. It is the result of quick considerations we had with @rasolca (together with potential work balance improvement over threads with different work distribution). For instance, the 2 columns per thread decision is a quick way for saying "enough work", but it is not really backed by any data observation.

We will have to benchmark this parameter in order to get a rough idea about if/how it affects performances. If we feel like it is the right time, I can do it and report here before merging this PR 😉

msimberg:
> This is just a slightly more thought default than the previous no-brainer one. It is the result of quick considerations we had with @rasolca (together with potential work balance improvement over threads with different work distribution). For instance, the 2 columns per thread decision is a quick way for saying "enough work", but it is not really backed by any data observation.
> 
> We will have to benchmark this parameter in order to get a rough idea about if/how it affects performances. If we feel like it is the right time, I can do it and report here before merging this PR wink

Perfect, thanks! This is enough for me (and now the reasoning is on record for our future selves).

include/dlaf/tune.h
albestro: See https://github.com/eth-cscs/DLA-
Future/pull/912/commits/895116962e314081234c7d8826d24067a31b3961

msimberg (approved):

rasolca (approved):

PR #913, opened 2023-06-23
minor: introduce parseFromString for bools 

albestro:
Following discussion in https://github.com/eth-cscs/DLA-Future/pull/880#discussion_r1221368140, this PR introduces `parseFromString` for `bool` types.

| option |  b | d | e |
| -- |  -- | -- | --  |
| unset DEBUG |  default  | default | default |
| export DEBUG |  false | true | default |
| export DEBUG=TRUE |  true | true | true |
| export DEBUG=FALSE  | false | false | false |
| export DEBUG=*  | ? | ? | default | 

Currently I went for the last option suggested E, but now the change between different solution B, D, E should be quite easy.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -8,6 +8,9 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 
+#include <algorithm>
+#include <array>
+#include <cctype>
 #include <cstdlib>
 #include <iostream>
 #include <memory>
@@ -121,6 +124,24 @@ struct parseFromString<SizeType> {
   static std::optional<SizeType> call(const std::string& var) {
     return std::stoll(var);
   };
+};
+
+template <>
+struct parseFromString<bool> {
+  static std::optional<std::size_t> call(const std::string& var) {
+    if (is_one_of_ignore_case(var, {"ON", "TRUE", "YES", "1"}))
+      return true;
+    if (is_one_of_ignore_case(var, {"OFF", "FALSE", "NO", "0"}))
+      return false;
+    return std::nullopt;
+  };
+
+private:
+  static bool is_one_of_ignore_case(std::string value, const std::array<std::string, 4>& values) {
+    std::transform(value.cbegin(), value.cend(), value.begin(),
+                   [](unsigned char c) { return std::toupper(c); });
+    return (values.cend()) != std::find(values.cbegin(), values.cend(), value);
+  }
 };
 
 template <class T>
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -128,7 +128,7 @@ struct parseFromString<SizeType> {
 
 template <>
 struct parseFromString<bool> {
-  static std::optional<std::size_t> call(const std::string& var) {
+  static std::optional<bool> call(const std::string& var) {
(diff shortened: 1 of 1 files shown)


src/init.cpp
```diff
@@ -135,7 +157,8 @@ void updateConfigurationValue(const pika::program_options::variables_map& vm, T&
   const std::string dlaf_env_var = "DLAF_" + env_var;
   char* env_var_value = std::getenv(dlaf_env_var.c_str());
   if (env_var_value) {
-    var = parseFromString<T>::call(env_var_value);
+    if (auto parsed_value = parseFromString<T>::call(env_var_value))
+      var = parsed_value.value();
```
albestro, line 161: We can decide to add a warning to the user here in the `else` branch.

rasolca: I would add the warning if the variable exists but an incorrect value is set.

albestro: Yes, exactly.

The first `if` checks variable existence. If it exists, we try to parse it with `parseFromString<T>` function that returns an `optional`:

- if it contains a value, it is because we were able to parse it (and understand it)
- otherwise (the suggested `else` branch) we warn the user that something went wrong.

rasolca: What I was suggesting is to already add the else branch with the warning.

albestro: See https://github.com/eth-cscs/DLA-Future/pull/913/commits/f87cc0c38f155346e4a7fd1f43bfbef5d5eb97b0

albestro:
Even if optional has become the return type for all type parsers (trivially), in this PR just the `bool` parser convey parsing failure information via optional. Indeed, others still throw an exception (e.g. `std::invalid_argument`, `std::out_of_range`).

src/init.cpp
```diff
@@ -135,7 +157,11 @@ void updateConfigurationValue(const pika::program_options::variables_map& vm, T&
...
+    else
+      std::cerr << "Environment variable " << dlaf_env_var << " has an invalid value (='"
+                << env_var_value << "').\n";
```
msimberg, line 164: How about we terminate instead? It's what I presume would happen soon after if e.g. `parseFromString<std::size_t>` throws because of an invalid value. It's the fastest way to get users(/us) to fix wrongly set environment variables instead of quietly continuing.

albestro: see https://github.com/eth-cscs/DLA-Future/pull/913/commits/589e5ac478b63ae1eaaf313552fc598d4016eeb2

msimberg: Thanks @albestro, looks good to me. In the future we might also want to say what are actually valid values, instead of just saying that it's invalid, but let's not overcomplicate things now.

msimberg (approved):

albestro:
cscs-ci run

rasolca (approved):
Looks good. Is this the new behaviour?

| variable | effect |
| -- |  -- |
| unset DEBUG | default |
| export DEBUG | terminate |
| export DEBUG=TRUE | true |
| export DEBUG=FALSE  | false |
| export DEBUG=* | terminate |

albestro:
Yes, good point. Thanks for pointing that out, so it remains documented 😉

PR #920, opened 2023-07-02
fetch tokens from middleware

finkandreas:
Soon static tokens will be removed, and tokens must be fetched explicitly when needed from the middleware. These tokens will be valid only for the duration of the job.


rasolca:
cscs-ci run

rasolca (approved):
Note: the commit in @finkandreas fork shows a CI failure. It is a bug as the doc-push action should only run in the main repo. See #921 for details.

PR #928, opened 2023-07-05
Update CI to use an allocation per job.

rasolca:
As suggested by @finkandreas it is better to allocate resources for each job, to reduce the amount of CI failures due to slurm.

Note: the docker image will be pulled once even if all the jobs require it. (I quote @finkandreas message)
> it will be pulled only once otherwise too. `sarus pull img_that_exists:tag`, will check the remote registry for the latest manifest, and if it's the same it will not pull again...

> Mostly a noop: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4455690602105886/4525297225819146/-/jobs/4595769743#L28

> and even if all 4 start simultaneously, and sarus itself does not support locking when pulling the same image concurrently (sarus would happily pull 4 times the same image, it works, but really pulls 4 times concurrently the same thing). However there is a locking mechanism in the implemenation of the custom executor, which locks pulling the same image to a single job, while the other jobs would wait until the pulling is done.

> Within CI:
Job 1 wants to pull --> requests lock --> locked --> download --> real container download, if not available --> release lock
Job 2 wants to pull --> requests lock --> locked --> download --> no more download, since job 1 already downloaded --> release lock
(that applies to pulling the same image, different images will be pulled simultaneously)


rasolca:
cscs-ci run

msimberg (approved):

PR #658, opened 2022-10-04
Remove explicit use of futures in tridiagonal solver

msimberg:
Fixes #588. Removes all explicit occurrences of futures, but does not yet get rid of the ones that come implicitly from `Matrix::read` (which returns a keep_future(shared_future)`).

~There's one TODO related to lifetime management of tiles that I haven't yet decided if I'll leave as it is in this PR or try to do something about it. Hence the draft.~ I've left a TODO for something that will be changed separately.

Issue #665, opened 2022-10-07
Explicitly release operation states early to release tiles for continuations


msimberg:
If we end up doing 1. or 2. from https://github.com/pika-org/pika/issues/479 we need to make changes in DLA-Future. Currently we rely on the pika implementation to release operation states/tiles early. The P2300 reference implementation does not release operation states early. This issue is meant to track whatever change we end up needing to do. I will most likely first go with an explicit sender adaptor that allows releasing the operation state early in tactical places.

msimberg:
Will be fixed by #1004.

PR #930, opened 2023-07-06
Add test phase in CI for stdexec configuration

aurianer:
- Add test phase for the stdexec configuration
- Early release of the tiles in unwrap to avoid hangs using `std::tuple` (previously handled by using pika's version of `ensure_started` after `transform` #658)

### DIFF
--- a/include/dlaf/common/unwrap.h
+++ b/include/dlaf/common/unwrap.h
@@ -13,6 +13,7 @@
 
 #include <functional>
 #include <type_traits>
+#include <utility>
 
 #include <pika/async_rw_mutex.hpp>
 #include <pika/execution.hpp>


rasolca:
cscs-ci run

msimberg (changes requested):
Thanks @aurianer for the investigation here.

Functionally I'm happy with this change but see inline comment before I'm ready to approve this.

Linking https://github.com/eth-cscs/DLA-Future/issues/665 as well since it's a related issue. I think the solution here is necessary, but possibly not sufficient for full control of tile lifetimes (as the remaining `ensure_started` problem hints at).

include/dlaf/common/unwrap.h
```diff
@@ -58,7 +59,17 @@ struct Unwrapping {
...
+        decltype(std::move(f)(Unwrapper<std::decay_t<Ts>>::unwrap(std::forward<Ts>(ts))...));
+    if constexpr (std::is_void_v<result_type>) {
+      std::move(f)(Unwrapper<std::decay_t<Ts>>::unwrap(std::forward<Ts>(ts))...);
+      std::tuple<Ts...>(std::forward<Ts>(ts)...);
+    }
+    else {
+      auto r = std::move(f)(Unwrapper<std::decay_t<Ts>>::unwrap(std::forward<Ts>(ts))...);
+      std::tuple<Ts...>(std::forward<Ts>(ts)...);
+      return r;
+    }
```
msimberg, line 72: I'd like to see this separated out into a separate function wrapper. Perhaps called `consume_rvalues` or similar (but we can bikeshed the name separately). The important thing is to separate independent functionality into independent function wrappers. I think you can mostly follow the pattern of unwrap except that you can simplify it significantly since you don't need customization for the consuming rvalues functionality.

Additionally, I'd like to see benchmarks of this change before it's merged. I wouldn't expect a big difference. I'd be sad and slightly surprised if performance is worse, but don't rule it out. Ideally we'd see a small, but measurable performance improvement from this.

include/dlaf/common/unwrap.h
```diff
@@ -12,6 +12,7 @@
 /// @file
 
 #include <functional>
+#include <type_traits>
```
msimberg, line 15:
```suggestion
```

msimberg: Sorry, that's a very good catch! I thought it was just a leftover from moving `consume_rvalues` from `unwrap.h` into a separate header.

Would you mind adding it back? It belongs there, as does `utility` (for `forward`). Separate PR is also fine.

msimberg (approved):
This looks good to me now, thanks!

I'd still like to wait to see the benchmark results, but assuming there are no surprises there this requires no further changes.

The missing includes from `unwrap.h` are not a blocker for this PR.

msimberg:
cscs-ci run

PR #933, opened 2023-07-11
Matrix Data Reshuffling (matrix re-distribute)

### LINKED ISSUE #919: Implement matrix redistribution algorithm

albestro:
Close #919

Simple implementation of a copy function that allows to re-distribute a matrix with a different tile size.

Currently, we concentrated on a small subset of configurations by adding some constraints, that should not represent a problem for the use-cases we'd like to try, but simplify the first implementation.

Constraints:
- `ranks(grid_in) == ranks(grid_out)`
- tileSize can be down/up scaled by an integral factor (each axis is independent, e.g. row axis down-scaled and col-axis up-scaled)

Note:
Current implementation uses `RetiledMatrix` (which simplifies the work a lot!), which we know it is a temporary solution waiting for the refactoring. For this reason, I had to relax the const-correctness of the input arguments of the new `copy` function, because currently `RetiledMatrix<const T>` is not supported.


include/dlaf/matrix/copy.h
```diff
@@ -70,4 +74,79 @@ void copy(Matrix<const T, Source>& source, Matrix<T, Destination>& dest) {
...
+void copy(Matrix<T, Device::CPU>& src,  // TODO this should be const
+          Matrix<T, Device::CPU>& dst, comm::CommunicatorGrid grid) {
+  namespace ex = pika::execution::experimental;
+
+  DLAF_ASSERT_MODERATE(equal_size(src, dst), src.size(), dst.size());
+  DLAF_ASSERT_MODERATE(equal_process_grid(src, grid), src.commGridSize(), grid.size());
+  DLAF_ASSERT_MODERATE(equal_process_grid(dst, grid), dst.commGridSize(), grid.size());
+
+  const TileElementSize tile_size_src = src.blockSize();
+  const TileElementSize tile_size_dst = dst.blockSize();
```
msimberg, line 98: I'm not quite sure about the constraints, but should this for now be:
```suggestion
  const TileElementSize tile_size_src = src.baseTileSize();
  const TileElementSize tile_size_dst = dst.baseTileSize();
```
with the possible addition of:
```suggestion
  DLAF_ASSERT(src.blockSize() == src.baseTileSize(), src.blockSize(), src.baseTileSize());
  DLAF_ASSERT(dst.blockSize() == dst.baseTileSize(), dst.blockSize(), dst.baseTileSize());
  const TileElementSize tile_size_src = src.blockSize();
  const TileElementSize tile_size_dst = dst.blockSize();
```

In the future I suppose we could make it:
```suggestion
  const TileElementSize block_size_src = src.blockSize();
  const TileElementSize block_size_dst = dst.blockSize();
```
with the assumption that we can retile a `Matrix` purely based on the block size, independently of what the current tile size is (not relevant until #905, and may not be relevant even after that...).

include/dlaf/matrix/copy.h
```diff
@@ -70,4 +74,79 @@ void copy(Matrix<const T, Source>& source, Matrix<T, Destination>& dest) {
...
+/// @pre equal_process_grid(src, dst)
+/// @pre mb = min(src.blockSize().rows(), dst.blockSize().rows())
+///      src.blockSize().rows() % mb == 0
+///      dst.blockSize().rows() % mb == 0
+/// @pre nb = min(src.blockSize().cols(), dst.blockSize().cols())
+///      src.blockSize().cols() % nb == 0
+///      dst.blockSize().cols() % nb == 0
+template <class T>
+void copy(Matrix<T, Device::CPU>& src,  // TODO this should be const
+          Matrix<T, Device::CPU>& dst, comm::CommunicatorGrid grid) {
```
msimberg, line 90: Any reason to leave this out except that redistribution isn't needed at the moment for GPU matrices:
```suggestion
template <class T, Device D>
void copy(Matrix<T, D>& src,  // TODO this should be const
          Matrix<T, D>& dst, comm::CommunicatorGrid grid) {
```
?
If you think it needs special cases it's ok to restrict it to CPU-only for now in my opinion, but I think it should be exactly the same for both.

msimberg (commented):
Just a couple of initial comments. Sorry if you were already planning to handle them.

include/dlaf/matrix/copy.h
msimberg: Thanks, makes sense. In that case I would absolutely rename the variable to avoid confusion.

msimberg: Yeah, also makes sense and I see now that I missed some points you were making. Note that I just opened https://github.com/eth-cscs/DLA-Future/pull/935 to handle part of #905. I currently assert that the block and tile sizes are equal, which is implicitly the same requirement that was there for `RetiledMatrix`.

Relaxing the constraints one notch would be to allow retiling already retiled matrices but based on tile size, i.e. only further subdividing the tiles. This is simple to implement, and I'm thinking this may even just work with the current implementation if we remove the assertion.

Relaxing the constraints further would be more complicated. The difficulty with choosing a different retiling is that the retiled matrix doesn't have access to the original matrix anymore which means that you one has to reconstruct tile dependencies. This is simpler if the new retiled tile size is a multiple of the old retiled tile size, and more complicated if it's not. We'll generally have an easier time if we can retile the original matrix instead of retiling a retiled matrix.

That's not to say that we can't or shouldn't do it, but at least from my current knowledge about it the cost seems higher than the value.

msimberg: Sorry for the wall of text. Given the constraint in #935 that a retiled matrix can't be retiled you might not need to check for that here and can instead rely on checks in `RetiledMatrix`/`retiledSubPipeline`.

Further, given the first relaxation would be
> Relaxing the constraints one notch would be to allow retiling already retiled matrices but based on tile size
the right thing to do in this PR is this one:
```suggestion
  const TileElementSize tile_size_src = src.baseTileSize();
  const TileElementSize tile_size_dst = dst.baseTileSize();
```

albestro: Added also a note https://github.com/eth-cscs/DLA-Future/pull/933/commits/5653f101dedd548c22a3e9068808fca5d16b83c3

include/dlaf/matrix/copy.h
```diff
@@ -70,4 +75,89 @@ void copy(Matrix<const T, Source>& source, Matrix<T, Destination>& dest) {
...
+  DLAF_ASSERT_MODERATE(equal_process_grid(dst, grid), dst.commGridSize(), grid.size());
+
+  // TODO Currently multiple tile per blocks cannot be tested, as Matrix does not support it yet.
+  DLAF_ASSERT_MODERATE(src.baseTileSize() == src.blockSize(), src.baseTileSize(), src.blockSize());
+  DLAF_ASSERT_MODERATE(dst.baseTileSize() == dst.blockSize(), dst.baseTileSize(), dst.blockSize());
+
+  // Note:
+  // From an algorithmic point of view it would be better to reason in terms of block instead of tiles,
+  // with the aim of reducing the number of communications.
+  // Current implementation reasons in terms of tiles due to a limitation for "recursively retiled"
```
albestro, line 105: Probably it would be enough even just

```suggestion
  // Current implementation reasons in terms of tiles due to a limitation for retiled
```

test/unit/matrix/test_matrix.cpp
```diff
@@ -1833,6 +1834,63 @@ TYPED_TEST(MatrixTest, GPUCopySubPipeline) {
 }
 #endif
 
+struct TestReshuffling {
+  const GlobalElementSize size;
+  const TileElementSize src_tilesize;
+  const TileElementSize dst_tilesize;
+};
+std::vector<TestReshuffling> sizes_reshuffling_tests{
+    TestReshuffling{{10, 10}, {3, 3}, {3, 3}},   // no-reshuffling
```
albestro, line 1843: Do you agree in keeping this managed without any specific fallback logic to a simpler case?

include/dlaf/matrix/copy.h
msimberg: Yeah, agreed on removing the "recursively" part. It's enough to have a "simple" (i.e. only once) retiled matrix as input to this algorithm to hit the limitations.

test/unit/matrix/test_matrix.cpp
rasolca: What do you mean?
As the `src_rank` is different the data has to be communicated anyway.

test/unit/matrix/test_matrix.cpp
```diff
@@ -1833,6 +1834,63 @@ TYPED_TEST(MatrixTest, GPUCopySubPipeline) {
...
+  const auto& [size, src_tilesize, dst_tilesize] = config;
+  const comm::Index2D origin_rank_src(std::max(0, grid.size().rows() - 1),
+                                      std::max(0, grid.size().cols() - 1));
+  matrix::Distribution dist_src(size, src_tilesize, grid.size(), grid.rank(), origin_rank_src);
+  const comm::Index2D origin_rank_dst(
+      std::min(grid.size().rows() - 1, dlaf::util::ceilDiv(grid.size().rows(), 2)),
+      std::min(grid.size().cols() - 1, dlaf::util::ceilDiv(grid.size().cols(), 2)));
+  matrix::Distribution dist_dst(size, dst_tilesize, grid.size(), grid.rank(), origin_rank_dst);
+
+  matrix::Matrix<T, Device::CPU> src_host(dist_src);  // TODO this should be const
```
rasolca, line 1859: How can you set the elements if const?

msimberg (approved):

msimberg:
cscs-ci run

rasolca:
There are deadlocks in the CUDA tests.

albestro:
cscs-ci run

albestro:
cscs-ci run

msimberg (approved):

PR #934, opened 2023-07-12
Add `MatrixRef` to take a sub-matrix from an existing `Matrix`

msimberg:
This adds a `MatrixRef` class which takes an existing `Matrix`, offset, and size to give sender access to the sub-distribution defined by offset and size. At least for this PR `MatrixRef` will stay a separate type. It's not clear to me how to cleanly make `MatrixRef` and `Matrix` the same type without ugliness in the implementation, and without confusing the guarantees that the types give.


msimberg:
cscs-ci run

include/dlaf/matrix/distribution.h
```diff
@@ -119,6 +119,17 @@ class Distribution {
...
+  /// Constructs a sub-distribution based on the given distribution @p dist with
+  /// an @p offset and @p size.
+  ///
+  /// @param[in] dist is the input distribution,
+  /// @param[in] offset is the offset of the new distribution relative to the input distribution,
+  /// @param[in] size is the size of the new distribution relative to the offset,
+  /// @pre origin.isValid()
+  /// @pre size.isValid()
+  /// @pre origin + size <= dist.size()
+  Distribution(Distribution dist, const GlobalElementIndex& offset, const GlobalElementSize& size);
```
msimberg, line 131: This currently takes `offset` and `size` as separate parameters. Some alternatives:

- Add a `SubDistributionSpec` (similar to `SubTileSpec`) and use that instead. I would not name it `SubMatrixSpec` here since we don't want to make `Distribution` "depend" on `Matrix` (it wouldn't be a real dependency, but a conceptual dependency). We could perhaps add a `using SubMatrixSpec = SubDistributionSpec` to `matrix.h` as well.
- Allow defaulting the `size` to use the rest of the distribution/matrix. With separate parameters this is trivial (e.g. `const std::optional<GlobalElementSize>& size = std::nullopt`). With a `SubDistributionSpec` it could be done either with an `optional` member in `SubDistributionSpec` or two constructors: `Distribution(Distribution dist, const GlobalElementIndex& offset)` and `Distribution(Distribution dist, const SubDistributionSpec& spec)`. The latter would probably be cleaner.

Thoughts?

include/dlaf/matrix/distribution.h
```diff
@@ -490,6 +501,27 @@ class Distribution {
             localElementDistanceFromLocalTile<Coord::Col>(begin.col(), end.col())};
   }
 
+  /// TODO
```
msimberg, line 504: Add docstring.

include/dlaf/matrix/matrix_ref.h
```diff
@@ -0,0 +1,175 @@
...
+  /// @pre origin + size <= mat.size()
+  MatrixRef(Matrix<const T, D>& mat, const GlobalElementIndex& offset, const GlobalElementSize& size)
+      : internal::MatrixBase(Distribution(mat.distribution(), offset, size)), mat_const_(mat),
+        offset_(offset) {}
+
+  // TODO: default, copy, move construction?
+  // - default: no, don't want empty MatrixRef
+  // - copy: implementable, still refer to the original matrix
+  // - move: implement as copy, i.e. still refer to original matrix?
+  MatrixRef() = delete;
```
msimberg, line 58: What should we allow here?

My preference would be to make `MatrixRef` behave like `std::reference_wrapper`, i.e.:
- no default constructor and thus no empty state
- regular constructor taking a `Matrix` etc.
- no move constructor since no empty state
- copy constructor

Preferences?

include/dlaf/matrix/matrix_ref.h
```diff
@@ -0,0 +1,175 @@
...
+
+  // TODO: Do we need access to the original matrix? e.g:
+  // Matrix& get() const noexcept { return mat_const_; }
```
msimberg, line 61: I wasn't sure if this is needed now or ever. If there's no clear short-term need for it I won't include it.

include/dlaf/matrix/distribution.h
```diff
@@ -490,6 +501,27 @@ class Distribution {
...
+  TileElementIndex tileElementOffsetFromSubDistribution(
+      const GlobalElementIndex& sub_offset, const Distribution& sub_distribution,
+      const GlobalTileIndex& sub_index) const noexcept {
+    DLAF_ASSERT(sub_index.isIn(sub_distribution.nrTiles()), sub_index, sub_distribution.nrTiles());
+    DLAF_ASSERT(isCompatibleSubDistribution(sub_offset, sub_distribution), "");
+    return {
+        sub_index.row() == 0 ? tileElementFromGlobalElement<Coord::Row>(sub_offset.row()) : 0,
+        sub_index.col() == 0 ? tileElementFromGlobalElement<Coord::Col>(sub_offset.col()) : 0,
+    };
+  }
```
msimberg, line 523: I've currently included a minimal set of helper functions as methods on `Distribution`. This is sufficient, together with `MatrixRef` storing the original `Matrix` (which gives access to its `Distribution`), `MatrixRef` inheriting `MatrixBase` (which gives access to the sub-distribution), and `MatrixRef` storing the offset between the original distribution and the sub-distribution, thus keeping track of the relationship between the two. This is why the helpers here take the sub-distribution and the offset. 

An alternative would be to have some kind of helper which keeps track of this relationship and performs necessary index transformations/calculations. Do you think this would be more readable and/or maintainable?

include/dlaf/matrix/distribution.h
msimberg: https://github.com/eth-cscs/DLA-Future/pull/934/commits/2010844f372b9bd83c9a6685bb7170c433f44da1

include/dlaf/matrix/distribution.h
```diff
@@ -490,6 +501,27 @@ class Distribution {
...
+  GlobalTileIndex globalTileIndexFromSubDistribution(const GlobalElementIndex& sub_offset,
+                                                     const Distribution& sub_distribution,
+                                                     const GlobalTileIndex& sub_index) const noexcept {
+    DLAF_ASSERT(sub_index.isIn(sub_distribution.nrTiles()), sub_index, sub_distribution.nrTiles());
+    DLAF_ASSERT(isCompatibleSubDistribution(sub_offset, sub_distribution), "");
+    const GlobalTileIndex tile_offset = globalTileIndex(sub_offset);
+    return tile_offset + common::sizeFromOrigin(sub_index);
+  }
+
+  TileElementIndex tileElementOffsetFromSubDistribution(
```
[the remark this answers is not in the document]
msimberg: https://github.com/eth-cscs/DLA-Future/pull/934/commits/2010844f372b9bd83c9a6685bb7170c433f44da1

include/dlaf/matrix/matrix_ref.h
```diff
@@ -0,0 +1,175 @@
...
+namespace dlaf::matrix {
+/// A @c MatrixRef represents a sub-matrix of a @c Matrix.
+///
+/// The class has reference semantics, meaning accesses to a @c MatrixRef and
+/// it's corresponding @c Matrix are interleaved if calls to read/readwrite are
+/// interleaved. Access to a @c MatrixRef and its corresponding @c Matrix is not
+/// thread-safe. A @c MatrixRef must outlive its corresponding @c Matrix.
+template <class T, Device D>
+class MatrixRef;
+
```
msimberg, line 30: Would this be better named `SubMatrixRef` since that's the primary use case? I'd argue that someone who receives a `MatrixRef` doesn't need to know that it's a sub-matrix of another `Matrix`, but I don't feel too strongly about it.

msimberg:
cscs-ci run

include/dlaf/matrix/matrix_ref.h
```diff
@@ -0,0 +1,175 @@
...
+    const auto ij_tile =
+        parent_dist.tileElementOffsetFromSubDistribution(offset_, this->distribution(), index);
+    return splitTile(std::move(tile_sender), SubTileSpec{ij_tile, tile_size});
+  }
+
+private:
+  Matrix<T, D>& mat_;
+  using MatrixRef<const T, D>::offset_;
+};
+}
```
msimberg, line 175: To do: explicitly instantiate?

msimberg: Done in https://github.com/eth-cscs/DLA-Future/pull/934/commits/a8c848f3e176fdf3ac0b2b77022629880b60c795.

msimberg:
cscs-ci run

include/dlaf/matrix/matrix_ref.h
msimberg: I've tentatively removed the TODO and commented out code, but leaving this comment unresolved until others have commented.

msimberg:
cscs-ci run

include/dlaf/matrix/distribution.h
albestro: I would be a fan of the first solution with `SubDistributionSpec` (and in case of need adding the aliasing as you suggested for `SubMatrixSpec`).

albestro (approved):
LGTM

msimberg:
cscs-ci run

rasolca (commented):
I would merge this PR and start using it. In this way we can fully realize if the API is ok or need tweaking.

But I would move `MatrixRef` to `internal` namespace, and move back in a later PR when we are confident about the API.

include/dlaf/matrix/matrix_ref.h
msimberg: I'm leaving this out. We'll add it if and when we need it.

include/dlaf/matrix/distribution.h
msimberg: Let's see in real usage if this needs refactoring, but I'm keeping this as is for now.

include/dlaf/matrix/matrix_ref.h
msimberg: Keeping the name for now until we know better. However, I'm moving this to `internal` until we're more confident with the functionality as suggested by @rasolca.

include/dlaf/matrix/distribution.h
msimberg: I've added
```
struct SubDistributionSpec {
    GlobalElementIndex origin;
    GlobalElementSize size;
};
```
and
```
using SubMatrixSpec = SubDistributionSpec;`
```
This lead to some `offset` to `origin` renamings as well. I'm not 100% confident that I've been consistent in using `origin` vs. `offset` (or that I even have a consistent rule to follow) so I'd ask you to keep an eye out and see if some namings don't make sense even after we merge this.

include/dlaf/matrix/matrix_ref.h
msimberg: I made it completely immovable to be conservative in the end because
1. It's not safe to pass copies into new tasks anyway and having it immovable _may_ avoid mistakes like that, and
2. One can always take a `MatrixRef<>&` where needed or create a new `MatrixRef`.
We can revisit this later if it turns out that it's more convenient to make it behave like `reference_wrapper` in some use cases.

msimberg:
cscs-ci run

rasolca (approved):

Issue #923, opened 2023-07-04
Optimize read-after-read access in `Matrix` subpipelines


msimberg:
See https://github.com/eth-cscs/DLA-Future/pull/898#discussion_r1238751326. In the worst case this may need support in `async_rw_mutex` in pika. Needs further investigation. Some investigation on where and if this actually could lead to a performance benefit may be useful.

PR #935, opened 2023-07-12
Refactor `RetiledMatrix` into `Matrix`

### LINKED ISSUE #905: Change `RetiledMatrix` to `Matrix`

msimberg:
Fixes #905.

I moved the creation of a `RetiledMatrix` into `Matrix` methods `retiledSubPipeline{,Const}` to mirror #898. Since the implementation is otherwise identical I've kept the "sub pipeline" name, but added `retiled` as a prefix (though the naming concerns from #898 apply here as well). I've removed the `RetiledMatrix` class and replaced it with `Matrix`. I've added an assertion to only allow retiling if the `Matrix` isn't already "retiled", i.e. if the tile size is the same as the block size. We can attempt to lift this restriction later if it's useful but I won't be doing that here. #923 also applies for retiled matrices and I'm not doing that in this PR either.

This should probably wait for #933 to be merged first, or maybe not...


msimberg:
cscs-ci run

albestro (approved):

rasolca (commented):
I have only a concern.
Algorithm implementations might not work with a retiled matrix as argument.

Therefore for each function that takes a matrix as argument we should either enhance the test and make sure it works or add an assertion.

Of course updating all the tests and fixing the algorithms does not belong to this PR.

Should we add an assertion at least in the public algorithms until they are tested and known to work?

Note: as a retiled matrix is now a `Matrix`, retiled of retiled will be needed. (e.g for complete functionality of #933)

msimberg:
> I have only a concern. Algorithm implementations might not work with a retiled matrix as argument.

Good point.

> Therefore for each function that takes a matrix as argument we should either enhance the test and make sure it works or add an assertion.
> 
> Of course updating all the tests and fixing the algorithms does not belong to this PR.
> 
> Should we add an assertion at least in the public algorithms until they are tested and known to work?

I think that would be good. I will add that here.
 
> Note: as a retiled matrix is now a `Matrix`, retiled of retiled will be needed. (e.g for complete functionality of #933)

Yeah, with the discussion in #933 I would include the redistributing `copy` to the list of algorithms above that currently require same block size and tile size. Then we can case by case relax the requirements/test what works with retiled matrices.

msimberg:
@rasolca, @albestro see https://github.com/eth-cscs/DLA-Future/pull/935/commits/c962eddf419c6366cf7b4b3c3135442b212c5252 for assertions on algorithms.

I added a helper function which I now called `retiled(MatrixLike)`. I'm not super happy with the name, but it was better than `unretiled`. I could go for a more explicit `equal_tile_and_block(_size)` or something in that direction as well. Suggestions welcome.

rasolca:
What about `single_tile_per_block(matrix)`?

include/dlaf/eigensolver/gen_to_std.h
```diff
@@ -38,6 +38,7 @@ namespace eigensolver {
 /// Note: B should be modifiable as the diagonal tiles might be temporarly modified during the calculation.
 /// @pre mat_a and mat_b have the same square size,
 /// @pre mat_a and mat_b have the same square block size,
+/// @pre mat_a and mat_b have the same tile tile and block sizes,
```
albestro, line 41: I have seen it a couple of times over this PR 😉 

```suggestion
/// @pre mat_a and mat_b have the same tile and block sizes,
```

include/dlaf/eigensolver/tridiag_solver.h
```diff
@@ -70,19 +77,26 @@ void tridiagSolver(Matrix<BaseType<T>, Device::CPU>& tridiag, Matrix<BaseType<T>
 ///                of the second column is not used.
 /// @param evals [out] (n x 1) local matrix holding the eigenvalues of the the symmetric tridiagonal
 ///              matrix
-/// @param evecs [out] (n x n) distributed matrix holding the eigenvectors of the the symmetric tridiagonal
+/// @param evecs [out] (n x n) distributed matrix holding the eigenvectors of the the symmetric
```
albestro, line 80:
```suggestion
/// @param evecs [out] (n x n) distributed matrix holding the eigenvectors of the symmetric
```

msimberg:
I'll go with `single_tile_per_block`. Thanks for the comments! Pushing soon...

include/dlaf/eigensolver/tridiag_solver.h
msimberg: I'll do this separately. There were at least a few other doc-formatting changes that I noticed were off while doing this.

include/dlaf/eigensolver/gen_to_std.h
msimberg: https://github.com/eth-cscs/DLA-Future/pull/935/commits/9c50411e7d5a942bee2953563b1e0800142ec054

msimberg:
See https://github.com/eth-cscs/DLA-Future/pull/935/commits/f3ecb55432dab102ed7c01a84ca66d4902bc1b26 for the renaming to `single_tile_per_block`. Would you prefer if I update the preconditions in the docs from
```
@pre mat has equal tile and block sizes
```
to
```
@pre mat has a single tile per block
```
to match?

rasolca:
> See [f3ecb55](https://github.com/eth-cscs/DLA-Future/commit/f3ecb55432dab102ed7c01a84ca66d4902bc1b26) for the renaming to `single_tile_per_block`. Would you prefer if I update the preconditions in the docs from
> 
> ```
> @pre mat has equal tile and block sizes
> ```
> 
> to
> 
> ```
> @pre mat has a single tile per block
> ```
> 
> to match?

No strong preference on my side.

rasolca (approved):

include/dlaf/multiplication/general.h
albestro, line 1: are these `single_tile_per_blocketiled` a leftover? otherwise, I'm missing something

include/dlaf/eigensolver/eigensolver.h
albestro, line 1: differently from other algorithms, doc has not been updated with precondition.

include/dlaf/permutations/general.h
albestro, line 1: also here the doc has not been updated with pre-conditions.

albestro (changes requested):

include/dlaf/eigensolver/eigensolver.h
msimberg: There are a number of other preconditions that are not documented. I'll do that as part of https://github.com/eth-cscs/DLA-Future/pull/935#discussion_r1262503563.

include/dlaf/multiplication/general.h
msimberg: Nope, just a typo... well spotted.

include/dlaf/permutations/general.h
msimberg: Same here, will update as part of https://github.com/eth-cscs/DLA-Future/pull/935#discussion_r1262503563. All other preconditions are also missing.

include/dlaf/multiplication/general.h
msimberg: Fixed.

albestro (approved):

include/dlaf/eigensolver/tridiag_solver.h
msimberg: #944.

PR #831, opened 2023-04-11
Remove the need of the distributed row permutation in the tridiagonal solver.

rasolca:
The distributed row permutation is not needed as the matrix can be setup in the correct order from the beginning.

Benchmarks will follow.
Depends on #819.

PR #936, opened 2023-07-14
TridiagSolver (local): embed row permutation in rank1 solver

albestro:
Distributed and local implementations have slightly diverged in the last months.

This PR addresses the main (and probably the only ones) differences. With #831 distributed implementation got row permutations and `setUnitDiag` embedded in rank1 solution, removing the need of explicitly doing so just before doing the gemm. Local implementation was still doing the permutation separately.

With this PR both `setUnitDiag` and row permutation happen during rank1 solver also for the local implementation.

### DIFF
--- a/include/dlaf/eigensolver/tridiag_solver/kernels.h
+++ b/include/dlaf/eigensolver/tridiag_solver/kernels.h
@@ -273,43 +273,6 @@ void initIndexTileAsync(SizeType tile_row, TileSender&& tile) {
   di::transformDetach(di::Policy<DefaultBackend_v<D>>(), initIndexTile_o, std::move(sender));
 }
 
-template <class T>
-void setUnitDiagonal(const SizeType& k, const SizeType& tile_begin,
-                     const matrix::Tile<T, Device::CPU>& tile);
-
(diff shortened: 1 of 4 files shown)


albestro:
Note:

Permutation is done in a slightly different way due to different optimization requirements between local and distributed. In the latter one, in order to avoid additional (and unnecessary) communications, the permutation is done in the first steps of the rank1-solver, and all other steps have to take this into account. On the other hand, in the local implementation the permutation is done in the last step, since there is no reason to do it immediately, keeping rank1-solver "clean" from permutations till the very end.

This also introduces a difference for `k == 2`, for which the permutations should happen even if it is handled differently from other cases. Since in local implementation is done at the very end, the aforementioned case has to handled differently by performing an ad-hoc permutation just before the early exit of the special case. It must be highlighted that these special cases are not tested extensively for the moment, and I'm going to open an issue to keep track of this.

rasolca (approved):

rasolca:
cscs-ci run

albestro:
cscs-ci run

msimberg (approved):

PR #938, opened 2023-07-14
Reimplementation of local band-to-tridiagonal

rasolca:
Implementation without `let_value`.
To me it looks simpler and more readable compared to other attempt with `let_value` (see #939), but introduce a little trick with a `shared_ptr` to allow to share the vector of tiles.

The new implementation is 45-50% faster (tested 20k, 30k and 40k).

PR #939, opened 2023-07-17
Reimplementation of local band-to-tridiagonal (let_value version)

rasolca:
Note: doesn't work yet (segfault)


rasolca:
Close in favour of #938.

PR #941, opened 2023-07-18
Avoid unnecessary allocation of `shared_ptr`s in `MemoryView`

msimberg:
I didn't run any benchmarks with this as it seems we're only using these in the `MemoryView` tests. It's likely that the `MemoryView(T* ptr, SizeType size)` constructor is used by the C API but I haven't checked if it makes a difference there. Similar motivation as #579.

### DIFF
--- a/include/dlaf/memory/memory_view.h
+++ b/include/dlaf/memory/memory_view.h
@@ -55,7 +55,8 @@ public:
   /// @param size The size (in number of elements of type @c T) of the existing allocation,
   /// @pre @p ptr+i can be deferenced for 0 < @c i < @p size.
   MemoryView(T* ptr, SizeType size)
-      : memory_(std::make_shared<MemoryChunk<ElementType, D>>(const_cast<ElementType*>(ptr), size)),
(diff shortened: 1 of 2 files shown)


msimberg:
cscs-ci run

include/dlaf/memory/memory_view.h
```diff
@@ -55,7 +55,8 @@ class MemoryView {
   /// @param size The size (in number of elements of type @c T) of the existing allocation,
   /// @pre @p ptr+i can be deferenced for 0 < @c i < @p size.
```
albestro, line 56: I know it is not related, but we might just take the chance to fix the doc.

```suggestion
  /// @pre @p ptr+i can be deferenced for 0 <= @c i < @p size.
```

albestro (approved):

include/dlaf/memory/memory_view.h
msimberg: https://github.com/eth-cscs/DLA-Future/pull/941/commits/eb80189c7b6371686fdd8e6ea635cb59f7fee452. Also fixed `deferenced` to `dereferenced`, and the same problems in `memory_chunk.h`.

rasolca (approved):

msimberg:
cscs-ci run

Issue #914, opened 2023-06-23
Permute (locally) to get the optimal matrix shape


rasolca:
The eigenvectors has to be permuted locally to achieve the optimal matrix shape to reduce the complexity of the GEMM.

This step can directly be done when apply the given rotation of the deflation.
Instead of in-place rotations -> rotations in the final location.

- [x] local implementation
  - https://github.com/eth-cscs/DLA-Future/pull/943
- [x] distributed implementation
  - https://github.com/eth-cscs/DLA-Future/pull/967
  - ...

PR #943, opened 2023-07-20
TridiagSolver (local): sort eigenvalues by column type for rank1 solver

albestro:
This address the local part of #914.

Main changes:
- added `i4` and `i5` index workspaces (not checked if there is any other way of re-using existing ones)
- rank1 now works just on first `k` eigenvectors, because deflated have been moved at the end beforehand
- adapt the documentation to new steps and indices

TODO
- [x] manually check that matrix shape for non-deflated is as expected (upper-dense-lower)
- [x] manually check that deflated are sorted by eigenvalue
- [x] check that comments are in-sync with new changes
- [x] ~code cleanup~

### DIFF
--- a/include/dlaf/eigensolver/tridiag_solver/merge.h
+++ b/include/dlaf/eigensolver/tridiag_solver/merge.h
@@ -338,7 +338,7 @@ SizeType stablePartitionIndexForDeflationArrays(const SizeType n, const ColType*
   //    initial <-- sorted by ascending eigenvalue in four groups (upper | dense | lower | deflated)
 
   // Note:
-  // This is the order how we want the eigenvectors to be sorted, since it leads to a nicer matrx
+  // This is the order how we want the eigenvectors to be sorted, since it leads to a nicer matrix
   // shape that allows to reduce the number of following operations (i.e. gemm)
   auto coltype_index = [](const ColType coltype) -> std::size_t {
     switch (coltype) {
--- a/include/dlaf/eigensolver/tridiag_solver/merge.h
+++ b/include/dlaf/eigensolver/tridiag_solver/merge.h
@@ -11,6 +11,7 @@
 
 #include <algorithm>
 #include <functional>
+#include <numeric>
 
 #include <pika/barrier.hpp>
 #include <pika/execution.hpp>
@@ -114,6 +115,7 @@ struct WorkSpace {
   Matrix<T, D> z1;
 
   Matrix<SizeType, D> i2;
+  Matrix<SizeType, D> i5;
 };
 
 template <class T>
@@ -124,6 +126,7 @@ struct WorkSpaceHost {
 
   Matrix<SizeType, Device::CPU> i1;
   Matrix<SizeType, Device::CPU> i3;
+  Matrix<SizeType, Device::CPU> i4;
 };
 
 template <class T, Device D>
@@ -140,6 +143,7 @@ struct WorkSpaceHostMirror {
   HostMirrorMatrix<T, D> z1;
 
   HostMirrorMatrix<SizeType, D> i2;
+  HostMirrorMatrix<SizeType, D> i5;
 };
 
 template <class T, Device D>
@@ -249,7 +253,7 @@ auto calcTolerance(const SizeType i_begin, const SizeType i_end, Matrix<const T,
 // The permutation will allow to keep the mapping between sorted eigenvalues and unsorted eigenvectors,
 // which is useful since eigenvectors are more expensive to permuted, so we can keep them in their initial order.
 //
-// @param n number of eigenvalues
+// @param n         number of eigenvalues
 // @param c_ptr     array[n] containing the column type of each eigenvector after deflation (initial order)
 // @param evals_ptr array[n] of eigenvalues sorted as in_ptr
 // @param in_ptr    array[n] representing permutation current -> initial (i.e. evals[i] -> c_ptr[in_ptr[i]])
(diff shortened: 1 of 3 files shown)


include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -681,22 +733,28 @@ void mergeSubproblems(const SizeType i_begin, const SizeType i_split, const Size
...
-  // - solve the rank-1 problem and save eigenvalues in `d0` and `d1` (copy) and eigenvectors in `e2`.
+  // - compute permutation `i4`: sorted by col type ---> deflated
+  // - solve rank-1 problem and save eigenvalues in `d0` and `d1` (copy) and eigenvectors in `e2` (sorted
+  // by coltype)
   // - set deflated diagonal entries of `U` to 1 (temporary solution until optimized GEMM is implemented)
   //
-  auto k = stablePartitionIndexForDeflation(i_begin, i_end, ws_h.c, ws_hm.i2, ws_h.i3) | ex::split();
+  auto k =
+      ex::ensure_started(stablePartitionIndexForDeflation(i_begin, i_end, ws_h.c, ws_hm.i2, ws_h.i3)) |
+      ex::split();
```
albestro, line 757: @msimberg do you think it is better, in general, to `ensure_started`/`start_detached` tasks? because this (but there are also others) for starting relied on it being a dependency for next tasks.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -433,6 +437,48 @@ auto applyDeflation(const SizeType i_begin, const SizeType i_end, RhoSender&& rh
...
+              return 0;
+            case ColType::Dense:
+              return 1;
+            case ColType::LowerHalf:
+              return 2;
+            case ColType::Deflated:
+              return 3;
+          }
+          return DLAF_UNREACHABLE(std::size_t);
+        };
```
albestro, line 464: I opted for this instead of adding values to the enum class because it represents an important implementation detail of the algorithm, and not of the enum class itself (constraint: they have to be contiguous, they have to start from 0 and they should represent the exact order how we want columns to be sorted).

If you feel it is better the other way, I can easily change this.

msimberg: For me this is ok, though I admit it took me a moment to understand what was going on here. You're essentially using the enum values as indices into an array. It'd be interesting if the compiler anyways sees through that `coltype_index` function sees that the mapping is `0` to `0`, `1` to `1` etc. "Interesting" but I'm not asking you to check.

No action required here (for me).

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -433,6 +437,48 @@ auto applyDeflation(const SizeType i_begin, const SizeType i_end, RhoSender&& rh
...
+
+        const ColType* types = coltype_tiles[0].get().ptr();
+        std::array<std::size_t, 4> offsets{0, 0, 0, 0};
+        std::for_each(types, types + n, [&offsets, &coltype_index](const auto& coltype) {
+          if (coltype != ColType::Deflated)
+            offsets[1 + coltype_index(coltype)]++;
+        });
```
msimberg, line 471: If I'm not mistaken this is a reduction. `std::reduce` doesn't help here because it requires the elements that are being reduced and the type into which we are reducing to be the same. You could rewrite this with `std::accumulate`, but I'll leave it up to you to judge if it's more readable: https://godbolt.org/z/Pxj6hq81v (alternative: https://godbolt.org/z/vKd66zc9c). It's not great that `offsets` has to be wrapped in `ref` so I'm not sure it's a win. The main win is that it says `accumulate` which tells me right away that it's a reduction.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -433,6 +437,48 @@ auto applyDeflation(const SizeType i_begin, const SizeType i_end, RhoSender&& rh
...
+        std::partial_sum(offsets.cbegin(), offsets.cend(), offsets.begin());
+
+        SizeType* i4 = index_tiles[0].ptr();
+        for (SizeType j = 0; j < n; ++j) {
+          const ColType& coltype = types[to_sizet(j)];
+          i4[offsets[coltype_index(coltype)]++] = j;
+        }
```
msimberg, line 478: Why `for_each` in the first loop and plain for loop in the second?

msimberg (approved):
Minor nitpicking comments that can be ignored. Looks good to me!

include/dlaf/eigensolver/tridiag_solver/merge.h
msimberg: Generally speaking, avoid `ensure_started` unless it's required for progress since it causes an extra heap allocation. However, as we've found out we can get into situations where eagerly starting work is required, and in those cases you of course have to use `ensure_started`.

include/dlaf/eigensolver/tridiag_solver/merge.h
albestro: Yes, thanks for the suggestion!

Actually I had already quickly checked if I had any `std`-way of doing this, but I was not able to come up with something really neat.

Even if I agree that with `std::accumulate` the name is a good hint for understanding what's going on, to me it seems that in this case it is not of much help because the init value is an array of initvalues, and the binary operator has to do everything exactly as in the current solution: select the right counter and increment it (+ boilerplate to make accumulate happy, e.g. return `std::ref` to the counters).

In the end, we would use `std::accumulate` just as a `std::for_each` with some more boilerplate...don't you think?

In any case, thank you for the different view on the problem!

include/dlaf/eigensolver/tridiag_solver/merge.h
albestro: It is exactly a LUT, but I didn't find a better way to create a static one (apart specifying values for `ColType` values in the `enum class` definition.

Note: the LUT creates an order that is slightly different from how they are defined in the `enum class`. 

```cpp
{Upper, Dense, Lower, Deflated} // as per LUT with switch
{Upper, Lower, Dense, Deflated} // as per enum class ColType definition
```

https://github.com/eth-cscs/DLA-Future/blob/7752e7b236951daab952c733413cf3d5175ae942/include/dlaf/eigensolver/tridiag_solver/coltype.h#L17-L22

This is part of why I didn't want to put an implementation detail in the `enum class` definition...it looked too risky to me (e.g. someone might decide to reorder/change the values in the definition).

include/dlaf/eigensolver/tridiag_solver/merge.h
msimberg: Yeah, agreed. I'm fine with leaving it as is.

include/dlaf/eigensolver/tridiag_solver/merge.h
msimberg: I feel like I'mmissing something obvious, but why is the order different between the enum and your indices? I understand that they're can go out of sync and that's ok, but is there a good reason for the particular order?

albestro: No, there is no reason for the current order of declaration in the enum. And that's why I didn't want to change that: there is no reason to order them differently, apart from an algorithmic detail.

I agree that can be changed for following the algorithmic requirement, but at that point I would have to find a way to constrain it/check it somehow (otherwise an "involuntary" change to the order would end up in a broken algorithm).

msimberg: Thanks @albestro. I was rather asking the opposite: why is the order in the algorithm not changed to be the same as in the enum? You're hinting at some algorithmic constraint that results in the current order, and if there is such a constraint it should probably be added to a comment in the implementation.

albestro: This PR sorts the eigenvectors in order to make the matrix storing them have a "nicer shape", so that general multiplication can be done more efficiently. Indeed, it sorts them as in

https://github.com/eth-cscs/DLA-Future/blob/3d21dc900bdb4fb5ce30ebbacded3230150f6623/include/dlaf/eigensolver/tridiag_solver/merge.h#L759-L763

so that later we will be able to perform the multiplication (deflated do not take part in this) as two sub-multiplication (top left and bottom right sub-matrices) skipping the zeros part, reducing the number of operations needed.

msimberg: Ok, I still feel like some comment at this point could be good, but the link to the explanation is already good enough for me. Thanks!

albestro:
For the moment I trivially duplicated the code for `stablePartitionIndexForDeflationArrays` (and its related "wrapper" for senders), creating an overload that accepts an additional output where to store the permutation that sorts by column type.

The original one has reason to exists just until the local and distributed version are diverging in what they do...and this will be true just for the next couple of months while we are implementing the new "gemm optimisation".

IMHO the effort in removing this duplication is not worth, but if you feel to, or you have a simple suggestion, just let me know and I will try to address it in this PR 😉

albestro:
cscs-ci run

@msimberg I don't remember if you were able to narrow down the last benchmark problems to this commit. In case you have some remembrance or evidence, let me know 😉 in the meanwhile, I will just check with the CI.

albestro:
cscs-ci run

rasolca (commented):
LGTM

rasolca (approved):

PR #944, opened 2023-07-20
Miscellaneous docstring fixes and updates

msimberg:
- Adds some missing preconditions
- Updates the bt_band_to_tridiag docstring to actually be a docstring (instead of regular comment)
- Adds a docstring for the distributed bt_band_to_tridiag
- Removes duplicate "the" from some docstrings

cf. https://github.com/eth-cscs/DLA-Future/pull/935#discussion_r1262498096

### DIFF
--- a/include/dlaf/eigensolver/eigensolver.h
+++ b/include/dlaf/eigensolver/eigensolver.h
@@ -35,6 +35,15 @@ namespace dlaf::eigensolver {
 /// @param mat contains the Hermitian matrix A
 /// @param eigenvalues is a N x 1 matrix which on output contains the eigenvalues
 /// @param eigenvectors is a N x N matrix which on output contains the eigenvectors
+/// @pre mat is not distributed
+/// @pre mat has a square size
+/// @pre mat has a square blocksize
+/// @pre mat has equal tile and block sizes
+/// @pre eigenvalues is not distributed
+/// @pre eigenvalues has equal tile and block sizes
+/// @pre eigenvectors is not distributed
+/// @pre eigenvectors has a square blocksize
+/// @pre eigenvectors has equal tile and block sizes
 template <Backend B, Device D, class T>
 void eigensolver(blas::Uplo uplo, Matrix<T, D>& mat, Matrix<BaseType<T>, D>& eigenvalues,
                  Matrix<T, D>& eigenvectors) {
@@ -69,6 +78,10 @@ void eigensolver(blas::Uplo uplo, Matrix<T, D>& mat, Matrix<BaseType<T>, D>& eig
 /// @return struct ReturnEigensolverType with eigenvalues, as a vector<T>, and eigenvectors as a Matrix
 /// @param uplo specifies if upper or lower triangular part of @p mat will be referenced
 /// @param mat contains the Hermitian matrix A
+/// @pre mat is not distributed
+/// @pre mat has a square size
+/// @pre mat has a square blocksize
+/// @pre mat has equal tile and block sizes
 template <Backend B, Device D, class T>
 EigensolverResult<T, D> eigensolver(blas::Uplo uplo, Matrix<T, D>& mat) {
   const SizeType size = mat.size().rows();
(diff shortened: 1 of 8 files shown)


include/dlaf/eigensolver/eigensolver.h
```diff
@@ -35,6 +35,15 @@ namespace dlaf::eigensolver {
 /// @param mat contains the Hermitian matrix A
 /// @param eigenvalues is a N x 1 matrix which on output contains the eigenvalues
 /// @param eigenvectors is a N x N matrix which on output contains the eigenvectors
```
msimberg, line 37: Some algorithms establish a precondition by talking about sizes in terms of `N` like here. Should we rather move this to a precondition and say `eigenvalues.rows() is equal to eigenvectors.rows()/cols()` or something like that? Or add the precondition _in addition_ to saying here that they're `N x 1` or `N x N`?

RMeli: I'd be in favor of moving this into a precondition. However, duplication of this information might lead to divergence.

albestro (approved):
LGTM

include/dlaf/eigensolver/eigensolver.h
rasolca: Not sure if doxigen likes this, but I would prefer to group the argument and its precondition to simplify the read. Moreover I would simplify the preconditions defining some variables. E.g. for the generalized eigensolver:
```
/// @param[in,out] mat_a contains the Hermitian matrix A, on output the value is destroyed
/// @pre mat_a is not distributed
/// @pre mat_a has size (N x N)
/// @pre mat_a has blocksize (NB x NB)
/// @pre mat_a has tilesize (NB x NB)
///
/// @param[in,out] mat_b contains the Hermitian positive definite matrix B on output the value is destroyed
/// @pre mat_b is not distributed
/// @pre mat_b has size (N x N)
/// @pre mat_b has blocksize (NB x NB)
/// @pre mat_b has tilesize (NB x NB)

/// @param[out] eigenvalues on output contains the eigenvalues
/// @pre eigenvalues is not distributed
/// @pre mat_b has size (N x 1)
/// @pre mat_b has blocksize (NB x 1)
/// @pre mat_b has tilesize (NB x 1)

/// @param[out] eigenvectors on output contains the eigenvectors
/// @pre eigenvectors is not distributed
/// @pre eigenvectors has size (N x N)
/// @pre eigenvectors has blocksize (NB x NB)
/// @pre eigenvectors has tilesize (NB x NB)
```
In this way it is much clearer what has to match among the different matrices.
What do you think?

rasolca: Exactly, giving "values" to the sizes is used to simplify writing the preconditions.
```
/// @pre mat_a has size (M x N)
...
/// @pre mat_b has size (N x M)
```
is immediately clear, while in the second case the sizes has to be reconstructed from the preconditions.
The first approach is more maintainable as if any of the precondition is relaxed it is easier to update the documentation.

msimberg: See https://github.com/eth-cscs/DLA-Future/pull/944/commits/54b6a7307d6502b52bd329553f1fe56d6ee7ed25 for potential new version. The grouping seems to work just fine:
2023-08-30-110239_1669x1221_scrot
Slightly unsure what would be best to put on preconditions for the grids, especially the 1d grids. Let me know what you think would be best.

rasolca: For the grid I would use "mat is distributed according to grid" as two grids with the same size might not be compatible (e.g. row-major vs col-major grid).

Note: `eigenvalues` is not distributed, each rank has a full copy of it.

msimberg: Updated all of `eigensolver/eigensolver.h` in https://github.com/eth-cscs/DLA-Future/pull/944/commits/5622f8bd804a0b57a306ba448df019d8e1b2b962. @rasolca, @albestro does that look good? If yes, I think that'll generalize nicely to the other algorithms.

rasolca: My only concern is about the grid part of the precondition.
As said in my previous comment:
`/// @pre mat is distributed on an (NG x MG) grid` is incorrect and I would keep the previous statement `/// @pre mat is distributed according to grid`.

The reason is that the first statement would allow:
- reordered grids (e.g. col vs row -major grids)
- grids built on top of reordered communicators
- grids built on top of communicators with different ranks

msimberg: Sorry about that. I never staged that change... One more try: https://github.com/eth-cscs/DLA-Future/issues/966.

include/dlaf/eigensolver/bt_reduction_to_band.h
```diff
@@ -64,15 +68,19 @@ void bt_reduction_to_band(const SizeType b, Matrix<T, device>& mat_c, Matrix<con
 /// (HH(j) is the House-Holder transformation (I - v tau vH)
 /// defined by the j-th element of tau and the HH reflector stored in the j-th column of the matrix V.
 ///
-/// @param mat_c contains the (m x n) matrix C (blocksize (mb x nb)), while on exit it contains Q C.
+/// @param mat_c contains the (m x n) matrix C (blocksize (mb x nb)), while on exit it contains Q C
+/// @pre @p mat_c is distributed according to @p grid
```
msimberg, line 72: In the old docs it said only `mat_c is distributed` (not according to grid). I'm assuming this was just unintentional, or can it be distributed some other way?

msimberg:
All right, I've now updated all the algorithms to talk about explicit matrix and block sizes.

include/dlaf/eigensolver/bt_reduction_to_band.h
rasolca: according to grid is correct

rasolca (approved):

PR #948, opened 2023-08-10
Update `gpu2ranks` scripts

msimberg:
On LUMI, it seems like exporting `CUDA_VISIBLE_DEVICES` alongside `ROCR_VISIBLE_DEVICES` will make no GPUs visible. This PR does the simple change of separating the launch scripts to be CUDA/HIP-specific. We may consider some adding some logic in the future to detect which one we're dealing with but I'm leaving this as it is for now.

Thanks @albestro for the investigation!

As a flyby this removes leftover `"Run command"` entry that was wrong, but also unused because it was overridden by the correct command later.


albestro (approved):
Thanks @msimberg! Not tested, but LGTM 😉

msimberg:
cscs-ci run

Issue #950, opened 2023-08-10
DLAF_WITH_PRECOMPILED_HEADERS=on fails after introduction of the C API


RMeli:
Reported by @albestro:

```
ialberto@daint101:/project/csstaff/ialberto/workspace/dla-future.master/build-gpu> cat ./configure.sh
CC=cc CXX=CC cmake ..\
  -DCMAKE_EXPORT_COMPILE_COMMANDS=on \
  -DCMAKE_BUILD_TYPE=Release\
  -DDLAF_WITH_PRECOMPILED_HEADERS=on\
  -DDLAF_WITH_MKL=on\
  -DDLAF_WITH_CUDA=on\
  -DDLAF_MPI_PRESET=slurm\
  -DMPIEXEC_NUMCORES=24\
  -DDLAF_TEST_RUNALL_WITH_MPIEXEC=on $@
ialberto@daint101:/project/csstaff/ialberto/workspace/dla-future.master/build-gpu> ./configure.sh
...
-- Configuring done (18.2s)
CMake Error in test/unit/c_api/eigensolver/CMakeLists.txt:
  Unable to resolve full path of PCH-header
  '/project/csstaff/ialberto/workspace/dla-future.master/build-gpu/src/CMakeFiles/dlaf.pch_exe.dir/cmake_pch.h'
  assigned to target test_eigensolver_c_api, although its path is supposed to
  be known!


CMake Error in test/unit/c_api/factorization/CMakeLists.txt:
  Unable to resolve full path of PCH-header
  '/project/csstaff/ialberto/workspace/dla-future.master/build-gpu/src/CMakeFiles/dlaf.pch_exe.dir/cmake_pch.h'
  assigned to target test_cholesky_c_api, although its path is supposed to be
  known!


CMake Error in test/unit/c_api/grid/CMakeLists.txt:
  Unable to resolve full path of PCH-header
  '/project/csstaff/ialberto/workspace/dla-future.master/build-gpu/src/CMakeFiles/dlaf.pch_exe.dir/cmake_pch.h'
  assigned to target test_grid_c_api, although its path is supposed to be
  known!


-- Generating done (2.8s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
ialberto@daint101:/project/csstaff/ialberto/workspace/dla-future.master/build-gpu> find . -name "cmake_pch.*"
./src/CMakeFiles/dlaf.pch_exe.dir/cmake_pch.hxx
[...]

PR #955, opened 2023-08-14
Fix precompiled headers with C files

msimberg:
Works around https://github.com/eth-cscs/DLA-Future/issues/950 with https://stackoverflow.com/questions/68735830/cmake-precompiled-headers-issue-with-mixed-c-c-project. Note that it's not enough to have a C header in the list of precompiled headers. The executable/library also has to have at least one C file for the `.h` precompiled header to be generated. In C-mode `-Wpedantic` also warns about an empty translation unit so I've added a dummy function declaration to the `dummy.c` file as well.


RMeli (approved):
Tested locally, LGTM. Thanks @msimberg for fixing this.

albestro (approved):

rasolca (approved):

PR #959, opened 2023-08-18
Correctly move callable `f` in `Unwrapping` call operator

msimberg:
The previous implementation was potentially buggy for cases when the `Unwrapping` call operator is called more than once, since `f` was for the actual call.

As far as I could tell we're not actually affected by this bug since we only ever call the `Unwrapping` call operator once.

This splits the call operator into two `&&` and `&` qualified overloads, with the latter not moving `f`. Note that the latter is not `const&` because that would fail with wrapped call operators that are not const-qualified. The `Duplicate` call operator was one such call operator. I've changed the latter to be const-qualified but I've also left the `Unwrapping` call operator as `&` in case we introduce other call operators that actually need to be non-const.

@aurianer you can apply the same change in #930 to `ConsumeRvalues`.

### DIFF
--- a/include/dlaf/common/unwrap.h
+++ b/include/dlaf/common/unwrap.h
@@ -56,9 +56,15 @@ struct Unwrapping {
   std::decay_t<F> f;
 
   template <typename... Ts>
-  auto operator()(Ts&&... ts)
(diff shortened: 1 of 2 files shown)


msimberg:
cscs-ci run

msimberg:
cscs-ci run

rasolca (approved):

albestro (approved):

PR #978, opened 2023-09-08
Use out-of-place trmm when available

msimberg:
I'll try to summarize the various stages of trmm evolution since it's a bit complicated to follow in the code:

| version  | `rocblas_Xtrmm` | `rocblas_Xtrmm_outofplace` |
| -- | -- | -- |
| v < 5.0.0  | only exists as two-parameter version | does not exist |
| 5.0.0 <= v < 5.6.0 | only exists as two-parameter version | only exists as three-parameter version which supports both in- and out-of-place (same as cuBLAS) |
| 5.6.0 <= v < develop | Same as above with supposedly a `ROCBLAS_V3` definition that should enable the three-parameter version, but which only seems to have an effect in documentation... | Same as above, but deprecated |
| develop <= v | only exists as three-parameter version which supports both in- and out-of-place (same as cuBLAS) | does not exist |

Given that

1. 5.6.0 is barely out (and not yet available in spack: https://github.com/spack/spack/pull/39673).
2. `trmm_outofplace` was deprecated in 5.6.X
3. We don't know yet exactly which version is going to actually remove `trmm_outofplace`

I would probably for the moment actually not put an upper bound on when to switch back from `trmm_outofplace` to `trmm`. It's a guessing game I suppose until they actually release 5.7.0 or 6.0.0 but it seems safer to assume that `trmm_outofplace` is going to actually be removed in the next major release rather than one minor release after it was deprecated.

I have _not_ tested this with rocblas 4.X (I'm not even sure if it works without these changes, if I remember correctly CMake support was badly broken before 5.something) but in theory it should work. I don't how much we care about that?

I also have not run large-scale runs with this change, so I don't know if it affects performance (smaller tests say no).


include/dlaf/blas/tile.h
```diff
@@ -137,22 +137,31 @@ DLAF_MAKE_GPUBLAS_SYHE_OP(Her2k, r2k);
...
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
 #elif defined(__GNUC__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #endif
-#endif
+
+// TODO: What will be the upper bound?
+#if HIP_VERSION < 50000000 || 50700000 <= HIP_VERSION
```
msimberg, line 153: I would probably make this simply
```suggestion
#if HIP_VERSION < 50000000
```
for now...

rasolca:
> I have _not_ tested this with rocblas 4.X (I'm not even sure if it works without these changes, if I remember correctly CMake support was badly broken before 5.something) but in theory it should work. I don't how much we care about that?

Neglect 4.X and require >= 5.

rasolca:
Please run the tests with the CI image to be sure that everything works correctly.

rasolca:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
The tests pass on hohgant. I've yet to verify that performance is unchanged.

msimberg:
There seems to be no measurable impact from this change.

rasolca (approved):

rasolca:
cscs-ci run

rasolca:
@msimberg This PR is based on a very old master which makes the CI fails.
Please update it.

PR #979, opened 2023-09-11
Remove `strategy: depend` clause as already in .run_common CI step

aurianer:
`strategy: depend` is already in the `.run_common` step. Each of the test steps already extend `.run_common` so I'm removing the duplicated clause


msimberg:
cscs-ci run

rasolca (approved):

PR #989, opened 2023-09-22
Fix wrong environment variable name in setup for `DLAF_BT_BAND_TO_TRIDIAG_HH_APPLY_GROUP_SIZE`

msimberg:
The `DLAF` prefix is added automatically by `updateConfigurationValue` so it shouldn't be included when calling `updateConfigurationValue`.

PR #990, opened 2023-09-22
Print tune parameters with `--dlaf:print-config`

msimberg:
From the `--dlaf:print-config` perspective, the tune parameters and config options are the same. Should we print them together (in one block) or separate them (with a heading?).

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -277,6 +277,8 @@ void initialize(const pika::program_options::variables_map& vm, const configurat
   if (vm.count("dlaf:print-config") > 0) {
     std::cout << "DLA-Future configuration options:" << std::endl;
     std::cout << cfg << std::endl;
+    std::cout << "DLA-Future tune parameters at startup:" << std::endl;
+    std::cout << getTuneParameters() << std::endl;
     std::cout << std::endl;
   }
 
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -155,6 +155,9 @@ struct parseFromCommandLine {
 template <class T>
 void updateConfigurationValue(const pika::program_options::variables_map& vm, T& var,
                               const std::string& env_var, const std::string& cmdline_option) {
+  DLAF_ASSERT(env_var.find("DLAF") == std::string::npos, env_var);
+  DLAF_ASSERT(cmdline_option.find("dlaf") == std::string::npos, cmdline_option);
+
   const std::string dlaf_env_var = "DLAF_" + env_var;
   char* env_var_value = std::getenv(dlaf_env_var.c_str());
   if (env_var_value) {
@@ -188,6 +191,9 @@ void updateConfiguration(const pika::program_options::variables_map& vm, configu
   cfg.mpi_pool = (pika::resource::pool_exists("mpi")) ? "mpi" : "default";
 
   // update tune parameters
+  //
+  // NOTE: Environment variables should omit the DLAF_ prefix and command line options the dlaf: prefix.
+  // These are added automatically by updateConfigurationValue.
   auto& param = getTuneParameters();
   updateConfigurationValue(vm, param.red2band_panel_nworkers, "RED2BAND_PANEL_NWORKERS",
                            "red2band-panel-nworkers");
--- a/src/tune.cpp
+++ b/src/tune.cpp
@@ -19,6 +19,7 @@ TuneParameters& getTuneParameters() {
 }
 
 std::ostream& operator<<(std::ostream& os, const TuneParameters& params) {
+  os << "  red2band_panel_nworkers = " << params.red2band_panel_nworkers << std::endl;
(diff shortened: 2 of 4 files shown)


src/init.cpp
```diff
@@ -277,6 +277,7 @@ void initialize(const pika::program_options::variables_map& vm, const configurat
   if (vm.count("dlaf:print-config") > 0) {
     std::cout << "DLA-Future configuration options:" << std::endl;
     std::cout << cfg << std::endl;
+    std::cout << getTuneParameters() << std::endl;
```
msimberg, line 280:
```suggestion
    std::cout << cfg;
    std::cout << getTuneParameters() << std::endl;
```
or
```suggestion
    std::cout << cfg << std::endl;
    std::cout "DLA-Future tune parameters:" << std::endl;
    std::cout << getTuneParameters() << std::endl;
```
?

msimberg:
cscs-ci run

src/init.cpp
rasolca: Option 3

msimberg: Updated to use option 3.

msimberg:
cscs-ci run

RMeli (approved):
LGTM!

Do we need a comment in `include/dlaf/tune.h` to remember to update `operator<<` when a new parameter is added?

msimberg:
> LGTM!
> 
> Do we need a comment in `include/dlaf/tune.h` to remember to update `operator<<` when a new parameter is added?

That's a good idea. See https://github.com/eth-cscs/DLA-Future/pull/990/commits/9e479c0e4a35cca77ea9c66611330e5cd5709aad. Does that look like a useful comment to you? I also added the same comment for configuration options.

RMeli (approved):
> Does that look like a useful comment to you?

Very useful, thanks. This is the kind of thing I tend to forget ;)

src/tune.cpp
```diff
@@ -18,4 +18,17 @@ TuneParameters& getTuneParameters() {
   return params;
 }
 
+std::ostream& operator<<(std::ostream& os, const TuneParameters& params) {
```
albestro, line 21: Is `red2band_panel_nworkers` missing, right?

albestro (changes requested):

src/tune.cpp
msimberg: Yep... fixed. Thanks!

msimberg:
cscs-ci run

albestro (approved):
I find the comment you added really useful!

We might also think to extend it adding a "guideline" of all the places that need to be updated when a parameter is added (e.g. add an entry in the option, environment variable, ...)

rasolca (approved):

msimberg:
cscs-ci run

msimberg:
This does not yet contain #989 so if all goes well this PR should now fail in CI with the assertions added in https://github.com/eth-cscs/DLA-Future/pull/990/commits/ddfbf0a5bfee299e4cee18408e0fe3c8db1d00c1. Will rebase once (if) it fails.

src/init.cpp
```diff
@@ -155,6 +155,9 @@ struct parseFromCommandLine {
 template <class T>
 void updateConfigurationValue(const pika::program_options::variables_map& vm, T& var,
                               const std::string& env_var, const std::string& cmdline_option) {
+  DLAF_ASSERT(env_var.find("DLAF") == std::string::npos, env_var);
+  DLAF_ASSERT(cmdline_option.find("dlaf") == std::string::npos, cmdline_option);
```
msimberg, line 159: I made these as conservative as possible at the moment, meaning `dlaf`/`DLAF` isn't allowed _anywhere_ in the strings. This is not meant to be a hard rule for the future, it only reflects the current state. If there's a need for having `DLAF`/`dlaf` somewhere in the names in the future we can relax this to e.g. only checking for prefixes.

msimberg:
> This does not yet contain #989 so if all goes well this PR should now fail in CI with the assertions added in [ddfbf0a](https://github.com/eth-cscs/DLA-Future/commit/ddfbf0a5bfee299e4cee18408e0fe3c8db1d00c1). Will rebase once (if) it fails.

https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5204882236#L1886

> 59: [ERROR] /DLA-Future/src/init.cpp:158 : void dlaf::internal::updateConfigurationValue(const pika::program_options::variables_map&, T&, const string&, const string&) [with T = long int; std::string = std::__cxx11::basic_string<char>]
> 59: env_var.find("DLAF") == std::string::npos
> 59: DLAF_BT_BAND_TO_TRIDIAG_HH_APPLY_GROUP_SIZE
> 59: 
> 59: terminate called without an active exception

msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #993, opened 2023-09-26
Wrap communicators in `RoundRobin` and `Pipeline` in `CommunicatorGrid`

msimberg:
Nothing to see here yet. Only opening for CI at this point. Better description to follow.

To do:
- [x] Add tests, currently none
- [x] Check debug bloat from newly added template instantiations
- [x] Check performance (https://confluence.cscs.ch/display/SCISWDEV/2023-11+communicator+grid+refactoring)
- [x] Decide on default number of pipelines to use (currently 3 to allow some overlap of algorithms, while still giving better performance than before these changes; the number is configurable at runtime)


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

ci/docker/debug-cuda-scalapack.yaml
```diff
@@ -25,6 +25,8 @@ spack:
       variants:
         - 'build_type=Release'
     pika:
+      version:
+        - main
```
msimberg, line 29: To do: require 0.19.0 once it's out.

include/dlaf/common/pipeline.h
```diff
@@ -32,25 +34,75 @@ class Pipeline {
...
+  ~Pipeline() {
+    release_parent_pipeline();
+  }
+
   /// Enqueue for the resource.
   ///
   /// @return a sender that will become ready as soon as the previous user releases the resource.
   /// @pre valid()
-  Sender operator()() {
+  ReadWriteSender operator()() {
```
msimberg, line 69: Rename `readwrite`?

include/dlaf/communication/kernels/p2p_allsum.h
```diff
@@ -30,6 +30,26 @@
 
 namespace dlaf::comm {
 
+namespace internal {
+// TODO: Clean this up into something a bit more sustainable. We were forcing senders to be
```
msimberg, line 34: To do: clean up.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

include/dlaf/tune.h
```diff
@@ -73,6 +77,11 @@ struct TuneParameters {
   SizeType eigensolver_min_band = 100;
   SizeType band_to_tridiag_1d_block_size_base = 8192;
   SizeType bt_band_to_tridiag_hh_apply_group_size = 64;
+
+  // TODO: The default parameter of CommunicatorGrid triggers the construction of TuneParameters, which
```
msimberg, line 81: To do.

test/include/dlaf_test/comm_grids/grids_6_ranks.h
```diff
@@ -78,14 +78,16 @@ class CommunicatorGrid6RanksCAPIEnvironment : public CommunicatorGrid6RanksEnvir
       if (world.size() == 1)
         return;
 
-      comm_grids.emplace_back(world, 3, 2, common::Ordering::RowMajor);
-      comm_grids.emplace_back(world, 3, 2, common::Ordering::ColumnMajor);
+      // TODO
```
msimberg, line 81: To do. See construction of `TuneParameters`.

test/unit/c_api/grid/test_grid_c_api.cpp
```diff
@@ -43,10 +43,13 @@ TEST(GridTest, GridScaLAPACKOrdering) {
 #endif
 
 TEST(GridTest, GridDLAFOrdering) {
+  // TODO
```
msimberg, line 46: To do.

test/unit/communication/test_all_reduce.cpp
```diff
@@ -27,8 +27,11 @@ using AllReduceInPlaceTest = SplittedCommunicatorsTest;
 
 using TypeParam = std::complex<double>;
 
+// TODO
```
msimberg, line 30: To do.

test/unit/communication/test_communicator_grid.cpp
```diff
@@ -19,6 +19,9 @@ using namespace dlaf::comm;
 using dlaf::common::Ordering;
 
 const auto valid_orderings = ::testing::Values(Ordering::RowMajor, Ordering::ColumnMajor);
+// TODO: Update TuneParameters to not fail if pika isn't initialized? Allow getting only the number of
```
msimberg, line 22: To do.

test/unit/communication/test_reduce.cpp
```diff
@@ -28,8 +28,11 @@ using ReduceInPlaceSplitAPITest = dlaf::comm::test::SplittedCommunicatorsTest;
 
 using TypeParam = std::complex<double>;
 
+// TODO
```
msimberg, line 31: To do.

msimberg:
cscs-ci run

include/dlaf/tune.h
```diff
@@ -63,16 +67,31 @@ struct TuneParameters {
...
   // - getOptionsDescription to add a corresponding command line option
+
+  TuneParameters() {
+    // Some parameters require the pika runtime to be initialized since they depend on the number of
+    // threads used by the runtime. We initialize them separately in the constructor after checking that
+    // pika is initialized.
+
+    // TODO: Check upfront whether pika has been initialized to be able to give a better error message.
+    // If pika has not been initialized getting the default thread pool will either trigger an assertion
+    // or a segfault.
```
msimberg, line 78: I've opened an issue for this on the pika side: https://github.com/pika-org/pika/issues/808.

msimberg:
cscs-ci run

msimberg:
To dos in the PR description are still relevant, but I think this would benefit from some reviews already now. Some notes:
- Preliminary benchmarks show that as you decrease the number of pipelines used in the communicator grid (until a minimum of two because of algorithmic requirements), performance of the eigensolvers approaches the "sync" benchmark results. The non-eigensolver miniapps show no dependence on the number of pipelines which makes sense.
- In https://github.com/eth-cscs/DLA-Future/pull/993/commits/a44f1c63a84e9fcae7fa080f01565091f82ac72b I've made `CommunicatorGrid` move only. If we suspect that it's going to cause annoying conflicts I can simply revert that commit to make the copy constructor and assignment simply deprecated.
- The number of pipelines to use in a communicator grid is a tune parameter now. However, constructing the tune parameters requires that pika is initialized, which means that communicator grid cannot use the default number of pipelines in e.g. tests unless pika is initialized. I've made many tests `MPIPIKA` tests to work around this. For the C API tests I've explicitly specified the number of pipelines to avoid the issue since they also test pika/DLAF initialization/shutdown.

msimberg:
Binary sizes have evolved like follows. I'm comparing the latest ancestor on `master` (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/commit/e778dc99130efb9e332c6126b98613531b3c7c9c) with the current latest commit on this branch (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/commit/bb4b5967ea86368a1076aa7c8a1736408be802c6). For the shared libraries the sizes have slightly grown.

`master` (including tests: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5288434329#L3084):
```
-rwxr-xr-x 1 root root 1.5G Oct 13 12:36 src/libDLAF.so
-rwxr-xr-x 1 root root  11M Oct 13 12:34 src/libdlaf.auxiliary.so
-rwxr-xr-x 1 root root 3.9M Oct 13 12:35 src/libdlaf.c_api.so
-rwxr-xr-x 1 root root 348M Oct 13 12:34 src/libdlaf.core.so
-rwxr-xr-x 1 root root 612M Oct 13 12:35 src/libdlaf.eigensolver.so
-rwxr-xr-x 1 root root  98M Oct 13 12:34 src/libdlaf.factorization.so
-rwxr-xr-x 1 root root 161M Oct 13 12:34 src/libdlaf.multiplication.so
-rwxr-xr-x 1 root root 134M Oct 13 12:34 src/libdlaf.permutations.so
-rwxr-xr-x 1 root root 106M Oct 13 12:34 src/libdlaf.solver.so
-rwxr-xr-x 1 root root 244M Oct 13 12:35 src/libdlaf.tridiagonal_eigensolver.so
```

This branch (including tests: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5289553001#L3095):
```
-rwxr-xr-x 1 root root 1.6G Oct 13 14:53 src/libDLAF.so
-rwxr-xr-x 1 root root  11M Oct 13 14:50 src/libdlaf.auxiliary.so
-rwxr-xr-x 1 root root 5.1M Oct 13 14:52 src/libdlaf.c_api.so
-rwxr-xr-x 1 root root 418M Oct 13 14:50 src/libdlaf.core.so
-rwxr-xr-x 1 root root 622M Oct 13 14:52 src/libdlaf.eigensolver.so
-rwxr-xr-x 1 root root 100M Oct 13 14:50 src/libdlaf.factorization.so
-rwxr-xr-x 1 root root 164M Oct 13 14:51 src/libdlaf.multiplication.so
-rwxr-xr-x 1 root root 135M Oct 13 14:51 src/libdlaf.permutations.so
-rwxr-xr-x 1 root root 108M Oct 13 14:50 src/libdlaf.solver.so
-rwxr-xr-x 1 root root 248M Oct 13 14:51 src/libdlaf.tridiagonal_eigensolver.so
```
There's a small increase in size in most libraries, and a larger increase in size in `libdlaf.core.so`, presumably because of the added instantiations.

See the full links above for a comparison of the test binary sizes. I think these are still within reason, but not ideal. What do you think?

Unrelated to the changes here, but I noticed the `with_temporary_tile` test binary is impressively large. I opened an issue to potentially look into that later, but it's low priority: https://github.com/eth-cscs/DLA-Future/issues/1013.

albestro (commented):
Nice and clean work!

Still have to fully understand the use-case for `pipeline.read()` and a I will have to study a bit more in detail a couple of new things that you used (namely `WithResultOf` and `std::exchange`).

Thanks!

include/dlaf/eigensolver/tridiag_solver/rot.h
```diff
@@ -205,9 +205,10 @@ void applyGivensRotationsToMatrixColumns(const SizeType i_begin, const SizeType
 /// @pre mat is distributed along rows the same way as comm_row
 /// @pre memory layout of @p mat is column major.
 template <class T, Device D, class GRSender>
-void applyGivensRotationsToMatrixColumns(comm::Communicator comm_row, const comm::IndexT_MPI tag,
-                                         const SizeType i_begin, const SizeType i_end,
-                                         GRSender&& rots_fut, Matrix<T, D>& mat) {
+void applyGivensRotationsToMatrixColumns([[maybe_unused]] comm::CommunicatorGrid& grid,
+                                         common::Pipeline<comm::Communicator>& comm_row_chain,
```
albestro, line 209: This comment just for highlighting it.

In the past we already pointed out this minor "problem" of not being able to access metadata of things "enclosed" in a sender (at the time it was a future, but it does not change much).

I don't think it is worth to over-complicate things, and I'm ok with this alternative.

rasolca (commented):
https://github.com/eth-cscs/DLA-Future/pull/993/files#diff-9fb38d01eaaba80087b718b707650bf8e68af04fe5c2d3acd4d1319a1b159e9eR1502-R1504

How do we deal with internal algorithm implementations?
1) We pass the grid and we get the pipeline from it.
   (In the eigensolver gemm case (link above) it will mean that each gemm has a different pipeline instead of the same (current implementation).
2) The required pipelines are added to the arguments (as for the eigensolver gemm case (link above)).
   Drawback: The internal algorithm has still access to other pipelines via the grid
3) Same as 2, but divide grid in grid_info and {communicators + pipelines} and the internal implementations require a grid_info and the pipelines needed.

I would go for 3.

test/unit/communication/test_communicator_grid.cpp
```diff
@@ -249,3 +206,49 @@ TEST_P(CommunicatorGridTest, Rank) {
 }
 
 INSTANTIATE_TEST_SUITE_P(Rank, CommunicatorGridTest, valid_orderings);
+
+TEST_P(CommunicatorGridTest, RoundRobin) {
+  Communicator world(MPI_COMM_WORLD);
+  auto grid_dims = computeGridDims(NUM_MPI_RANKS);
+  const std::vector<std::size_t> test_npipelines{1};  //, 3, 10};
```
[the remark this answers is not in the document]
msimberg: Absolutely, thanks for spotting this. I've reenabled the `3` and `10` tests.

src/communication/communicator_grid.cpp
```diff
@@ -8,13 +8,16 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 
+#include <pika/config.hpp>
```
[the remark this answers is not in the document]
msimberg: I think it's a leftover from when I used `with_result_of` from pika initially, or similar... It doesn't belong here so I'm removing it.

test/unit/common/test_pipeline.cpp
```diff
@@ -157,3 +194,591 @@ TEST(PipelineDestructor, DestructionWithDependency) {
...
+  auto checkpoint2 = sub_pipeline() | ex::then([&](auto wrapper) {
+                       EXPECT_EQ(wrapper.get().get(), 28);
+                       EXPECT_TRUE(first_access_done);
+                       EXPECT_TRUE(second_access_done);
+                       EXPECT_FALSE(third_access_done);
+                       third_access_done = true;
+                     });
+
+  tt::sync_wait(ex::when_all(std::move(checkpoint0), std::move(checkpoint1), std::move(checkpoint2)));
+}
```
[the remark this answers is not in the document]
msimberg: Good point. I've added a check of `third_access_done` but also of `first...` and `second...` as sanity checks to the end of all the `test_pipeline` tests that do a similar check, not only this one.

include/dlaf/common/round_robin.h
```diff
@@ -13,28 +13,46 @@
...
       pool_.emplace_back(args...);
   }
 
+  RoundRobin(RoundRobin&&) = default;
+  RoundRobin(const RoundRobin&) = delete;
+  RoundRobin& operator=(RoundRobin&&) = default;
+  RoundRobin& operator=(const RoundRobin&) = delete;
+
   T& currentResource() {
+    DLAF_ASSERT(curr_index_ < pool_.size(), curr_index_, pool_.size());
```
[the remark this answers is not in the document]
msimberg: Updated this, https://github.com/eth-cscs/DLA-Future/pull/993/files/bb4b5967ea86368a1076aa7c8a1736408be802c6#diff-97018c62681fbe88215cbdd275c5177efbb4d7af6ed889fcd91cfbf11871a2d7, and the assertion for the `pool_` not being empty to `DLAF_ASSERT_HEAVY`. Or did you consciously not comment on `DLAF_ASSERT(!pool_.empty(), "")` because you'd prefer not to use `HEAVY` there?

ci/docker/debug-cuda-scalapack.yaml
msimberg: Thanks, I did update the requirement in `CMakeLists.txt` and `package.py`, but forgot to remove these :upside_down_face:. These are back to normal now.

include/dlaf/common/pipeline.h
msimberg: Renamed in the latest commits. However, I also kept a deprecated `operator()` in https://github.com/eth-cscs/DLA-Future/pull/993/commits/13c1d2a2c40248c8191b6f36d75d6b315f5366de in case it's useful for @albestro's changes. I'll remove it if we don't see a need for it.

include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -1462,8 +1463,7 @@ void mergeDistSubproblems(comm::CommunicatorGrid grid,
   //
   // Note: i_split is unique
   const comm::IndexT_MPI tag = to_int(i_split);
-  applyGivensRotationsToMatrixColumns(grid.rowCommunicator(), tag, i_begin, i_end, std::move(rots),
-                                      ws.e0);
+  applyGivensRotationsToMatrixColumns(grid, row_task_chain, tag, i_begin, i_end, std::move(rots), ws.e0);
```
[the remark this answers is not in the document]
msimberg: Discussed this with @albestro briefly, with his argument for not hiding it inside the algorithm being that it's clear from the algorithm structure that `applyGivensRotationsToMatrixColumns` requires a sub pipeline (please correct me if I'm misrepresenting your view on this).

My counterargument is that it is not the caller's responsibility to know whether `applyGivensRotationsToMatrixColumns` requires a sub pipeline (schedules communication in subtasks) or if it suffices to pass a non-sub pipeline (all communication scheduled directly). If we call `applyGivensRotationsToMatrixColumns` with `row_task_chain.sub_pipeline()` and then change the implementation we may have taken a sub pipeline unnecessarily if we don't remember to update the call. It may be unlikely that we forget to update it, but given the two choices I don't see a clear benefit of moving the call to `sub_pipeline` out of the function.

msimberg:
cscs-ci run

include/dlaf/eigensolver/tridiag_solver/merge.h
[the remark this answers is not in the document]
msimberg: Decided to keep the `sub_pipeline` call inside the algorithm.

include/dlaf/eigensolver/tridiag_solver/rot.h
msimberg: Closing in favour of the discussion starting at https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5469465492. We will change this to not pass the full `CommunicatorGrid`.

include/dlaf/common/round_robin.h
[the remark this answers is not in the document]
msimberg: Updated this and the the other assertions in `round_robin.h` based on discussion. Summary: keep the first assertion in `currentResource` and `nextResource` as `DLAF_ASSERT` because they are runtime preconditions (the vector can be empty), but make the second, dependent, assertion in `nextResource` `DLAF_ASSERT_HEAVY` (because if the first assertion does not fail, the second should also only fail if the implementation is wrong, not by user error).

include/dlaf/permutations/general.h
```diff
@@ -148,7 +148,7 @@ void permute(comm::CommunicatorGrid grid, common::Pipeline<comm::Communicator>&
 /// This overload clones the row communicator (if Coord::Col) or column communicator (if Coord::Row) of
 /// @p grid internally.
 template <Backend B, Device D, class T, Coord coord>
-void permute(comm::CommunicatorGrid grid, SizeType i_begin, SizeType i_end,
+void permute(comm::CommunicatorGrid& grid, SizeType i_begin, SizeType i_end,
              Matrix<const SizeType, D>& perms, Matrix<const T, D>& mat_in, Matrix<T, D>& mat_out) {
   common::Pipeline<comm::Communicator> sub_task_chain(grid.subCommunicator(orthogonal(coord)).clone());
```
msimberg, line 153: To do: this should not use a bare `clone`. Also check other places for leftover `clone`s.

include/dlaf/common/index2d.h
```diff
@@ -24,6 +24,9 @@
 
 namespace dlaf {
 
+// TODO: Combine with below? What to call this?
+enum class TODOCoord { Row, Col, Full };
```
msimberg, line 28:
```suggestion
enum class CommunicatorType { Row, Col, Full };
```
? Other suggestions for names?

And possibly
```
enum class CommunicatorType { Row = static_cast<int>(Coord::Row), Col = static_cast<int>(Coord::Col), Full };
```

include/dlaf/communication/broadcast_panel.h
```diff
@@ -53,10 +54,11 @@ std::pair<SizeType, comm::IndexT_MPI> transposedOwner(const matrix::Distribution
 ///                     on other ranks it is the destination panel
 /// @param serial_comm  where to pipeline the tasks for communications.
 /// @pre Communicator in @p serial_comm must be orthogonal to panel axis
-template <class T, Device D, Coord axis, matrix::StoreTransposed storage,
+template <class T, Device D, Coord axis, TODOCoord C2, matrix::StoreTransposed storage,
           class = std::enable_if_t<!std::is_const_v<T>>>
 void broadcast(comm::IndexT_MPI rank_root, matrix::Panel<axis, T, D, storage>& panel,
-               common::Pipeline<comm::Communicator>& serial_comm) {
+               comm::CommunicatorPipeline<C2>& serial_comm) {
+  // TODO: Constrain axis vs C2?
```
msimberg, line 61: To do.

include/dlaf/communication/broadcast_panel.h
```diff
@@ -68,11 +70,26 @@ void broadcast(comm::IndexT_MPI rank_root, matrix::Panel<axis, T, D, storage>& p
...
     if (rank == rank_root)
-      ex::start_detached(scheduleSendBcast(serial_comm(), panel.read(index)));
+      ex::start_detached(scheduleSendBcast(serial_comm.readwrite(), panel.read(index)));
     else
-      ex::start_detached(scheduleRecvBcast(serial_comm(), rank_root, panel.readwrite(index)));
+      ex::start_detached(scheduleRecvBcast(serial_comm.readwrite(), rank_root, panel.readwrite(index)));
+  }
+}
+
+// TODO: Move somewhere else?
```
msimberg, line 79: To do.

include/dlaf/communication/communicator_grid.h
```diff
@@ -16,97 +16,134 @@
...
+  CommunicatorPipeline<TODOCoord::Col> col_communicator_pipeline() {
+    return col_pipelines_.nextResource().sub_pipeline();
+  }
+
+  template <Coord C> auto communicator_pipeline() {
+    if constexpr (C == Coord::Row)
+      return row_communicator_pipeline();
+    else
+      return col_communicator_pipeline();
+  }
```
msimberg, line 124: `Coord` or `TODOCoord` or both as options?

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// TODO: Move this to separate header?
+/// TAG for strong-typing basic_coords.
+struct TAG_MPI;
+
+/// 2D index strong-typed for MPI.
+using Index2D = common::Index2D<IndexT_MPI, TAG_MPI>;
+/// 2D size strong-typed for MPI.
+using Size2D = common::Size2D<IndexT_MPI, TAG_MPI>;
```
msimberg, line 33: To do: move to common communication header?

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+  CommunicatorPipeline &operator=(const CommunicatorPipeline &) = delete;
+
+  /// Return the rank of the current process in the CommunicatorPipeline.
+  ///
+  /// @return the 2D coordinate representing the position in the grid that this
+  /// pipeline belongs to
+  Index2D rank() const noexcept { return rank_; }
+
+  /// Return the size of the grid.
+  Size2D size() const noexcept { return size_; }
```
msimberg, line 77: Should these be independent or dependent on `TODOCoord`? Should there be two sets of functions: one for (potentially) 1D index/size (only `Row` and `Col`) and one for the full grid?

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+
+  /// Return rank in the grid with all ranks given the 2D index.
+  IndexT_MPI rankFullCommunicator(const Index2D &index) const noexcept {
+    return common::computeLinearIndex<IndexT_MPI>(
+        internal::FULL_COMMUNICATOR_ORDER, index, {size_.rows(), size_.cols()});
+  }
+
+  /// TODO: Update docs.
```
msimberg, line 85: To do.

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+  /// Enqueue for exclusive read-write access to the resource.
+  ///
+  /// @return a sender that will become ready as soon as the previous user
+  /// releases the resource.
+  /// @pre valid()
+  /// TODO: Rename?
```
msimberg, line 91: To do: `exclusive`? `serial`? something else?

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+    return pipeline_.readwrite();
+  }
+
+  /// Enqueue for shared read-only access to the resource.
+  ///
+  /// @return a sender that will become ready as soon as the previous user
+  /// releases the resource.
+  /// @pre valid()
+  /// TODO: Rename?
+  ReadOnlySender read() { return pipeline_.read(); }
```
msimberg, line 104: To do: `shared`? something else?

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+using CommunicatorPipelineReadOnlyWrapper =
+    typename internal::CommunicatorPipeline::ReadOnlyWrapper;
+using CommunicatorPipelineReadWriteWrapper =
+    typename internal::CommunicatorPipeline::ReadWriteWrapper;
+using CommunicatorPipelineReadOnlySender =
+    typename internal::CommunicatorPipeline::ReadOnlySender;
+using CommunicatorPipelineReadWriteSender =
+    typename internal::CommunicatorPipeline::ReadWriteSender;
+
+template <TODOCoord Coord> class CommunicatorPipeline {
```
msimberg, line 48: Default:
```suggestion
template <TODOCoord Coord = TODOCoord::Full> class CommunicatorPipeline {
```
?

include/dlaf/permutations/general.h
msimberg: This was the only `clone`. Updated to use the pipeline members of `COmmunicatorGrid`.

include/dlaf/communication/communicator_grid.h
msimberg: Keep only the current version since it doesn't make sense to parameterize with the option to choose `Full` (one would not need to mix `Full` with `Row`/`Col` anywhere).

include/dlaf/communication/communicator_pipeline.h
msimberg: Keeping it explicit with no default.

include/dlaf/communication/broadcast_panel.h
msimberg: Leaving where it is. It can't be a lambda with C++17 because of the template parameter.

include/dlaf/common/index2d.h
msimberg: I've renamed it to `CommunicatorType` but leaving the `static_cast`s out for now since there's no immediate need for it. If we need to ensure they use the same values we can enforce that later.

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+  /// belongs to.
+  Size2D size_2d() const noexcept {
+    return size_;
+  }
+
+  /// Return rank in the grid with all ranks given the 2D index.
+  IndexT_MPI rankFullCommunicator(const Index2D& index) const noexcept {
+    return common::computeLinearIndex<IndexT_MPI>(internal::FULL_COMMUNICATOR_ORDER, index,
+                                                  {size_.rows(), size_.cols()});
+  }
```
msimberg, line 85: I think these would need another pair of eyes. In particular, should e.g. `size` be dependent on the `CommunicatorType`:
```
IndexT_MPI size() const noexcept {
  if constexpr (communicator_type == CommunicatorType::Row) {
    return size_.cols();
  } else if (communicator_type == CommunicatorType::Col) {
    return size_.rows();
  } else {
    return size_.rows() * size_.cols();
  }
}
```
or is even that wrong and I should be getting it from the communicator itself during construction?

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
Ready for another round of reviews. Open questions are:
- Behaviour of rank/size member functions on `CommunicatorPipeline`. I'm not entirely sure I have the right behaviour there, see https://github.com/eth-cscs/DLA-Future/pull/993#pullrequestreview-1729404029.
- Naming of `read`/`readwrite` on `CommunicatorPipeline`. We don't have to rename them, and being consistent with `Matrix`/`Pipeline` is probably a good idea, but we have an opportunity to rename them to something more appropriate for communicators.
- Default number of pipelines. I currently have it at 3 to allow _some_ overlap between algorithms, while still getting most of the performance benefit of running with synchronization between algorithms.

msimberg:
cscs-ci run

include/dlaf/permutations/general.h
```diff
@@ -109,15 +110,16 @@ void permute(SizeType i_begin, SizeType i_end, Matrix<const SizeType, D>& perms,
 /// @pre @p mat_out has blocksize (NB x NB)
 /// @pre @p mat_out has tilesize (NB x NB)
 ///
-/// Note: The Pipeline<> API allows to use permute() within other algorithms without having to clone communicators
-///       internally.
+/// Note: The CommunicatorPipeline API allows to use permute() within other algorithms without having to
+///       clone communicators internally.
```
albestro, line 114: This comment here caught my eye, because it was not straightforward for me remember how this algorithm worked.

I did some checks, and with the help of @rasolca, we realised that the implementation fo this algorithm actually need some cleanup too. See the comment in the implementation for reference, and in case, we can remove this note here.

include/dlaf/permutations/general/impl.h
```diff
@@ -288,7 +289,7 @@ void all2allData(common::Pipeline<comm::Communicator>& sub_task_chain, int nrank
         pika::this_thread::experimental::sync_wait(ex::when_all_vector(std::move(all_comms)));
       };
 
-  ex::when_all(sub_task_chain(), std::forward<SendCountsSender>(send_counts_sender),
+  ex::when_all(sub_task_chain.readwrite(), std::forward<SendCountsSender>(send_counts_sender),
```
albestro, line 292: This implementation is currently exploiting a `pipeline.readwrite()` call to mimic what now can be done with a `sub_pipeline()`.

Indeed, currently we pass a sender of a Communicator, and within the task we "manually" reuse it, as if it was a sequence of `read()` calls on a `sub_pipeline()`.

albestro (changes requested):
About the "naming" of CommunicatorPipeline calls, my first thought is that `readwrite`/`read` are not meaningful at first sight for a Communicator resource.

You already pointed out something along the lines of `exclusive`/`shared`, which I think are quite on point.

include/dlaf/common/pipeline.h
```diff
@@ -32,25 +34,79 @@ class Pipeline {
...
+
+  Pipeline& operator=(Pipeline&& other) {
+    if (this != &other) {
+      pipeline = std::exchange(other.pipeline, std::nullopt);
+      nested_sender = std::exchange(other.nested_sender, std::nullopt);
+    }
+
+    return *this;
+  };
+
```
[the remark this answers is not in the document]
msimberg: Good catch, should definitely be `noexcept`. Updated.

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+  CommunicatorPipeline& operator=(const CommunicatorPipeline&) = delete;
+
+  /// Return the rank of the current process in the CommunicatorPipeline.
+  IndexT_MPI rank() const noexcept {
+    return rankFullCommunicator(rank_);
+  }
+
+  /// Return the size of the grid.
+  IndexT_MPI size() const noexcept {
+    return size_.rows() * size_.cols();
```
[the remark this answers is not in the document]
msimberg: Updated in https://github.com/eth-cscs/DLA-Future/pull/993/commits/c32ca21bdde62de20ba3f6f13d98a76152d25afe.

include/dlaf/communication/communicator_grid.h
```diff
@@ -16,47 +16,55 @@
...
   /// @param size with @p size[0] rows and @p size[1] columns,
   /// @param comm must be valid during construction.
-  CommunicatorGrid(Communicator comm, const std::array<IndexT_MPI, 2>& size, common::Ordering ordering)
-      : CommunicatorGrid(comm, size[0], size[1], ordering) {}
+  CommunicatorGrid(Communicator comm, const std::array<IndexT_MPI, 2>& size, common::Ordering ordering,
+                   std::size_t npipelines = getTuneParameters().communicator_grid_num_pipelines)
+      : CommunicatorGrid(comm, size[0], size[1], ordering, npipelines) {}
+
+  CommunicatorGrid(CommunicatorGrid&&) = default;
+  CommunicatorGrid& operator=(CommunicatorGrid&&) = default;
```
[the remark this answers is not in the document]
msimberg: Agreed. Updated in https://github.com/eth-cscs/DLA-Future/pull/993/commits/e411415f8394f7fdf4daebad41d2ee43fefade88.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -1036,9 +1039,9 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
     return {std::move(mat_trid), std::move(mat_v)};
   }
 
-  auto comm = ex::just(grid.fullCommunicator().clone());
+  auto mpi_chain = grid.full_communicator_pipeline();
   // Need a pipeline of comm for broadcasts.
-  common::Pipeline<comm::Communicator> comm_bcast(grid.fullCommunicator().clone());
+  auto mpi_chain_bcast = grid.full_communicator_pipeline();
```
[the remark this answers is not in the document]
msimberg: I added a comment in https://github.com/eth-cscs/DLA-Future/pull/993/commits/47adaa0e66a303256248e444b346fe1b65d93aa3. Could you please check if it explains the situation clearly enough for you?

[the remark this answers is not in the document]
rasolca: Someone might ask why you need 2 pipelines...
I would add in someway in the comment, 1 works 2 it's better...

rasolca (commented):
At this stage it looks good.

include/dlaf/eigensolver/band_to_tridiag/mc.h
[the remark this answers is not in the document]
msimberg: Good point, I'll add that as well!

include/dlaf/communication/communicator_pipeline.h
msimberg: Renamed to `exclusive`.

include/dlaf/communication/communicator_pipeline.h
msimberg: Renamed to `shared`.

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+  using ReadOnlyWrapper = typename PipelineType::ReadOnlyWrapper;
+  using ReadWriteWrapper = typename PipelineType::ReadWriteWrapper;
+  using ReadOnlySender = typename PipelineType::ReadOnlySender;
+  using ReadWriteSender = typename PipelineType::ReadWriteSender;
+
+  static constexpr CommunicatorType communicator_type = CT;
+
+  /// Create a CommunicatorPipeline by moving in the resource (it takes the
+  /// ownership).
+  explicit CommunicatorPipeline(Communicator comm, Index2D rank = {0, 0}, Size2D size = {0, 0})
```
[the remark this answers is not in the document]
msimberg: I've updated both to default to `{-1, -1}`. However, I noticed that `Index2D` defaults to `{-1, -1}` while `Size2D` doesn't have a default constructor. Do we want to default the size (here or in `index2d.h`) to `{0, 0}` instead?

test/unit/communication/test_comm_matrix.cpp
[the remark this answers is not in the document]
msimberg: Good point. As far as I can tell no. Should I update it to get the pipeline from the grid instead, or just remove the grid?

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -325,7 +325,8 @@ void QR_Tfactor<backend, device, T>::call(matrix::Panel<Coord::Col, T, device>&
   // at this point each rank has its partial result for each column
   // so, let's reduce the results (on all ranks, so that everyone can independently compute T factor)
   if (true)  // TODO if the column communicator has more than 1 tile...but I just have the pipeline
```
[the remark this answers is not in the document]
msimberg: https://github.com/eth-cscs/DLA-Future/pull/993/commits/c68e274ab4379b87cb3a184b8604d8f23a8e1497

include/dlaf/communication/communicator_pipeline.h
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  /// Return the 2D size of the current process in the grid that this pipeline
+  /// belongs to.
+  Size2D size_2d() const noexcept {
+    return size_;
+  }
+
+  /// Return rank in the grid with all ranks given the 2D index.
+  IndexT_MPI rankFullCommunicator(const Index2D& index) const noexcept {
```
rasolca, line 82: why not snake_case?

msimberg: Because of copy-paste. This comes from `CommunicatorGrid`. Good catch!

include/dlaf/permutations/general/impl.h
msimberg: I had a quick look at this and indeed it seems like it was an easy change. At least the permutations tests pass for me locally. This was changed in https://github.com/eth-cscs/DLA-Future/pull/993/commits/b11bbc37a9a88ecebd9764df21f08d22ce0076c8.

include/dlaf/permutations/general.h
msimberg: I changed the implementtion, but I'm not sure the comment is inaccurate even with the changed implementation (unless I'm misunderstanding what it refers to). I'm not sure the comment adds much either so I'd be fine to remove it as well. What would you prefer?

include/dlaf/communication/communicator_pipeline.h
msimberg: Updated in https://github.com/eth-cscs/DLA-Future/pull/993/commits/4927179823baf6ffa10357482e47390fab145ddf. Note that I'm not changing the camelCase members in `CommunicatorGrid` here.

include/dlaf/communication/communicator_pipeline.h
msimberg: I think I've answered my own question along the way. I'm now getting the 1d size and rank from the `Communicator` before putting it into a `Pipeline`. Does that sound right?

msimberg:
cscs-ci run

msimberg:
I've resolved what I could resolve, but there are still some open questions. Please have another look at the unresolved comments.

include/dlaf/eigensolver/band_to_tridiag/mc.h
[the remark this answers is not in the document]
albestro: A summary from your two comments (feel free to adjust it)

> The algorithm would still work if we only defined one pipeline below and used that for both p2p communication and broadcasts (since mpi_chain is used fully before mpi_chain_bcast, i.e. not interleaved).
> However, using two pipelines is beneficial since it allows the two different types of communications to be submitted concurrently. This is true in case there are actually two resources (pipelines.size() = 2), otherwise it will fallback to the "single" pipeline case anyway (ending up serialising the calls).

test/unit/communication/test_comm_matrix.cpp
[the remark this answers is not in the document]
albestro: I would just drop the `grid`...

include/dlaf/communication/communicator_pipeline.h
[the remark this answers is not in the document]
albestro: IMO `{-1, -1}` is more meaningful than `{0, 0}`, since it immediately tells you that the coordinate is invalid and, AFAIU, when you pass a Communicator not obtained from a Grid, both `rank` and `size` should not be used at all, and being invalid feel safer to me.

include/dlaf/permutations/general.h
albestro: I'm not sure, but I would remove it in any case. I mean, otherwise we should write this to every algorithm.

With your work in this PR, I think it should be "clear" from the arguments what guarantees a function gives or not. Isn't it?

msimberg: Yep, I agree!

msimberg: https://github.com/eth-cscs/DLA-Future/pull/993/commits/286628ba436c1b00e295cf4254af92034a81ee4f

test/unit/communication/test_comm_matrix.cpp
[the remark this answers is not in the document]
msimberg: https://github.com/eth-cscs/DLA-Future/pull/993/commits/bb4c341a36755bc2e4b0ecd6aecabcb2470ff777

include/dlaf/communication/communicator_pipeline.h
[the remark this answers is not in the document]
msimberg: All right, well they're both `{-1, -1}` now so marking this as resolved in that case.

include/dlaf/eigensolver/band_to_tridiag/mc.h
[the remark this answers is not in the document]
msimberg: I'm ok with the suggestion from @albestro as well, or the one that's there now. @rasolca do you have a preference?

include/dlaf/common/index2d.h
msimberg: Marking this as resolved, but feel free to unresolve if you disagree with the current naming.

msimberg:
cscs-ci run

include/dlaf/eigensolver/band_to_tridiag/mc.h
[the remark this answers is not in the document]
rasolca: Ok with @albestro suggestion.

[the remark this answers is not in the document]
msimberg: https://github.com/eth-cscs/DLA-Future/pull/993/commits/e16a63b102f6cd0f849b6c18bfe78a3ca916a757

msimberg:
I've now marked all open comments as resolved. If you would happen to approve this now without further comments, @rasolca please don't merge it yet as I would like to run the benchmarks with the final version first.

albestro (approved):

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
The test failure is a timeout while creating codecov reports: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5625542709.

albestro:
> The test failure is a timeout while creating codecov reports: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5625542709.

@rasolca is already investigating this. it is happening recently and from the log it is not yet clear what's going on, considering that it seems that all ranks finished with a "done" message (in a time range of a couple of seconds max). almost surely not related to this PR. 😉

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
`alby/trisolver-dist-opt-step3` is now merged into this PR. #997 and #998 are ahead in the queue to be merged.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

rasolca (approved):

PR #1002, opened 2023-10-06
Fix coverage tests

rasolca:
Umpire added the compiler option `-fcompare-debug-second` which breaks coverage.

Force Umpire version 2022.03.1 to avoid the problem. 

Note: Opened issue in Umpire: https://github.com/LLNL/Umpire/issues/852

### DIFF
--- a/ci/docker/debug-cpu.yaml
+++ b/ci/docker/debug-cpu.yaml
@@ -28,3 +28,5 @@ spack:
       variants:
         - 'build_type=Debug'
         - 'malloc=system'
+    umpire:
+      require: '@2022.03.1'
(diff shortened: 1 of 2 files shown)


rasolca:
cscs-ci run

msimberg (approved):
:grimacing: 

Is this another reason the coverage reports have been weird lately? Definitely a good catch!

rasolca:
@msimberg This is the main reason why the coverage reports were weird.

albestro:
For future reference.

https://github.com/LLNL/Umpire/commit/bce72924e6d902b7a59e94dd22b1c4fe3fb4dfa0 in v2023.06.0 and v2022.10.0 adds `-f-compare-debug-second` with `INTERFACE`, so it gets propagated to dependents.

albestro (approved):
As said, I would have added a comment in the environment in order to keep track of this problem (e.g. linking to https://github.com/LLNL/Umpire/issues/852). Since it might be expensive (it would require re-building the build image), I added a reference in this PR and I leave to you to decide if it is enough or not.

PR #1009, opened 2023-10-11
Add error message for invalid grid context in C API

### LINKED ISSUE #1008: Improve error message if grid is missing in C API
If for some reason the `dlaf_context` in the C API does not correspond to a valid grid, currently the error message is the following:

```
_Map_base::at
```

I immediately know what's up because I implemented the C API, but this would be very unclear to external users.

We need to capture the `std::out_of_range` exception thrown by `at`

https://github.com/eth-cscs/DLA-Future/blob/c970ce57311d2161714f5461ec018ee724addad0/src/c_api/eigensolver/eigensolver.h#L46

and print out a proper error message.

RMeli:
If an invalid DLA-Future context is passed to the C API, the following error is printed
```
[Error] No DLA-Future grid for context 123456789.
[Info]  Make sure you called dlaf_create_grid() or dlaf_create_grid_from_blacs().
```
and the program terminates.

Close #1008.

### DIFF
--- a/src/c_api/utils.cpp
+++ b/src/c_api/utils.cpp
@@ -17,6 +17,7 @@
 
 #include "grid.h"
 #include "utils.h"
+
 struct DLAF_descriptor make_dlaf_descriptor(const int m, const int n, const int i, const int j,
                                             const int desc[9]) {
   DLAF_ASSERT(i == 1, i);
--- a/src/c_api/utils.cpp
+++ b/src/c_api/utils.cpp
@@ -53,11 +53,11 @@ dlaf::comm::CommunicatorGrid& grid_from_context(int dlaf_context) {
     return dlaf_grids.at(dlaf_context);
   }
   catch (const std::out_of_range& e) {
-    std::stringstream s;
-    s << "[Error] No DLA-Future grid for context " << dlaf_context << ".\n";
-    s << "[Info]  Make sure you called dlaf_create_grid() or dlaf_create_grid_from_blacs().\n";
+    std::stringstream ss;
+    ss << "[ERROR] No DLA-Future grid for context " << dlaf_context << ". ";
+    ss << "Did you forget to call dlaf_create_grid() or dlaf_create_grid_from_blacs()?\n";
 
-    std::cerr << s.str() << std::flush;
+    std::cerr << ss.str() << std::flush;
 
     std::terminate();
   }
--- a/src/c_api/utils.cpp
+++ b/src/c_api/utils.cpp
@@ -8,8 +8,6 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 
-#include "utils.h"
-
 #include <iostream>
 #include <stdexcept>
 
@@ -18,7 +16,7 @@
 #include <dlaf_c/utils.h>
 
 #include "grid.h"
-
+#include "utils.h"
 struct DLAF_descriptor make_dlaf_descriptor(const int m, const int n, const int i, const int j,
                                             const int desc[9]) {
   DLAF_ASSERT(i == 1, i);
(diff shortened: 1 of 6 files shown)


RMeli:
cscs-ci run

RMeli:
cscs-ci run

rasolca (approved):

src/c_api/utils.h
```diff
@@ -8,6 +8,8 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 
+#include <stdexcept>
+
```
RMeli, line 12:
```suggestion
```

RMeli:
cscs-ci run

src/c_api/utils.cpp
```diff
@@ -43,3 +48,18 @@ dlaf::common::Ordering char2order(const char order) {
...
 }
+
+dlaf::comm::CommunicatorGrid& grid_from_context(int dlaf_context) {
+  try {
+    return dlaf_grids.at(dlaf_context);
+  }
+  catch (const std::out_of_range& e) {
+    std::stringstream s;
+    s << "[Error] No DLA-Future grid for context " << dlaf_context << ".\n";
+    s << "[Info]  Make sure you called dlaf_create_grid() or dlaf_create_grid_from_blacs().\n";
```
msimberg, line 59: This is major bikeshedding, but should we make this 
```suggestion
    s << "[ERROR] No DLA-Future grid for context " << dlaf_context << ". ";
    s << "Make sure you called dlaf_create_grid() or dlaf_create_grid_from_blacs().\n";
```
?

We're not super consistent about the style of error messages but maybe it's good to not introduce a new style here. We have at the moment:
- No special prefix, i.e. print message and terminate.
- `[ERROR] <message>` and terminate.
- `[{CU,ROC}{SOLVER,BLAS} ERROR] <message>` and terminate.

msimberg (changes requested):

src/c_api/utils.cpp
RMeli: Good point. I think I had SIRIUS' output in mind when I wrote this.

What I was trying to achieve is not to conflate the error message (i.e. there is no grid corresponding to the context), and the suggestion of a possible cause. What about leaving the `\n` in the first line but print the second one without `[ERROR]`? Or phrasing it differently, as a question:

```
[ERROR] No DLA-Future grid for context 12345. Did you forget to call dlaf_create_grid() or dlaf_create_grid_from_blacs()?
```

Although it might sound a bit condescending, I don't know.

src/c_api/utils.cpp
```diff
@@ -10,10 +10,15 @@
...
 
+#include <iostream>
+#include <stdexcept>
+
 #include <dlaf/communication/communicator_grid.h>
 #include <dlaf_c/desc.h>
 #include <dlaf_c/utils.h>
 
+#include "grid.h"
+
```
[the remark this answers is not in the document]
rasolca: Can you try `IncludeIsMainRegex: "_disabled$"` to trick clang-format to disable main header?

https://github.com/eth-cscs/DLA-Future/blob/master/.clang-format#L126

src/c_api/utils.cpp
rasolca: Not sure how error message are printed, please check asserts and replicate it.

RMeli: That's what I did (although I mistakenly wrote `Error` instead of `ERROR`)

https://github.com/eth-cscs/DLA-Future/blob/c970ce57311d2161714f5461ec018ee724addad0/include/dlaf/common/assert.h#L46

But the question here is how to add the additional information (which is not part of the error, but I think it might be useful for an external user since it points to the most likely cause).

msimberg (approved):

src/c_api/utils.cpp
RMeli: Cool, let's go with that then.

albestro (approved):

PR #1014, opened 2023-10-16
Added srun_args option to miniapps.py

rasolca:
The option is useful to specify the `--uenv` option when there is the need to mount a stackinator squashfs without modifying the `system.py` configuration each time.

Note:
- removed marconi-m100 system (EOL)
- added clariden

### DIFF
--- a/scripts/systems.py
+++ b/scripts/systems.py
@@ -37,7 +37,7 @@ cscs["daint-mc"] = {
     "Allowed rpns": [1, 2],
     "Multiple rpn in same job": True,
     "GPU": False,
-    "Run command": "srun -u -n {total_ranks} --cpu-bind=core -c {threads_per_rank}",
+    "Run command": "srun -u {srun_args} -n {total_ranks} --cpu-bind=core -c {threads_per_rank}",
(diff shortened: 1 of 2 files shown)


scripts/systems.py
```diff
@@ -157,24 +156,22 @@
 """,
 }
 
-# NOTE: Here is assumed that `gpu2ranks_slurm_hip` is in PATH!
+# NOTE: Here is assumed that `gpu2ranks_slurm_cuda` is in PATH!
 #       modify "Run command" if it is not the case.
-cscs["hohgant-amdgpu"] = {
+cscs["clariden-nvgpu"] = {
```
msimberg, line 161: Would it make sense to do
```
cscs["clariden-nvgpu"] = cscs["hohgant-nvgpu"]
```
and same for `*-amdgpu` (assuming I see correctly that they actually are the same)?

There may be good reasons for keeping them separate as well (being able to easily make one-off changes to one or the other system) so not at all insisting on this.

msimberg (approved):

albestro (approved):
LGTM

PR #1016, opened 2023-10-17
Instrument distributed eigensolver with HDF5

RMeli:
Since in previous discussion we decided that HDF5 is only used internally for debugging, I added HDF5 output for all stages of the distributed standard eigensolver. This expands the previous implementation, where only the input/output of the tridiagonal eigensolver is available.

The `debug_` options are explicitly _not_ added to `operator<<` in `tune.cpp`.

I need this for debugging, but please feel free to close the PR if it is not otherwise useful to have this functionality.


RMeli:
cscs-ci run

msimberg:
Revisit early next year.

RMeli:
cscs-ci run

RMeli:
I'm updating this PR because I need it. Please ignore. 

(However, having it in `master` would be helpful for easier bisections if needed again in the future).

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -1034,6 +1035,17 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
   // Should be dispatched to local implementation if (1x1) grid.
   DLAF_ASSERT(grid.size() != comm::Size2D(1, 1), grid);
 
+#ifdef DLAF_WITH_HDF5
+  static size_t num_b2t_calls = 0;
```
msimberg, line 1039: Possibly:
```suggestion
  static std::atomic<std::size_t> num_b2t_calls = 0;
```
and possibly move this into a function-local static in a separate function defined in a `cpp` file so there's only one static. Alternatively to either or both of the above simply document the restrictions somewhere (no concurrent calls, or overlapping indices for different types).

include/dlaf/eigensolver/eigensolver/impl.h
```diff
@@ -60,28 +60,33 @@ void Eigensolver<B, D, T>::call(comm::CommunicatorGrid& grid, blas::Uplo uplo, M
   if (uplo != blas::Uplo::Lower)
     DLAF_UNIMPLEMENTED(uplo);
 
-  auto mat_taus = reduction_to_band<B>(grid, mat_a, band_size);
-  auto ret = band_to_tridiagonal<Backend::MC>(grid, uplo, band_size, mat_a);
-
 #ifdef DLAF_WITH_HDF5
+  static size_t num_eigensolver_calls = 0;
```
msimberg, line 64: Same comment as above for `band_to_tridiag`.

include/dlaf/eigensolver/tridiag_solver/impl.h
```diff
@@ -363,6 +364,17 @@ void TridiagSolver<B, D, T>::call(comm::CommunicatorGrid& grid, Matrix<T, Device
   if (evecs.size().isEmpty())
     return;
 
+#ifdef DLAF_WITH_HDF5
+  static size_t num_tridiag_solver_calls = 0;
```
msimberg, line 368: Same as above.

include/dlaf/eigensolver/band_to_tridiag/mc.h
msimberg: As I was scrolling further I saw the warning with the option:
```
WARNING: just a single execution can be dumped on disk, and any subsequent call fails
```
I think this already implies that there can't be concurrent (or even multiple sequential) calls to the algorithm so not sure how important the `atomic` and `static` comment above from me is...

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -1524,6 +1536,14 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
                                                    mat_trid.readwrite(index)));
   }
 
+#ifdef DLAF_WITH_HDF5
+  if (getTuneParameters().debug_dump_band_to_tridiagonal_data) {
+    file->write(mat_trid, "/tridiaginal");
```
msimberg, line 1541:
```suggestion
    file->write(mat_trid, "/tridiagonal");
```

include/dlaf/eigensolver/band_to_tridiag/mc.h
RMeli: The warning is a leftover from the initial implementation by @albestro. I added the `static` variable to allow for multiple (sequential) calls (by writing to a different file).

msimberg: Oh right, you said as much. In that case I suppose the warning should be removed and then I stand by the above comment, at least the part about moving the counter into a separate function so that it's type-independent (I think the `atomic` is less important right now, it's unlikely to be called concurrently).

include/dlaf/matrix/hdf5.h
```diff
@@ -53,6 +53,32 @@ const H5::PredType& hdf5_datatype<std::complex<T>>::type = hdf5_datatype<T>::typ
...
+// Type to string mappings
+template<typename T>
+struct TypeToString {
+  static inline const std::string value = typeid(T).name();
+};
+
+template<>
+struct TypeToString<float> {
+  static inline const std::string value = "s";
+};
```
msimberg, line 65: Not worth much bikeshedding as long as this stays internal. For reference:

- We use this for miniapps: https://github.com/eth-cscs/DLA-Future/blob/08d13255892e6c3e6e7023434770c6db1f788543/miniapp/include/dlaf/miniapp/options.h#L114-L132. It's similar, but not exactly the same, and at some point unifying it may be useful.

This should include `typeinfo`. On top of that I'd probably do:
```suggestion
#include <string_view>

// Type to string mappings
template<typename T>
struct TypeToString {
  static inline constexpr std::string_view value = typeid(T).name();
};

template<>
struct TypeToString<float> {
  static inline constexpr std::string_view value = "s";
};
```

and later:
```
template <typename T>
inline constexpr std::string_view TypeToString_v = TypeToString<T>::value;
```
to have a shorthand for it.

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

include/dlaf/matrix/hdf5.h
RMeli: Apart from the `ElementType` information, I think I addressed all the comments. Thank you for the feedback.

include/dlaf/matrix/hdf5.h
```diff
@@ -15,6 +15,7 @@
 #include <complex>
 #include <cstdint>
 #include <string>
+#include <string_view>
```
msimberg, line 18:
```suggestion
#include <string_view>
#include <typeinfo>
```
because of (https://en.cppreference.com/w/cpp/language/typeid):
> The header [<typeinfo>](https://en.cppreference.com/w/cpp/header/typeinfo) must be [included](https://en.cppreference.com/w/cpp/standard_library#Including_headers) or [imported](https://en.cppreference.com/w/cpp/standard_library#Importing_headers)(since C++20) or a [standard library module](https://en.cppreference.com/w/cpp/standard_library#Importing_modules) imported(since C++23) before using typeid, otherwise the program is ill-formed.

include/dlaf/eigensolver/eigensolver/impl.h
```diff
@@ -60,28 +62,35 @@ void Eigensolver<B, D, T>::call(comm::CommunicatorGrid& grid, blas::Uplo uplo, M
   if (uplo != blas::Uplo::Lower)
     DLAF_UNIMPLEMENTED(uplo);
 
-  auto mat_taus = reduction_to_band<B>(grid, mat_a, band_size);
-  auto ret = band_to_tridiagonal<Backend::MC>(grid, uplo, band_size, mat_a);
-
 #ifdef DLAF_WITH_HDF5
+  static std::atomic<size_t> num_eigensolver_calls = 0;
+  std::stringstream fname;
+  fname << "eigensolver-" << matrix::internal::TypeToString<T>::value << "-"
```
msimberg, line 68:
```suggestion
  fname << "eigensolver-" << matrix::internal::TypeToString_v<T> << "-"
```

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -1129,6 +1132,19 @@ Matrix<T, Device::CPU> ReductionToBand<B, D, T>::call(comm::CommunicatorGrid& gr
   auto mpi_col_chain = grid.col_communicator_pipeline();
   auto mpi_col_chain_panel = grid.col_communicator_pipeline();
 
+#ifdef DLAF_WITH_HDF5
+  static std::atomic<size_t> num_reduction_to_band_calls = 0;
+  std::stringstream fname;
+  fname << "reduction_to_band-" << matrix::internal::TypeToString<T>::value << "-"
```
msimberg, line 1138:
```suggestion
  fname << "reduction_to_band-" << matrix::internal::TypeToString_v<T> << "-"
```

msimberg (changes requested):
Only minor final comments, looks good to me otherwise. Thanks for making the changes!

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -1034,6 +1038,19 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
   // Should be dispatched to local implementation if (1x1) grid.
   DLAF_ASSERT(grid.size() != comm::Size2D(1, 1), grid);
 
+#ifdef DLAF_WITH_HDF5
+  static std::atomic<size_t> num_b2t_calls = 0;
+  std::stringstream fname;
+  fname << "band_to_tridiag-" << matrix::internal::TypeToString<T>::value << "-"
```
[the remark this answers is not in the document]
RMeli: Yes, thank you! I messed up with `git stash`, and I did not re-apply all the changes properly, it seems.

include/dlaf/matrix/hdf5.h
RMeli: TIL, thanks!

msimberg (approved):

RMeli:
cscs-ci run

rasolca:
cscs-ci run

PR #969, opened 2023-09-05
Enable Gemm (local) to be used with MatrixRef

albestro:
This PR partially address #915 (just local variant)

- `gemm` (local) implementation working with MatrixRef (only, at the moment) with related tests for both full and sub-matrix test-cases
- introduces `scal` tile extension
- basic refactoring(=adaptation) of `local_matrix` and `multipliable` assert + introduction of `same_process_grid`
- refactor test helper for sub-matrices (`subValues` and `mixValues`)
- minor: `MatrixBase` string representation also print information about `source_rank_index` and `offset`

TODO
+ [x] Check preconditions
+ [x] Extend the test
+ [x] Move the new function in the right place (see following note)
+ [x] Evaluate how to deal with `multipliable` assertion (separate it from algorithm details)
+ [x] Test also with ROCm (mainly if `scal` workaround of using `gemm` with nullptr works there too)

Note: `GeneralSub` was intended to work on submatrices, while now the `sub-matrix` concept is hidden by `MatrixRef`. For this reason, the new gemm helper might just be considered a full general multiplication, and it might be moved in `multiplication::general::General` (or something like that).

PR #1021, opened 2023-10-27
Enable `copy` to be used with MatrixRef

albestro:
Basic implementation (mainly copy-paste) for a copy method involving `MatrixRef`s.

Currently `Matrix` and `MatrixRef` are still two independent objects, and this implies that we would have to duplicate all the code for the `Matrix` (more or less, e.g. `MatrixRef` might have an offset which creates some new corner cases) the same for `MatrixRef`.

This represents a problem
- from the code point of view: with duplicated code, exponential number of combinations (if A,B,C are 3 matrix-like objects there are 8 combinations of `Matrix` vs `MatrixRef` calls), trivial template solution might costs a lot in terms of compile times and binaries sizes.
- but also for tests: should I take almost all `Matrix` tests and replicate them for `MatrixRef`?!

Nothing new, but since it does not sound useful to spend time replicating things like that, in this PR I started with just the basic implementation of the `copy` and as soon as we take a decision on how to proceed, I will try to apply here too. 😉 

Waiting for your ideas and suggestions!

EDIT:
After a discussion with others, we opted for using `copy` as a "testbed" also for the sub-matrix/pipeline mechanisms, as a "real" use-case. This means that test for `copy` will be more "extensive" wrt other algorithms (e.g. gemm) and we will have:
- tests for Matrix: just full matrix
- tests for MatrixRef and Pipeline (i.e. Matrix)
  - sub-matrix
  - full-matrix

For the rest of the algorithms (so `gemm` included), we will just test the with MatrixRef both full and sub-matrices cases (where applicable).


msimberg:
Just for the record, I will benchmark these changes: https://github.com/eth-cscs/DLA-Future/compare/master...msimberg:DLA-Future:matrix-ref-algorithm-input-shortcut-trivial-ref. Hopefully by early next week we'll know more about if we can possibly only instantiate algorithms with `MatrixRef` and then you could update this PR to do the same.

test/unit/matrix/test_copy.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+::testing::Environment* const comm_grids_env =
+    ::testing::AddGlobalTestEnvironment(new CommunicatorGrid6RanksEnvironment);
+
+template <typename Type>
+struct MatrixCopyTest : public TestWithCommGrids {};
+
+TYPED_TEST_SUITE(MatrixCopyTest, MatrixElementTypes);
+
+// TODO local test
```
albestro, line 43: note-to-self: cover also non-distributed test-cases

include/dlaf/matrix/copy.h
```diff
@@ -77,6 +78,32 @@ void copy(Matrix<const T, Source>& source, Matrix<T, Destination>& dest) {
   copy(source.distribution().localNrTiles(), LocalTileIndex(0, 0), source, LocalTileIndex(0, 0), dest);
 }
 
+namespace internal {
+template <class T, Device Source, Device Destination>
+void copy(MatrixRef<const T, Source>& src, MatrixRef<T, Destination>& dst) {
+  namespace ex = pika::execution::experimental;
+
+  if constexpr (Source == Destination) {
+    if (&src == &dst)
```
msimberg, line 87: I don't know if this needs to be resolved in this PR, but since this compares `MatrixRef` addresses and not `Matrix` addresses this could in the worst case lead to deadlocks if the `MatrixRef`s refer to the same `Matrix` (`read` and `readwrite` into the same task). I think one step in the right direction would be to have `operator==` on `MatrixRef` (which compares the addresses of the contained `Matrix` and sub distributions), but I'm not sure if it covers all possible issues.

I'd be fine with dealing with this outside of this PR.

msimberg (approved):

include/dlaf/matrix/copy.h
albestro: That's a good point. I did a no-brainer copy-paste, but actually it hides a dangerous thing. How should we address this? Immediately or just open an issue? @rasolca @msimberg

rasolca: I think that Submatrix pipeline should be considered as well, so it is better to work on it separately.
However it should be taken care as the tridiagonal solver might be impacted by it.
(device <-> host mirror copies in the case of multicore are skipped as they are the same matrix, however after the cleanup they will be two equal ref but different objects)

msimberg: https://github.com/eth-cscs/DLA-Future/issues/1035

test/unit/matrix/test_copy.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+bool isFullMatrix(const Distribution& dist_full, const GlobalElementIndex& sub_origin,
+                  const GlobalElementSize& sub_size) noexcept {
+  return sub_origin == GlobalElementIndex{0, 0} && sub_size == dist_full.size();
+}
+
+bool isInSub(const GlobalElementIndex& ij, const GlobalElementIndex& origin,
+             const GlobalElementSize& sub_size) noexcept {
+  return ij.row() >= origin.row() && ij.col() >= origin.col() &&
+         ij.isIn(sizeFromOrigin(origin + sub_size));
```
rasolca, line 187: Might be helpful to add this helper to indices directly?

rasolca (commented):
Please start to use `snake_case` for functions.

test/unit/matrix/test_copy.cpp
albestro: see 38f96c9

include/dlaf/matrix/copy.h
albestro: Thanks @msimberg for opening the issue!

I'll not resolve this comment hoping it will be easier in the future to spot it in case of need of reference.

albestro:
cscs-ci run

test/unit/matrix/test_copy.cpp
```diff
@@ -0,0 +1,365 @@
...
+template <class OutsideElementGetter, class InsideElementGetter>
+auto mixValues(const GlobalElementIndex& offset, const GlobalElementSize& sub_size,
+               InsideElementGetter&& insideValues, OutsideElementGetter&& outsideValues) {
+  return [outsideValues, insideValues, offset, sub_size](const GlobalElementIndex& ij) {
+    if (ij.isInSub(offset, sub_size))
+      return insideValues(ij - common::sizeFromOrigin(offset));
+    else
+      return outsideValues(ij);
+  };
+}
```
albestro, line 187: These helpers might be useful elsewhere, and I will probably move them in some "common" place so that they can be used in other tests too. (Probably this will happen in #969)

albestro:
cscs-ci run

rebased on new master (for getting matrix base snake_case) and addressed snake_case in implementation too.

after this it is ready to be merged 👍🏻

PR #1022, opened 2023-10-27
Enable Gemm (distributed) to be used with MatrixRef

albestro:
Implementation has been copied from the GeneralSub case, but it has been refined and cleaned up (thanks to MatrixRef functionalities). More importantly, the GeneralSub implementation was constrained to just work with square sub-matrices, while now this constraint has been removed.

TODO
- [x] #969 (change base after it gets merged)
- [x] #1047
- [x] Add documentation
- [x] Add tests
- [x] Use new snake_case functions in implementation

### DIFF
--- a/include/dlaf/multiplication/general.h
+++ b/include/dlaf/multiplication/general.h
@@ -67,23 +67,15 @@ void generalMatrix(const blas::Op opA, const blas::Op opB, const T alpha, Matrix
 template <Backend B, Device D, class T>
 void generalMatrix([[maybe_unused]] comm::CommunicatorGrid grid,
                    common::Pipeline<comm::Communicator>& row_task_chain,
-                   common::Pipeline<comm::Communicator>& col_task_chain, const SizeType a,
-                   const SizeType b, const T alpha, MatrixRef<const T, D>& mat_a,
-                   MatrixRef<const T, D>& mat_b, const T beta, MatrixRef<T, D>& mat_c) {
-  DLAF_ASSERT(equal_process_grid(mat_a, grid), mat_a, grid);
-  DLAF_ASSERT(equal_process_grid(mat_b, grid), mat_a, grid);
-  DLAF_ASSERT(equal_process_grid(mat_c, grid), mat_a, grid);
+                   common::Pipeline<comm::Communicator>& col_task_chain, const T alpha,
+                   MatrixRef<const T, D>& mat_a, MatrixRef<const T, D>& mat_b, const T beta,
+                   MatrixRef<T, D>& mat_c) {
+  DLAF_ASSERT(matrix::equal_process_grid(mat_a, grid), mat_a, grid);
+  DLAF_ASSERT(matrix::equal_process_grid(mat_b, grid), mat_a, grid);
+  DLAF_ASSERT(matrix::equal_process_grid(mat_c, grid), mat_a, grid);
 
-  using matrix::multipliable_sizes;
-  DLAF_ASSERT(multipliable_sizes(mat_a.size(), mat_b.size(), mat_c.size()),
-              "Multiplication incompatible matrix sizes.", mat_a.size(), mat_b.size(), mat_c.size());
-  DLAF_ASSERT(multipliable_sizes(mat_a.blockSize(), mat_b.blockSize(), mat_c.blockSize()),
-              "Multiplication incompatible tile sizes.");
-  DLAF_ASSERT(mat_c.size().isEmpty() || multipliable_sizes(mat_a.distribution().tileSize({0, 0}),
-                                                           mat_b.distribution().tileSize({0, 0}),
-                                                           mat_c.distribution().tileSize({0, 0})),
-              "Multiplication incompatible tile sizes in first row/col. "
-              "(Are you using a matrix with offset not aligned with tile?)");
+  DLAF_ASSERT_HEAVY(matrix::multipliable(mat_a, mat_b, mat_c, blas::Op::NoTrans, blas::Op::NoTrans),
+                    mat_a, mat_b, mat_c);
 
   internal::General<B, D, T>::callNN(row_task_chain, col_task_chain, alpha, mat_a, mat_b, beta, mat_c);
 }
--- a/include/dlaf/multiplication/general.h
+++ b/include/dlaf/multiplication/general.h
@@ -68,28 +68,22 @@ void generalMatrix(const blas::Op opA, const blas::Op opB, const T alpha, Matrix
 /// C = alpha * A * B + beta * C
 ///
 /// @param  mat_a contains the input matrix A.
-/// @pre @p mat_a is distributed according to @p grid
-///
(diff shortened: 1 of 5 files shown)


include/dlaf/multiplication/general/impl.h
```diff
@@ -65,6 +65,82 @@ void General<B, D, T>::callNN(const blas::Op opA, const blas::Op opB, const T al
...
 
+template <Backend B, Device D, class T>
+void General<B, D, T>::callNN(common::Pipeline<comm::Communicator>& row_task_chain,
+                              common::Pipeline<comm::Communicator>& col_task_chain, const T alpha,
+                              MatrixRef<const T, D>& mat_a, MatrixRef<const T, D>& mat_b, const T beta,
+                              MatrixRef<T, D>& mat_c) {
+  namespace ex = pika::execution::experimental;
+
+  if (mat_c.size().isEmpty())
+    return;
```
rasolca, line 76: The case k=0 is not handled correctly.
The matrix C is not scaled accordingly to beta.

rasolca (changes requested):

test/unit/multiplication/test_multiplication_general.cpp
```diff
@@ -162,7 +251,9 @@ std::vector<GemmConfig> sub_gemm_configs = {
     // single-tile
     {blas::Op::NoTrans, blas::Op::NoTrans, 8, 8, 11, 10, 9, 13, {{2, 1}}, {{1, 1}}, {{0, 0}}},
     // multi-tile
-    {blas::Op::NoTrans, blas::Op::NoTrans, 12, 20, 11, 3, 4, 5, {{7, 1}}, {{11, 10}}, {{4, 2}}},
+    // TODO check this, I suspect a problem/limit with panel offset
```
albestro, line 254: note-to-self: after a quick check it seems a limitation of panel. it needs more investigation

albestro:
Currently, "empty" test-cases are not working: this is probably due to `alignSubRankIndex` that does not consider this edge case.

test/unit/multiplication/test_multiplication_general.cpp
albestro: See #1047

include/dlaf/multiplication/general.h
```diff
@@ -20,11 +20,28 @@
...
 #include <dlaf/matrix/index.h>
 #include <dlaf/matrix/matrix.h>
+#include <dlaf/matrix/matrix_ref.h>
 #include <dlaf/multiplication/general/api.h>
 #include <dlaf/util_matrix.h>
 
 namespace dlaf::multiplication::internal {
 
+template <Backend B, Device D, class T>
+void generalMatrix([[maybe_unused]] comm::CommunicatorGrid grid,
```
albestro, line 30: note-to-self: add documentation

include/dlaf/multiplication/general.h
```diff
@@ -20,11 +20,45 @@
...
+/// @pre @p mat_b is distributed according to @p grid
+///
+/// @param  mat_c On entry it contains the input matrix C. On exit matrix tiles in the range will be
+///         overwritten with the result, while others are left untouched.
+/// @pre @p mat_c is distributed according to @p grid
+///
+/// @pre multipliable_sizes(mat_a.size(), mat_b.size(), mat_c.size(), opA, opB)
+/// @pre multipliable_sizes(mat_a.tile_size(), mat_b.tile_size(), mat_c.tile_size(), opA, opB)
+/// @pre multipliable_sizes(mat_a.tile_size_of({0, 0}), mat_b.tile_size_of({0, 0}),
+///      mat_c.tile_size_of({0, 0}), opA, opB)
```
albestro, line 45: Some doubts about how to expose pre-conditions:
- high level `multipliable` vs more verbose `multipliable_sizes`
- how to expose rank-alignment requirement (currently just NN is implemented, this requirement might likely be op dependent)
- should we expose here also the constraint imposed by panel (i.e. blocksize == tilesize).
  On one side it is a good idea to document, so that the user knows without inspecting code (after getting an assert triggered), on the other hand it would easily become difficult to keep track of all the places where a constraint like this is involved (e.g. it gets fixed/improved and you have to update all docs, it might not be so straightforward to do for many reasons)

include/dlaf/multiplication/general.h
```diff
@@ -64,6 +64,39 @@ void generalMatrix(const blas::Op opA, const blas::Op opB, const T alpha, Matrix
...
+template <Backend B, Device D, class T>
+void generalMatrix([[maybe_unused]] comm::CommunicatorGrid grid,
+                   common::Pipeline<comm::Communicator>& row_task_chain,
+                   common::Pipeline<comm::Communicator>& col_task_chain, const T alpha,
+                   MatrixRef<const T, D>& mat_a, MatrixRef<const T, D>& mat_b, const T beta,
+                   MatrixRef<T, D>& mat_c) {
+  DLAF_ASSERT(matrix::equal_process_grid(mat_a, grid), mat_a, grid);
+  DLAF_ASSERT(matrix::equal_process_grid(mat_b, grid), mat_a, grid);
+  DLAF_ASSERT(matrix::equal_process_grid(mat_c, grid), mat_a, grid);
```
msimberg, line 92: Why are these not simply:
```suggestion
  DLAF_ASSERT(matrix::same_process_grid(mat_a, mat_b), mat_a, mat_b);
  DLAF_ASSERT(matrix::same_process_grid(mat_b, mat_c), mat_b, mat_c);
```
? I guess the `grid` kind of additionally acts as a tag here to say that this is a distributed matrix multiplication? But then the pipelines already kind of do that as well...

msimberg (approved):

include/dlaf/multiplication/general.h
albestro: Good point! Just updated in https://github.com/eth-cscs/DLA-Future/pull/1022/commits/4f9514d1f3e2da25ba11729f8d3e09e5524e2090

Moreover, if I'm not mistaken, the requirement could be even slightly more relaxed (at least for the current implementation of the case we support).

@rasolca what do you think?

rasolca (approved):

include/dlaf/multiplication/general.h
albestro: Ok, so I will leave it like this for now.

include/dlaf/multiplication/general.h
albestro: About the slightly more relaxed requirement, I was wrong.

About the check, as soon as we will get #993 we can add the check about the "matching" between matrices grid and pipelines.

For what concerns this PR, I think this change is enough 😉

albestro:
cscs-ci run

include/dlaf/multiplication/general.h
msimberg: Should I update something in #993 now that this is merged or would you do the change you're talking about separately?

PR #1023, opened 2023-11-01
Prepare for `pika::start/finalize` cleanup in pika 0.21.0

msimberg:
https://github.com/pika-org/pika/pull/825 will change the return types of `pika::start` and `pika::finalize` from `int` to `void`, because the return value has no significance. The only functions that retain the `int` return value are `pika::init` and `pika::stop` because they can return the return value of `pika_main`, if `pika_main` is used (otherwise `0`).

This prepares DLA-Future in a compatible way for the change. Since the return value of `pika::finalize` never was anything but `0` I'm ignoring the return value regardless of pika version. For `pika::start` I'm keeping the check of the return value for pika 0.20.0 and older because there are cases where non-zero values signal an error. From pika 0.21.0 all errors are reported as exceptions and the return type is `void`.

### DIFF
--- a/src/c_api/init.cpp
+++ b/src/c_api/init.cpp
@@ -27,8 +27,13 @@ void dlaf_initialize(int argc_pika, const char** argv_pika, int argc_dlaf, const
     pika::init_params params;
     params.rp_callback = dlaf::initResourcePartitionerHandler;
     params.desc_cmdline = desc;
+    // After pika 0.21.0 pika::start reports errors only by exception and returns void
+#if PIKA_VERSION_FULL >= 0x001500
+    pika::start(argc_pika, argv_pika, params);
+#else
     auto pika_started = pika::start(nullptr, argc_pika, argv_pika, params);
     DLAF_ASSERT(pika_started, pika_started);
+#endif
 
     // DLA-Future initialization
     dlaf::initialize(argc_dlaf, argv_dlaf);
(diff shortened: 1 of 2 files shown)


msimberg:
cscs-ci run

src/c_api/init.cpp
```diff
@@ -27,8 +27,13 @@ void dlaf_initialize(int argc_pika, const char** argv_pika, int argc_dlaf, const
     pika::init_params params;
     params.rp_callback = dlaf::initResourcePartitionerHandler;
     params.desc_cmdline = desc;
+    // After pika 0.21.0 pika::start reports errors only by exception and returns void
+#if PIKA_VERSION_FULL >= 0x001500
```
RMeli, line 31: For other reviewers:
```c++
#define PIKA_VERSION_FULL                                                                          \
 ((PIKA_VERSION_MAJOR << 16) | (PIKA_VERSION_MINOR << 8) | PIKA_VERSION_PATCH)
```

RMeli (approved):

src/c_api/init.cpp
msimberg: Yep... not the most intuitive, but we will unfortunately not be changing this in pika simply because of compatibility.

rasolca (commented):
Quick question:
Do the changes affect only `pika::start` and `pika::finalize`, while `pika::init` remains the same?

msimberg:
> Quick question: Do the changes affect only `pika::start` and `pika::finalize`, while `pika::init` remains the same?

Correct. Since `pika::init` == `pika::start + pika::stop`, and `pika::stop` still returns the (possible) return value of `pika_main`, `pika::init` will also keep returning that.

rasolca (approved):

PR #1025, opened 2023-11-02
Make sure GCC 12 is installed in GCC 12 CI configuration

msimberg:
Testing... but it looks like the "GCC 12" CI configuration was actually using GCC 11. Opening this for debugging and potential fixes. @rasolca and others, please feel free to comment if you have thoughts on how this should be fixed properly.

### DIFF
--- a/ci/cpu/gcc12_release_cxx20.yml
+++ b/ci/cpu/gcc12_release_cxx20.yml
@@ -5,7 +5,7 @@ cpu gcc12 cxx20 release deps:
   extends: .build_deps_common
   variables:
     EXTRA_APTGET: "g++-12 gfortran-12"
-    COMPILER: gcc@12.1.0
+    COMPILER: gcc@12
     CXXSTD: 20
     SPACK_ENVIRONMENT: ci/docker/release-cpu.yaml
     USE_MKL: "ON"
--- a/ci/cpu/gcc12_release_cxx20.yml
+++ b/ci/cpu/gcc12_release_cxx20.yml
@@ -4,6 +4,7 @@ include:
 cpu gcc12 cxx20 release deps:
   extends: .build_deps_common
   variables:
+    EXTRA_APTGET: "g++-12 gfortran-12"
     COMPILER: gcc@12.1.0
     CXXSTD: 20
     SPACK_ENVIRONMENT: ci/docker/release-cpu.yaml
--- a/ci/cpu/clang12_release.yml
+++ b/ci/cpu/clang12_release.yml
@@ -5,7 +5,7 @@ cpu clang12 release deps:
   extends: .build_deps_common
   variables:
     EXTRA_APTGET: "clang-12 libomp-12-dev"
-    COMPILER: clang@12.0.1
+    COMPILER: clang@12
     USE_MKL: "ON"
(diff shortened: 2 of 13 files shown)


msimberg:
cscs-ci run

ci/cpu/gcc12_release_cxx20.yml
```diff
@@ -4,6 +4,7 @@ include:
 cpu gcc12 cxx20 release deps:
   extends: .build_deps_common
   variables:
+    EXTRA_APTGET: "g++-12"
```
rasolca, line 7: I would add `gfortran-12` as well

rasolca (changes requested):
@albestro can you comment on the fact that spack doesn't complain that the compiler version we ask for doesn't exist?

ci/cpu/gcc12_release_cxx20.yml
albestro: I'm going to give a check and I'll report here 😉

msimberg:
cscs-ci run

albestro:
> @albestro can you comment on the fact that spack doesn't complain that the compiler version we ask for doesn't exist?

Apparently, we are just asking for a package preference

https://github.com/eth-cscs/DLA-Future/blob/a7900e800000df65c3ed742eb459003da7cca178/ci/docker/build.Dockerfile#L60

which is not a requirement like `packages:all:require:'%${COMPILER}'`.

The question for which I still don't have an answer is: how does the config scope "merging" work in case we add something in the global config and then we have something also in the environment file? I'm going to have a look.

albestro:
> > @albestro can you comment on the fact that spack doesn't complain that the compiler version we ask for doesn't exist?
> 
> Apparently, we are just asking for a package preference
> 
> https://github.com/eth-cscs/DLA-Future/blob/a7900e800000df65c3ed742eb459003da7cca178/ci/docker/build.Dockerfile#L60
> 
> which is not a requirement like `packages:all:require:'%${COMPILER}'`.
> 
> The question for which I still don't have an answer is: how does the config scope "merging" work in case we add something in the global config and then we have something also in the environment file? I'm going to have a look.

https://github.com/spack/spack/pull/36526 This is something important to know that @haampie pointed out. It is a "workaround", but as far as we keep using arrays instead of string, it should work as expected™.

With arrays, we are going to get a set of all rules available in different scopes. Each entry in any scope is considered as an individual rule. If they are not compatible all together, we will get a concretisation error.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
It looks like the GCC 11 version was wrong as well (we specify 11.2.0 but the available version is 11.4.0). I've updated that in the latest commit.

albestro:
> It looks like the GCC 11 version was wrong as well (we specify 11.2.0 but the available version is 11.4.0). I've updated that in the latest commit.

As discussed offline, we might end up indicating just the major version for the compiler (it should just work™), which is actually what the CI general wants (GCC12, CLANG15, ...).

msimberg:
cscs-ci run

rasolca (approved):

PR #1027, opened 2023-11-03
Fix build problem with clang due to nested structure definition

albestro:
Fix #1017 

@gulivarese can you please check if this works now?

Issue #1017, opened 2023-10-19
DLA-FUTURE doesn't compile with apple-clang and clang 17


gulivarese:
```
[...]

PR #1028, opened 2023-11-03
Add clang 15 CI configuration

msimberg:
Given #1017 and #1027, I would suggest that we add at least a clang 15 CI configuration. With the base image of ubuntu 22.04 clang 15 is the newest we can get, but we can also go with clang 16 if we use a different base image for this configuration.

https://github.com/eth-cscs/DLA-Future/pull/853#issuecomment-1551027826 is still an issue, so I've added a separate environment where BLAS threading is disabled. https://github.com/spack/spack/issues/38509 is still open on the spack side. If that gets resolved we can eventually enable threading also on this CI configuration.

### DIFF
--- a/ci/.gitlab-ci.yml
+++ b/ci/.gitlab-ci.yml
@@ -1,6 +1,7 @@
 include:
   - local: 'ci/cpu/clang12_release.yml'
   - local: 'ci/cpu/clang13_release_cxx20.yml'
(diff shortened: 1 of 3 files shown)


msimberg:
cscs-ci run

msimberg:
cscs-ci run

rasolca (approved):

PR #1032, opened 2023-11-08
Fix CI configurations

rasolca:
#1024 #1025 and #1028 generated some conflicts in the master brach (CI on master is currently failing)
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/pipelines/1065150395


rasolca:
cscs-ci run

ci/cpu/clang14_release_stdexec.yml
```diff
@@ -0,0 +1,30 @@
+include:
+  - local: 'ci/common-ci.yml'
+
+cpu clang14 stdexec release deps:
+  extends: .build_deps_common
+  variables:
+    EXTRA_APTGET: "clang-14 libomp-14-dev"
+    COMPILER: clang@14
```
rasolca, line 8: @msimberg stdexec package contains: `conflicts("%clang@:13")`

rasolca:
cscs-ci run

ci/cpu/clang14_release_stdexec.yml
msimberg: That's wrong :disappointed: let's go with clang 14 here then. We can update the conflict separately in spack.

msimberg: Update: the constraint in the stdexec package file is subtly wrong, but intentionally so. The issue is that clang prior to version 14 does not support coroutines correctly when used together with libstdc++ (see e.g. https://github.com/facebook/folly/pull/1694#issuecomment-1157255386). Since we can't model the standard library (and its version) in spack, the spack package uses the conservative option and just requires clang 14 or newer.

rasolca:
cscs-ci run

ci/docker/build.Dockerfile
```diff
@@ -104,3 +100,6 @@ RUN spack -e ci config add "packages:dla-future:variants:cxxstd=${CXXSTD}"
 # 4. Install only the dependencies of this (top level is our package)
 ARG NUM_PROCS
 RUN spack -e ci install --jobs ${NUM_PROCS} --fail-fast --only=dependencies
+
+# make ctest executable available.
+RUN export PATH=$PATH:`spack -e ci location -i cmake`
```
albestro, line 105: Not checked the result, but probably it needs

```suggestion
RUN export PATH=$PATH:`spack -e ci location -i cmake`/bin
```

rasolca:
cscs-ci run

PR #1036, opened 2023-11-15
Use static priority scheduler and `pika::execution::thread_priority::boost` priority threads for MPI

msimberg:
Changes the priority for MPI tasks to `thread_priority::boost`, to give higher priority to the submission of MPI communication, but then fall back to normal priority for polling. Also changes to the `static_priority` scheduler for the MPI pool since it only has a single thread. I haven't benchmarked this in isolation and doubt that it makes much of a difference, but it's a bit more appropriate than the default `local_priority_fifo` since no work stealing needs to happen.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

rasolca (approved):

PR #1038, opened 2023-11-15
Make C API functions `noexcept`

msimberg:
Adds a macro `DLAF_NOEXCEPT_CPP` (bikeshedding on the name welcome) which expands to `noexcept` in C++ mode and nothing in C mode. I've applied this to the C API function declarations in headers in `dlaf.c_api`. I've made the function definitions unconditionally `noexcept` since they're always built in C++ mode.

I have not touched the test wrappers since they are in `.c` files and built in C mode. I have also not done anything to the blacs declarations since they're just declarations.

### DIFF
--- a/include/dlaf_c/utils.h
+++ b/include/dlaf_c/utils.h
@@ -13,9 +13,11 @@
 /// @file
 
 #ifdef __cplusplus
-#define DLAF_EXTERN_C extern "C"
+#define DLAF_EXTERN_C     extern "C"
+#define DLAF_NOEXCEPT_CPP noexcept
 #else
 #define DLAF_EXTERN_C
+#define DLAF_NOEXCEPT_CPP
 #endif
 
 #ifdef __cplusplus
@@ -39,4 +41,5 @@ typedef double complex dlaf_complex_z;  ///< Double precision complex number
 /// @param desc ScaLAPACK descriptor
 /// @return DLA-Future descriptor
 DLAF_EXTERN_C struct DLAF_descriptor make_dlaf_descriptor(const int m, const int n, const int i,
-                                                          const int j, const int desc[9]);
+                                                          const int j,
+                                                          const int desc[9]) DLAF_NOEXCEPT_CPP;
--- a/include/dlaf_c/utils.h
+++ b/include/dlaf_c/utils.h
@@ -13,11 +13,11 @@
 /// @file
 
 #ifdef __cplusplus
-#define DLAF_EXTERN_C     extern "C"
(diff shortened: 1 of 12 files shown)


msimberg:
cscs-ci run

include/dlaf_c/utils.h
```diff
@@ -13,9 +13,11 @@
 /// @file
 
 #ifdef __cplusplus
-#define DLAF_EXTERN_C extern "C"
+#define DLAF_EXTERN_C     extern "C"
+#define DLAF_NOEXCEPT_CPP noexcept
```
RMeli, line 17: Since you asked for it... ;) I think the `CPP` is a bit redundant, what about simply using `DLAF_NOEXCEPT`? I put the `C` in `DLAF_EXTERN_C` because it's part of the definition: `extern "C"` -> `DLAF_EXTERN_C`, `noexcept` -> `DLAF_NOEXCEPT`.

But I don't really mind TBH.

msimberg: I did indeed!

Yeah, I wasn't quite happy with the name, but I wanted _some_ suffix since it doesn't unconditionally expand to `noexcept`. `_CPP` was the first thing that came to mind since it applies in C++ mode. Then again, we're never going to have an unconditional `DLAF_NOEXCEPT` on its own since that should just be `noexcept` directly, so I could live with plain `DLAF_NOEXCEPT`.

On a related note, perhaps these should be `DLAF_INTERNAL_*` as well since they're not meant for external consumption?

RMeli: > we're never going to have an unconditional `DLAF_NOEXCEPT` on its own since that should just be `noexcept` directly

Yes, this was basically my thinking, that's why I feel `_CPP` is redundant. But as I said, it does not bother me.

albestro (approved):

include/dlaf_c/utils.h
msimberg: Note, I haven't renamed anything yet (i.e. it's still `DLAF_NOEXCEPT_CPP`). Looking for opinions from others still.

msimberg: @rasolca and @albestro agree to changing it to `DLAF_NOEXCEPT`. Will update.

msimberg:
cscs-ci run

include/dlaf_c/utils.h
msimberg: Updated to `DLAF_NOEXCEPT`.

rasolca (approved):

RMeli (approved):

PR #1039, opened 2023-11-15
Improve eigensolver tests

rasolca:
Eigensolver tests now check that the eigenvalues are sorted. (tridiagonal, full, generalized)

Added an eigensolver test with identity matrix.

(This new test is enough to reproduce one of the CP2K problems with `202310/develop`)

### DIFF
--- a/test/include/dlaf_test/eigensolver/test_eigensolver_correctness.h
+++ b/test/include/dlaf_test/eigensolver/test_eigensolver_correctness.h
@@ -58,6 +58,11 @@ void testEigensolverCorrectness(const blas::Uplo uplo, Matrix<const T, Device::C
     return allGather(blas::Uplo::General, mat_e.get(), grid...);
   }();
 
+  // eigenvalues are contiguous in the mat_local buffer
+  BaseType<T>* evals_start = mat_evalues_local.ptr({0, 0});
+  BaseType<T>* evals_end = evals_start + m;
+  EXPECT_TRUE(std::is_sorted(evals_start, evals_end));
+
   MatrixLocal<T> workspace({m, m}, reference.blockSize());
 
   dlaf::common::internal::SingleThreadedBlasScope single;
--- a/test/unit/eigensolver/test_eigensolver.cpp
+++ b/test/unit/eigensolver/test_eigensolver.cpp
@@ -66,8 +66,14 @@ const std::vector<std::tuple<SizeType, SizeType, SizeType>> sizes = {
     {34, 8, 3},  {32, 6, 3}                                   // m > mb, sub-band
 };
 
+const std::vector<std::tuple<SizeType, SizeType, SizeType>> sizes_id = {
+    // {m, mb, eigensolver_min_band}
+    {8, 4, 4},
+    {34, 8, 4},
+};
+
 template <class T, Backend B, Device D, Allocation allocation, class... GridIfDistributed>
-void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb,
+void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb, bool identity,
                      GridIfDistributed... grid) {
   constexpr bool isDistributed = (sizeof...(grid) == 1);
   const LocalElementSize size(m, m);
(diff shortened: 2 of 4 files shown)


rasolca:
cscs-ci run

test/unit/eigensolver/test_eigensolver.cpp
```diff
@@ -163,9 +194,14 @@ TYPED_TEST(EigensolverTestGPU, CorrectnessDistributed) {
       for (auto [m, mb, b_min] : sizes) {
         getTuneParameters().eigensolver_min_band = b_min;
         testEigensolver<TypeParam, Backend::GPU, Device::GPU, Allocation::do_allocation>(uplo, m, mb,
-                                                                                         grid);
+                                                                                         false grid);
```
RMeli, line 197:
```suggestion
                                                                                         false, grid);
```
CI is failing.

test/unit/eigensolver/test_eigensolver.cpp
```diff
@@ -81,7 +87,14 @@ void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb,
         return Matrix<T, Device::CPU>(LocalElementSize(m, m), block_size);
     }
     ();
-    matrix::util::set_random_hermitian(reference);
+    if (identity)
+      matrix::util::set(reference, [](const GlobalElementIndex& ij) {
+        if (ij.row() == ij.col())
+          return T{1};
+        return T{0};
+      });
```
RMeli, line 95: Given this is the identity matrix, would it be worth moving it to a re-usable function (`matrix::util::set_identity`)?

test/include/dlaf_test/eigensolver/test_eigensolver_correctness.h
```diff
@@ -58,6 +58,11 @@ void testEigensolverCorrectness(const blas::Uplo uplo, Matrix<const T, Device::C
     return allGather(blas::Uplo::General, mat_e.get(), grid...);
   }();
 
+  // eigenvalues are contiguous in the mat_local buffer
+  BaseType<T>* evals_start = mat_evalues_local.ptr({0, 0});
+  BaseType<T>* evals_end = evals_start + m;
+  EXPECT_TRUE(std::is_sorted(evals_start, evals_end));
```
msimberg, line 64: This makes me wonder if it would make sense to provide iterators/begin/end at least for some of our matrix/tile types (obviously only where it makes sense, layout concerns may make this harder than it seems)? Definitely not for this PR, but worth at least tracking in a low priority issue?

test/unit/eigensolver/test_eigensolver.cpp
```diff
@@ -66,8 +66,14 @@ const std::vector<std::tuple<SizeType, SizeType, SizeType>> sizes = {
...
 
+const std::vector<std::tuple<SizeType, SizeType, SizeType>> sizes_id = {
+    // {m, mb, eigensolver_min_band}
+    {8, 4, 4},
+    {34, 8, 4},
+};
+
 template <class T, Backend B, Device D, Allocation allocation, class... GridIfDistributed>
-void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb,
+void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb, bool identity,
```
msimberg, line 76:
```suggestion
void testEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType mb, matrix_type_or_something_like_that type,
```

with 
```
enum class matrix_type... {
    identity,
    random_hermitian
};
```
?

test/include/dlaf_test/eigensolver/test_eigensolver_correctness.h
rasolca: This works because we are dealing with a `MatrixLocal` (used only in tests) used as a column vector.

test/unit/eigensolver/test_eigensolver.cpp
rasolca: 😅

rasolca:
cscs-ci run

msimberg (approved):

RMeli (approved):

albestro (approved):

test/include/dlaf_test/eigensolver/test_eigensolver_correctness.h
msimberg: It's clear to me why this works. My comment was about allowing to write `std::is_sorted(std::begin(mat_evalues_local), std::end(mat_evalues_local))`. But perhaps you're saying that the number of cases where this will work are too few so it's not worth trying to provide that functionality?

PR #1043, opened 2023-11-20
Fix compilation flags

rasolca:
Use `-pedantic` instead of `-pedantic-errors`.

Note: in the CI with `-Werror` pedantic warnings are still reported as error.


rasolca:
cscs-ci run

msimberg (approved):

PR #1045, opened 2023-11-20
Restructure the code such that it can be compiled with nvcc 11.2

rasolca:
Tridiagonal eigensolver GPU kernels do not require header files which include pika anymore.

This allow to compile the library with CUDA 11.2.

Note: I noticed some evident missing asserts, and I added them.


rasolca:
cscs-ci run

rasolca:
Some include guards has to be fixed. But please review the changes.

rasolca:
cscs-ci run

albestro (approved):
LGTM

include/dlaf/eigensolver/tridiag_solver/impl.h
```diff
@@ -94,8 +94,7 @@ template <class T>
 void solveLeaf(Matrix<T, Device::CPU>& tridiag, Matrix<T, Device::CPU>& evecs) {
   const SizeType ntiles = tridiag.distribution().nrTiles().rows();
   for (SizeType i = 0; i < ntiles; ++i) {
-    stedcAsync<Device::CPU>(tridiag.readwrite(LocalTileIndex(i, 0)),
-                            evecs.readwrite(LocalTileIndex(i, i)));
+    stedcAsync<>(tridiag.readwrite(LocalTileIndex(i, 0)), evecs.readwrite(LocalTileIndex(i, i)));
```
msimberg, line 97:
```suggestion
    stedcAsync(tridiag.readwrite(LocalTileIndex(i, 0)), evecs.readwrite(LocalTileIndex(i, i)));
```
here and later?

include/dlaf/eigensolver/tridiag_solver/kernels.cu.h
```diff
@@ -0,0 +1,60 @@
+//
```
msimberg, line 1: Should this file not be just `kernels.h` (or `gpu_kernels.h`, or `gpu/kernels.h`)? It's not using any device functionality as far as I can tell. I would reserve the `*.cu*` for files that actually use device constructs.

include/dlaf/eigensolver/tridiag_solver/kernels.h
```diff
@@ -15,188 +15,126 @@
...
-void givensRotationOnDevice(SizeType len, T* x, T* y, T c, T s, whip::stream_t stream);
-
-#define DLAF_GIVENS_ROT_ETI(kword, Type)                                                     \
-  kword template void givensRotationOnDevice(SizeType len, Type* x, Type* y, Type c, Type s, \
-                                             whip::stream_t stream)
-
-DLAF_GIVENS_ROT_ETI(extern, float);
-DLAF_GIVENS_ROT_ETI(extern, double);
-
+using gpu::givensRotationOnDevice;
```
msimberg, line 138: Just curious, not necessarily asking for a change: what's the reason for putting this in a nested namespace but then pulling it in here anyway?

msimberg (changes requested):
Only a few minor comments/questions.

include/dlaf/eigensolver/tridiag_solver/kernels.h
rasolca: When compiling without gpu support, the compiler complains that gpu is not a known namespace. E.g. in

https://github.com/eth-cscs/DLA-Future/blob/master/include/dlaf/eigensolver/tridiag_solver/rot.h#L183

include/dlaf/eigensolver/tridiag_solver/kernels.cu.h
rasolca: I just wanted to be nice to reviewer avoiding moving kernel.cu.... 😛

include/dlaf/eigensolver/tridiag_solver/kernels.h
msimberg: Ah, I mean: why is it not defined in this namespace to start with instead of pulling it in with `using ...`?

rasolca: That function is a kernel launcher so it belongs to the gpu namespace.
However as it doesn't need an extra wrapper to call it (e.g. no Tile argument) and it cannot be used with the gpu namespace, `using` is the nicer solution.

msimberg: All right, thanks for the explanation. Makes sense.

msimberg (approved):

rasolca:
cscs-ci run

PR #1046, opened 2023-11-21
Remove rocthrust and rocprim from dependencies

rasolca:
Depends on #1045.

As #1045 removes the last usage of these 2 dependencies (only some unused include files where present) I remove the dependencies as well.


rasolca:
cscs-ci run

rasolca:
cscs-ci run

spack/packages/dla-future/package.py
```diff
@@ -93,15 +94,18 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
             depends_on(
                 "rocblas amdgpu_target={0}".format(val), when="amdgpu_target={0}".format(val)
             )
+            depends_on("whip amdgpu_target={0}".format(val), when="amdgpu_target={0}".format(val))
```
RMeli, line 97: Nitpick: Should we start moving to f-strings, at least where we do changes? (If yes, I'd change `val` to something more self-explanatory like `gpu_target` or similar.)

albestro (approved):

spack/packages/dla-future/package.py
```diff
@@ -76,9 +76,10 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("whip +rocm", when="+rocm")
 
     depends_on("rocblas", when="+rocm")
-    depends_on("rocprim", when="+rocm")
     depends_on("rocsolver", when="+rocm")
-    depends_on("rocthrust", when="+rocm")
+
+    depends_on("rocprim", when="@:0.3 +rocm")
+    depends_on("rocthrust", when="@:0.3 +rocm")
```
albestro, line 82: ?

```suggestion
    depends_on("rocprim", when="@:0.3.1 +rocm")
    depends_on("rocthrust", when="@:0.3.1 +rocm")
```

spack/packages/dla-future/package.py
```diff
@@ -93,15 +94,18 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
             depends_on(
                 "rocblas amdgpu_target={0}".format(val), when="amdgpu_target={0}".format(val)
             )
+            depends_on("whip amdgpu_target={0}".format(val), when="amdgpu_target={0}".format(val))
             depends_on(
-                "rocprim amdgpu_target={0}".format(val), when="amdgpu_target={0}".format(val)
+                "umpire amdgpu_target={0}".format(val), when="amdgpu_target={0}".format(val)
             )
+
+    with when("@:0.3 +rocm"):
```
albestro, line 102: In case, it applies here too.

```suggestion
    with when("@:0.3.1 +rocm"):
```

spack/packages/dla-future/package.py
rasolca: No, patch releases are for bug fixing, and adding and removing dependencies is not part of it. (Note that a version change of a dependency can be included in a patch release)
Note that v0.3.1 is not a master release, but it branched out (version_0.3) and only includes bug-fixing PRs.

spack/packages/dla-future/package.py
rasolca: I would fix the whole package if we update the style.
Let me open an issue to address it before v0.4.0

spack/packages/dla-future/package.py
albestro: Ok, so next release will be for sure `0.4.0`, right?

rasolca: Can be `0.3.2`, but without #1045.

msimberg (approved):

spack/packages/dla-future/package.py
RMeli: #1053

RMeli (approved):

PR #1047, opened 2023-11-21
Panel: add limited functionality for Matrix with offset

albestro:
Panel currently has a lot of functionalities, but they do not take into account the use-case of a Panel bound to a Matrix with an offset (i.e. top-left tile is not complete).

Since we are going to refactor Panel using new facilities, this PR aims at just making it work in limited conditions also in case Matrix has an offset.

In particular, range utilities (`setRangeStart`, `setRangeEnd`, ...) are disabled in case the matrix which the panel bounds to has an offset not aligned with tile.

Morevoer, this PR adds a couple of asserts where tile and block size difference might not work: specifically for `Panel` and `Matrix`.

Thanks @rasolca for supporting me in doing this on-spot change.

TODO:
- [x] add a basic test for checking that `Panel` works also with `MatrixRef` (with offset)

### DIFF
--- a/include/dlaf/matrix/matrix.h
+++ b/include/dlaf/matrix/matrix.h
@@ -296,7 +296,10 @@ public:
   }
 
 protected:
-  Matrix(Distribution distribution) : internal::MatrixBase{std::move(distribution)} {}
+  Matrix(Distribution distribution) : internal::MatrixBase{std::move(distribution)} {
+    DLAF_ASSERT((distribution.offset() == GlobalElementIndex{0, 0}), "not supported",
+                distribution.offset());
+  }
   struct SubPipelineTag {};
   Matrix(Matrix& mat, const SubPipelineTag) noexcept : MatrixBase(mat.distribution()) {
     setUpSubPipelines(mat);
--- a/include/dlaf/matrix/panel.h
+++ b/include/dlaf/matrix/panel.h
@@ -24,6 +24,7 @@
 #include <dlaf/matrix/index.h>
 #include <dlaf/matrix/matrix.h>
 #include <dlaf/matrix/matrix_base.h>
+#include <dlaf/matrix/matrix_ref.h>
 #include <dlaf/types.h>
 
 namespace dlaf {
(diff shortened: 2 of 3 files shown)


test/unit/matrix/test_panel.cpp
```diff
@@ -719,3 +720,49 @@ TYPED_TEST(PanelStoreTransposedTest, OffsetTileUnalignedCol) {
...
+  const GlobalElementIndex sub_origin;
+  const GlobalElementSize sub_size;
+};
+
+const std::vector<ConfigSub> sub_configs{
+    {{3, 4}, {4, 2}, {12, 20}},
+    {{3, 5}, {7, 1}, {12, 11}},
+};
+
+TYPED_TEST(PanelTest, MatrixWithOffset) {
```
albestro, line 735: The test is not very extended, since I didn't want to spend to much time on this considering it will be changed soon. Let me know if you'd like to see it a bit more extended than this.

albestro:
cscs-ci run

include/dlaf/matrix/matrix.h
```diff
@@ -296,7 +296,10 @@ class Matrix<const T, D> : public internal::MatrixBase {
   }
 
 protected:
-  Matrix(Distribution distribution) : internal::MatrixBase{std::move(distribution)} {}
+  Matrix(Distribution distribution) : internal::MatrixBase{std::move(distribution)} {
+    DLAF_ASSERT((distribution.offset() == GlobalElementIndex{0, 0}), "not supported",
+                distribution.offset());
```
msimberg, line 301: This is fine by me, and I think conveys the purpose quite well, but we do have `DLAF_UNIMPLEMENTED` as well. If I understand correctly, this restriction is hopefully temporary so not worth spending time on it, but perhaps in the future some macro-unification/development needs to happen.

include/dlaf/matrix/panel.h
```diff
@@ -305,8 +304,12 @@ struct Panel<axis, const T, D, StoreTransposed::No> {
 protected:
   using ReadWriteSenderType = typename BaseT::ReadWriteSenderType;
 
+  bool isBoundToMatrixWithSubTileOffset() const noexcept {
```
msimberg, line 307: `WithSubTileOffset`: why not just with `WithNonZeroOffset` or, abbreviated, `WithOffset`?

msimberg (approved):

include/dlaf/matrix/panel.h
albestro: Do you mean going for `isBoundToMatrixWithOffset` or just `WithSubTileOffset`?

One reason I would go for being more verbose (former one), is because I want to keep it clearly separated from the panel functionality where you can have sub-tile offset (i.e. `setRangeStart`) while this refers to the fact that the panel might be bound to a matrix with a sub-tile offset (and `setRange` would not work for the panel). Not sure it helps in clarifying...

include/dlaf/matrix/matrix.h
albestro: I had a quick look about how we currently use it, and just provide values of what is not provided. I'm not sure it will be clear for the user what it is not implemented without looking at the code.

```suggestion
   if (dist_matrix_.block_size() != dist_matrix_.tile_size())
     DLAF_UNIMPLEMENTED(dist_matrix_.block_size(), dist_matrix_.tile_size());
```

Would you for an additional message like next?

```suggestion
   if (dist_matrix_.block_size() != dist_matrix_.tile_size())
     DLAF_UNIMPLEMENTED("block_size != tile_size", dist_matrix_.block_size(), dist_matrix_.tile_size());
```

rasolca (approved):

include/dlaf/matrix/panel.h
msimberg: I meant the former, i.e. `isBoundToMatrixWithOffset`. In other words, my question was: why is `SubTile` in the name? What information does it convey?

albestro:
cscs-ci run

PR #1050, opened 2023-11-22
Unify docs actions and automate release documentation generation

rasolca:


.github/workflows/doc.yml
```diff
@@ -18,4 +23,22 @@ jobs:
...
-          cd doc && make doc
+          cd doc && make doc && cd ..
+      - name: Push Docs master
+        if: "github.repository == 'eth-cscs/DLA-Future' && github.ref == 'refs/heads/master'"
+        run: |
+          mv doc/html master
+          COMMIT_MSG=`git log -n 1 --pretty=format:%s`
+          git config user.name "Doc CI Action" && git config user.email "rasolca@users.noreply.github.com"
+          git fetch --no-tags --depth=1 origin +refs/heads/docs:refs/heads/docs
+          git symbolic-ref HEAD refs/heads/docs && git reset
```
RMeli, line 34: Nitpick: can these commands be lifted into their own step, since they are shared?

RMeli (approved):
LGTM

albestro (approved):
LGTM

.github/workflows/doc.yml
rasolca: I thought about it, but this IMO this solution was the cleaner.
As for master I need to get the commit message before switching the branch the if clauses become a mess.

RMeli: Fair. I think the `if` clause could simply define the `TAG` (either the actual tag or master) and `COMMIT_MSG` variables, and then everything else would be the same? But more work than it's worth, so happy to leave it as-is.

PR #1054, opened 2023-11-23
Revise spack environments

rasolca:
Spack changed the way in which providers are selected, which impact our scalapack CI configs.

E.g. https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5602008308 where netlib-scalapack is used instead of MKL.

### DIFF
--- a/ci/docker/debug-cpu-stdexec.yaml
+++ b/ci/docker/debug-cpu-stdexec.yaml
@@ -18,9 +18,18 @@ spack:
       true
 
   specs:
-    - dla-future@master +miniapps +ci-test +ci-check-threads ^intel-mkl threads=openmp ^mpich ^pika+stdexec ^stdexec@git.48c52df0f81c6151eecf4f39fa5eed2dc0216204=main
+    - dla-future@master build_type=Debug +miniapps +ci-test +ci-check-threads
 
   packages:
     all:
       variants:
+        - 'build_type=Release'
+    pika:
+      require:
+        - '+stdexec'
+        - 'build_type=Debug'
+        - 'malloc=system'
+    stdexec:
+      require:
+        - '@git.48c52df0f81c6151eecf4f39fa5eed2dc0216204=main'
         - 'build_type=Debug'
--- a/ci/docker/debug-cpu.yaml
+++ b/ci/docker/debug-cpu.yaml
@@ -18,15 +18,22 @@ spack:
       true
 
   specs:
-    - dla-future@master build_type=Debug +miniapps +ci-test +ci-check-threads ^openblas threads=openmp ^mpich
+    - dla-future@master build_type=Debug +miniapps +ci-test +ci-check-threads
 
   packages:
     all:
       variants:
         - 'build_type=Release'
+    blas:
+      require:: 'openblas'
+    lapack:
+      require:: 'openblas'
     pika:
-      variants:
+      require:
         - 'build_type=Debug'
         - 'malloc=system'
+    # For correct code coverage
+    # https://github.com/eth-cscs/DLA-Future/pull/1002
     umpire:
-      require: '@2022.03.1'
+      require:
+        - '@2022.03.1'
--- a/ci/docker/debug-cuda-scalapack.yaml
+++ b/ci/docker/debug-cuda-scalapack.yaml
@@ -18,13 +18,22 @@ spack:
       true
 
   specs:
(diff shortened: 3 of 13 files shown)


rasolca:
cscs-ci run

ci/docker/debug-cpu-stdexec.yaml
```diff
@@ -18,9 +18,18 @@ spack:
       true
 
   specs:
-    - dla-future@master +miniapps +ci-test +ci-check-threads ^intel-mkl threads=openmp ^mpich ^pika+stdexec ^stdexec@git.48c52df0f81c6151eecf4f39fa5eed2dc0216204=main
+    - dla-future@master build_type=Debug +miniapps +ci-test +ci-check-threads
```
RMeli, line 21: In `common.yaml`, `threads=openmp` is only a variant of `intel-mkl`. My current understanding is that variants have low precedence. Do we need
```
intel-mkl:
  require:
  - "threads=openmp"
```
to ensure the same behavior?

ci/docker/debug-cpu.yaml
```diff
@@ -18,15 +18,22 @@ spack:
...
+    - dla-future@master build_type=Debug +miniapps +ci-test +ci-check-threads
 
   packages:
     all:
       variants:
         - 'build_type=Release'
+    blas:
+      require:: 'openblas'
+    lapack:
+      require:: 'openblas'
```
RMeli, line 30: Related to comment above for `intel-mkl`.
```suggestion
    blas:
      require: 'openblas threads=openmp'
    lapack:
      require: 'openblas threads=openmp'
```
?

ci/docker/debug-cuda-scalapack.yaml
```diff
@@ -18,13 +18,22 @@ spack:
...
+    - dla-future@master build_type=Debug +cuda +miniapps +scalapack +ci-test +hdf5
 
   packages:
     all:
       variants:
         - 'build_type=Release'
+    blas:
+      require:: 'openblas'
+    lapack:
+      require:: 'openblas'
```
RMeli, line 30:
```suggestion
    blas:
      require: 'openblas threads=openmp'
    lapack:
      require: 'openblas threads=openmp'
```
?

ci/docker/debug-cuda.yaml
```diff
@@ -18,15 +18,17 @@ spack:
       true
 
   specs:
-    - dla-future@master build_type=Debug +cuda +miniapps +ci-test ^openblas threads=openmp ^mpich
+    - dla-future@master build_type=Debug +cuda +miniapps +ci-test
```
RMeli, line 21: Is
```
    blas:
      require: 'openblas threads=openmp'
    lapack:
      require: 'openblas threads=openmp'
```
missing in this YAML file? I think this now defaults to `intel-mkl`.

rasolca: Ups...

rasolca:
@RMeli Let me answer all your comments here:
I had a chat with @albestro discussing the best method and we concluded the following:

1) the virtual packages just specify which provider you want to use. The options for the specific packages go in the specs of the package itself. In this way we can specify `threads=openmp` for both openblas and intel-mkl in `common.yaml`

2) use array instead of string in `require:` (only exception in virtual packages)
in this way specs combine and are not overwritten.

3) prefer `variant:` in `common.yaml` unless it is an hard requirement (e.g. mpich version 3.4.2 is an hard requirement otherwise sarus mpi replacement is not working.)
In the CI, `variant:` should not be problematic as we always start with a fresh spack.
Moreover, being just a preference it is possible for the environments to specify a requirement which is in contrast with the `common.yaml` defaults.

RMeli:
> The options for the specific packages go in the specs of the package itself.

Ok, I see what it is going on now, thanks for the clarification.

> In the CI, `variant:` should not be problematic as we always start with a fresh spack.

Sounds good.

Thanks for the clarifications!

ci/docker/debug-cuda.yaml
RMeli: This is now fixed.

rasolca:
cscs-ci run

msimberg (approved):

albestro (approved):

rasolca:
cscs-ci run

albestro (approved):

albestro:
Good catch @rasolca 🧐 😄

PR #1057, opened 2023-11-30
debug CI

rasolca:
Just trying to debug problems like:
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5645709351


rasolca:
cscs-ci run

rasolca:
cscs-ci run

rasolca:
cscs-ci run

msimberg:
I don't know if it helps, but it seems to also have happened on a non-codecov configuration: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5676897631.

rasolca:
As the timeout is showing up more often, I would try to upstream this changes.

msimberg (approved):
@rasolca no objection to this. Do I understand correctly that setting `SLURM_WAIT=0` just means wait forever for the job to finish (https://slurm.schedmd.com/srun.html#OPT_wait)? And then we rely on the gitlab job timeout to kill the job instead if it hangs?

rasolca:
> @rasolca no objection to this. Do I understand correctly that setting `SLURM_WAIT=0` just means wait forever for the job to finish (https://slurm.schedmd.com/srun.html#OPT_wait)? And then we rely on the gitlab job timeout to kill the job instead if it hangs?

Not fully correct.
When one of the processes terminates, slurm expect all other processes to terminate within the wait time, otherwise it kills them. `SLURM_WAIT=0` just disable this behaviour. If the job reaches the time limit it is still killed.

PR #1060, opened 2023-12-04
Output CSV data for all miniapps

biddisco:
Certain benchmarks may benefit from additional plotting info present in the output, rather than embedded in the directory name or launch script - this PR is an extension of #809 - it adds CSV support to all the miniapps, rather than just a selected subset.


rasolca (approved):

rasolca:
cscs-ci run

PR #1061, opened 2023-12-04
miniapp benchmarks: add a template for job launch to allow srun alternative

biddisco:
The systems have a "launch command" template that is filled out with the job name etc and can be customized for other batch/launch methods

The current scripts assume that srun is used, but some systems may require a different method of invocation

### DIFF
--- a/scripts/miniapps.py
+++ b/scripts/miniapps.py
@@ -10,7 +10,7 @@
 
 from itertools import product
 from math import ceil, sqrt
-from os import makedirs
+from os import makedirs, environ
 from os.path import expanduser, isfile
 from time import sleep
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


scripts/miniapps.py
```diff
@@ -134,8 +139,24 @@ def submitJobs(self, run_dir, debug=False):
...
             return
 
-        print(f"Submitting : {job_file}")
-        system(f"sbatch --chdir={job_path} {job_file}")
+        launch_cmd = self.system["Launch command"]
+        # fstring substitution of vars in launch_cmd
+        launch_cmd = launch_cmd.format(job_path=job_path, job_file=job_file)
+        # get default SHELL if not specified
+        shell = environ["SHELL"]
+        if shell == None:
```
RMeli, line 147:
```suggestion
        if shell is None:
```

scripts/miniapps.py
```diff
@@ -134,8 +139,24 @@ def submitJobs(self, run_dir, debug=False):
...
+            raise ValueError(f"no $SHELL environemt variable found")
+        print(f"Submitting : {launch_cmd} using shell {shell}")
+
+        process = Popen(
+            f"{launch_cmd}",
```
RMeli, line 152:
```suggestion
            launch_cmd,
```

rasolca (approved):

PR #1068, opened 2023-12-11
Print warning when MPI pool is unintentionally not used

### LINKED ISSUE #165: No warning for "mpi" pool not used in Cholesky Factorization
At the moment the distributed version of the Cholesky factorization silently checks if a dedicated resource for `mpi` is available and in case it uses it.

Due to possible changes in HPX API (see https://github.com/eth-cscs/DLA-Future/pull/144#issuecomment-589531532), it is risky since we may not be aware of a behavior change.

I suggest adding a warning message on the MPI pool check, in case it is not available when there are multiple ranks.

msimberg:
Fixes #165.

Warns if using more than one rank, no MPI pool, and `--dlaf:no-mpi-pool` was _not_ used. This would indicate that the resource partitioner was not correctly initialized to use the MPI pool.

Also fixes parsing of command line options in tests. When `pika_main` takes `argc` and `argv` it contains all the command line options that were not yet parsed by pika, including the added DLA-Future command line options. It's not really ideal that this is even possible, but I'm not sure yet how to deal with this better.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -190,6 +190,16 @@ void updateConfiguration(const pika::program_options::variables_map& vm, configu
                            "umpire-device-memory-pool-initial-bytes");
   cfg.mpi_pool = (pika::resource::pool_exists("mpi")) ? "mpi" : "default";
 
+  // Warn if not using MPI pool without --dlaf:no-mpi-pool
+  int ntasks;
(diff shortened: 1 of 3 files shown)


rasolca (approved):

rasolca:
cscs-ci run

msimberg:
cscs-ci run

RMeli (approved):

albestro (approved):

PR #1083, opened 2024-01-17
Workarounds for broken complex type operator overloads in HIP 5.7 onwards

msimberg:
Nothing to see here yet. This is an ugly, hacky attempt at not using the broken upstream operators, but I'm unsure if 1. it works correctly, 2. it generates the same/similar code, and 3. if it's maintainable. Opening as PR to see how it works in CI. I know it compiles with HIP 5.2.3, nothing more.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
I've just discovered that it seems like spack patches the issue in HIP: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/hip/0015-reverting-operator-mixup-fix-for-slate.patch. The upstream comment about use of the operators being discouraged still applies of course, and they haven't reverted the change that broke things upstream. That said I'm probably going to lower the priority of this a bit, and we can perhaps think longer term about a "whipblas" interface.

msimberg:
I'll close this as #1195 proposes a simpler fix. We can always revisit the more complicated approach.

PR #1087, opened 2024-01-30
Add pkgconfig dependency to Spack package

RMeli:
`pkgconfig` is required in `FindLAPACK`, but it is not listed as a dependency in our Spack package.

The issue popped up when adding the `+dlaf` variant to CP2K in Spack CI: https://github.com/spack/spack/pull/42346.

### DIFF
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -17,12 +17,30 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
 
     license("BSD-3-Clause")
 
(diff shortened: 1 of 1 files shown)


spack/packages/dla-future/package.py
```diff
@@ -46,6 +46,7 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
     )
 
     depends_on("cmake@3.22:", type="build")
+    depends_on("pkgconfig", type="build")
```
rasolca, line 49: Shouldn't this be needed only when not using MKL, or am I missing something?

RMeli: No, I think you are right. Is there an easy way to express this with `when=`?

rasolca (approved):

rasolca:
cscs-ci run

PR #1088, opened 2024-01-31
Fix use of MPI GPU RDMA variables

msimberg:
The variables were inconsistently named, leading to the CMake variable not working. I've also made the CMake variable dependent on `DLAF_WITH_HIP`. At this point I've simply made them consistent, but I'd also like to propose a rename of the variable since it applies to both HIP and CUDA.

Some options:
- `DLAF_WITH_MPI_GPU_SUPPORT`
- `DLAF_WITH_MPI_GPU`
- `DLAF_WITH_GPUDIRECT`
- `DLAF_WITH_GPU_RDMA`
- etc.

I think I quite like the first option as it's quite straightforward to understand and uses generic enough terms (e.g. I just learned that GPUDirect seems to be an NVIDIA marketing term: https://developer.nvidia.com/gpudirect).


RMeli:
+1 for `DLAF_WITH_MPI_GPU_SUPPORT`

src/communication/kernels/reduce.cpp
```diff
@@ -8,8 +8,8 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 
-#ifdef DLAF_WITH_CUDA_RDMA
-#warning "Reduce is not using CUDA_RDMA."
+#ifdef DLAF_WITH_MPI_GPU_SUPPORT
+#warning "Reduce is not using MPI_GPU_SUPPORT."
```
msimberg, line 12: @albestro do you think we still need this warning? It's not so good with `-Werror`. If we still think it's possible to make reduce support GPU memory directly we should perhaps track it in an issue instead?

albestro: I'm ok with removing it.

It was just a placeholder to make the user aware that even if MPI with GPU support is enabled and it works, actually `reduce` is not using it.

In case we decide to track that with an issue, it won't be more hidden than it is now. I'm not even sure it is something we want to track.

rasolca: Not sure why the reduce was not implemented.
The only requirement is to have contiguous memory (but the requirement is present in the CPU version as well)

rasolca (approved):

RMeli (approved):

src/communication/kernels/reduce.cpp
albestro: So I would go for tracking it with an issue so we can eventually implement it. 😄

albestro: https://github.com/eth-cscs/DLA-Future/issues/1094

msimberg: Thanks @albestro!

albestro (approved):

msimberg:
cscs-ci run

PR #1089, opened 2024-02-01
Add support for logarithmic Y axis to trmm plotting functions

msimberg:
I've also updated the default to `None` instead of `False` simply to be consistent with the other function definitions (not because I prefer `None` over `False`).

### DIFF
--- a/scripts/postprocess.py
+++ b/scripts/postprocess.py
@@ -785,7 +785,8 @@ def gen_trsm_plots_weak(
 
 def gen_trmm_plots_strong(
     df,
-    logx=False,
+    logx=None,
+    logy=None,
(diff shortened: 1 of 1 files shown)


rasolca (approved):
Note: `False` is different from `None`:

- `None` uses the default (https://github.com/eth-cscs/DLA-Future/blob/fd6a8833b27d539a45562381110cc611d85e0b02/scripts/postprocess.py#L23-L24) values that are set by the command line options (https://github.com/eth-cscs/DLA-Future/blob/fd6a8833b27d539a45562381110cc611d85e0b02/scripts/postprocess.py#L395-L403).
- `False` means always linear scale

PR #982, opened 2023-09-12
Add test stage for ROCm configuration

msimberg:
This won't work yet...

- [ ] Add testing also for stdexec configuration

PR #1090, opened 2024-02-01
Initialize HIP before MPI for older Cray MPICH versions

msimberg:
I should've done this a long time ago, but was reminded again when running benchmarks on clariden. On LUMI Cray seems to have fixed the issue of MPI init making HIP devices disappear. If HIP is initialized before MPI, everything seems to work normally. On clariden this is still an issue. This puts a `hipInit` into the `mpi_init` struct which makes it effective in all the miniapps. If needed we can consider doing the same for tests with #982.

### DIFF
--- a/include/dlaf/communication/init.h
+++ b/include/dlaf/communication/init.h
@@ -16,6 +16,11 @@
 
 #include <mpi.h>
 
+#ifdef DLAF_WITH_HIP
+#include <hip/hip_runtime.h>
+#include <whip.hpp>
+#endif
+
 #include <pika/mutex.hpp>
 
 #include <dlaf/communication/error.h>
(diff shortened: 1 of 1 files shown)


msimberg:
cscs-ci run

include/dlaf/communication/init.h
```diff
@@ -26,6 +31,12 @@ namespace comm {
 struct mpi_init {
   /// Initialize MPI to MPI_THREAD_MULTIPLE
   mpi_init(int argc, char** argv) noexcept {
+    // On older Cray MPICH versions initializing HIP after MPI leads to HIP not seeing any devices. Hence
```
rasolca, line 34: Do you know which version is ok and which is problematic?
It would be nice to put it in the comment.

rasolca (approved):

PR #1093, opened 2024-02-05
Remove `eval` from wrapper scripts

msimberg:
Don't use `eval` to avoid surprising variable expansion. Variable expansion can instead be opted into by explicitly wrapping a command in `eval` or `bash -c`.


scripts/gpu2ranks_ompi
```diff
@@ -26,4 +26,4 @@ then
   fi
 fi
 
-eval "$@"
+"$@"
```
rasolca, line 29: I would remove this file completely, as it is not used anymore.

rasolca (approved):

scripts/gpu2ranks_ompi
msimberg: Good point. Removed in https://github.com/eth-cscs/DLA-Future/pull/1093/commits/a44eacc6eba3a1424179f784951467c99590dc91.

PR #1096, opened 2024-02-09
Add option to force use of contiguous buffers when using `DLAF_WITH_MPI_GPU_SUPPORT`

msimberg:
Based on #1088.

Introduces another dependent CMake option `DLAF_WITH_MPI_GPU_SUPPORT_FORCE_CONTIGUOUS` which I propose to default to `ON` when `DLAF_WITH_MPI_GPU_SUPPORT` is enabled, given that it seems to always perform better than not forcing contiguous buffers (small improvement with HIP, huge improvement with CUDA). I'd say long-term we can consider if it should just always be used without a CMake option, but I think at this early stage when we're still figuring out what works where I think it's nice to have an option for it.

Independent of the option we could also consider making both `DLAF_WITH_MPI_GPU_SUPPORT` and `DLAF_MPI_GPU_SUPPORT_FORCE_CONTIGUOUS` runtime options, but that would require larger changes and is a separate discussion.

In this PR I'm only forcing GPU buffers to be contiguous if the option is enabled. CPU buffers are left as before.

I've currently forced the use of contiguous buffers at the call-site of `withTemporaryTile`. This could also be done inside `withTemporaryTile` but at least in theory `withTemporaryTile` is not meant only for communication so I don't know if it makes sense. Communication is, however, the only use case for it right now. I'm open to thoughts on what you think is cleaner.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

src/communication/kernels/broadcast.cpp
```diff
@@ -50,8 +50,13 @@ template <class CommSender, class TileSender>
   // The input tile must be copied to the temporary tile used for the send, but
   // the temporary tile does not need to be copied back to the input since the
   // data is not changed by the send. A send does not require contiguous memory.
+  constexpr bool require_contiguous =
```
rasolca, line 53: Do you mean
```suggestion
  constexpr auto require_contiguous =
```
?

rasolca (changes requested):

src/communication/kernels/broadcast.cpp
msimberg: Yep... good catch.

msimberg: https://github.com/eth-cscs/DLA-Future/pull/1096/commits/0c0fea8fb7db11ec65a2b154aa8f8f8b920931fd

rasolca:
cscs-ci run

CMakeLists.txt
```diff
@@ -41,6 +41,10 @@ option(DLAF_WITH_HIP "Enable HIP support" OFF)
 cmake_dependent_option(
   DLAF_WITH_MPI_GPU_SUPPORT "Enable MPI GPU support" OFF "DLAF_WITH_CUDA OR DLAF_WITH_HIP" OFF
 )
+cmake_dependent_option(
+  DLAF_WITH_MPI_GPU_SUPPORT_FORCE_CONTIGUOUS "Force GPU buffers to be contigous before communicating" ON
```
albestro, line 45:
```suggestion
  DLAF_WITH_MPI_GPU_SUPPORT_FORCE_CONTIGUOUS "Force GPU buffers to be contiguous before communicating" ON
```

rasolca (approved):

CMakeLists.txt
albestro: Is there any rationale behind the decision of having it defaulted to ON? I know benchmark results, just wondering if it is somehow future-proof or not...

albestro (approved):
LGTM, just a couple of comments.

CMakeLists.txt
msimberg: No rationale other than what you already mentioned. The parent option `DLAF_WITH_MPI_GPU_SUPPORT` is anyway off by default so I thought this doesn't hurt. I would consider the whole thing experimental and for that reason I don't feel too bad about defaulting `DLAF_WITH_MPI_GPU_SUPPORT_FORCE_CONTIGUOUS` to on without more benchmarking results.

msimberg: https://github.com/eth-cscs/DLA-Future/pull/1096/commits/1afdf643f33bf8a27b071611f480fc83b0aef4b1 and bonus: https://github.com/eth-cscs/DLA-Future/pull/1096/commits/d5bc59e221bdbbcf7021ad79856d5afd3bbbb663.

PR #1100, opened 2024-02-19
Use ninja in spack package

msimberg:


msimberg:
I was hoping this would slightly improve build times (it does for pika because the generated build files expose more parallelism) but that doesn't seem to be the case in CI (at least looking at the last few builds there is no noticeable difference). I'm still proposing this because:
- pika (the spack package) already depends on ninja
- the output of ninja is much more pleasant to look at
- on slow filesystems the meta-operations of ninja vs make are much faster (e.g. determining if anything needs to be rebuilt)

albestro (approved):

PR #1104, opened 2024-02-22
Fix data race in `test_pipeline`

msimberg:
Reading the size of the `accesses` vector races with pushing items into the vector. The pushing is done under a lock, but reading the size was not done under a lock, and doing so is considered a data race (there are no guarantees on what the size will return while the vector is being modified). This only applies to the case when subpipelines are accessed in new tasks, not inline, but I've added the lock for both cases to keep things simple. See https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/6223967358 (view the full log) for the report by thread sanitizer.

### DIFF
--- a/test/unit/common/test_pipeline.cpp
+++ b/test/unit/common/test_pipeline.cpp
@@ -810,7 +810,11 @@ TEST(SubPipeline, RandomAccess) {
       // of senders stored in accesses. No accesses should be done yet at this point since they depend
       // on the first access completing.
       ASSERT_GE(state.spawn_count, 1);
-      ASSERT_GE(state.spawn_count, state.accesses.size());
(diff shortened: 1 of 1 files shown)


msimberg:
cscs-ci run

albestro (approved):

msimberg:
The test failure was jfrog being unavailable: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/6233220389#L1345. I've restarted the failed job.

RMeli (approved):

rasolca (approved):

Issue #1013, opened 2023-10-16
Attempt to decrease binary size of `with_temporary_tile` test


msimberg:
The `with_temporary_tile` test is currently by far the largest test, at 175 MB (with CUDA): https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/5289553001#L3156. While not critical, it may be worth looking into whether it's possible to decrease the size within reason.

PR #1105, opened 2024-02-23
miniapp communication

rasolca:
- Moved the template scheduling functions to internal headers (used by the miniapp as well)
- Renamed the scheduling functions according to the new naming conventions.
- implemented communication miniapp which test the current default settings and the relevant combinations, to allow a comparison to find the best combination.


include/dlaf/eigensolver/tridiag_solver/merge.h
```diff
@@ -25,6 +25,7 @@
 #include <dlaf/communication/communicator_pipeline.h>
 #include <dlaf/communication/index.h>
 #include <dlaf/communication/kernels.h>
+#include <dlaf/communication/kernels/internal/broadcast.h>
```
rasolca, line 28: Note: needed for https://github.com/eth-cscs/DLA-Future/blob/691794b11e472ee337ccb9e19122a42ae35406f6/include/dlaf/eigensolver/tridiag_solver/merge.h#L1269-L1270 as the function objects has been moved in the internal file.

msimberg:
@rasolca looks reasonable to me. Just to check my understanding, what you're benchmarking is a broadcast from rank 0 to all other ranks:

- with the default options for contiguous/non-contiguous
- with all the combinations of CPU/GPU memory, contiguous/non-contiguous send/recv

Probably my ignorance, but do you need ensure that each rank will do as many receives as the root rank does sends? Or since it's a broadcast, do you not need to do a receive on every rank?

albestro (commented):
LGTM.

rasolca:
> @rasolca looks reasonable to me. Just to check my understanding, what you're benchmarking is a broadcast from rank 0 to all other ranks:
> 
>     * with the default options for contiguous/non-contiguous
> 
>     * with all the combinations of CPU/GPU memory, contiguous/non-contiguous send/recv
 
Exactly

> Probably my ignorance, but do you need ensure that each rank will do as many receives as the root rank does sends? Or since it's a broadcast, do you not need to do a receive on every rank?

The idea is to allocate a local matrix with the same size on all the ranks. I still have a couple of open TODO when creating the matrix. It might be that I'm still using a distributed matrix by mistake.

rasolca:
Summary of the first benchs (A100):

- bcast p2p
  - force contiguous is faster for CPU communication as well (especially bcast, but for p2p as well if the ranks are on different node).
  - copying CPU to GPU and communicate is even faster
  - non contiguous GPU communication is terribly 
- reduce/all_reduce:
  - on a single node GPU communication is slightly faster
  - on two nodes copy to CPU and communicate is faster

rasolca:
Note: Compilation of the new miniapp is very slow.

rasolca:
cscs-ci run

rasolca:
cscs-ci run

msimberg:
> Note: Compilation of the new miniapp is very slow.

Likely due to the same reason as #1013. I have been getting increasingly annoyed by compile times of the other miniapps recently as well, though I don't think it's necessarily gotten worse. It may be worth bumping this up my to do list, or if someone else feels motivated they could look into it.

CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ cmake_dependent_option(
 )
 cmake_dependent_option(
   DLAF_WITH_MPI_GPU_FORCE_CONTIGUOUS "Force GPU buffers to be contiguous before communicating" ON
-  "DLAF_WITH_MPI_GPU_SUPPORT" OFF
+  "DLAF_WITH_MPI_GPU_AWARE" OFF
```
msimberg, line 46: #1107.

include/dlaf/communication/kernels/all_reduce.h
```diff
@@ -32,11 +32,11 @@ namespace dlaf::comm {
 /// signals completion when the reduction is done. The output sender tile must
 /// be writable so that the received and reduced data can be written to it. The
 /// output tile is sent by the returned sender.
-template <class T, Device D>
-[[nodiscard]] dlaf::matrix::ReadWriteTileSender<T, D> scheduleAllReduce(
+template <class T, Device D_in, Device D_out>
```
msimberg, line 35: Camel case?
```suggestion
template <class T, Device DIn, Device DOut>
```

include/dlaf/communication/kernels/broadcast.h
```diff
@@ -29,55 +29,28 @@
...
-}
 
 /// Schedule a broadcast send.
 ///
 /// The returned sender signals completion when the send is done. If the input
 /// tile is movable it will be sent by the returned sender. Otherwise a void
 /// sender is returned.
 template <class T, Device D, class Comm>
-[[nodiscard]] pika::execution::experimental::unique_any_sender<> scheduleSendBcast(
+[[nodiscard]] pika::execution::experimental::unique_any_sender<> schedule_bcast_send(
```
msimberg, line 39: No objection to the name change, I just wanted to ask what the motivation for the change was? (I'm referring to `bcast send` vs `send bcast`, not the change of case).

include/dlaf/communication/kernels/internal/all_reduce.h
```diff
@@ -0,0 +1,152 @@
...
+#include <dlaf/communication/communicator_pipeline.h>
+#include <dlaf/communication/kernels/all_reduce.h>
+#include <dlaf/communication/message.h>
+#include <dlaf/communication/rdma.h>
+#include <dlaf/matrix/tile.h>
+#include <dlaf/sender/transform_mpi.h>
+#include <dlaf/sender/with_temporary_tile.h>
+
+namespace dlaf::comm::internal {
+template <class T, Device D_in, Device D_out>
```
msimberg, line 32:
```suggestion
template <class T, Device DIn, Device DOut>
```
?

miniapp/miniapp_communication.cpp
```diff
@@ -0,0 +1,610 @@
...
+}
+
+template <Backend B, class CommPipeline, class T>
+void benchmark_all_reduce(int64_t run_index, const Options& opts, Communicator& world,
+                          CommPipeline&& pcomm, Matrix<const T, Device::CPU>& matrix_ref) {
+  using dlaf::comm::scheduleAllReduce;
+
+  constexpr Device D = DefaultDevice_v<B>;
+  Matrix<T, D> matrix = get_matrix_on_device_sync<D>(matrix_ref);
+  Matrix<T, D> matrix_out = get_matrix_on_device_sync<D>(matrix_ref);
```
msimberg, line 164: Is the reason for preferring `get_matrix_on_device_sync` over `MatrixMirror` that you actually want to force `matrix` and `matrix_out` to be new matrices, independent of the device?

msimberg (commented):
Looks good to me. Only some superficial comments and questions.

include/dlaf/communication/kernels/broadcast.h
rasolca: To have a consistent naming among the different collectives.
`scheduleSendBcast` vs `scheduleReduceSend`
I chose the second one, but can go with the first option as well if you think is better.

include/dlaf/communication/kernels/internal/all_reduce.h
rasolca: They are variable.
In the beginning we were using `Device device` and `Backend backend` in templates, but at some point we switched to `Device D` and `Backend B`. TBH I don't remember the reason for the change.

miniapp/miniapp_communication.cpp
rasolca: Exactly, I want to avoid an extra CPU copy if the resulting matrix should be on GPU.

include/dlaf/communication/kernels/broadcast.h
msimberg: That makes sense, thanks!

include/dlaf/communication/kernels/internal/all_reduce.h
msimberg: Ah, my understanding has been "they are template parameters and thus CamelCase" overriding "they are variable names and thus snake_case". I may have contributed to the current situation. And at least when I looked now in `coding.md` I actually don't see any explicit mention of it so I may just have misunderstood.

If we're changing them to follow the "variable names" convention, then it should be `d_in`?

I don't mind either way.

miniapp/miniapp_communication.cpp
msimberg: I clarified this directly with @rasolca and this is a combination of:
- Always wanting to work with a copy of the reference, and
- Wanting to avoid an unnecessary copy in the GPU case.

The reference is always on CPU. With `MatrixMirror` one would have first one copy of the reference to CPU, and then another copy from CPU to GPU. The CPU-side copy created for `MatrixMirror` is not used. With `get_matrix_on_device_sync` one can copy directly from the reference to the GPU.

On CPU there's no difference.

Thanks for the explanation @rasolca!

include/dlaf/communication/kernels/p2p_allsum.h
```diff
@@ -39,8 +39,8 @@ namespace dlaf::comm {
 /// `in` is a sender of a read-only tile that, together with the one received from `rank_mate`, will be
 /// summed in `out` (on both ranks)
 template <Backend B, class CommSender, class SenderIn, class SenderOut>
-[[nodiscard]] auto scheduleAllSumP2P(CommSender&& comm, IndexT_MPI rank_mate, IndexT_MPI tag,
-                                     SenderIn&& in, SenderOut&& out) {
+[[nodiscard]] auto schedule_sum_p2p(CommSender&& comm, IndexT_MPI rank_mate, IndexT_MPI tag,
```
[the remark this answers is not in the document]
rasolca: IMO it is redundant.

msimberg (approved):

PR #1109, opened 2024-03-12
Don't assume miniapps are in `miniapp` subdirectory

msimberg:
With this change it's easier to use miniapps from e.g. a spack-installed DLA-Future (which puts the miniapps in the `bin` subdirectory). One simply has to be explicit about the full path to the miniapps after this change.

### DIFF
--- a/scripts/gen_dlaf_strong-gpu.py
+++ b/scripts/gen_dlaf_strong-gpu.py
@@ -20,7 +20,7 @@ import systems
 
 system = systems.cscs["daint-gpu"]
 
-dlafpath = "<path_to_dlaf_build_dir>"
+dlafpath = "<path_to_dlaf_miniapp_dir>"
 matrixrefpath = "<path_to_h5_refs>"
 
 run_dir = "~/ws/runs/strong"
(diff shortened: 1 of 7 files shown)


msimberg:
@rasolca at least I couldn't find any logic related to falling back to not using the `miniapp` subdirectory or similar, but I may have missed it. There is a check that the executable exists, which is good, and which is how I first noticed that I was using the wrong directory.

albestro (approved):
Thanks for doing this!

Just a minor comment: `{build_dir}` can be now replaced with something like `{miniapp_bin_dir}` or even just `{miniapp_dir}`.

msimberg:
> Thanks for doing this!
> 
> Just a minor comment: `{build_dir}` can be now replaced with something like `{miniapp_bin_dir}` or even just `{miniapp_dir}`.

Great point! I quite like `miniapp_dir`. On the other hand `build_dir` is used for all the other libraries as well. I think they still mostly assume the tests/miniapps come from a build directory, but I wouldn't be sad if they use `miniapp_dir` as well. I'll change it with a few :+1:.

msimberg:
cscs-ci run

RMeli (approved):

rasolca (approved):

PR #1110, opened 2024-03-13
Fix warning about strict prototypes in precompiled headers `dummy.c`

msimberg:
clang starting with version 15 warns about C functions that don't declare their arguments (with `-Wstrict-prototypes`, see https://godbolt.org/z/YbjY14d74). Explicitly specifying that `dummy` takes no arguments with `void` silences the warning.

### DIFF
--- a/src/dummy.c
+++ b/src/dummy.c
@@ -11,4 +11,4 @@
 // target.
 
 // Silences -Wpedantic: ISO C forbids an empty translation unit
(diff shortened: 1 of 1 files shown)


msimberg:
cscs-ci run

RMeli (approved):
Thanks, LGTM.

rasolca (approved):

PR #1111, opened 2024-03-15
Generate ETI files

msimberg:
Nothing to see here yet. It's not a clear win to split the ETI files into smaller files (some builds become faster, but there is overhead), but checking to see what the effect is in CI.


msimberg:
Generally speaking, rebuilds of single algorithms become faster since they can be faster, and full rebuilds become slower, since they had enough parallelism anyway. I'm leaving the PR as a draft, but I'm not going to do any further work on it for the time being as the benefit is not entirely clear.

msimberg:
Closing just to tidy up my open PRs. This can always be revived later. I'm not deleting the branch.

PR #1112, opened 2024-03-19
Add miniapps as tests

msimberg:
Adds the algorithms miniapps as tests. Does not add `miniapp_laset` as it's not really testing anything too useful, and it messes with the `check-threads` setup (which checks if OS threads are being spawned that shouldn't be spawned). The miniapps are run with 6 ranks, `--grid-rows 3 --grid-cols 2`, and otherwise default options.

This splits `DLAF_addTest` into two separate functions: one which adds a test given a target (`DLAF_addTargetTest`) which is used to add the miniapps since they already have CMake targets, and `DLAF_addTest` which creates an executable and then calls `DLAF_addTargetTest`. The behaviour of `DLAF_addTest` remains unchanged.

This also adds a `CATEGORY` parameter that can be added to tests as a label. This is added alongside the `RANK_*` labels. The `CATEGORY` defaults to `UNIT` for "unit tests", and I set it to `MINIAPP` for the miniapps. I'm fine with making the choice explicit and/or renaming/adding the categories. I then use the `RANK` and `CATEGORY` labels to generate jobs the same way `RANK` is used currently on `master`. Combinations that have no tests are not added as CI jobs.

Finally, many miniapps were hanging on the CUDA configurations, where they run with only one worker thread and one MPI thread. I've changed the `waitLocalTiles` calls to `pika::wait` calls before `MPI_Barrier` is called to make sure that really nothing is running anymore and deadlocks are avoided. I've only changed them to `pika::wait` after the algorithms are called, but for consistency it might make sense to use `pika::wait` anywhere before `MPI_Barrier` is called?


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

ci/ctest_to_gitlab.sh
```diff
@@ -98,20 +98,28 @@ JOB_TEMPLATE="
...
-    JOB=`echo "$JOB_TEMPLATE" | sed "s|{{LABEL}}|$label|g" \
-                              | sed "s|{{NTASKS}}|$N|g" \
-                              | sed "s|{{CPUS_PER_TASK}}|$C|g"`
+        # Skip label combinations that match no tests
+        if ! ctest -N -L $category_label -L $rank_label | egrep --quiet "^Total Tests: [1-9][0-9]?$"; then
+            continue
+        fi
 
-    JOBS="$JOBS$JOB"
+        JOB=`echo "$JOB_TEMPLATE" | sed "s|{{CATEGORY_LABEL}}|$category_label|g" \
```
RMeli, line 116: Bikeshedding: what about dropping the `CATEGORY_` prefix, to make the job name a bit less verbose and more readable? Probably personal preference, but I prefer `MINIAPP_RANK_6` to `CATEGORY_MINIAPP_RANK_6` (i.e. the `CATEGORY` part is a bit redundant IMO, since it will be added to all tests).

msimberg: Yep, that sounds reasonable I think. I initially just wanted to keep it as straightforward of a mapping from CTest labels to CI job names, but I'd be ok with dropping the prefix there.

Just checking: are you suggesting to drop the `CATEGORY_` prefix from the job name only, or also from the CTest label? For the latter it's still handy to be explicit; `CATEGORY` and `RANK` essentially act as keys in key-value pairs. The filtering is a bit more straightforward I think if `CATEGORY` stays in the label name.

RMeli: > are you suggesting to drop the `CATEGORY_` prefix from the job name only, or also from the CTest label?

Initially I thought it would be nice to drop it from both, but then I realized you were parsing the label as you point out. That's why I suggested to drop it only from CI, just for having a slightly less bloated output in the GitLab UI.

msimberg: Thoughts from the others?

albestro (approved):

ci/ctest_to_gitlab.sh
msimberg: https://github.com/eth-cscs/DLA-Future/pull/1112/commits/dfb60874b642cc05bf34f549e7898e1c6f7cd0b1

RMeli (approved):

msimberg:
> Not happy about spreading `pika::wait`s everywhere in the miniapps. I'd prefer another solution. Would scheduling and `sync_wait`ing an Ibarrier work?

I don't know, but I can try. It would need `IBarrier`-ing all the communicators in that case.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

miniapp/miniapp_reduction_to_band.cpp
```diff
@@ -158,6 +160,11 @@ struct reductionToBandMiniapp {
         // wait and barrier for all ranks
         matrix.waitLocalTiles();
         taus.waitLocalTiles();
+        for (std::size_t i = 0; i < comm_grid.num_pipelines(); ++i) {
+          sync_wait(comm_grid.full_communicator_pipeline().exclusive());
+          sync_wait(comm_grid.row_communicator_pipeline().exclusive());
+          sync_wait(comm_grid.col_communicator_pipeline().exclusive());
+        }
```
msimberg, line 167: Waiting for communicators does not strictly seem to be required in all the miniapps, but I would add it to all of them in any case. Whether or not it's required probably depends on the internal communication patterns.

In any case, the above snippet seems to work fine in CI now. I think we could either:
- add a helper function available only for tests/miniapps with the signature `void wait_all_communicators(CommunicatorGrid&)` which does the above, or
- add a member function, similar to `Matrix::waitLocalTiles`, to `CommunicatorGrid` which does the above

Do you have preferences? Other ideas? I think the latter is nice in terms of encapsulation, but I'm not sure we want to encourage users to use it (yet...?).

ci/.gitlab-ci.yml
```diff
@@ -1,18 +1,18 @@
...
+    #  - local: 'ci/cpu/clang12_release.yml'
+    #  - local: 'ci/cpu/clang13_release_cxx20.yml'
+    #  - local: 'ci/cpu/clang14_release_stdexec.yml'
+    #  - local: 'ci/cpu/clang15_release.yml'
+    #  - local: 'ci/cpu/clang16_release.yml'
+    #  - local: 'ci/cpu/gcc11_codecov.yml'
+    #  - local: 'ci/cpu/gcc11_release.yml'
+    #  - local: 'ci/cpu/gcc11_release_stdexec.yml'
+    #  - local: 'ci/cpu/gcc11_debug_stdexec.yml'
+    #  - local: 'ci/cpu/gcc12_release_cxx20.yml'
```
rasolca, line 12: Should be re-enabled.

msimberg: Reenabled.

msimberg:
cscs-ci run

msimberg:
Another two hermitian matrix multiplication hangs in https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/7841914424 and https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/7841944540. I've restarted the jobs.

miniapp/miniapp_reduction_to_band.cpp
rasolca: 2nd option looks OK.

msimberg:
cscs-ci run

miniapp/miniapp_reduction_to_band.cpp
msimberg: I've added a helper `CommunicatorGrid::wait_all_communicators` in the latest commit. Unlike the snippet above which simply waits for the communicators (which does not ensure that communication has completed) I've made `wait_all_communicators` actually do an `MPI_Ibarrier`. At the same time I've removed the blocking call to `MPI_Barrier` at the end of the algorithm miniapps.

I have not changed _all_ `MPI_Barrier` calls to use `wait_all_communicators`. I think it makes sense, but if you agree to that change I will do that in a follow-up PR. What do you think?

albestro (approved):
LGTM

ci/ctest_to_gitlab.sh
```diff
@@ -98,20 +98,30 @@ JOB_TEMPLATE="
...
-                              | sed "s|{{CPUS_PER_TASK}}|$C|g"`
-
-    JOBS="$JOBS$JOB"
+for rank_label in `ctest --print-labels | egrep -o "RANK_[1-9][0-9]?"`; do
+    for category_label in `ctest --print-labels | egrep -o "CATEGORY_[A-Z]+"`; do
+        N=`echo "$rank_label" | sed "s/RANK_//"`
+        C=$(( THREADS_PER_NODE / N ))
+
+        # Skip label combinations that match no tests
+        if ! ctest -N -L $category_label -L $rank_label | egrep --quiet "^Total Tests: [1-9][0-9]?$"; then
```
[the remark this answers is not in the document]
msimberg: No, I don't think so. Probably just me getting stuck on using grep here for some reason. I've attempted your version in https://github.com/eth-cscs/DLA-Future/pull/1112/commits/02bf96a293356ca67a877cc07d0b0d2253a7ba8b. It looks like it should do the right thing and is indeed much simpler to understand.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

miniapp/miniapp_reduction_to_band.cpp
```diff
@@ -97,6 +97,8 @@ struct Options
 struct reductionToBandMiniapp {
   template <dlaf::Backend backend, typename T>
   static void run(const Options& opts) {
+    using pika::this_thread::experimental::sync_wait;
+
```
RMeli, line 101:
```suggestion
```

miniapp/miniapp_triangular_multiplication.cpp
```diff
@@ -51,6 +51,7 @@ using dlaf::comm::CommunicatorGrid;
 using dlaf::common::Ordering;
 using dlaf::matrix::Matrix;
 using dlaf::matrix::MatrixMirror;
+using pika::this_thread::experimental::sync_wait;
```
RMeli, line 54:
```suggestion
```

RMeli (approved):
LGTM.

miniapp/miniapp_reduction_to_band.cpp
msimberg: Good catch, thanks! Fixed in https://github.com/eth-cscs/DLA-Future/pull/1112/commits/a49c151f29c8e52fe7b7c75859f2afdef2145943.

miniapp/miniapp_triangular_multiplication.cpp
msimberg: Fixed in https://github.com/eth-cscs/DLA-Future/pull/1112/commits/a49c151f29c8e52fe7b7c75859f2afdef2145943.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

albestro:
I was experiencing hangs on `miniapp_eigensolver` just after the new `transform_mpi` #1125. I don't think that's the direct cause of them, but it seems like it exposed a problem.

I was able to replicate the hang on both daint (with 8 nodes), and on todi (with 4 and 8 nodes). I did some basic investigation, and miniapp `check` was the discriminant between hang (enabled) or no hang (disabled). Actually in one of my tests I tried disabling just `hermitian_multiplication` communications and it was enough to not make it hang anymore.

As suggested by @msimberg talking about this, I tried adding `pika::wait()` in the `dlaf::finalize()` and actually that resolved hangs in my problem reproducing configurations. Moreover, I opted for giving a try to `wait_all_communicators` changes (alone) from this PR, and also them are enough to fix hangs in my test-case.

msimberg:
cscs-ci run

PR #1117, opened 2024-04-08
Check for include guards in all header files in repo, not only in `include` directory

msimberg:
Fixes missing `#pragma once` in a couple of headers.

### DIFF
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Check include guards
         if: always()
         run: |
-          find . test -type f '(' -name "*.h" -o -name "*.h.in" ')'             \
+          find . test -type f '(' -name "*.h" -o -name "*.h.in" ')'                   \
             | xargs -I{} sh -c                                                        \
                 "egrep -sq '^#pragma once' {} || echo {}"                             \
             > pragma-once.check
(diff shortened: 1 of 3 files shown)


.github/workflows/check_format.yml
```diff
@@ -47,7 +47,7 @@ jobs:
       - name: Check include guards
         if: always()
         run: |
-          find include test -type f '(' -name "*.h" -o -name "*.h.in" ')'             \
+          find . test -type f '(' -name "*.h" -o -name "*.h.in" ')'             \
```
albestro, line 50: minor: can you fix the alignment of the line continuation backslash?

albestro (approved):

RMeli (approved):
LGTM.

Note: the C++ Core Guidelines suggest _not_ to use `pragma once` (but it is less error prone, and easier to check).

.github/workflows/check_format.yml
msimberg: Yep, good catch: https://github.com/eth-cscs/DLA-Future/pull/1117/commits/cc12ac586e0a6eb6855ad27b5ee870a4d3c72387.

msimberg:
cscs-ci run

rasolca:
> Note: the C++ Core Guidelines suggest _not_ to use `pragma once` (but it is less error prone, and easier to check).

We discussed it when we decided to use it. Most of the compilers support this feature (https://en.wikipedia.org/wiki/Pragma_once) so it is not a problem.

rasolca (approved):

PR #1119, opened 2024-04-12
Update CI to use pika 0.24.0

aurianer:


msimberg:
Could you rebase this on latest master before doing a CI run?

msimberg:
cscs-ci run

msimberg:
I've updated this to use an upstream spack `develop` SHA which has 0.24.0 (https://github.com/spack/spack/pull/43624). Credits to @aurianer for all the actual work for the release!

msimberg:
cscs-ci run

msimberg:
With the spack commit bump we were affected by https://github.com/spack/spack/pull/41373, which moves the `spack develop` build directory to `/tmp`. This lead to failures because the parent directory of the new build directory location was not yet created when we try to symlink it into a known location. I've worked it around by forcing the parent directory to be created before symlinking. An alternative could be to change the build directory location in the spack configuration (if I understand correctly that should be possible). Let me know what you prefer.

albestro (approved):

PR #1121, opened 2024-04-17
Update and export project version in CMake

RMeli:
Currently, trying to use
```cmake
find_package(DLAF 0.4.0 REQUIRED)
```
results in the following error:
```
CMake Error at CMakeLists.txt:32 (find_package):
  Could not find a configuration file for package "DLAF" that is compatible
  with requested version "0.4.0".

  The following configuration files were considered but not accepted:

    /home/rmeli/spack/opt/spack/linux-ubuntu23.10-zen4/gcc-11.4.0/dla-future-0.4.0-coebvqe54dvcjliwldugouxjxo7vv6jv/lib/cmake/DLAF/DLAFConfig.cmake, version: unknown
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -354,13 +354,10 @@ configure_package_config_file(
 )
 
 write_basic_package_version_file(
-  DLAFConfigVersion.cmake
-  VERSION ${PACKAGE_VERSION}
-  COMPATIBILITY AnyNewerVersion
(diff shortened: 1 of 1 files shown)


msimberg (approved):
Excellent catch, looks good to me.

RMeli:
cscs-ci run

RMeli:
Since we are still on `0.x`, I like the idea of releasing `0.4.1`. It would be the cleanest solution IMO, as we can probably avoid start adding quite a bit of patches to the Spack package.

albestro (approved):

rasolca (approved):

Issue #1122, opened 2024-04-22
Make link libraries private if possible


RMeli:
While playing around with DLA-Future for the Fortran interface, I realized that MKL was linked directly into my application, despite my application not using it. Linking ScaLAPACK and LAPACK against `dlaf.prop_private` instead of `dlaf.prop`, solved the issue.

We should investigate which libraries can be linked to `dlaf.prop_private` instead of `dlaf.prop`.

PR #1123, opened 2024-04-22
Add ScaLAPACK explicitly

RMeli:
Building a Spack stack on Santis results in the following issue:
```
CMake Error at /user-environment/linux-sles15-neoverse_v2/gcc-12.3.0/dla-future-0.4.0-yybav6v7epu7cwhjqfzd7asrgwopoqmq/lib64/cmake/DLAF/DLAF-Targets.cmake:69 (set_targ
et_properties):
  The link interface of target "DLAF::dlaf.prop" contains:

    DLAF::SCALAPACK

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /user-environment/linux-sles15-neoverse_v2/gcc-12.3.0/dla-future-0.4.0-yybav6v7epu7cwhjqfzd7asrgwopoqmq/lib64/cmake/DLAF/DLAFConfig.cmake:38 (include)
  CMakeLists.txt:626 (find_package)
```

I haven't yet fully investigated this issue, but possibly related to #1122.


RMeli:
cscs-ci run

msimberg:
Given the error in https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/6687067202#L3097, I think this is actually not necessary, at least not in that form... Are you building with nvpl on santis? Perhaps it's not actually not finding scalapack for some reason? At least the target should be correctly created if it is found: https://github.com/eth-cscs/DLA-Future/blob/64b6ddb88c629eb175da5032251786fdbe5554be/cmake/FindSCALAPACK.cmake#L35.

RMeli:
Yes, this is definitely not the correct solution. I'll need to investigate further. I have only encountered the issue on Santis, but I'm building with `openblas`/`netlib-scalapack`.

PR #1125, opened 2024-04-29
Use pika's `transform_mpi` and polling support

### LINKED ISSUE #536: Use pika's `transform_mpi` in DLA-Future's `transformMPI`
#497 removed the MPI executor and replaced it with a custom sender adaptor. Ideally we should be using pika's `transform_mpi` which uses MPI polling.

biddisco:
Fixes #536.

This is a squashed commit containing multiple changes

completion_modes:
pika supports different completion modes that may be used as an alternative to the previous dlaf:: transformMPI mechanism that used yield_while to wait on an MPI request.

The completion modes may be set via the environment variable PIKA_MPI_COMPLETION_MODE=<numeric value>
which by default will select the one chosen by pika/dlaf developers known to give good results across a broad range of use cases.

polling pool:
Enabling or disabling the creation of a pool dedicated to mpi traffic is controlled by the environment variable
PIKA_MPI_ENABLE_POOL=<boolean 0 or 1>
Creation and detruction of the pool is handled by pika and there is no user action required (other than setting startup flags via env vars)

By default the creation of a pool is disabled and the user  must explicitly request it. Flags in the completion mode control behaviour - the user is directed to pika documentation for an explanation. Lower bits control task placement
```
            request_inline = 1,
            completion_inline = 2,    
            high_priority = 4,  
```
and higher bits control the polling method  
```
            yield_while = 0,
            suspend_resume =8,
            new_task = 16,              
            continuation = 24,        
            mpix_continuation = 32,
```
By default the completion mode is set to 30 
`default_mode = continuation + completion_inline + high_priority,    // 24 + 2 + 4 = 30`

Please see pika source/docs for full explanation

An additional flag is also available to fine tune polling, but should not be touched in general.

Change: the transform mpi code does not need to return an MPI_SUCCESS value, the return value from mpi_transform has been removed to simplify code and an error is set using senders set_error if any mpi call fails. Should mpi_transform calls thnat return other value be required, this code can be reinstated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,7 +163,7 @@ if(DLAF_WITH_SCALAPACK)
 endif()
 
 # ----- pika
-find_package(pika 0.19.1 REQUIRED)
+find_package(pika 0.30.0 REQUIRED)
 
 # ----- BLASPP/LAPACKPP
 find_package(blaspp REQUIRED)


biddisco:
Note that this PR depends on the mpi_polling pr in pika so probably won't work with pika master

msimberg:
I've just merged https://github.com/pika-org/pika/pull/1102 so the changes are on pika `main` now.

include/dlaf/sender/transform_mpi.h
```diff
@@ -11,6 +11,10 @@
 
 #include <type_traits>
 
+#ifdef EXTRA_MPI_TYPES_DEBUGGING
+#include <pika/debugging/demangle_helper.hpp>
+#endif
+#include <pika/debugging/print.hpp>
```
msimberg, line 17: I understand that these can be useful for debugging, but I would prefer if we leave these out at this point, especially since you're planning to change the debugging/logging facilities in pika sometime soon.

include/dlaf/sender/transform_mpi.h
```diff
@@ -19,9 +23,14 @@
 #include <dlaf/communication/communicator_pipeline.h>
 #include <dlaf/sender/transform.h>
 #include <dlaf/sender/when_all_lift.h>
+//
+#include <pika/mpi.hpp>
```
msimberg, line 27: Move this up to after `pika/execution.hpp`.

include/dlaf/sender/transform_mpi.h
```diff
@@ -70,17 +74,41 @@ struct MPICallHelper {
...
-      pika::util::yield_while(is_request_completed);
+      pika::util::yield_while([req]() { return !mpid::poll_request(req); });
     }
     else {
-      auto r = std::move(f)(dlaf::common::internal::unwrap(ts)..., &req);
+      /*auto r = */ std::move(f)(dlaf::common::internal::unwrap(ts)..., &req);
+      (internal::consumeCommunicatorWrapper(ts), ...);
+      pika::util::yield_while([req]() { return !mpid::poll_request(req); });
+    }
+  }
```
msimberg, line 84: I suppose we can keep this on this branch until we've verified that going through pika's `transform_mpi` is equivalent or better, but then I'd remove this? If we can avoid it we should not use things from `pika::mpi::detail`.

include/dlaf/sender/transform_mpi.h
```diff
@@ -89,10 +117,34 @@ template <typename F, typename Sender,
...
+                 mpi::transform_mpi(dlaf::common::internal::ConsumeRvalues{MPICallHelper{std::move(f)}});
+        });
+    return ex::make_unique_any_sender(std::move(snd1));
+#else
+    PIKA_DETAIL_DP(dla_debug<5>, debug(str<>("MPI fn\n")));
+    auto snd1 =
+        std::forward<Sender>(sender) |
+        mpi::transform_mpi(dlaf::common::internal::ConsumeRvalues{MPICallHelper{std::forward<F>(f)}});
+    return ex::make_unique_any_sender(std::move(snd1));
+#endif
```
msimberg, line 146: Same comment here: let's verify that pika's `transform_mpi` is as good or better than DLAF's `yield_while` and then get rid of DLAF's version? It'd also avoid an unnecessary type-erasure.

include/dlaf/sender/transform_mpi.h
```diff
@@ -147,29 +199,6 @@ class PartialTransformMPI : private PartialTransformMPIBase<F> {
...
-  template <typename Sender>
-  friend auto operator|(Sender&& sender, PartialTransformMPIDetach pa) {
-    return pika::execution::experimental::start_detached(transformMPI(std::move(pa.f_),
-                                                                      std::forward<Sender>(sender)));
-  }
-};
-
-template <typename F>
-PartialTransformMPIDetach(F&& f) -> PartialTransformMPIDetach<std::decay_t<F>>;
-
```
msimberg, line 172: I think we can remove these in a separate PR if they're unused. I can do so.

src/init.cpp
```diff
@@ -55,6 +56,10 @@ template <>
 struct Init<Backend::MC> {
   static void initialize(const configuration& cfg) {
     memory::internal::initializeUmpireHostAllocator(cfg.umpire_host_memory_pool_initial_bytes);
+    // install mpi polling loop
+    // pika::mpi::experimental::init(false);
+    pika::mpi::experimental::init(false, true);
```
msimberg, line 61: It's not required for pika to initialize MPI, is it? In e.g. CP2K we're not going to get to initialize it anyway, so I would avoid doing it here.

src/init.cpp
```diff
@@ -94,9 +99,14 @@ template <>
 struct Init<Backend::GPU> {
   static void initialize(const configuration& cfg) {
     const int device = 0;
+    // setup polling on default pool, enable exceptions and init mpi internals
+    pika::mpi::experimental::init(false, true);
+    pika::mpi::experimental::register_polling();
+
```
msimberg, line 105: This is not necessary. The CPU backend is always initialized and takes care of registering for polling.

src/init.cpp
```diff
@@ -94,9 +99,14 @@ template <>
...
     memory::internal::initializeUmpireDeviceAllocator(cfg.umpire_device_memory_pool_initial_bytes);
     initializeGpuPool(device, cfg.num_np_gpu_streams_per_thread, cfg.num_hp_gpu_streams_per_thread);
-    pika::cuda::experimental::detail::register_polling(pika::resource::get_thread_pool("default"));
+    pika::cuda::experimental::detail::register_polling(
+        pika::resource::get_thread_pool(pika::cuda::experimental::get_pool_name()));
```
msimberg, line 109: Can we remove this change from here and deal with CUDA separately?

src/init.cpp
```diff
@@ -188,21 +198,21 @@ void updateConfiguration(const pika::program_options::variables_map& vm, configu
...
+  // if (mpi_initialized) {
+  //   int ntasks;
+  //   DLAF_MPI_CHECK_ERROR(MPI_Comm_size(MPI_COMM_WORLD, &ntasks));
+  //   if (ntasks != 1 && cfg.mpi_pool == "default" && !vm["dlaf:no-mpi-pool"].as<bool>()) {
+  //     std::cerr << "Warning! DLA-Future is not using the \"mpi\" pika thread pool for "
+  //                  "MPI communication but --dlaf:no-mpi-pool is not set. This may "
+  //                  "indicate a bug in DLA-Future or pika. Performance may be degraded."
+  //               << std::endl;
+  //   }
+  // }
```
msimberg, line 215: Assuming we can now sufficiently control the MPI pool creation through pika, you can remove these options completely.

src/init.cpp
```diff
@@ -364,26 +374,17 @@ ScopedInitializer::~ScopedInitializer() {
...
-
-  // Create a thread pool with a single core that we will use for all
-  // communication related tasks
-  rp.create_thread_pool("mpi", pika::resource::scheduling_policy::static_priority, mode);
-  rp.add_resource(rp.numa_domains()[0].cores()[0].pus()[0], "mpi");
+    pool_mode = mpi::pool_create_mode::force_no_create;
+
+  namespace mpix = pika::mpi::experimental;
+  // create a pool for mpi if necessary
+  mpix::create_pool(mpix::get_pool_name(), pool_mode);
```
msimberg, line 388: What do you think about moving this logic completely to pika and not initializing the MPI pool through the resource partitioner callback? In the long term I think it makes little sense to have DLA-Future be responsible for creating the MPI pool if the rest of the MPI functionality is in pika as well. The difficulty is: how does pika know if the MPI pool is needed or not (in the generic case, i.e. if it's not running for DLA-Future)... In the ideal case doing the polling on the default pool would be as performant as a separate pool, then we wouldn't need to bother with this. Perhaps worth coming back to this once we know a bit more about performance on Grace-Hopper?

test/unit/communication/test_comm_sender.cpp
```diff
@@ -41,10 +41,7 @@ void test_transform_mpi() {
 
   auto send = just(send_buf.data(), size, dtype, send_rank, tag, comm) | transformMPI(MPI_Isend);
   auto recv = just(recv_buf.data(), size, dtype, recv_rank, tag, comm) | transformMPI(MPI_Irecv);
-  sync_wait(when_all(std::move(send), std::move(recv)) | then([](int e1, int e2) {
-              DLAF_MPI_CHECK_ERROR(e1);
-              DLAF_MPI_CHECK_ERROR(e2);
-            }));
+  sync_wait(when_all(std::move(send), std::move(recv)) | then([]() {}));
```
msimberg, line 44: If we're really planning on leaving out the return value of the function passed to `transform_mpi`, you can remove the `then` from here (same further down). It serves no purpose.

src/init.cpp
rasolca: Not sure which options are available to initialize MPI pool in pika.
But I will keep the warning in the case non optimal options were chosen.

include/dlaf/sender/transform_mpi.h
```diff
@@ -70,17 +74,41 @@ struct MPICallHelper {
...
+    using result_type = decltype(std::move(f)(dlaf::common::internal::unwrap(ts)...));
+    if constexpr (std::is_void_v<result_type>) {
+      std::move(f)(dlaf::common::internal::unwrap(ts)...);
+      (internal::consumeCommunicatorWrapper(ts), ...);
+    }
+    else {
+      auto r = std::move(f)(dlaf::common::internal::unwrap(ts)...);
       (internal::consumeCommunicatorWrapper(ts), ...);
-      pika::util::yield_while(is_request_completed);
       return r;
```
msimberg, line 104: If we had a scope guard type of thing (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/n4853.html#scopeguard.exit) in DLA-Future we could write this roughly as:
```
      scope_exit s{[&] { (internal::consumeCommunicatorWrapper(ts), ...); });  
      return std::move(f)(dlaf::common::internal::unwrap(ts)...);
```
without the need to branch for void and non-void cases.

I'm not expecting a change here, but I just wanted to write this down here as a reminder for when we get closer to merging this PR.

include/dlaf/sender/transform_mpi.h
```diff
@@ -90,10 +119,26 @@ template <typename F, typename Sender,
           typename = std::enable_if_t<pika::execution::experimental::is_sender_v<Sender>>>
 [[nodiscard]] decltype(auto) transformMPI(F&& f, Sender&& sender) {
   namespace ex = pika::execution::experimental;
-
-  return ex::transfer(std::forward<Sender>(sender), dlaf::internal::getMPIScheduler()) |
-         ex::then(dlaf::common::internal::ConsumeRvalues{MPICallHelper{std::forward<F>(f)}}) |
-         ex::drop_operation_state();
+  namespace mpi = pika::mpi::experimental;
+  namespace mpid = pika::mpi::experimental::detail;
```
msimberg, line 123:
```suggestion
```

include/dlaf/sender/transform_mpi.h
```diff
@@ -90,10 +119,26 @@ template <typename F, typename Sender,
...
+               mpi::transform_mpi(dlaf::common::internal::ConsumeRvalues{MPICallHelper{std::move(f)}});
+      });
+  return ex::make_unique_any_sender(std::move(snd1));
+#else
+  PIKA_DETAIL_DP(dla_debug<5>, debug(str<>("MPI fn\n")));
+  auto snd1 =
+      std::forward<Sender>(sender) |
+      mpi::transform_mpi(dlaf::common::internal::ConsumeRvalues{MPICallHelper{std::forward<F>(f)}});
+  return ex::make_unique_any_sender(std::move(snd1));
+#endif
```
msimberg, line 141:
```suggestion
  return ex::make_unique_any_sender(
      std::forward<Sender>(sender) |
      mpi::transform_mpi(dlaf::common::internal::ConsumeRvalues{MPICallHelper{std::forward<F>(f)}}));
```
(modulo formatting).

include/dlaf/sender/transform_mpi.h
msimberg: https://github.com/eth-cscs/DLA-Future/pull/1202

src/init.cpp
```diff
@@ -37,7 +38,7 @@ std::ostream& operator<<(std::ostream& os, const configuration& cfg) {
      << std::endl;
   os << "  umpire_device_memory_pool_initial_bytes = " << cfg.umpire_device_memory_pool_initial_bytes
      << std::endl;
-  os << "  mpi_pool = " << cfg.mpi_pool << std::endl;
+  os << "  mpi_pool = " << pika::mpi::experimental::get_pool_name() << std::endl;
```
msimberg, line 41: This doesn't hurt, but should we actually add the pool name also to pika's runtime configuration, so that `--pika:dump-config` prints the pool name, instead of leaving it up to DLA-Future to print it? We can also have both.

src/init.cpp
```diff
@@ -106,13 +113,20 @@ template <>
...
     memory::internal::initializeUmpireDeviceAllocator(cfg.umpire_device_memory_pool_initial_bytes);
     initializeGpuPool(device, cfg.num_np_gpu_streams_per_thread, cfg.num_hp_gpu_streams_per_thread,
                       cfg.num_gpu_blas_handles, cfg.num_gpu_lapack_handles);
     pika::cuda::experimental::detail::register_polling(pika::resource::get_thread_pool("default"));
   }
 
   static void finalize() {
+    if (pika::mpi::detail::environment::is_mpi_initialized()) {
+      pika::mpi::experimental::stop_polling();
+    }
```
msimberg, line 129: We don't need to do it for both GPU and MC, it's enough to do it once for MC (it will always be initialized/finalized):
```suggestion

    memory::internal::initializeUmpireDeviceAllocator(cfg.umpire_device_memory_pool_initial_bytes);
    initializeGpuPool(device, cfg.num_np_gpu_streams_per_thread, cfg.num_hp_gpu_streams_per_thread,
                      cfg.num_gpu_blas_handles, cfg.num_gpu_lapack_handles);
    pika::cuda::experimental::detail::register_polling(pika::resource::get_thread_pool("default"));
  }

  static void finalize() {
```

src/init.cpp
```diff
@@ -59,9 +60,15 @@ template <>
 struct Init<Backend::MC> {
   static void initialize(const configuration& cfg) {
     memory::internal::initializeUmpireHostAllocator(cfg.umpire_host_memory_pool_initial_bytes);
+    if (pika::mpi::detail::environment::is_mpi_initialized()) {
+      pika::mpi::experimental::start_polling(pika::mpi::experimental::exception_mode::no_handler);
+    }
```
msimberg, line 65: Ideally, I'd have these removed (see https://github.com/pika-org/pika/pull/1267#discussion_r1804692945). If they stay I think this and the `stop_polling` should have the conditional removed:
```suggestion
      pika::mpi::experimental::start_polling(pika::mpi::experimental::exception_mode::no_handler);
```
There should be no situation where MPI isn't already initialized at this point (if anything we can add an assertion, but then it shouln't be calling `is_mpi_initialized` from pika's detail namespace, but something from DLA-Future or `MPI_Initialized` directly).

src/init.cpp
```diff
@@ -230,21 +244,21 @@ void updateConfiguration(const pika::program_options::variables_map& vm, configu
   updateConfigurationValue(vm, cfg.umpire_device_memory_pool_initial_bytes,
                            "UMPIRE_DEVICE_MEMORY_POOL_INITIAL_BYTES",
                            "umpire-device-memory-pool-initial-bytes");
-  cfg.mpi_pool = (pika::resource::pool_exists("mpi")) ? "mpi" : "default";
+  //  cfg.mpi_pool = (pika::resource::pool_exists("mpi")) ? "mpi" : "default";
```
msimberg, line 247:
```suggestion
```

src/init.cpp
```diff
@@ -230,21 +244,21 @@ void updateConfiguration(const pika::program_options::variables_map& vm, configu
...
+  // if (mpi_initialized) {
+  //   int ntasks;
+  //   DLAF_MPI_CHECK_ERROR(MPI_Comm_size(MPI_COMM_WORLD, &ntasks));
+  //   if (ntasks != 1 && cfg.mpi_pool == "default" && !vm["dlaf:no-mpi-pool"].as<bool>()) {
+  //     std::cerr << "Warning! DLA-Future is not using the \"mpi\" pika thread pool for "
+  //                  "MPI communication but --dlaf:no-mpi-pool is not set. This may "
+  //                  "indicate a bug in DLA-Future or pika. Performance may be degraded."
+  //               << std::endl;
+  //   }
+  // }
```
msimberg, line 261:
```suggestion
```

test/unit/communication/test_comm_sender.cpp
msimberg:
```suggestion
  sync_wait(when_all(std::move(send), std::move(recv)));
```

test/unit/communication/test_comm_sender.cpp
```diff
@@ -65,8 +62,7 @@ TEST(Bcast, Polling) {
   double val = (comm.rank() == root_rank) ? 4.2 : 1.2;
   std::vector<double> buf(static_cast<std::size_t>(size), val);
 
-  sync_wait(just(buf.data(), size, dtype, root_rank, comm) | transformMPI(MPI_Ibcast) |
-            then([](int e) { DLAF_MPI_CHECK_ERROR(e); }));
+  sync_wait(just(buf.data(), size, dtype, root_rank, comm) | transformMPI(MPI_Ibcast) | then([]() {}));
```
msimberg, line 65:
```suggestion
  sync_wait(just(buf.data(), size, dtype, root_rank, comm) | transformMPI(MPI_Ibcast));
```

msimberg (changes requested):

msimberg:
cscs-ci run

msimberg:
cscs-ci run

include/dlaf/init.h
```diff
@@ -38,19 +38,18 @@ struct configuration {
...
-  std::size_t umpire_host_memory_pool_next_block_bytes = 1 << 30;
+  std::size_t umpire_host_memory_pool_initial_block_bytes = 1 << 16;
+  std::size_t umpire_host_memory_pool_next_block_bytes = 1 << 16;
   std::size_t umpire_host_memory_pool_alignment_bytes = 16;
   double umpire_host_memory_pool_coalescing_free_ratio = 1.0;
   double umpire_host_memory_pool_coalescing_reallocation_ratio = 1.0;
-  std::size_t umpire_device_memory_pool_initial_block_bytes = 1 << 30;
-  std::size_t umpire_device_memory_pool_next_block_bytes = 1 << 30;
+  std::size_t umpire_device_memory_pool_initial_block_bytes = 1 << 16;
+  std::size_t umpire_device_memory_pool_next_block_bytes = 1 << 16;
```
msimberg, line 47: Revert these please. If you'd like to discuss changing the defaults open a separate PR and/or issue.

include/dlaf/schedulers.h
```diff
@@ -50,6 +51,6 @@ auto getBackendScheduler(
 
 inline auto getMPIScheduler() {
   return pika::execution::experimental::thread_pool_scheduler{
-      &pika::resource::get_thread_pool(getConfiguration().mpi_pool)};
+      &pika::resource::get_thread_pool(pika::mpi::experimental::get_pool_name())};
 }
```
msimberg, line 55: This function can now be removed entirely (along with the `pika/mpi.hpp` include) since `transform_mpi` internally handles the transfer to/from the MPI pool.

include/dlaf/sender/transform_mpi.h
```diff
@@ -41,42 +42,20 @@ void consumeCommunicatorWrapper(T&) {}
 /// callable. The wrapper then waits for the request to complete with
 /// yield_while.
 ///
-/// This could in theory be a lambda inside transformMPI.  However, clang at
-/// least until version 12 fails with an internal compiler error with a trailing
-/// decltype for SFINAE. GCC has no problems with a lambda.
```
msimberg, line 46: Did you test that this comment no longer applies?

include/dlaf/sender/transform_mpi.h
```diff
@@ -89,12 +68,11 @@ MPICallHelper(F&&) -> MPICallHelper<std::decay_t<F>>;
 template <typename F, typename Sender,
           typename = std::enable_if_t<pika::execution::experimental::is_sender_v<Sender>>>
 [[nodiscard]] decltype(auto) transformMPI(F&& f, Sender&& sender) {
-  using dlaf::internal::continues_on;
   namespace ex = pika::execution::experimental;
-
-  return continues_on(std::forward<Sender>(sender), dlaf::internal::getMPIScheduler()) |
-         ex::then(dlaf::common::internal::ConsumeRvalues{MPICallHelper{std::forward<F>(f)}}) |
-         ex::drop_operation_state();
```
msimberg, line 97: I would keep the `drop_operation_state` to make it explicit that we're releasing everything used until this point.

include/dlaf/sender/transform_mpi.h
```diff
@@ -89,12 +68,11 @@ MPICallHelper(F&&) -> MPICallHelper<std::decay_t<F>>;
...
+  namespace mpi = pika::mpi::experimental;
+  namespace dci = dlaf::common::internal;
```
msimberg, line 73: Maybe just:
```suggestion
  using pika::mpi::experimental::transform_mpi;
  using dlaf::common::internal::MPICallHelper;
```
?

test/unit/communication/test_transform_mpi.cpp
```diff
@@ -65,11 +65,7 @@ TEST_F(TransformMPITest, PromiseGuardManagement) {
 
     int message;
     whenAllLift(&message, 1, MPI_INT, 1, 0, chain.exclusive()) | transformMPI(MPI_Irecv) |
-        ex::then([&sent_guard](auto mpi_err_code) {
-          EXPECT_EQ(MPI_SUCCESS, mpi_err_code);
-          sent_guard = true;
-        }) |
-        ex::ensure_started();
+        ex::then([&sent_guard](/*auto mpi_err_code*/) { sent_guard = true; }) | ex::ensure_started();
```
msimberg, line 68:
```suggestion
        ex::then([&sent_guard]() { sent_guard = true; }) | ex::ensure_started();
```

msimberg (changes requested):

msimberg:
cscs-ci run

ci/common-ci.yml
```diff
@@ -96,3 +96,5 @@ stages:
   stage: test
   trigger:
     strategy: depend
+  variables:
+    PIKA_MPI_ENABLE_POOL: 1
```
msimberg, line 100: We have to see if there's anything we can do to not require this. At the moment this is a requirement to not have tests hang...

ci/docker/common.yaml
```diff
@@ -46,3 +46,6 @@ packages:
     # Force git as non-buildable to allow deprecated versions in environments
     # https://github.com/spack/spack/pull/30040
     buildable: false
+  pika:
+    require:
+      - '@git.release-0.30.X=0.30.0'
```
msimberg, line 51: To do:
- bump minimum required pika version in spack package to 0.30.0 (or 0.31.0 depending on what issues we still find)
- bump spack commit to include at least 0.30.0
- remove this `require`

msimberg:
cscs-ci run

include/dlaf/init.h
msimberg: These still need to be reverted.

include/dlaf/sender/transform_mpi.h
msimberg: I'm still wondering about the removal of the comment.

src/init.cpp
msimberg: Any comments on this?

src/init.cpp
```diff
@@ -70,9 +71,20 @@ struct Init<Backend::MC> {
         cfg.umpire_host_memory_pool_initial_block_bytes, cfg.umpire_host_memory_pool_next_block_bytes,
         cfg.umpire_host_memory_pool_alignment_bytes, cfg.umpire_host_memory_pool_coalescing_free_ratio,
         cfg.umpire_host_memory_pool_coalescing_reallocation_ratio);
+    // install mpi polling loop
+    int mpi_initialized;
+    DLAF_MPI_CHECK_ERROR(MPI_Initialized(&mpi_initialized));
+    if (mpi_initialized) {
+      pika::mpi::experimental::start_polling(pika::mpi::experimental::exception_mode::no_handler);
+    }
```
msimberg, line 79: Sorry for not noticing this before. Perhaps this could be moved into the if in https://github.com/eth-cscs/DLA-Future/blob/e9a661d7190c2c27841d341ac5b8bae38b60abfb/src/init.cpp#L374-L381? Since the MPI init isn't specific to neither `Backend::MC` nor `Backend::GPU`.

Same for stopping the polling?

msimberg (changes requested):
I think you will still need the `PIKA_MPI_ENABLE_POOL` flag for CI.

Could you please look through the old comments as well. There were a few that I still had questions about.

Finally, could you update the PR description to match what's actually in pika 0.30.0 in the end?

msimberg:
cscs-ci run

ci/common-ci.yml
```diff
@@ -34,7 +34,7 @@ stages:
     reports:
       dotenv: build.env
   variables:
-    SPACK_SHA: develop-2024-10-06
+    SPACK_SHA: develop-2024-11-10
```
msimberg, line 37: Interestingly spack concretizes the environment with pika 0.29.0 even though 0.30.0 is available in 2024-11-10... We do need a change in the spack package anyway to require 0.30:. I'll push a change for that.

msimberg:
cscs-ci run

ci/common-ci.yml
msimberg: https://github.com/eth-cscs/DLA-Future/pull/1125/commits/43763dc2fe86ff12a730ad7d86186128e79e70ae

msimberg:
cscs-ci run

ci/common-ci.yml
msimberg: Note, these are the test results without an MPI pool: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/pipelines/1542591315. In other words, the CPU tests are fine (on the 6-rank tests, each rank has 6 threads) but the GPU tests aren't (each rank only has 2 threads on daint-gpu). Should we set `PIKA_MPI_ENABLE_POOL: 1` only for the GPU pipelines so we use both? Should we try to find out if we can fix things to run without a pool (yes, but when)?

include/dlaf/init.h
```diff
@@ -43,14 +43,13 @@ struct configuration {
...
```
msimberg, line 47: @biddisco word of warning in case you benchmark on todi: having them this small really kills performance. As an example, the eigensolver on a 10k input matrix goes from ~3 seconds to ~9 seconds on two GPUs.

include/dlaf/sender/transform_mpi.h
biddisco: I played around, but would prefer to leave this until a later PR

src/init.cpp
biddisco: unfortunately, there are some dlaf unit tests that don't use mpi and we need some kind of conditional check here

include/dlaf/sender/transform_mpi.h
msimberg: That's ok, but then why is the comment removed?

src/init.cpp
msimberg: Ah, yeah, makes sense. In that case, please don't use the internal pika API for checking if polling should be started.

msimberg: Never mind, I see you're not using `pika::mpi::detail::environment::is_mpi_initialized()` anymore. Thank you!

msimberg:
cscs-ci run

ci/common-ci.yml
msimberg: The pool is enabled for CUDA configurations.

albestro (changes requested):

include/dlaf/sender/transform_mpi.h
```diff
@@ -85,16 +65,16 @@ struct MPICallHelper {
...
-  namespace ex = pika::execution::experimental;
-
-  return continues_on(std::forward<Sender>(sender), dlaf::internal::getMPIScheduler()) |
-         ex::then(dlaf::common::internal::ConsumeRvalues{MPICallHelper{std::forward<F>(f)}}) |
-         ex::drop_operation_state();
+  using dlaf::common::internal::ConsumeRvalues;
+  using pika::execution::experimental::drop_operation_state;
+  using pika::mpi::experimental::transform_mpi;
+  return std::forward<Sender>(sender)                                        //
+         | transform_mpi(ConsumeRvalues{MPICallHelper{std::forward<F>(f)}})  //
```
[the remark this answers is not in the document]
biddisco: Quite the opposite - I want clang-format to do the alignment of the pipe operators for me, - especially on multi line multi pipe operations, so adding an eol comment forces clang-format to keep each sender on a new line without adding two lines of `// clang-format off` and having to manually line things up myself

src/init.cpp
```diff
@@ -55,6 +57,11 @@ bool& initialized() {
   return i;
 }
 
+int& mpi_initialized() {
+  static int i = 0;
```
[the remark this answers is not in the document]
biddisco: The initialized check isn't an index. it's really just a flag, I don't think using `comm::IndexT` makes the code more readable - if mpi had a flag type, then ....

ci/docker/common.yaml
```diff
@@ -46,3 +46,6 @@ packages:
     # Force git as non-buildable to allow deprecated versions in environments
     # https://github.com/spack/spack/pull/30040
     buildable: false
+  pika:
+    require:
+      - '@git.release-0.30.X=0.30.1'
```
msimberg, line 51: Removing this since we've made one round through CI with what became pika 0.30.1.
```suggestion
```

msimberg (changes requested):
@biddisco one more place to change for the new version requirement: https://github.com/eth-cscs/DLA-Future/blob/9988b211bf67f764c4ef052610f64d3e5a45b5a6/CMakeLists.txt#L166. Could you update that to 0.30.0 as well please?

msimberg (approved):
Thanks @biddisco!

The remaining open comments I don't feel strongly enough about, and can be ignored from my perspective.

msimberg:
cscs-ci run

msimberg:
One of the test failures is the usual hermitian matrix multiplication failure (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8445202094), which is already fixed on master, but not yet on this branch. That's nothing to worry about as far as I can tell.

The other test failure is a timeout on the generalized eigensolver unit test (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8445317800). It's not the first time and likely not the last time we see something like that. We should be skeptical about the failure and keep an eye on it, but I'd be surprised if that particular timeout is now caused by the changes in this PR, so my recommendation would be to rerun and hope that it's not a regression.

include/dlaf/sender/transform_mpi.h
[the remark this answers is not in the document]
rasolca: I agree with John. Using fake comments to give hints to clang-format is better than disable it. In case of a complete reformat `// clang-format off` lines have to be handled manually.

src/init.cpp
rasolca: As DLAF has no settings for it but it comes form pika, IMO it is the best option.
On the other hand is pika dump-config invoked together with DLAF? Knowing if MPI pool was used is quite important.
Anyway, I would address this in another PR.

albestro (approved):
LGTM

I'd leave my comments with change requests open just as reminder to fix them in #1218.

PR #1133, opened 2024-05-10
Debugging MKL linking

msimberg:
Not for merging.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
Backstory in https://github.com/spack/spack/pull/43673, but short story is that that PR fixed things for fortran applications while breaking DLA-Future which now links to both `libmkl_intel_lp64.so` and `libmkl_gf_lp64.so`. The former because of our CMake/spack package.py and the latter because of dependencies using what spack suggests to use.

PR #1146, opened 2024-05-21
github actions use ubuntu 2024.04

rasolca:


.github/workflows/check_format.yml
```diff
@@ -24,9 +24,9 @@ jobs:
       - name: Install tools
         run: |
           sudo apt-get update
-          sudo apt-get install --no-install-recommends clang-format-15 python3
-          pip3 install black==23.03.0
-          pip3 install cmakelang==0.6.13
+          sudo apt-get install --no-install-recommends clang-format-15 python3 python3-pip
+          pip3 install --break-system-packages black==23.03.0
+          pip3 install --break-system-packages cmakelang==0.6.13
```
msimberg, line 29: Just out of curiosity, because `--break-system-packages` sounds scary, which system packages does this "break" (presumably, overwrite/override or something like that)?

msimberg (approved):

.github/workflows/check_format.yml
msimberg: Thanks, looks reasonable.

No need to consider for this PR, but for the pika CI environments I've started using pipx (https://github.com/pypa/pipx) for installing tools like this. It seems to work quite nicely and doesn't touch system directories (as far as I understand). This may be an alternative to consider in the future. However, take this with a big grain of salt as I don't really understand the python packaging ecosystem that well...

PR #1147, opened 2024-05-22
Improve release script

RMeli:
* Add support for HTTPS (alongside SSH) as `git` operations protocol
* Add check on `git status` (no un-committed changes allowed)

For DLA-Future-Fortran I used [`gh`](https://cli.github.com/) instead of [`hub`](https://hub.github.com/) used here. `gh` is an official GitHub product, and can simplify authentication. Do we want to switch to it? See [gh-vs-hub](https://github.com/cli/cli/blob/trunk/docs/gh-vs-hub.md) for more context.

### DIFF
--- a/scripts/roll_release.sh
+++ b/scripts/roll_release.sh
@@ -53,7 +53,7 @@ echo "Sanity checking release"
 sanity_errors=0
 
 printf "Checking that the git repository is in a clean state... "
-if git status --porcelain ; then
+if [[ `git status --porcelain` ]] ; then
     echo "ERROR"
     sanity_errors=$((sanity_errors + 1))
 else
--- a/scripts/roll_release.sh
+++ b/scripts/roll_release.sh
@@ -121,7 +121,7 @@ else
     git tag --annotate "${VERSION_FULL_TAG}" --message="${VERSION_TITLE}"
 fi
 
-remote=$(git remote -v | grep github.com:eth-cscs\/DLA-Future.git | cut -f1 | uniq)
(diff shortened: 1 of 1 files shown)


rasolca (approved):

rasolca:
For the `gh` vs `hub` question, I think it is better to have a separate PR.

scripts/roll_release.sh
```diff
@@ -52,6 +52,14 @@ echo "Sanity checking release"
 
 sanity_errors=0
 
+printf "Checking that the git repository is in a clean state... "
+if [[ `git status --porcelain` ]] ; then
```
msimberg, line 56: Not a blocking comment: I would probably find something like
```suggestion
if [[ $(git status --porcelain | wc -l) -eq 0 ]] ; then
```
a bit less magic but it may be due to my (lack of) bash experience. The `if` and `else` branches would need to be swapped of course for this... (`-eq` instead of `==` only because of other uses in the script, I would actually prefer `==`).

Also note the use of `$()` instead of backticks (just to match other uses in the script). Again, only my subjective opinion, I don't mind either way, just thought I'd point it out.

msimberg (approved):

RMeli:
> For the gh vs hub question, I think it is better to have a separate PR.

Do you want me to open the PR and discuss there, or open an issue/discussion?

rasolca:
> > For the gh vs hub question, I think it is better to have a separate PR.
> 
> Do you want me to open the PR and discuss there, or open an issue/discussion?

An issue or a discussion during the weekly meeting...

PR #1157, opened 2024-05-31
Release script improvements

msimberg:
- Check that user is authenticated before proceeding with release
- Be explicit about repo (otherwise `gh` will not make the release if there's no default repo set, as I had the first time I tried this version of the script)

I've tested the first point, but not the second.

@RMeli will you port this over to DLA-Future-Fortran or would you like a PR?

### DIFF
--- a/scripts/roll_release.sh
+++ b/scripts/roll_release.sh
@@ -15,6 +15,7 @@
 
 set -o errexit
 
+REPO="eth-cscs/DLA-Future"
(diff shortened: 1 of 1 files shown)


RMeli:
PRs always welcome. ;) But I can definitely take care of this port.

RMeli (approved):
LGTM

rasolca (approved):

PR #1162, opened 2024-06-06
fix grep and sed usage

### LINKED ISSUE #1158: Improvements for release script
- https://github.com/eth-cscs/DLA-Future/blob/master/scripts/roll_release.sh#L72-L74 doesn't work on macos.
- The same check should be added for DOCUMENTATION.md

rasolca:
Closes #1158

### DIFF
--- a/scripts/roll_release.sh
+++ b/scripts/roll_release.sh
@@ -27,6 +27,7 @@ RELEASE_DATE=$(date '+%Y-%m-%d')
 
 GREP_VERSION_FULL="$(echo ${VERSION_FULL} | sed s/\\./\\\\./g)"
 GREP_VERSION_FULL_TAG="$(echo ${VERSION_FULL_TAG} | sed s/\\./\\\\./g)"
+GREP_VERSION_TITLE="$(echo ${VERSION_TITLE} | sed s/\\./\\\\./g)"
 
 if ! which gh >/dev/null 2>&1; then
     echo "GitHub CLI not installed on this system (see https://cli.github.com). Exiting."
--- a/scripts/roll_release.sh
+++ b/scripts/roll_release.sh
@@ -16,14 +16,17 @@
 set -o errexit
 
 REPO="eth-cscs/DLA-Future"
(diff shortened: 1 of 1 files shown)


scripts/roll_release.sh
```diff
@@ -16,15 +16,18 @@
...
+VERSION_MAJOR=$(sed -n 's/project(DLAF VERSION \([0-9][0-9]*\)\.\([0-9][0-9]*\)\.\([0-9][0-9]*\))/\1/p' CMakeLists.txt)
+VERSION_MINOR=$(sed -n 's/project(DLAF VERSION \([0-9][0-9]*\)\.\([0-9][0-9]*\)\.\([0-9][0-9]*\))/\2/p' CMakeLists.txt)
+VERSION_PATCH=$(sed -n 's/project(DLAF VERSION \([0-9][0-9]*\)\.\([0-9][0-9]*\)\.\([0-9][0-9]*\))/\3/p' CMakeLists.txt)
 VERSION_FULL="${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}"
 VERSION_FULL_TAG="v${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}"
 VERSION_TITLE="DLA-Future ${VERSION_FULL}"
 CURRENT_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
 RELEASE_DATE=$(date '+%Y-%m-%d')
 
+GREP_VERSION_FULL="$(echo ${VERSION_FULL} | sed s/\\./\\\\./g)"
```
rasolca, line 28: Need to escape dots otherwise grep matches every char.

rasolca:
@RMeli, do you want the same changes on DLAFF?

RMeli:
Since getting an instance on `castor` I don't use macOS anymore for actual work, but I think it would be good to keep the two scripts more or less in sync. Do you want to open a PR? Otherwise, I'll port it over when this is merged.

rasolca:
> Since getting an instance on `castor` I don't use macOS anymore for actual work, but I think it would be good to keep the two scripts more or less in sync. Do you want to open a PR? Otherwise, I'll port it over when this is merged.

Will open a PR when merged.
Anyway non escaping dots when grepping is a bug everywhere...

msimberg (approved):

PR #1164, opened 2024-06-10
Instrument distributed generalized eigensolver and Cholesky decomposition with HDF5

RMeli:
* Follow-up to #1016 
* Add HDF5 input/output to `miniapp_gen_eigensolver`


RMeli:
cscs-ci run

RMeli:
cscs-ci run

rasolca (commented):
Looks good, need rebase due to generalized eigensolver changes

RMeli:
cscs-ci run

include/dlaf/eigensolver/gen_eigensolver/impl.h
```diff
@@ -40,15 +46,39 @@ template <Backend B, Device D, class T>
 void GenEigensolver<B, D, T>::call(comm::CommunicatorGrid& grid, blas::Uplo uplo, Matrix<T, D>& mat_a,
                                    Matrix<T, D>& mat_b, Matrix<BaseType<T>, D>& eigenvalues,
                                    Matrix<T, D>& eigenvectors, const Factorization factorization) {
+#ifdef DLAF_WITH_HDF5
+  static std::atomic<size_t> num_gen_eigensolver_calls = 0;
+  std::stringstream fname;
+  fname << "generalized-eigensolver-" << matrix::internal::TypeToString_v<T> << "-"
+        << std::to_string(num_gen_eigensolver_calls) << ".h5";
+  std::optional<matrix::internal::FileHDF5> file;
```
rasolca, line 54: I think it need to provide different names depending on `factorization`.

include/dlaf/eigensolver/gen_eigensolver/impl.h
```diff
@@ -40,15 +46,39 @@ template <Backend B, Device D, class T>
...
+
+  if (getTuneParameters().debug_dump_generalized_eigensolver_data) {
+    file = matrix::internal::FileHDF5(grid.fullCommunicator(), fname.str());
+    file->write(mat_a, "/input-a");
+    file->write(mat_b, "/input-b");
```
rasolca, line 59: And maybe we can name it `/input_b` or `/input-b-factorized` depending on `factorization`. What do you think?

include/dlaf/eigensolver/gen_eigensolver/impl.h
RMeli: I don't think it's needed because every call will update the same `static std::atomic`, but considering also your suggestion below it might be good to name them differently. I'd suggest to keep the same counter though, since it is easier to track synchronous calls (C/Fortran API).

RMeli:
cscs-ci run

miniapp/miniapp_gen_eigensolver.cpp
```diff
@@ -94,7 +122,18 @@ struct GenEigensolverMiniapp {
     GlobalElementSize matrix_size(opts.m, opts.m);
     TileElementSize block_size(opts.mb, opts.mb);
 
-    ConstHostMatrixType matrix_a_ref = [matrix_size, block_size, &comm_grid]() {
+    // Default capture & is needed to suppress warning of unused opts when DLAF_WITH_HDF5 is not defined.
+    // [[maybe_unused]] is not supported in lambda captures
```
msimberg, line 126: No big preference either way, but I slightly prefer an explicit version with an `#else` branch using `silenceUnusedWarningFor`. This would avoid the use ofa catch-all `&`. Not a blocking comment.

msimberg (approved):

miniapp/miniapp_gen_eigensolver.cpp
RMeli: I forgot about this! Thanks for pointing it out. Let's change it.

rasolca (approved):

RMeli:
cscs-ci run

rasolca (approved):

PR #1168, opened 2024-06-17
Rename pxhegvx to pxhegvd

RMeli:
ScaLAPACK only implements `pxhegvx`, but the correct name for our implementation is `pxhegvd` (see LAPACK naming conventions).

This is a breaking change that will require changes to the Fortran API and its use by downstream projects (currently CP2K).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## DLA-Future 0.Y.Z
+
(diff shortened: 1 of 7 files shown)


RMeli:
cscs-ci run

RMeli:
cscs-ci run

rasolca (approved):

PR #1177, opened 2024-08-16
CA-Red2Band: Communication-Avoiding Reduction To Band

albestro:
First basic implementation without all required bells and whistles.

- It is not fully optimised
  - it does more communications than needed
  - it performs superfluous computation, e.g. it applies HH even when they are identities by construction
- It does not manage edge-cases (e.g. incomplete tiles)
- It does not work sub-band

Changelog:
- Some old reduction to band functions have been moved to a common place
- Some old reduction to band functions (and related, e.g. t-factor) have been fixed to deal with distributed matrices where we want to work just on the local part


albestro:
cscs-ci run

albestro:
cscs-ci run

PR #1179, opened 2024-09-04
Add helper for printing Umpire pool statistics

msimberg:
This is the non-production ready debugging version that I used to print Umpire pool statistics for debugging. It needs cleanup to be ready for merging, but more importantly if we want something like this in master we should:
- decide how to trigger the printing and when (when finalizing DLA-Future, after every algorithm, with or without synchronization, environment variable)
- decide which statistics to show and in what form

### DIFF
--- a/include/dlaf/memory/memory_chunk.h
+++ b/include/dlaf/memory/memory_chunk.h
@@ -26,6 +26,8 @@ namespace dlaf {
 namespace memory {
 
 namespace internal {
+void print_cuda_stats(std::string_view label);
+
 umpire::Allocator& getUmpireHostAllocator();
 void initializeUmpireHostAllocator(std::size_t initial_bytes);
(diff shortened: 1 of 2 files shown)


albestro:
Thanks! We might think of taking chance of this also for tracking memory usage in our benchmarks (which should just add to requirements "easy to be parsed").

msimberg:
> Thanks! We might think of taking chance of this also for tracking memory usage in our benchmarks (which should just add to requirements "easy to be parsed").

I think this is a very good idea. I could see at least three different directions:
- trigger printing of statistics at the end of each algorithm
- explicitly call a generic function to print statistics at the end of all miniapps
- above, but implement a specialized format only for miniapps that's easier to parse

msimberg:
Closing just to tidy up my open PRs. This can always be revived later. I'm not deleting the branch.

PR #1180, opened 2024-09-05
Update to use `sockets` instead of `numa_domains` with pika's resource partitioner

msimberg:
Fixes a breaking change introduced in pika 0.28.0.

I've taken the opportunity to fix a TODO that required a newer version of pika: check whether pika has been initialized before accessing the resource partitioner.

PR #1181, opened 2024-09-05
Fix blaspp/lapackpp deprecations of enum conversions (Uplo, Op, ...) from and to string

### LINKED ISSUE #1165: Replace lapackpp's `matrixtype2str` with `to_c_string` or `to_string`
The former is deprecated and will be removed in 2025-05.

albestro:
Close #1165

From `blaspp@2024.05.31` [release notes](https://github.com/icl-utk-edu/blaspp/releases/tag/v2024.05.31)

> - Updated enum parameters to have to_string, from_string; deprecate `<enum>`2str, str2`<enum>`

### DIFF
--- a/include/dlaf/lapack/enum_output.h
+++ b/include/dlaf/lapack/enum_output.h
@@ -37,7 +37,7 @@ inline std::ostream& operator<<(std::ostream& out, const Norm& norm) {
 }
 
 inline std::ostream& operator<<(std::ostream& out, const lapack::MatrixType& mtype) {
-  out << lapack::matrixtype2str(mtype);
+  out << to_c_string(mtype);
   return out;
 }
 
--- a/include/dlaf/blas/enum_parse.h
+++ b/include/dlaf/blas/enum_parse.h
@@ -1,0 +1,25 @@
+//
+// Distributed Linear Algebra with Future (DLAF)
+//
+// Copyright (c) 2018-2024, ETH Zurich
+// All rights reserved.
+//
+// Please, refer to the LICENSE file in the root directory.
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+#pragma once
+
+/// @file
+
+#include <blas/util.hh>
+
+namespace dlaf::internal {
+
+inline blas::Uplo char2uplo(const char uplo_c) {
+  blas::Uplo uplo;
+  blas::from_string(std::to_string(uplo_c), &uplo);
+  return uplo;
+}
+
+}
--- a/include/dlaf/blas/enum_parse.h
+++ b/include/dlaf/blas/enum_parse.h
@@ -12,13 +12,15 @@
 
 /// @file
 
+#include <string>
+
 #include <blas/util.hh>
 
 namespace dlaf::internal {
 
 inline blas::Uplo char2uplo(const char uplo_c) {
   blas::Uplo uplo;
-  blas::from_string(std::to_string(uplo_c), &uplo);
+  blas::from_string(std::string(&uplo_c, 1), &uplo);
   return uplo;
 }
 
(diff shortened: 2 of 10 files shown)


include/dlaf/lapack/enum_output.h
```diff
@@ -37,7 +37,7 @@ inline std::ostream& operator<<(std::ostream& out, const Norm& norm) {
 }
 
 inline std::ostream& operator<<(std::ostream& out, const lapack::MatrixType& mtype) {
-  out << lapack::matrixtype2str(mtype);
+  out << to_c_string(mtype);
```
msimberg, line 40: In which version was this introduced? I.e. do we need a lower bound on blaspp/lapackpp versions?

msimberg (commented):
Thank you for doing this! Just a question about versions...

include/dlaf/lapack/enum_output.h
albestro: Good point!

This seems to be the [commit](https://github.com/icl-utk-edu/blaspp/commit/62ed424fbe0d675070160dbeeb6846dfe0fd4505) where they did the change, which is tagged with `2024.05.31`.

Looking at [2024.05.31 release notes](https://github.com/icl-utk-edu/blaspp/releases/tag/v2024.05.31), the change is actually reported

> Updated enum parameters to have to_string, from_string; deprecate <enum>2str, str2<enum>

I'm going to add the conflict to the spack package 😉

msimberg:
Just for reference, the spack commit used in https://github.com/eth-cscs/DLA-Future/pull/1180 now pulls in a newer version of blaspp/lapackpp and the build fails because of the warnings, so we should try merging this one quite quickly as well.

include/dlaf/lapack/enum_output.h
albestro: @msimberg @rasolca 

do you think it is reasonable to use this?

```python
depends_on("blaspp@2024.05.31:", when="@0.6.1:")
```

We are going to probably release directly `0.7.0`, but in case we end up for some reason releasing a minor version, this would be more correct. Isn't it? We can also use `master`, but it looks more error-prone because we would have to remember to change that accordingly with the next release.

Any preference or suggestion?

rasolca: Isn't something like `depends_on("blaspp@:2024.05.31", when="@:0.6.0")` also needed?

albestro: Yep, you're right! We can anticipate the actual deprecation too.

In the deprecation messages [here](https://github.com/icl-utk-edu/blaspp/blob/fb004632feddfbe4df6e06de8ec29406153b6e32/include/blas/util.hh#L175) and [here](https://github.com/icl-utk-edu/blaspp/blob/fb004632feddfbe4df6e06de8ec29406153b6e32/include/blas/util.hh#L275) they report `2025.05`, so probably we should add something like

```python
conflicts("^blasspp@2025.05:", when="@:0.6.0")
```

which differently from your suggestion would work even if they will make another 2024 release (and we should ship this PR with our next first release, either minor or major, but this sounds more than reasonable).

msimberg: > @msimberg @rasolca
> 
> do you think it is reasonable to use this?
> 
> ```python
> depends_on("blaspp@2024.05.31:", when="@0.6.1:")
> ```

Yep. We can always change it to `when="@0.7:"` when we decide what the next version will be.

> ```python
> conflicts("^blasspp@2025.05:", when="@:0.6.0")
> ```

:+1:

albestro: see https://github.com/eth-cscs/DLA-Future/pull/1181/commits/7072da2199a1c351b572d79f4e3dadfd761b0480

albestro:
cscs-ci run

msimberg (approved):

rasolca (approved):

albestro:
cscs-ci run

msimberg:
I restarted the one failed test job: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/7764034639#L282. This looks like a filesystem issue perhaps?

PR #1143, opened 2024-05-17
Workaround for bad GPUDirect performance with unaligned GPU buffers

msimberg:
Opening this as a draft for reference. I think we should wait for responses from both the Umpire developers (https://github.com/LLNL/Umpire/issues/881) and HPE before deciding if and what workaround to apply. This typically, but not always, gives reasonable performance after only one warmup iteration, and the warmup iteration isn't ridiculously slow compared to the best case. However, this always allocates at least 2MiB per allocation from Umpire and can end up wasting quite a lot of memory for small tiles. As an example the `gen_to_std` miniapp can look like this on current master:
```
[0]
[0] 17.3253s 495.804GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[1]
[1] 11.5633s 742.859GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[2]
[2] 2.86979s 2993.22GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[3]
[3] 0.0939851s 91396.8GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[4]
[4] 2.95547s 2906.45GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[5]
[5] 0.0937317s 91643.8GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[6]
[6] 0.0919855s 93383.6GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[7]
[7] 0.0930948s 92270.8GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[8]
[8] 0.0933742s 91994.7GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[9]
[9] 0.0922234s 93142.6GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
```
and most of the time looks like this on this PR:
```
[0]
[0] 0.318221s 26993.7GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[1]
[1] 0.0949778s 90441.5GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[2]
[2] 0.0906252s 94785.2GFlop/s dL (20480, 20480) (1024, 1024) (2, 2) 72 GPU
[3]
[...]

PR #1183, opened 2024-09-06
Use custom coalescing heuristic and introduce a number of new configuration options for memory pools

msimberg:
This is intended to replace https://github.com/eth-cscs/DLA-Future/pull/1143 as a more permanent solution.

This introduces:
- configuration options for controlling the "next" allocation size (i.e. all blocks after the first one) and alignment size of pools
- a custom coalescing heuristic that works around https://github.com/LLNL/Umpire/issues/906
- configuration options that go along with the custom heuristic

I'm 100% sure I haven't chosen the best names for the new options, but I'd like to discuss them with you first before making any further changes. Additionally, I'm 100% aware that the coalescing options add complexity. My intention with them is primarily to have as debugging options, not necessarily something most users would change. The default behaviour is very "vanilla", but in case we see situations where too much memory is used or similar we can use these options to try to tune how umpire frees/coalesces blocks in the pool.

As a semi-related change I've also namespaced the Umpire pools with a `DLAF_` prefix since we used quite generic names. I suppose we'd notice if the names clash, but I'd rather be on the safe side and avoid clashes (e.g. if the non-thread safe pool is accessed without synchronization I expect bad things to happen).


msimberg:
cscs-ci run

msimberg:
Context for current names: https://github.com/LLNL/Umpire/blob/bf2df8d46f6e8918b7402a45b0f95127c3121478/src/umpire/strategy/QuickPool.hpp#L46-L62.

src/init.cpp
```diff
@@ -129,6 +143,13 @@ struct parseFromString<SizeType> {
   }
 };
 
+template <>
+struct parseFromString<double> {
+  static std::optional<SizeType> call(const std::string& var) {
```
RMeli, line 148:
```suggestion
  static std::optional<double> call(const std::string& var) {
```

include/dlaf/init.h
```diff
@@ -38,7 +38,15 @@ struct configuration {
   std::size_t num_np_gpu_streams_per_thread = 3;
   std::size_t num_hp_gpu_streams_per_thread = 3;
   std::size_t umpire_host_memory_pool_initial_bytes = 1 << 30;
+  std::size_t umpire_host_memory_pool_next_bytes = 1 << 30;
```
msimberg, line 41: If I were to rename these I'd probably do:
```suggestion
  std::size_t umpire_host_memory_pool_initial_block_bytes = 1 << 30;
  std::size_t umpire_host_memory_pool_next_block_bytes = 1 << 30;
```
Is that an improvement? Other ideas?

msimberg:
cscs-ci run

include/dlaf/init.h
msimberg: I've updated to the suggestion above.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

rasolca (approved):

PR #1194, opened 2024-09-23
Partial back transformation of eigenvectors

RMeli:
Select the eigenvectors corresponding to the N lowest eigenvalues (from index `0` to `last-eval-index`) on which the back transformation is applied. The tridiagonal solver is still solved in full, therefore only the cost of the back transformation is gained.

TODO:
- [x] Partial triangular multiplication (final step of generalized eigensolver)
- [x] C API (and C API tests)
  - [x] Standard
  - [x] Generalised


RMeli:
cscs-ci run

RMeli:
cscs-ci run

msimberg (changes requested):
Very nice overall. I still have to do some thinking about the `MatrixRef/Matrix` interaction with sub-pipelines...

test/unit/eigensolver/test_gen_eigensolver.cpp
```diff
@@ -8,7 +8,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 
-#include <functional>
+#include <cmath>
```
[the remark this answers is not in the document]
RMeli: You should thank `clangd`, which also changed this one automatically, I think. I don't recall doing this myself, so I'll double check. I noticed in other occasions that the `#include` fixes when auto-formatting are a bit iffy...

test/include/dlaf_test/matrix/util_matrix_local.h
```diff
@@ -100,11 +101,21 @@ MatrixLocal<T> allGather(blas::Uplo uplo, Matrix<const T, Device::CPU>& source)
...
+MatrixLocal<T> allGather(blas::Uplo uplo, Matrix<const T, Device::CPU>& source) {
+  return allGather_helper<T>(uplo, source);
+}
+
+template <class T>
+MatrixLocal<T> allGather(blas::Uplo uplo,
+                         dlaf::matrix::internal::MatrixRef<const T, Device::CPU>& source) {
+  return allGather_helper<T>(uplo, source);
+}
+
```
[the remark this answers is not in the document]
RMeli: The intention was to constrain, but I agree here we can just template the matrix type in this case.

miniapp/miniapp_bt_band_to_tridiag.cpp
```diff
@@ -208,11 +218,12 @@ int main(int argc, char** argv) {
...
-    ("n",   value<SizeType>()   ->default_value(4096), "Matrix E columns")
-    ("mb",  value<SizeType>()   ->default_value( 256), "Matrix E block rows")
-    ("nb",  value<SizeType>()   ->default_value( 512), "Matrix E block columns")
-    ("b",   value<SizeType>()   ->default_value(  64), "Band size")
+    ("m",               value<SizeType>()   ->default_value(2048), "Matrix E rows")
+    ("n",               value<SizeType>()   ->default_value(4096), "Matrix E columns")
+    ("mb",              value<SizeType>()   ->default_value( 256), "Matrix E block rows")
+    ("nb",              value<SizeType>()   ->default_value( 512), "Matrix E block columns")
+    ("b",               value<SizeType>()   ->default_value(  64), "Band size")
+    ("last-eval-index", value<SizeType>()                        , "Index of last eigenvalue of interest/eigenvector to transform")
```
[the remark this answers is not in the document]
RMeli: Yes, If you read it as `last-evaluation-index`, others will too. I'll be more explicit.

I like the idea of `first + N`, but I was following the ScaLAPACK convention which uses the actual indices too.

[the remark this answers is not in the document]
msimberg: > I like the idea of `first + N`, but I was following the ScaLAPACK convention which uses the actual indices too.

Thanks, makes sense. That sounds like a good motivation to stick with first/last.

As reference for others, the related question about what to pass to the API is here: https://github.com/eth-cscs/DLA-Future/pull/1194#discussion_r1781261041.

miniapp/miniapp_bt_band_to_tridiag.cpp
```diff
@@ -153,27 +167,23 @@ struct BacktransformBandToTridiagMiniapp {
...
-                    << "backend, " << backend << ", " << opts.info << std::endl;
+          std::cout << "CSVData-2, " << "run, " << run_index << ", " << "time, " << elapsed_time << ", "
+                    << "GFlops, " << gigaflops << ", " << "type, "
+                    << dlaf::internal::FormatShort{opts.type}.value << ", " << "matrixsize, "
+                    << mat_e_host.size().rows() << ", " << "blocksize, " << mat_e_host.blockSize().rows()
+                    << ", " << "band_size, " << opts.b << ", " << "comm_rows, "
+                    << comm_grid.size().rows() << ", " << "comm_cols, " << comm_grid.size().cols()
+                    << ", " << "threads, " << pika::get_os_thread_count() << ", " << "backend, "
+                    << backend << ", " << opts.info << ", " << "first eigenvalue index, " << 0l << ", "
+                    << "last eigenvalue index, " << opts.last_eval_idx << std::endl;
```
rasolca, line 186: I think you are using the bugged clang-format available in ubuntu 24.04

miniapp/miniapp_eigensolver.cpp
```diff
@@ -279,21 +290,24 @@ using dlaf::matrix::Tile;
...
+                      const SizeType last_eval_idx) {
   const Index2D rank_result{0, 0};
 
-  // 1. Compute the norm of the original matrix in A (largest eigenvalue)
-  const GlobalElementIndex last_ev(evalues.size().rows() - 1, 0);
-  const GlobalTileIndex last_ev_tile = evalues.distribution().globalTileIndex(last_ev);
-  const TileElementIndex last_ev_el_tile = evalues.distribution().tileElementIndex(last_ev);
-  const auto norm_A = std::max(std::norm(sync_wait(evalues.read(GlobalTileIndex{0, 0})).get()({0, 0})),
-                               std::norm(sync_wait(evalues.read(last_ev_tile)).get()(last_ev_el_tile)));
+  // 1. Get largest eigenvalue (amongst computed eigenvalues)
```
rasolca, line 297: outdated comment

include/dlaf/auxiliary/norm.h
```diff
@@ -27,9 +27,10 @@ namespace dlaf::auxiliary {
 ///
 /// @note @p uplo == blas::uplo::Upper not yet implemented
 ///
-/// @pre @p A is distributed according to @p grid
-/// @pre @p A has blocksize (NB x NB)
-/// @pre @p A has tilesize (NB x NB)
+/// @pre @p A has size (M x N) and is distributed according to @p grid
+/// @pre @p A has blocksize (MB x NB)
+/// @pre @p A has tilesize (MB x NB)
+/// @pre if uplo == blas::uplo::Upper or uplo == blas::uplo::Lower Ashould be square, i.e. M == N and MB == NB.
```
RMeli, line 33:
```suggestion
/// @pre if uplo == blas::uplo::Upper or uplo == blas::uplo::Lower A should be square, i.e. M == N and MB == NB.
```

include/dlaf/auxiliary/norm/mc.h
```diff
@@ -112,13 +112,9 @@ dlaf::BaseType<T> Norm<Backend::MC, Device::CPU, T>::max_G(comm::CommunicatorGri
 
   const auto& distribution = matrix.distribution();
 
-  DLAF_ASSERT(square_size(matrix), matrix);
-  DLAF_ASSERT(square_blocksize(matrix), matrix);
-
```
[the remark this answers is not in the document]
RMeli: Thanks @rasolca.

include/dlaf/eigensolver/eigensolver.h
```diff
@@ -52,9 +52,15 @@ namespace dlaf {
...
+///
+/// @param[in] first_eigenvalue_index is the index of the first eigenvalue to compute
+/// @pre @p first_eigenvalue_index == 0
+/// @param[in] last_eigenvalue_index is the index of the last eigenvalue to compute
+/// @pre @p last_eigenvalue_index < N
 template <Backend B, Device D, class T>
 void hermitian_eigensolver(blas::Uplo uplo, Matrix<T, D>& mat, Matrix<BaseType<T>, D>& eigenvalues,
-                           Matrix<T, D>& eigenvectors) {
+                           Matrix<T, D>& eigenvectors, SizeType first_eigenvalue_index,
+                           SizeType last_eigenvalue_index) {
```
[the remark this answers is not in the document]
RMeli: > However, I'm more favorable to [first, last), i.e last excluded.

ScaLAPACK uses `[first, last]`, do we want to differ? However it is Fortran indexing so it is less ugly (i.e. the last index can be `N`, not `N-1`).

include/dlaf/eigensolver/eigensolver.h
```diff
@@ -70,8 +76,47 @@ void hermitian_eigensolver(blas::Uplo uplo, Matrix<T, D>& mat, Matrix<BaseType<T
   DLAF_ASSERT(single_tile_per_block(mat), mat);
   DLAF_ASSERT(single_tile_per_block(eigenvalues), eigenvalues);
   DLAF_ASSERT(single_tile_per_block(eigenvectors), eigenvectors);
+  DLAF_ASSERT(first_eigenvalue_index == 0, first_eigenvalue_index);  // TODO: remove this restriction
```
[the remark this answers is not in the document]
RMeli:
```suggestion
  DLAF_ASSERT(first_eigenvalue_index == 0, first_eigenvalue_index);
```

miniapp/miniapp_bt_band_to_tridiag.cpp
RMeli: I updated to `18.1.8` as in CI.

include/dlaf/util_matrix.h
```diff
@@ -441,6 +439,16 @@ void set_random_hermitian_with_offset(Matrix<T, Device::CPU>& matrix, const Size
   }
 }
 
+template <class T, Device D>
+dlaf::matrix::internal::SubMatrixSpec sub_matrix_spec_slice_cols(const Matrix<T, D>& matrix,
+                                                                 SizeType first_col_index,
+                                                                 SizeType last_col_index) {
+  DLAF_ASSERT(first_col_index >= 0, first_col_index);
+  DLAF_ASSERT(last_col_index < matrix.size().cols(), last_col_index, matrix.size().cols());
```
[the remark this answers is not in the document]
RMeli: This condition does not currently hold for zero-sized matrices (that we have in the tests), since `last_col_index` is inclusive and therefore the sub-matrix dimension is computed as `last_col_index - first_col_index + 1`. This is a good argument to make `last_col_index` exclusive as @rasolca suggested.

include/dlaf/eigensolver/eigensolver.h
```diff
@@ -11,8 +11,6 @@
 
 /// @file
 
-#include <utility>
-
```
msimberg, line 15: inshpect rightfully complains about this header being removed. It's required for `std::move` (and possibly more).

RMeli: Yep, thanks. I need to prevent the tooling I have to change the source, it’s proving quite un-reliable.

include/dlaf/eigensolver/eigensolver.h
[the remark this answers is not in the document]
msimberg: I don't have a strong opinion on this, but from the options so far it looks to me like [first, last) seems like the sensible option because:
- it's still two indices like scalapack, but
- it's still the more natural formulation for 0-based indices

I think the difference to scalapack's indexing isn't too bad, but I'm not the typical user... Do we create massive confusion if the C API uses [first, last) with 0-based indices and the Fortran API [first, last] with 1-based indices?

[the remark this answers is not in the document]
RMeli: We discussed this yesterday at the meeting, and the consensus was essentially what you suggest here. Using `[first, last)` is much cleaner in C/C++, and the Fortran API can still expose `[first, last]`.

> Do we create massive confusion if the C API uses [first, last) with 0-based indices and the Fortran API [first, last] with 1-based indices?

I don't think so. I think differentiating between C/C++ and Fortran will make it less confusing for the different users.

https://github.com/eth-cscs/DLA-Future/pull/1194/commits/f566f8980cbbbf326222daa7688eab11d56bb826 takes care of most of the changes already; I also changed `first`/`last` with `begin`/`end`.

include/dlaf/util_matrix.h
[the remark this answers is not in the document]
RMeli: Should be fixed by the new conditions, with `[begin, end)`.

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

include/dlaf/matrix/matrix_ref.h
```diff
@@ -55,12 +55,30 @@ class MatrixRef<const T, D> : public internal::MatrixBase {
...
+  ///
+  /// All accesses to the sub-pipelined matrix are sequenced after previous accesses and before later
+  /// accesses to the original matrix, independently of when tiles are accessed in the sub-pipelined
+  /// matrix.
+  ///
+  /// @pre blockSize() is divisible by @p tiles_per_block
+  /// @pre blockSize() == tile_size()
+  Matrix<const T, D> retiledSubPipelineConst(const LocalTileSize& tiles_per_block) {
+    return mat_const_.retiledSubPipelineConst(tiles_per_block);
+  }
```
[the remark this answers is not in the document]
RMeli: This was discussed offline with @msimberg. The return type will remain `Matrix`, and the `retiledSubPipeline`/`setUpRetiledSubPipeline` functions will be updated to work with the sub-distribution.

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

include/dlaf/matrix/matrix.h
```diff
@@ -34,6 +34,12 @@ namespace matrix {
 
 namespace internal {
 
+template <class T, Device D>
+class MatrixRef;
+
+// template <class T, Device D>
+// class MatrixRef;
```
msimberg, line 41:
```suggestion
```

include/dlaf/matrix/matrix_ref.h
```diff
@@ -20,7 +20,15 @@
...
-namespace dlaf::matrix::internal {
+namespace dlaf::matrix {
+
+// Pre-declaration of MatrixRef
+// MatrixRef is a friend of Matrix
+// template <class T, Device D>
+// class Matrix;
+
+namespace internal {
+
```
msimberg, line 31: Can be reverted?

include/dlaf/matrix/matrix_ref.h
```diff
@@ -34,7 +42,7 @@ template <class T, Device D>
 class MatrixRef;
 
 template <class T, Device D>
-class MatrixRef<const T, D> : public internal::MatrixBase {
+class MatrixRef<const T, D> : public MatrixBase {
```
msimberg, line 45: Still necessary?

include/dlaf/matrix/matrix_ref.h
```diff
@@ -193,4 +277,5 @@ DLAF_MATRIX_REF_ETI(extern, double, Device::GPU)
 DLAF_MATRIX_REF_ETI(extern, std::complex<float>, Device::GPU)
 DLAF_MATRIX_REF_ETI(extern, std::complex<double>, Device::GPU)
 #endif
-}
+}  // namespace internal
+}  // namespace dlaf::matrix
```
msimberg, line 281: Revert?

include/dlaf/multiplication/hermitian.h
```diff
@@ -31,10 +73,10 @@ namespace dlaf {
...
 /// the lower triangular part which represent the Hermitian matrix (depending on the value of uplo)
 /// are accessed in read-only mode (the elements are not modified),
-/// @pre @p mat_b is not distributed
-/// @pre @p mat_b has size (N x M)
-/// @pre @p mat_b has blocksize (NB x NB)
-/// @pre @p mat_b has tilesize (NB x NB)
+/// @pre @p mat_a is not distributed
+/// @pre @p mat_a has size (N x M)
+/// @pre @p mat_a has blocksize (NB x NB)
+/// @pre @p mat_a has tilesize (NB x NB)
```
msimberg, line 79: Nice catch!

include/dlaf/util_math.h
```diff
@@ -106,6 +107,23 @@ constexpr ArithmeticT generic_integer_op(const TA a, const TB b, BinaryOp op) {
...
 
+/// Convert a percentage to an index.
+///
+/// @tparam ArithmeticT must be a floating point type
+/// @param size the size of the array
+/// @pre @p size >= 0,
+/// @param percent index corresponding to the first @percent percent of elements
+/// @pre @p percent >= 0 and @p percent <= 100,
+/// @return the end index corresponding to the first @percent percent of elements
+template <typename ArithmeticT>
```
msimberg, line 118: Purely aesthetic, but I think we mostly use `template <class ArithmeticT>`.

miniapp/miniapp_bt_reduction_to_band.cpp
```diff
@@ -166,34 +190,30 @@ struct BacktransformBandToTridiagMiniapp {
...
-                    << "backend, " << backend << ", " << opts.info << std::endl;
+          std::cout << "CSVData-2, " << "run, " << run_index << ", " << "time, " << elapsed_time << ", "
+                    << "GFlops, " << gigaflops << ", " << "type, "
+                    << dlaf::internal::FormatShort{opts.type}.value << ", " << "matrixsize, "
+                    << mat_e_host.size().rows() << ", " << "blocksize, " << mat_e_host.blockSize().rows()
+                    << ", " << "band_size, " << opts.b << ", " << "comm_rows, "
+                    << comm_grid.size().rows() << ", " << "comm_cols, " << comm_grid.size().cols()
+                    << ", " << "threads, " << pika::get_os_thread_count() << ", " << "backend, "
+                    << backend << ", " << opts.info << ", " << "eigenvalue index begin, " << 0l << ", "
+                    << "eigenvalue index end, " << opts.eval_idx_end << std::endl;
```
msimberg, line 216: Just wondering, did we not check the formatting of this before, or did it get reformatted for some other reason? I think we might want to do a `// clang-format off` here with the old formatting?

scripts/gen_dlaf_strong-gpu.py
```diff
@@ -96,20 +136,21 @@ def createAndSubmitRun(run_dir, nodes_arr, dtype, **kwargs):
...
-        **full_kwargs,
+        **fullbt_kwargs,
     )
     run.add(
         mp.bt_red2band,
         params={"rpn": rpn, "m_sz": m_szs, "mb_sz": mb_szs, "band": 128, "n_sz": None},
-        **full_kwargs,
+        **fullbt_kwargs,
     )
+    # TODO: Partial trsm for Cholesky
```
msimberg, line 146: To do?

scripts/postprocess.py
```diff
@@ -240,14 +239,20 @@ def _parse_line_based(fout, bench_name, nodes):
...
         alg_name = bench_name[0 : bench_name.find("_dlaf")]
 
-        if alg_name in ["chol", "hegst", "trsm", "trmm"]:
+        print(alg_name)
+
+        if alg_name in ["chol", "hegst", "trmm"]:
             pstr_res = "[{run_index:d}] {time:g}s {perf:g}GFlop/s{matrix_type:optional_text} ({matrix_rows:d}, {matrix_cols:d}) ({block_rows:d}, {block_cols:d}) ({grid_rows:d}, {grid_cols:d}) {:d}{backend:optional_text}"
-        if alg_name in ["red2band", "band2trid", "bt_band2trid", "bt_red2band"]:
+        if alg_name in ["trsm"]:
+            pstr_res = "[{run_index:d}] {time:g}s {perf:g}GFlop/s{matrix_type:optional_text} ({matrix_rows:d}, {matrix_cols:d}) ({first_eval_idx:d}, {last_eval_idx:d}) ({block_rows:d}, {block_cols:d}) ({grid_rows:d}, {grid_cols:d}) {:d}{backend:optional_text}"
```
msimberg, line 247: This is not a blocking comment, but I want to point out that the new entries here will prevent parsing of old benchmark results. Ok to simply require using an old commit of the parsing scripts for that?

scripts/postprocess.py
```diff
@@ -327,6 +332,12 @@ def _parse_line_based(fout, bench_name, nodes):
                 run_index += 1
 
             data.append(dict(rd))
+        else:
+            print(fout, bench_name, nodes, pstr_res)
+
+    print(len(data))
+    # if(len(data) == 0):
+    #    raise RuntimeError("No data could be parsed!")
```
msimberg, line 340: ?

scripts/systems.py
```diff
@@ -231,6 +231,45 @@
...
+export PIKA_SMALL_STACK_SIZE=0x40000
+
+# Debug
+module list &> modules_{bs_name}.txt
+printenv > env_{bs_name}.txt
+
+# Commands
+""",
+}
+
```
msimberg, line 272: Overlap with https://github.com/eth-cscs/DLA-Future/pull/1209? We can also rename `santis` to `todi`? See PR description:
> Not changing santis to e.g. todi or daint... none of the names make much sense right now since systems are still being created and renamed

test/include/dlaf_test/eigensolver/test_gen_eigensolver_correctness.h
```diff
@@ -37,11 +34,14 @@ namespace dlaf::test {
...
                                    Matrix<const T, Device::CPU>& reference_b,
-                                   dlaf::EigensolverResult<T, D>& ret, GridIfDistributed&... grid) {
+                                   dlaf::EigensolverResult<T, D>& ret, SizeType eval_index_begin,
+                                   SizeType eval_index_end, GridIfDistributed&... grid) {
   using dlaf::matrix::MatrixMirror;
   using dlaf::matrix::test::allGather;
   using dlaf::matrix::test::MatrixLocal;
 
+  DLAF_ASSERT(eval_index_begin == 0l, eval_index_begin);
+
```
msimberg, line 44: Is this asserted in teh algorithm itself as well? If yes, maybe we can leave it out here?

test/unit/matrix/test_retiled_matrix_ref.cpp
```diff
@@ -0,0 +1,276 @@
...
+  };
+
+  auto el1_shifted_cols = [&spec, &el1](const GlobalElementIndex& index) {
+    auto shifted_idx = GlobalElementIndex{index.row(), index.col() + spec.size.cols()};
+    return el1(shifted_idx);
+  };
+
+  // TODO: Remove
+  // DLAF_ASSERT(spec.origin.row() == 0, spec.origin.row());
+  // DLAF_ASSERT(spec.origin.col() == 0, spec.origin.col());
```
msimberg, line 101: To do?

test/unit/matrix/test_retiled_matrix_ref.cpp
```diff
@@ -0,0 +1,276 @@
...
+struct RetiledMatrixRefTest : public TestWithCommGrids {};
+
+TYPED_TEST_SUITE(RetiledMatrixRefTest, MatrixElementTypes);
+
+const std::vector<
+    std::tuple<LocalElementSize, TileElementSize, LocalTileSize, GlobalElementIndex, GlobalElementSize>>
+    local_sizes_tests({
+        // size, tile_size (target), tiles_per_block (target), distribution origin (ref), distribution size (ref)
+        {{8, 8}, {2, 2}, {2, 2}, {0, 0}, {4, 4}},
+        // {{8, 8}, {2, 2}, {2, 2}, {2, 2}, {4, 4}},
```
msimberg, line 58: To do?

test/unit/matrix/test_retiled_matrix.cpp
```diff
@@ -78,7 +78,7 @@ const std::vector<std::tuple<LocalElementSize, TileElementSize, LocalTileSize>>
     {{4, 4}, {4, 4}, {1, 1}},
 });
 
-TYPED_TEST(RetiledMatrixTest, LocalConstructor) {
+TYPED_TEST(RetiledMatrixLocalTest, LocalConstructor) {
```
msimberg, line 81: Why was this renamed? Am I missing something obvious? :thinking:

test/unit/eigensolver/test_gen_eigensolver.cpp
```diff
@@ -163,21 +165,25 @@ void testGenEigensolver(const blas::Uplo uplo, const SizeType m, const SizeType
...
-      testGenEigensolver<TypeParam, Backend::MC, Device::CPU, Allocation::do_allocation,
-                         Factorization::do_factorization>(uplo, m, mb);
-      testGenEigensolver<TypeParam, Backend::MC, Device::CPU, Allocation::use_preallocated,
-                         Factorization::do_factorization>(uplo, m, mb);
-      testGenEigensolver<TypeParam, Backend::MC, Device::CPU, Allocation::do_allocation,
-                         Factorization::already_factorized>(uplo, m, mb);
-      testGenEigensolver<TypeParam, Backend::MC, Device::CPU, Allocation::use_preallocated,
-                         Factorization::already_factorized>(uplo, m, mb);
+
+      // Test with all evals and half of them
```
msimberg, line 176: Is it allowed to ask for 0 eigenvectors/values (or should it at least fail gracefully)? If yes, is it tested?

test/unit/c_api/eigensolver/test_gen_eigensolver_c_api_wrapper.h
```diff
@@ -20,78 +20,171 @@ DLAF_EXTERN_C int C_dlaf_symmetric_generalized_eigensolver_s(
     const int dlaf_context, const char uplo, float* a, const struct DLAF_descriptor desca, float* b,
     const struct DLAF_descriptor descb, float* w, float* z, const struct DLAF_descriptor descz);
 
+DLAF_EXTERN_C int C_dlaf_symmetric_generalized_eigensolver_partial_spectrum_s(
+    const int dlaf_context, const char uplo, float* a, const struct DLAF_descriptor desca, float* b,
+    const struct DLAF_descriptor descb, float* w, float* z, const struct DLAF_descriptor descz,
+    const SizeType eigenvalues_index_begin, const SizeType eigenvalues_index_end);
+
```
msimberg, line 27: These definitely look like they'd enjoy being generated with so many variants, but I suppose this might be the last major change in the API for a while? Definitely not asking you to change it in this PR, if at all, just thinking out loud...

test/unit/c_api/eigensolver/test_gen_eigensolver_c_api.cpp
```diff
@@ -8,12 +8,14 @@
...
 
-#include <functional>
+#include <optional>
 #include <tuple>
 #include <typeinfo>
 #include <utility>
 #include <vector>
 
+#include <unistd.h>
+
```
msimberg, line 18: What does this header provide?

.gitignore
```diff
@@ -9,6 +9,7 @@
 .clangd
```
msimberg, line 9: This could all be moved to a separate PR, but not objecting to the changes here either.

msimberg (changes requested):
I have only requests for cosmetic changes and some questions. This is looking very good!

scripts/systems.py
RMeli: Yes, this is for my own convenience right now (to run the benchmarks) but will be removed when #1209 is merged.

test/unit/c_api/eigensolver/test_gen_eigensolver_c_api_wrapper.h
RMeli: YES! I think this is the very last thing I will add manually: if we need to add something else, then this definitely needs to be generated.

I might already try to generate the Fortran interface with FYPP (which I already use for tests).

test/unit/c_api/eigensolver/test_gen_eigensolver_c_api.cpp
RMeli: This is still my annoying LPS...

include/dlaf/matrix/distribution.h
```diff
@@ -723,6 +727,7 @@ class Distribution {
   // Note: safe to use in constructors.
   void set_default_sizes() noexcept;
 
+  GlobalElementIndex origin_ = {0, 0};
   GlobalElementIndex offset_;
```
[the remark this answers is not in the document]
RMeli: This is was a leftover for previous experimentation. Removed.

include/dlaf/multiplication/hermitian.h
RMeli: I think it's fine to add them here, but if you prefer to have this in a separate PR I can do that too, I don't mind.

include/dlaf/util_math.h
RMeli: I tried to follow the DLAF convention, but this slipped. Changed to `class`.

miniapp/miniapp_bt_reduction_to_band.cpp
RMeli: Not sure. I don't think I used the bogus version. If we want one entry per line I can revert to that, and disable formatting.

scripts/gen_dlaf_strong-gpu.py
RMeli: Well spotted, thanks. So far I only tested the standard eigensolver, so this was not yet needed (but the miniapp has already been updated).

scripts/postprocess.py
RMeli: This is a very good point. I can add an option to trigger the old behavior if it makes things easier going forward.

scripts/postprocess.py
RMeli: This was for debugging, sorry. As mentioned, cleanup is still underway.

.gitignore
RMeli: https://github.com/eth-cscs/DLA-Future/pull/1215

Thanks for the quick merge @rasolca!

test/unit/matrix/test_retiled_matrix_ref.cpp
RMeli: Yes, but I will remove it for this PR since it is not technically needed for the partial back-transformation and the PR is too big already.

test/unit/matrix/test_retiled_matrix.cpp
RMeli: The dependencies test should also be changed to a local one. Happy to move this to a separate PR if needed (they are essentially un-related with this PR).

test/unit/eigensolver/test_gen_eigensolver.cpp
RMeli: No, currently it fails because of checks on indices access. Thinking about it, I think it should be supported as a "dirty" way to ask only for eigenvalues and no eigenvector. (I say "dirty" because completely removing the eigenvectors would require a separate API for some functions, where you still have to pass in the full eigenvector matrix.) I'm wondering if this should simply be a corner case, where we do a quick return after solving the tridiagonal system?

test/unit/matrix/test_retiled_matrix.cpp
msimberg: Ah, thanks! Yes, that makes sense now. If you don't mind, a separate PR would be great along with this explanation.

scripts/postprocess.py
msimberg: Still not requesting this change for this PR, but if you want to look into it, you might want to have a look at the `optional_parser` in `postprocess.py`. We added that the previous time we changed the output format. It will parse the field to `None` if it's not there (though I see we don't even use the potentially-parsed field at the moment...).

miniapp/miniapp_bt_reduction_to_band.cpp
msimberg: In my subjective opinion the old version with one entry per line was easier to read, so I'm in favour of that.

include/dlaf/multiplication/hermitian.h
msimberg: I'm fine with leaving them here.

test/unit/matrix/test_retiled_matrix.cpp
RMeli: Moved to https://github.com/eth-cscs/DLA-Future/pull/1216. I also moved there the small documentation fixes, to de-clutter a bit more this PR.

test/include/dlaf_test/eigensolver/test_gen_eigensolver_correctness.h
RMeli: Good point, I will remove it.

include/dlaf/matrix/matrix_ref.h
```diff
@@ -127,12 +145,30 @@ class MatrixRef : public MatrixRef<const T, D> {
...
+  ///
+  /// All accesses to the sub-pipelined matrix are sequenced after previous accesses and before later
+  /// accesses to the original matrix, independently of when tiles are accessed in the sub-pipelined
+  /// matrix.
+  ///
+  /// @pre blockSize() is divisible by @p tiles_per_block
+  /// @pre blockSize() == tile_size()
+  Matrix<T, D> retiledSubPipeline(const LocalTileSize& tiles_per_block) noexcept {
+    return mat_.retiledSubPipeline(tiles_per_block);
+  }
```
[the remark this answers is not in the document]
RMeli: Resolving this, but leaving the other comment open as one point to focus on during the review.

RMeli:
cscs-ci run

include/dlaf/matrix/matrix.h
```diff
@@ -205,7 +213,8 @@ class Matrix<const T, D> : public internal::MatrixBase {
   using ConstTileType = Tile<const ElementType, D>;
   using TileDataType = internal::TileData<ElementType, D>;
   using ReadOnlySenderType = ReadOnlyTileSender<T, D>;
-  friend Matrix<ElementType, D>;
+  using ReadWriteSenderType = ReadWriteTileSender<T, D>;
+  friend class internal::MatrixRef<const ElementType, D>;
```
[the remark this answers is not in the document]
RMeli: Yes, this is necessary to get access to the added protected `readwrite` functions of `MatrixRef<const T, D>`, which is called within `Matrix<const T, D>::setUpRetiledSubPipelines`.  Maybe there is a better way of doing things?

test/unit/eigensolver/test_gen_eigensolver.cpp
RMeli: Actually, this limitation was only coming from the correctness test. I'll update the tests accordingly.

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

RMeli:
cscs-ci run

scripts/postprocess.py
RMeli: I actually need this to compare with the `master` branch, therefore I implemented it as you suggest. Thanks for pointing it out.

include/dlaf/matrix/matrix.h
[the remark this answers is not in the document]
msimberg: Nothing comes to mind immediately... we can always refactor later if we realize there's a simpler way of doing it.

RMeli:
cscs-ci run

ci/ctest_to_gitlab.sh
```diff
@@ -95,7 +95,7 @@ JOB_TEMPLATE="
   variables:
     SLURM_CPUS_PER_TASK: {{CPUS_PER_TASK}}
     SLURM_NTASKS: {{NTASKS}}
-    SLURM_TIMELIMIT: '20:00'
+    SLURM_TIMELIMIT: '55:00'
```
rasolca, line 98: So much longer?

RMeli: We can probably reduce it to 50, but unfortunately yes, tests are quite a bit longer (`RANK_6` for `cuda gcc11 codecov test` is 45 minutes). The reason is that the eigensolver tests are now 4x: I test the original API and the new API is tested with 0, half, and the full eigenspectrum. Maybe we can reduce the coverage, especially for the C API (only testing the old API and the new one with half of the eigenvectors)?

include/dlaf/matrix/matrix.h
```diff
@@ -297,23 +306,36 @@ class Matrix<const T, D> : public internal::MatrixBase {
...
   Matrix(Distribution distribution) : internal::MatrixBase{std::move(distribution)} {
     DLAF_ASSERT((distribution.offset() == GlobalElementIndex{0, 0}), "not supported",
                 distribution.offset());
   }
+
+  template <class MatrixLike>
+  Matrix(MatrixLike& mat, const LocalTileSize& tiles_per_block) noexcept
+      : MatrixBase(mat.distribution(), tiles_per_block) {
+    setUpRetiledSubPipelines(mat, tiles_per_block);
+  }
```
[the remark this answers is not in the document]
RMeli: I'm going to resolve this conversation because it appears that the comment time is reported as "last week" and the code is also marked as "outdated". I already made this constructor protected again, as part of the big cleanup last week (making it public was for initial testing purposes).

include/dlaf/multiplication/hermitian/api.h
```diff
@@ -10,16 +10,20 @@
...
+using matrix::internal::MatrixRef;
+
 template <Backend B, Device D, class T>
 struct Hermitian {
   static void call_LL(const T alpha, Matrix<const T, D>& mat_a, Matrix<const T, D>& mat_b, const T beta,
                       Matrix<T, D>& mat_c);
 
   static void call_LL(comm::CommunicatorGrid& grid, const T alpha, Matrix<const T, D>& mat_a,
-                      Matrix<const T, D>& mat_b, const T beta, Matrix<T, D>& mat_c);
+                      MatrixRef<const T, D>& mat_b, const T beta, Matrix<T, D>& mat_c);
```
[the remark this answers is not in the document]
RMeli: The hermitian multiplication is only used in the checks for the miniapps, an only use the distributed version. Since this is all in the `internal` namespace, I did not change everything to use `MatrixRef`. I agree the inconsistency is not great, I'll fix it.

include/dlaf/multiplication/hermitian.h
```diff
@@ -21,6 +21,48 @@
...
+    }
+  }
+  else {
+    DLAF_ASSERT(matrix::multipliable(mat_b, mat_a, mat_c, blas::Op::NoTrans, blas::Op::NoTrans), mat_a,
+                mat_b, mat_c);
+    DLAF_UNIMPLEMENTED(side);
+  }
+}
+
+}
```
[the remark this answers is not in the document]
RMeli: Internally, I changed everything to use a `MatrixRef` but only for the matrix we are interested in. Therefore, this is the only function that is available for `MatrixRef` and it is called by the version using `Matrix` (so there is no code duplication). I'm fine with supporting all the different combinations of `Matrix`/`MatrixRef` but it seemed a bit overkill (it will add a combinatorial number of tests).

scripts/postprocess.py
```diff
@@ -215,7 +214,23 @@ def _parse_optional_text(text):
         return None
 
 
-additional_parsers = dict(optional_text=_parse_optional_text)
+@with_pattern(r"(|\s+\S+)")
+def _parse_optional_int(text):
+    text = text.strip()
+
+    for c in ["(", ",", ")"]:
+        text = text.replace(c, "")
```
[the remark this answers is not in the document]
RMeli: Very junky way to match a plain integer, or the following format which appears in the mini apps: `(NUMBER,` and `NUMBER)`. I can update the decorating regexp to do this properly (this is what I originally planned to do, but forgot).

include/dlaf/multiplication/hermitian.h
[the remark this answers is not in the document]
RMeli: After discussing in person with @rasolca yesterday, we agreed that this is probably fine since `MatrixRef` is in the `internal` namespace and all the changes are internal too. Essentially, I only modified what is needed in the internal namespace, while leaving the exposed API (with `Matrix`) unchanged. Templating over matrices will force us to test all different combinations of `Matrix`/`MatrixRef`, which is not ideal.

ci/ctest_to_gitlab.sh
RMeli: @rasolca suggested that the C API tests could be separated out (like the miniapps in #1112).

scripts/postprocess.py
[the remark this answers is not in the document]
RMeli: Parsing fixed as discussed online.

msimberg (approved):
No further comments from. Thanks for pushing this through!

rasolca (approved):

rasolca:
cscs-ci run

PR #1200, opened 2024-10-04
Add environment variable to print configuration

### LINKED ISSUE #1186: Add environment variable to print configuration
We currently have the command line option `--dlaf:print-config` which dumps the configuration used by DLA-Future. Is there any opposition to adding a corresponding `DLAF_PRINT_CONFIG` environment variable (or similar, but I think this name makes most sense given the existing command line option) with the same effect, that could be used e.g. when running CP2K?

msimberg:
Fix #1186.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -29,7 +29,6 @@
 
 namespace dlaf {
 std::ostream& operator<<(std::ostream& os, const configuration& cfg) {
-  os << "  print_config = " << cfg.print_config << std::endl;  // TODO: Is this silly?
   os << "  num_np_gpu_streams_per_thread = " << cfg.num_np_gpu_streams_per_thread << std::endl;
   os << "  num_hp_gpu_streams_per_thread = " << cfg.num_hp_gpu_streams_per_thread << std::endl;
   os << "  num_gpu_blas_handles = " << cfg.num_gpu_blas_handles << std::endl;
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -29,6 +29,7 @@
 
 namespace dlaf {
 std::ostream& operator<<(std::ostream& os, const configuration& cfg) {
+  os << "  print_config = " << cfg.print_config << std::endl;  // TODO: Is this silly?
   os << "  num_np_gpu_streams_per_thread = " << cfg.num_np_gpu_streams_per_thread << std::endl;
   os << "  num_hp_gpu_streams_per_thread = " << cfg.num_hp_gpu_streams_per_thread << std::endl;
   os << "  num_gpu_blas_handles = " << cfg.num_gpu_blas_handles << std::endl;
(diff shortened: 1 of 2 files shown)


msimberg:
cscs-ci run

src/init.cpp
```diff
@@ -29,6 +29,7 @@
 
 namespace dlaf {
 std::ostream& operator<<(std::ostream& os, const configuration& cfg) {
+  os << "  print_config = " << cfg.print_config << std::endl;  // TODO: Is this silly?
```
msimberg, line 32: This will always print 1 since it'll only be printed when it's 1... Leave it out?

src/init.cpp
```diff
@@ -185,8 +186,13 @@ void updateConfigurationValue(const pika::program_options::variables_map& vm, T&
...
   const std::string dlaf_cmdline_option = "dlaf:" + cmdline_option;
-  if (vm.count(dlaf_cmdline_option)) {
-    var = parseFromCommandLine<T>::call(vm, dlaf_cmdline_option);
+  if constexpr (std::is_same_v<T, bool>) {
+    var = var || vm.count(dlaf_cmdline_option) > 0;
+  }
+  else {
+    if (vm.count(dlaf_cmdline_option)) {
+      var = parseFromCommandLine<T>::call(vm, dlaf_cmdline_option);
+    }
```
msimberg, line 195: This is a bit funny but a result of the `--dlaf:print-config` option being a flag, not an option that takes `true/false`. This also assumes now that if we have binary configuration options, the corresponding command line option is simply a flag that toggles the option to true.

Do you think this is ok or would you prefer changing the command line option to be similar to the environment variable (which accepts `true/1/on/false/...`)?

msimberg:
Two open questions so not ready for merging, but ready for review.

src/init.cpp
RMeli: Yeah, I'd ditch it.

src/init.cpp
RMeli: Do we have many things that are ON by default and potentially need to be turned OFF? Those are the ugly cases with the flag approach, which I think it's otherwise great if there is a majority of OFF->ON.

msimberg: I think the only other flag we have is `--dlaf:no-mpi-pool` which is binary, and negated for good measure. However, pika's MPI polling will replace this hopefully soon so I wouldn't consider this flag for this discussion.

msimberg:
cscs-ci run

albestro (approved):

RMeli (approved):

rasolca (approved):

rasolca:
Merging other PRs created a conflict

msimberg:
cscs-ci run

msimberg:
> Merging other PRs created a conflict

I've fixed the conflict.

src/init.cpp
msimberg: Agreed in meeting that this is ok behaviour.

rasolca:
cscs-ci run

PR #1202, opened 2024-10-17
Remove unused `transformMPIDetach`, `transformMPILift`, and `transformMPILiftDetach`

msimberg:
We could do the same for a few `transform*` variants as well, but they're actually used still, so we'd need to do some rewriting. I think generally the explicit `start_detached(...)` is clearer to understand (and easier to grep for), so I'm starting by removing the unused `transformMPI*` variants.


msimberg:
cscs-ci run

rasolca (approved):

PR #1203, opened 2024-10-18
Small changes in the plotting scripts

aurianer:
- Adds complex and double in the caption
- Limit the caption to use the base directory name and not the absolute path (otherwise the caption of the plot was unreadable)

### DIFF
--- a/scripts/postprocess.py
+++ b/scripts/postprocess.py
@@ -380,7 +380,7 @@ def parse_jobs_cmdargs(description):
     parser.add_argument(
         "--path",
         action="append",
-        help="Plot results from this directory.",
+        help="Plot results from this directory. You can pass this option several times.",
     )
     parser.add_argument(
         "--distinguish-dir",
--- a/scripts/postprocess.py
+++ b/scripts/postprocess.py
@@ -353,13 +353,23 @@ def parse_jobs(data_dirs, distinguish_dir=False):
         data_dirs = [data_dirs]
     data = []
     for data_dir in data_dirs:
+        bench_name_postfix = os.path.basename(os.path.dirname(data_dir))
+        if os.path.basename(data_dir) == "d":
+            bench_name_postfix += "-double"
+        elif os.path.basename(data_dir) == "z":
+            bench_name_postfix += "-complex"
+        else:
+            print(
+                "The base of the --path provided is not a z or d folder, the results",
+                f"for complex and double might be aggreted in the plot. Path provided: {data_dir}",
+            )
         for subdir, dirs, files in os.walk(os.path.expanduser(data_dir)):
             for f in files:
                 if f.endswith(".out"):
                     nodes = float(os.path.basename(subdir))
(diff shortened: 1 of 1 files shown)


scripts/postprocess.py
```diff
@@ -353,13 +353,23 @@ def parse_jobs(data_dirs, distinguish_dir=False):
         data_dirs = [data_dirs]
     data = []
     for data_dir in data_dirs:
+        bench_name_postfix = os.path.basename(os.path.dirname(data_dir))
```
msimberg, line 356: Possible alternative: strip the common prefix from all `data_dirs`. If the basename happens to be the but some other parent component of the path is different, then you'd still be able to distinguish the paths. Worst case is you still end up with the full path in the name, but at least they'll be different. However, this is enough of a corner case that I don't care too much and would be happy with a simple solution (like now) here.

rasolca: possible alternative:
`--path=<path>[:<label>]`
In case label is not specified path is used as label.
```
for data_dir_label in data_dirs:
    dir_label = data_dir_label.split(":")
    data_dir = dir_label[0]
    if len(dir_label) == 1:
        label = data_dir
    else:
        label = dir_label[1]
```

albestro: probably with @rasolca solution there is a bit more control without too much complication.

scripts/postprocess.py
```diff
@@ -353,13 +353,23 @@ def parse_jobs(data_dirs, distinguish_dir=False):
...
+        if os.path.basename(data_dir) == "d":
+            bench_name_postfix += "-double"
+        elif os.path.basename(data_dir) == "z":
+            bench_name_postfix += "-complex"
+        else:
+            print(
+                "The base of the --path provided is not a z or d folder, the results",
+                f"for complex and double might be aggreted in the plot. Path provided: {data_dir}",
```
albestro, line 364:
```suggestion
                f"for complex and double might be aggregated in the plot. Path provided: {data_dir}",
```

scripts/postprocess.py
msimberg: @rasolca's option sounds pretty nice to me...

scripts/postprocess.py
```diff
@@ -352,14 +352,29 @@ def parse_jobs(data_dirs, distinguish_dir=False):
     if not isinstance(data_dirs, list):
         data_dirs = [data_dirs]
     data = []
-    for data_dir in data_dirs:
+    for data_dir_label in data_dirs:
+        dir_label = data_dir_label.split(":")
```
msimberg, line 356: I don't know what the most appropriate error checking/reporting thing is for python, but could we at least say `assert(1 <= len(dir_label) <= 2)` or whatever the python way of doing that is?

msimberg (approved):
Looks good to me now, thanks! Just a suggestion for some error checking.

msimberg (approved):

PR #1204, opened 2024-10-18
Add a plot_strong script to allow to generate all the plots for strong scaling

aurianer:
I think I would have gained quite some time if I had this kind of scripts in the repo. Also because passing the root directory where z/ and d/ are is actually plotting an aggregation of both which leads to high variance.
The script doesn't have defaults so it forces the user to set at least 2 variables (the base_paths: list of path to the data we want to plot, root of z/ and d/ directories + the output path where we want the plots).

### DIFF
--- a/scripts/plot_strong.sh
+++ b/scripts/plot_strong.sh
@@ -1,5 +1,17 @@
 #!/usr/bin/env bash
 
+#
+# Distributed Linear Algebra with Future (DLAF)
+#
+# Copyright (c) 2018-2024, ETH Zurich
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


scripts/plot_strong.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash -l
```
msimberg, line 1:
```suggestion
#!/usr/bin/env bash
```

msimberg (approved):
No particular comments besides the bash/python question. For consistency a python script would absolutely be better, but given that this isn't user-facing I'm happy here with anything that makes our life easier.

rasolca (approved):

PR #1205, opened 2024-10-22
Explicitly set sbatch `--no-requeue` in benchmarking sbatch scripts

msimberg:


RMeli (approved):
I feel this should be the default on ALPS, but LGTM.

rasolca (approved):

PR #1206, opened 2024-10-31
Don't use `SCOPED_TRACE` in `test_multiplication_hermitian`

msimberg:
The pika task may yield while the `SCOPED_TRACE` is in scope, and if it resumes on a different thread there will be segfaults. This PR removes the use of `SCOPED_TRACE` in that test.

It's still unclear what exactly in pika ends up yielding, since it's not expected to yield. That will be investigated separately.

Part of #603.

PR #1207, opened 2024-11-05
Test hemm no yield async rw mutex

msimberg:
Companion PR to #1206. Not for merging. I'm only testing a custom version of pika to see if the segfault goes away.


msimberg:
cscs-ci run

msimberg:
Closing. My naive change in pika did not avoid the yielding, but I think I have a reproducer for the other yield that happens in pure pika, so I won't be needing this PR anymore (or at the moment).

PR #1209, opened 2024-11-07
Small updates to system definitions for benchmarking

msimberg:
- Updates `MIMALLOC_LARGE_OS_PAGES` to `MIMALLOC_ALLOW_LARGE_OS_PAGES` (the latter works also for older versions)
- Adds `FI_MR_CACHE_MONITOR=disabled` to santis and lumi-gpu because of potential hangs without it
- Adds `DLAF_BT_BAND_TO_TRIDIAG_HH_APPLY_GROUP_SIZE=128` for santis and lumi-gpu
- Adds `DLAF_BAND_TO_TRIDIAG_1D_BLOCK_SIZE_BASE=2048` for lumi-gpu
- Sets `DLAF_UMPIRE_DEVICE_MEMORY_POOL_ALIGNMENT_BYTES` to the large page size on santis
- Removes `MPICH_OPT_THREAD_SYNC=0` since we now have new enough versions of MPICH available
- Uses `--gpus-per-task` instead of `gpu2ranks_slurm_cuda` on santis

I did not change the following:
- Not adding `PIKA_MPI_COMPLETION_MODE` yet since the values will change with 0.30.0 and possibly again with 0.31.0... and the default should long term be sufficient
- Not changing santis to e.g. todi or daint... none of the names make much sense right now since systems are still being created and renamed
- Not adding `FI_CXI_RDZV_THRESHOLD=262144` since in the end I think this only had a minor to no effect, but mentioning it here for searchability; we can add it later if we find it useful


scripts/systems.py
```diff
@@ -219,9 +216,12 @@
 #SBATCH --no-requeue
 
 # Env
-export MPICH_OPT_THREAD_SYNC=0 # Required to work around MPICH bug
+export FI_MR_CACHE_MONITOR=disabled
+export MPICH_GPU_SUPPORT_ENABLED=1
 export MIMALLOC_EAGER_COMMIT_DELAY=0
-export MIMALLOC_LARGE_OS_PAGES=1
+export MIMALLOC_ALLOW_LARGE_OS_PAGES=1
+export DLAF_BT_BAND_TO_TRIDIAG_HH_APPLY_GROUP_SIZE=128
```
RMeli, line 223: I know that `gen_dlaf_strong-gpu.py` is just an example, but I just wanted to point out that this variable is also set there directly as an extra flag for the miniapps. Should it be removed from there since there is a risk of divergence?

msimberg: I'm mostly indifferent. Should we put some other "placeholder" flag to replace the group size in the example script just to show what "extra flags" it's referring to?

msimberg: I removed them in https://github.com/eth-cscs/DLA-Future/pull/1209/commits/6f88fe2c0f2b12fb08f85aa1be061f2a23e054c9. Please unresolve if you disagree.

scripts/systems.py
```diff
@@ -198,15 +197,13 @@
 """,
 }
 
-# NOTE: Here is assumed that `gpu2ranks_slurm_cuda` is in PATH!
-#       modify "Run command" if it is not the case.
 cscs["santis"] = {
```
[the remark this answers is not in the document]
msimberg: Fine by me, but will do so in a separate PR then.

scripts/systems.py
```diff
@@ -87,7 +87,7 @@
 # Env
 export MPICH_MAX_THREAD_SAFETY=multiple
 export MIMALLOC_EAGER_COMMIT_DELAY=0
-export MIMALLOC_LARGE_OS_PAGES=1
+export MIMALLOC_ALLOW_LARGE_OS_PAGES=1
```
albestro, line 90: In the generator scripts we had 64 for `daint-mc` (which corresponds to the default value) and 128 for `daint-gpu`.

I'm ok leaving it out for `daint-mc` (even if it does not harm specifying it), but for `daint-gpu` I think it is missing.

```suggestion
export MIMALLOC_ALLOW_LARGE_OS_PAGES=1
export DLAF_BT_BAND_TO_TRIDIAG_HH_APPLY_GROUP_SIZE=128
```

I know that Piz Daint is going to be removed sooner rather than later, it's absolutely a minor comment.

albestro (approved):

PR #1213, opened 2024-11-15
TFactor: simplification and cleanup

albestro:
Some preparatory work for TFactor optimisations (both CPU and GPU coming next).

Main changes:
- head of V vector is no more treated differently (i.e. consider just the lower part): since it is well-formed (i.e. it has zeros in the upper part) we exploit that and we simplify the code (not skipping anymore computations with zeros involved).
- taus are copied all together in the tile diagonal instead of setting them one by one

TODO:
- [ ] check that this does not introduce any performance regression

### DIFF
--- a/include/dlaf/factorization/qr/t_factor_impl.h
+++ b/include/dlaf/factorization/qr/t_factor_impl.h
@@ -45,147 +45,150 @@ struct Helpers {};
 
 template <class T>
 struct Helpers<Backend::MC, Device::CPU, T> {
-  template <class TSender>
-  static auto set0(TSender&& t) {
+  static auto prepareT(matrix::ReadOnlyTileSender<T, Device::CPU> taus,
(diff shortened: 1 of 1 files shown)


albestro:
### Piz Daint MC
image
image
image
image

albestro:
### GPU

image
image
image
image

albestro:
I've just done a quick benchmark campaign. It has been done on daint mc and gpu (IMHO it does not really represent a problem) but it is not very extensive (high results variability in some cases).

Despite this, from this "quick and dirty" results I don't see any relevant problem in terms of performance (as expectable).

msimberg (approved):
I can't comment on the algorithmic changes, but from the pika perspective it looks good to me.

rasolca (approved):

rasolca:
cscs-ci run

PR #1214, opened 2024-11-15
TFactor: bulkerify GEMV (both MC and GPU)

albestro:
In this PR we try to increase parallelisation of GEMV step of TFactor, which is by construction serial (all results goes into the same block tile), by using workspaces for storing partial results and then reducing them before the final TRMV step.

Algorithmically, the main change is that `stepGEMV` loop has been replaced with a single `stepGEMVAll`, and the concept has been applied in a similar way for both backends, MC and GPU:

- MC: `pika::bulk` splits input tiles over multiple task, each one stores the partial results in their workspace and just one at the end does the reduction.
- GPU: similarly to the CPU, the work is forked over different pika tasks, each one computing a partial result on a different GPU stream. These tasks then join into a single task which performs the reduction.

In order to implement this solution, workspaces for intermediate results have been added (there is another option which does not require any additional workspace nor reduction, and that we might explore for MC in another PR).

TODO:
- [x] This work is based on #1213 
- [x] check with @msimberg if we can do something different instead of the "fork" (e.g. a bulk for the GPU but each pika thread should get a different cuda stream)
- [x] define better how many workspaces are needed (currently a random choice)
- [x] profile and test performance of this solution
- [x] Verify if there is anything to be "ported" from #798
- [x] workers default value should be backend dependent

Close #798.

EDIT: since this is conceptually similar to #798 and it is going to be closed as soon as this gets merged, I migrated here the doc fixes happened there.


albestro:
cscs-ci run

just to check if there is any major problems

rasolca (changes requested):
LGTM.
Please name new functions in snake case, and rename existing internal functions if possible.

include/dlaf/eigensolver/bt_reduction_to_band/impl.h
```diff
@@ -162,6 +163,14 @@ void BackTransformationReductionToBand<backend, device, T>::call(
   dlaf::matrix::Distribution dist_t({mb, total_nr_reflector}, {mb, mb});
   matrix::Panel<Coord::Row, T, device> panelT(dist_t);
 
+  const auto dist_ws = [&]() {
+    using dlaf::factorization::internal::getTFactorNWorkers;
+    const SizeType nworkspaces = to_SizeType(std::max<std::size_t>(0, getTFactorNWorkers() - 1));
```
msimberg, line 168: Naive question: 0 workspaces is ok? I guess this is the case when you reduce directly into the final tile without any intermediate workspaces, and the number of workers is 1?

include/dlaf/factorization/qr/api.h
```diff
@@ -10,9 +10,12 @@
 
 #pragma once
 
+#include <vector>
+
 #include <pika/execution.hpp>
 
 #include <dlaf/communication/communicator_pipeline.h>
+#include <dlaf/factorization/qr/internal/get_tfactor_nworkers.h>
```
msimberg, line 18: Needed here?

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -72,9 +71,9 @@ struct Helpers<Backend::MC, Device::CPU, T> {
                          });
   }
 
-  static matrix::Tile<T, Device::CPU> gemvLoop(const matrix::Tile<const T, Device::CPU>& tile_v,
-                                               const matrix::Tile<const T, Device::CPU>& taus,
-                                               matrix::Tile<T, Device::CPU> tile_t) noexcept {
+  static matrix::Tile<T, Device::CPU> loop_GEMV(const matrix::Tile<const T, Device::CPU>& tile_v,
```
msimberg, line 74: Just a style question:
```suggestion
  static matrix::Tile<T, Device::CPU> loop_gemv(const matrix::Tile<const T, Device::CPU>& tile_v,
```
?

msimberg (approved):
Only a few questions, nothing blocking.

include/dlaf/factorization/qr/api.h
albestro: Good that you raised it 😄 This is a choice that I wanted to quickly discuss.

Workspaces tiles have to be allocated outside, and there should be enough tiles for potentially all the workers. So, we are sure that the user needs to check how many workers there are through this getter, so I opted for including it together with the declarations of the API where the workspace is going to be passed.

include/dlaf/eigensolver/bt_reduction_to_band/impl.h
albestro: Yes, exactly.

There is another point to raise here: it is not said that all workspaces allocated will be used. It is legit to think that some are not going to used always (e.g. the panel is going to shrink in size over iterations so less workers are enough), but it might also happen that some tiles are never going to be used because the biggest size of the panel is not enough to have all the workers involved.

I preferred to not involve any logic about how many workers are actually going to be involved and kept on the safe side allocating one tile for any possible worker thread that might ever be involved given the configuration.

I might see the argument about possible memory waste, but at the same time I don't see it as a threatening or compelling topic.

include/dlaf/factorization/qr/t_factor_impl.h
albestro: see 05dc48d

include/dlaf/factorization/qr/internal/get_tfactor_nworkers.h
```diff
@@ -0,0 +1,32 @@
...
+#include <cstddef>
+#include <cstdint>
+
+#include <pika/runtime.hpp>
+
+#include <dlaf/tune.h>
+
+namespace dlaf::factorization::internal {
+
+inline size_t getTFactorNWorkers() noexcept {
```
rasolca, line 22: What about this to be snake case?

albestro: see https://github.com/eth-cscs/DLA-Future/pull/1214/commits/36f179d170396472eadd5bd9b3ad1b2a9b43b8c7

include/dlaf/factorization/qr/api.h
msimberg: I think internally we can explicitly include the required header? I think it'd be a different discussion if we considered a "public API header"-to-"names" mapping, but for internal functionality I think directly including the header that actually provides a function is simplest. I'm not insisting on this at all though, that's just my subjective opinion.

include/dlaf/tune.h
```diff
@@ -55,6 +55,10 @@ namespace dlaf {
 ///     Enable dump of tridiagonal solver input/output data to "tridiagonal.h5" file that will before
 ///     created in the working folder (it should not exist before the execution).
 ///     Set with environment variable DLAF_DEBUG_DUMP_TRIDIAG_SOLVER_DATA.
+/// - tfactor_nworkers:
+///     The maximum number of threads/stream to use for computing tfactor (e.g. which is used for
```
msimberg, line 59:
```suggestion
///     The maximum number of threads/streams to use for computing tfactor (e.g. which is used for
```

include/dlaf/factorization/qr/api.h
albestro: it makes totally sense! see https://github.com/eth-cscs/DLA-Future/pull/1214/commits/80797c1842d7d822b971e60868d51d694e2305be

rasolca (approved):

msimberg (approved):

albestro:
cscs-ci run

albestro:
Converted to draft in order to prevent merging, since I'm still doing some checks.

albestro:
cscs-ci run

albestro:
Apart the fixes committed between the two last CI runs, which fix for sure some problems, one major difference between the CI runs is that the oldest one ran on Eiger while the latest ran on Daint (because of the rebase on master).

The oldest one on Eiger had some failures that are concerning, specifically all the `CPU UNIT_RANK_6`, which failed due to a timeout on Eigensolver/GenEigensolver MC Distributed test ([example](https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8669017378#L2057))

I will try to investigate why that's the case and why it haven't happened on Daint.

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -41,66 +49,129 @@
...
+
+                          // reduce ws_T in tile_t
+                          if (worker_id == 0) {
+                            for (std::size_t other_worker = 1; other_worker < nworkers; ++other_worker) {
+                              tile::internal::add(T(1), workspaces[other_worker - 1], tile_t);
+                            }
+                          }
+                        }) |
+                    // Note: drop the barrier sent by the bulk and return tile_t
+                    ex::then([&tile_t](auto&&) mutable { return std::move(tile_t); });
```
msimberg, line 170: If you want to more closely match what the comment above says:
```suggestion
                    ex::then([&tile_t](auto) mutable { return std::move(tile_t); });
```
The barrier isn't releasing anything else so it's not critical to drop it as soon as possible. I just want to make it clear that with `auto&&` the barrier will not be destroyed in the lambda, but potentially much later.

I also realize now that the name `drop_value` that I've chosen for the algorithm in pika to do this is not clear enough (so much that it even misled me: https://github.com/eth-cscs/DLA-Future/pull/1257). It should likely be `ignore_value`, at least with the current semantics. If it actually consumed the values it could be called `drop_value`...

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -147,53 +211,117 @@ struct Helpers<Backend::MC, Device::CPU, T> {
...
+    const auto workers_params = split_tfactor_work(hh_tiles.size());
+    const std::size_t nworkers = workers_params.nworkers;
+    const std::size_t batch_size = workers_params.batch_size;
+    DLAF_ASSERT(workspaces.size() >= nworkers - 1, workspaces.size(), nworkers - 1);
+
+    for (std::size_t worker_id = 0; worker_id < nworkers; ++worker_id) {
+      const std::size_t begin = worker_id * batch_size;
+      const std::size_t end = std::min(hh_tiles.size(), (worker_id + 1) * batch_size);
+
+      if (end - begin <= 0)
```
msimberg, line 258: Are you expecting to cover cases where `end < begin` as well? If yes, perhaps do
```suggestion
      if (end <= begin)
```
since they're unsigned integers. If not, perhaps explicitly do
```suggestion
      if (end - begin == 0)
```
instead to make it clear that the value can't be negative? Though I think I prefer `end <= begin` in any case since it'll keep working with signed and unsigned integers.

msimberg (changes requested):

include/dlaf/factorization/qr/t_factor_impl.h
albestro: Actually, the comment sense was more towards "ignoring" the barrier value, not in the sense of releasing it but to actually do not care about it anymore, because I just wanted to return the `tile_t` alone.

Indeed, I didn't thought about what was happening to the barrier 😅 so in this case as you said it is not going to hurt in any way probably, but I agree it is better to change as you suggested.

Thanks for pointing that out!

include/dlaf/factorization/qr/t_factor_impl.h
albestro: AFAICT it should never happen `end < begin`, but the logic is not straightforward so it might be worth adding some check.

If you look just below, you can see that there is an implicit assumption about this

https://github.com/eth-cscs/DLA-Future/blob/0d2bd4f41f485f249f1c139898914823cedc2ed8/include/dlaf/factorization/qr/t_factor_impl.h#L262

include/dlaf/factorization/qr.h
```diff
@@ -33,26 +34,84 @@ namespace dlaf::factorization::internal {
...
+/// get_tfactor_nworkers), each tile should have the same size as @param tile_t
+/// @param mpi_col_task_chain where internal communications are issued
+///
+/// @pre reflectors in hh_panel are well formed (1s on the diagonal and 0s in the upper part)
+/// @pre hh_panel.getWidth() <= t.get().size().rows && hh_panel.size().getWidth() <= t.get().size().cols()
 template <Backend backend, Device device, class T>
 void computeTFactor(matrix::Panel<Coord::Col, T, device>& hh_panel,
                     matrix::ReadOnlyTileSender<T, Device::CPU> taus,
                     matrix::ReadWriteTileSender<T, device> t,
+                    std::vector<matrix::ReadWriteTileSender<T, device>> workspaces,
```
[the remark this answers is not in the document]
albestro: If I get it correctly, https://github.com/eth-cscs/DLA-Future/pull/1214#discussion_r1869233789 is on topic. The main idea is that in this way the caller can manage the workspace by re-using it over multiple calls.

include/dlaf/factorization/qr/t_factor_impl.h
albestro: Sorry, I missed my conclusion:

- I like `end == begin`
- I would probably add a moderate/heavy assert with `end >= begin`

what would you go for?

msimberg: Thanks for the update (https://github.com/eth-cscs/DLA-Future/pull/1214/commits/37bedc3ff3820ecce1ec88de3b16e776b7c82ce0)! If you now go for the `end == begin` version, I would still add an assertion that `end >= begin` as well. If it's not true, things would blow up spectacularly later.

msimberg: Sorry, I posted my above comment without seeing your new comment! I think I already answered your question, but to be explicit: I would add the assert.

Making the condition for the if-clause `end <= begin` only makes sense if you actually expect `end` to sometimes be strictly smaller than `begin`, but if I understand correctly, you don't expect that to ever happen.

include/dlaf/eigensolver/bt_reduction_to_band/impl.h
```diff
@@ -207,8 +219,15 @@ void BackTransformationReductionToBand<backend, device, T>::call(
 
     const LocalTileIndex taus_index{Coord::Row, k};
     const LocalTileIndex t_index{Coord::Col, k};
-    dlaf::factorization::internal::computeTFactor<backend>(panelV, mat_taus.read(taus_index),
-                                                           panelT.readwrite(t_index));
+
+    auto workspace_fit = select(panelWS, panelWS.iteratorLocal());
+    if (is_last)
+      for (auto& tile_ws : workspace_fit)
+        tile_ws = splitTile(std::move(tile_ws), {{0, 0}, {nr_reflectors, nr_reflectors}});
```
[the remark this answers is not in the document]
albestro: see https://github.com/eth-cscs/DLA-Future/pull/1214/commits/65637804a5efe4699778e2ce87d5a7caa1a3a5cf

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -41,66 +49,129 @@
...
+    const std::size_t nworkers = workers_params.nworkers;
+    const std::size_t batch_size = workers_params.batch_size;
+    DLAF_ASSERT(workspaces.size() >= nworkers - 1, workspaces.size(), nworkers - 1);
+
+    const auto hp_scheduler = di::getBackendScheduler<Backend::MC>(thread_priority::high);
+    return ex::when_all(ex::when_all_vector(std::move(hh_tiles)), std::move(taus), std::move(tile_t),
+                        ex::when_all_vector(std::move(workspaces))) |
+           di::continues_on(hp_scheduler) |
+           ex::let_value([hp_scheduler, nworkers, batch_size](auto&& hh_tiles, auto&& taus,
+                                                              auto&& tile_t, auto&& workspaces) {
```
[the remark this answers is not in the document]
albestro: see https://github.com/eth-cscs/DLA-Future/pull/1214/commits/fc565a3a3be7a6d409bf12db5f76b894743e1620

include/dlaf/factorization/qr/t_factor_impl.h
albestro: see https://github.com/eth-cscs/DLA-Future/pull/1214/commits/66150a4634f1bac74d34819e2586500ae1aebed4

include/dlaf/factorization/qr/t_factor_impl.h
albestro: see https://github.com/eth-cscs/DLA-Future/pull/1214/commits/021c642f7c6260e2a5e58476f6faa2690613900e and https://github.com/eth-cscs/DLA-Future/pull/1214/commits/2eae9d5cf37244cf8604d6fe76c592b4e0c1c10b

include/dlaf/factorization/qr/t_factor_impl.h
```diff
@@ -41,66 +49,129 @@
...
+             const std::chrono::duration<double> barrier_busy_wait = std::chrono::microseconds(1000);
```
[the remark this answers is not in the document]
albestro: see https://github.com/eth-cscs/DLA-Future/pull/1214/commits/021c642f7c6260e2a5e58476f6faa2690613900e

include/dlaf/factorization/qr.h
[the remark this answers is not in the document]
msimberg: Answering my own question after a call with @albestro: It _is_ using a round-robin panel for the tile senders, the senders are just accessed by the caller of `computeTFactor`. From the call we concluded that it may be worth passing the panel itself to `computeTFactor` and let it access the senders directly. This may be useful to avoid some `splitTile`s and unnecessary sender accesses that are currently happening in the caller(s) of `computeTFactor`. It remains to be seen if it's actually an improvement (in organization and/or performance). Thanks @albestro for the explanation and for looking into this!

include/dlaf/factorization/qr/internal/get_tfactor_nworkers.h
```diff
@@ -0,0 +1,38 @@
...
+#include <cstdint>
+
+#include <pika/runtime.hpp>
+
+#include <dlaf/tune.h>
+
+namespace dlaf::factorization::internal {
+
+template <Backend B>
+size_t get_tfactor_nworkers() noexcept {
```
msimberg, line 23:
```suggestion
std::size_t get_tfactor_nworkers() noexcept {
```

src/init.cpp
```diff
@@ -291,6 +291,9 @@ void updateConfiguration(const pika::program_options::variables_map& vm, configu
   // These are added automatically by updateConfigurationValue.
   auto& param = getTuneParameters();
   // clang-format off
+  updateConfigurationValue(vm, param.tfactor_nthreads, "TFACTOR_NTHREADS", "tfactor-nthreads");
+  updateConfigurationValue(vm, param.tfactor_nstreams, "TFACTOR_NSTREAMS", "tfactor-nstreams");
```
msimberg, line 295:
```suggestion
  updateConfigurationValue(vm, param.tfactor_num_threads, "TFACTOR_NUM_THREADS", "tfactor-num-threads");
  updateConfigurationValue(vm, param.tfactor_num_streams, "TFACTOR_NUM_STREAMS", "tfactor-num-streams");
```
? I see we've been inconsistent about this already, so could go either way :cry: We've already got `nworkers` for tune parameters, and `num_streams/handles` for config options...

msimberg: Thanks for the change @albestro. I suppose we might have to consider whether we should separately change the other `*nworkers` tune parameters to `*num_workers` as well. It's a breaking change, but perhaps worth again at this point. One can still set both variables for now to be compatible with both?

include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -1063,7 +1075,14 @@ Matrix<T, Device::CPU> ReductionToBand<B, D, T>::call(Matrix<T, D>& mat_a, const
     // TODO probably the first one in any panel is ok?
```
albestro, line 1075: Random thought: just saw this comment and wondering if we should have a look and see if we can "merge" this workspace into the other one? @rasolca

albestro: I will leave it for another PR 😉

albestro:
cscs-ci run

include/dlaf/factorization/qr/api.h
```diff
@@ -11,6 +11,7 @@
 #pragma once
 
 #include <complex>
+#include <vector>
```
msimberg, line 14: Nit: I suppose you could remove this? It was only for the vector<TileSender>, right? Also in `include/dlaf/factorization/qr.h`?

msimberg (changes requested):
Only a minor include-nit, otherwise this looks good to me.

@albestro make sure to merge latest master since it's got the new CI changes now.

rasolca (approved):

albestro:
cscs-ci run

msimberg (approved):

albestro:
cscs-ci run

PR #1218, opened 2024-11-21
Consume `CommunicatorPipelineSharedWrapper` in `transform_mpi`

albestro:
Currently `CommunicatorPipelineExclusiveWrapper` is the only object that gets consumed just after an MPI operation is posted.

This should happen also for `CommunicatorPipelineSharedWrapper` for the same exact reason.


albestro:
cscs-ci run

not yet tested...let's see what happens

include/dlaf/sender/transform_mpi.h
```diff
@@ -23,13 +23,18 @@
 
 namespace dlaf::comm::internal {
 
-/// This helper "consumes" a CommunicatorPipelineExclusiveWrapper ensuring that after this call
-/// the one passed as argument gets destroyed. All other types left as they are
-/// by the second overload.
+/// This helper "consumes" a CommunicatorPipelineSharedWrapper or a CommunicatorPipelineSharedWrapper
```
msimberg, line 26:
```suggestion
/// This helper "consumes" a CommunicatorPipelineExclusiveWrapper or a CommunicatorPipelineSharedWrapper
```

albestro: this is a new level of error 😂 new achievement ✅

thanks for spotting it

albestro:
cscs-ci run

rasolca (approved):

albestro:
@msimberg can you check the latest changes to the comments? are they ok or would you like to do some change? thanks!

rasolca:
🚨 https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8517190037

albestro:
> 🚨 https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8517190037

Possibly even more scary...that error is from a test which should in principle not be affected by these changes...

https://github.com/eth-cscs/DLA-Future/blob/5362f7f4ff25676c7067d3414e52693f87efa2ff/test/unit/communication/test_reduce.cpp#L33-L51

Indeed, it is testing the synchronous communication. Isn't it?

test/unit/eigensolver/test_gen_eigensolver.cpp
```diff
@@ -9,7 +9,6 @@
 //
 
 #include <fstream>
-#include <numbers>
```
msimberg, line 12: Do we have any idea how this was not caught by CI?

msimberg (approved):
Changes to `transform_mpi.h` look good.

Also good catch with `<numbers>`, but wondering how that did not cause compilation failure in CI? Header available in libstdc++ but exposes nothing in C++17 mode?

test/unit/eigensolver/test_gen_eigensolver.cpp
albestro: I just realised that using an old compiler (namely, the good old gcc@9 on Daint).

> Header available in libstdc++ but exposes nothing in C++17 mode?

Apparently, yes. Some evidence:
- https://godbolt.org/z/zoWbjoG58
- [`<numbers>` in GCC](https://github.com/gcc-mirror/gcc/blob/be8d1a358e3abc50c14a1d7b1cfee82fe6f6aa3c/libstdc%2B%2B-v3/include/std/numbers#L39)
- [`<numbers>` in LLVM](https://github.com/llvm/llvm-project/blob/c1ad064dd3b9a1bf503bb8cd6d0d0418a05d10e8/libcxx/include/numbers#L65)

albestro: Just to unlink it from this PR, do we want to open an issue and keep track of that? or we are not even interested since there seems no way of checking that?

albestro: As quickly discussed with @rasolca, we are not going to keep track of this.

CI didn't realise it because by using sufficiently new compilers the header is available anyway, but since it was actually unused, it was not triggering any compilation error.

So, it should either be
- compiler error because not available in the cpp standard used
- an unused header

The former is enforced by the compiler, the latter might be detected by a more thorough check about actually unused headers.

msimberg: Thanks for looking into this! Yeah, I suppose as long as we don't have a proper include-what-you-use check that also removes includes we can't do much about these kind of issues.

PR #1192, opened 2024-09-19
CI move to ALPS (daint-mc -> eiger)

rasolca:
Depends on #1197

PR #1224, opened 2024-12-02
Enable MPI pool on all CIs

rasolca:


rasolca:
cscs-ci run

msimberg:
Any particular reason? Debugging? Consistency? I have nothing much against this, but just looking to understand the reasoning (as I think it can be good to test with it both on and off).

rasolca:
Noticed during rebase of #1192 that pool was enabled only on CUDA pipelines with no clear reason.

IMO:
- Default behavior should be with pool.
- If we want to test without pool we can add or modify some pipeline (and should be tested on both mc and gpu)

msimberg (changes requested):
Already discussed separately with @rasolca, but making the request explicit: I'd like to keep at least one CI configuration with the pool disabled just as a sanity check, since the pool is off by default.

PR #1227, opened 2024-12-04
scripts: let the user know if submitting with sbatch failed

albestro:
Currently the script submitting benchmark jobs is not reporting anything from the sbatch submission, resulting in the user now knowing if the job has been queued or not.

With this PR, in case of error the output is reported to the user("soft failure", just error reported without making the launching script to terminate).

```
(dla-future)[todi][ialberto@todi-ln001 dla-future.master]$ python3 scripts/gen-dlaf-strong-panel-workers.py
Submitting : sbatch --chdir=/capstor/scratch/cscs/ialberto/20241204-dlaf-strong-panel-workers-local-40/d/0.25 /capstor/scratch/cscs/ialberto/20241204-dlaf-strong-panel-workers-local-40/d/0.25/job_dlaf.sh using shell /usr/local/bin/bash
Submitting : sbatch --chdir=/capstor/scratch/cscs/ialberto/20241204-dlaf-strong-panel-workers-local-50/d/0.25 /capstor/scratch/cscs/ialberto/20241204-dlaf-strong-panel-workers-local-50/d/0.25/job_dlaf.sh using shell /usr/local/bin/bash
Job submission failed with return code 1
sbatch: error: QOSMaxSubmitJobPerUserLimit
sbatch: error: Batch job submission failed: Job violates accounting/QOS policy (job submit limit, user's size and/or time limits)

```

Any suggestions/preference is welcome. Some possible discussion points:
- should we report the error code at all?
- add a success message when everything works

### DIFF
--- a/scripts/miniapps.py
+++ b/scripts/miniapps.py
@@ -157,6 +157,17 @@ class JobText:
             stdout=PIPE,
             stderr=PIPE,
         )
+
+        try:
+            raw_stdout, raw_stderr = process.communicate(timeout=10)
+            retcode = process.returncode
+            if retcode != 0:
+                print(f"Job submission failed with return code {retcode}")
+                print(raw_stderr.strip())
+        except TimeoutExpired:
+            process.kill()
+            print("Job submission timed out")
+
         # sleep to not overload the scheduler
         sleep(1)
 
--- a/scripts/miniapps.py
+++ b/scripts/miniapps.py
@@ -158,15 +158,11 @@ class JobText:
             stderr=PIPE,
         )
 
-        try:
-            raw_stdout, raw_stderr = process.communicate(timeout=10)
-            retcode = process.returncode
-            if retcode != 0:
-                print(f"Job submission failed with return code {retcode}")
-                print(raw_stderr.strip())
-        except TimeoutExpired:
-            process.kill()
-            print("Job submission timed out")
+        raw_stdout, raw_stderr = process.communicate()
+        print(raw_stdout, end="")
+        print(raw_stderr, end="")
+        if process.returncode != 0:
+            print(f"Job submission FAILED.")
 
         # sleep to not overload the scheduler
         sleep(1)


albestro:
After a chat with @rasolca about this, I slightly changed the behaviour.

@rasolca recalled that there are two use-cases:
- submitting via sbatch
- launching via bash script (e.g. local runs without slurm queue)

This PR is trying to address job submission to the queue, which does not happen in the latter case.

- Using it with slurm will end up printing "submitting" and either nothing more or the full `sbatch` output in case of error.
- Using it in local (i.e. without queuing) it will print out the full output, both stdout and stderr.

In case of error, either from the sbatch queueing command or the bash script (e.g. just last command failing?), the return code is checked and a failure message is printed.

Waiting for your comments for last refinements.

scripts/miniapps.py
```diff
@@ -157,6 +157,13 @@ def submitJobs(self, run_dir, debug=False):
             stdout=PIPE,
             stderr=PIPE,
         )
+
+        raw_stdout, raw_stderr = process.communicate()
+        print(raw_stdout, end="")
+        print(raw_stderr, end="")
+        if process.returncode != 0:
+            print(f"Job submission FAILED.")
```
msimberg, line 165: In the case that one is submitting locally without slurm, will this mean that if there's an actual miniapp failure, the script will stop submitting jobs? That might not be the desired behaviour, but then I never use that so I'm not sure what the desired behaviour is.

@biddisco's input would be very useful here.

msimberg (commented):
Thanks @albestro for taking this on. I've also suffered in the past from errors not being reported.

scripts/miniapps.py
albestro: No, actually we just report the output/error but we don't stop in any case the execution.

I never used it locally, so I might be missing something. In the local case, the return code depends on the last command executed by the bash script, and just if the last one fails the "FAILED" message is printed, but no "FAILED" printed if any of the "middle" runs fails.

msimberg: Ah, good point, I misread the code... In that case, probably no concern here from my side.

msimberg (approved):

RMeli (approved):

PR #1228, opened 2024-12-05
Enable timestamps in CI logs

msimberg:
Enables timestamps in CI logs (according to https://docs.gitlab.com/runner/configuration/feature-flags.html). Raw logs become slightly more verbose. The GitLab UI renders the timestamps in a dark gray (same color as the line number) separately from the actual log output (see e.g. https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/8557603567).

I find this is useful to see if e.g. a test hung for a long time before eventually being killed by slurm, or if it ran for only a few seconds before being killed. Without timestamps one only sees the total time taken for a full command.

### DIFF
--- a/ci/common-ci.yml
+++ b/ci/common-ci.yml
@@ -6,6 +6,9 @@ stages:
   - build
   - test
 
+variables:
+  FF_TIMESTAMPS: true
+
 ##
 ## BUILDS
 ##
--- a/ci/ctest_to_gitlab.sh
+++ b/ci/ctest_to_gitlab.sh
@@ -28,6 +28,9 @@ stages:
   - upload
 
 variables:
+  FF_TIMESTAMPS: true
+
+variables:
   SLURM_EXCLUSIVE: ''
   SLURM_EXACT: ''
   SLURM_CONSTRAINT: $SLURM_CONSTRAINT
@@ -77,6 +80,9 @@ image: $IMAGE
 
 stages:
   - test
+
+variables:
+  FF_TIMESTAMPS: true
 
 variables:
   SLURM_EXCLUSIVE: ''


ci/common-ci.yml
```diff
@@ -6,6 +6,9 @@ stages:
   - build
   - test
 
+variables:
+  FF_TIMESTAMPS: true
+
```
msimberg, line 11: To do: will this be inherited by the pipeline generated by `ctest_to_gitlab.sh`? Probably not, but check it.

msimberg:
cscs-ci run

ci/common-ci.yml
msimberg: It won't. Added to `ctest_to_gitlab.sh` as well in https://github.com/eth-cscs/DLA-Future/compare/4234fc91ed6c209be7fd0bcb58a605db28a5d716..22e82b065dbc497ed2b3077a3c1ef6cae76df9a4.

msimberg:
cscs-ci run

ci/common-ci.yml
rasolca: Did you try https://github.com/eth-cscs/DLA-Future/blob/ci/alps_1/ci/common-ci.yml#L110-L111 ?

msimberg: I did not! Thanks for pointing that out, I wasn't aware that existed. Trying it out in https://github.com/eth-cscs/DLA-Future/pull/1228/commits/2438e960bad38a256fa981b39e0db760bfc170a8.

msimberg:
cscs-ci run

ci/common-ci.yml
msimberg: https://github.com/eth-cscs/DLA-Future/commit/2438e960bad38a256fa981b39e0db760bfc170a8 did not seem to have an effect, but I don't know if I'm using it wrong. Do you know if there are any special requirements on what variables are forwarded?

rasolca: Had a look to the doc:
The correct forward should be `yaml_variables` but should be true by default.
Worth a try explicitly setting it.
https://docs.gitlab.com/ee/ci/yaml/#triggerforward

msimberg: Yep, thanks. Let's see: https://github.com/eth-cscs/DLA-Future/pull/1228/commits/a10b945d6776c0550c5a1cfca1d1f00e3a5fe414.

msimberg:
cscs-ci run

ci/common-ci.yml
rasolca: 😞 Let's rollback to 22e82b0. Sorry for the noise.

msimberg: No worries, I at least learned about a new gitlab feature :wink: It's now back to the original version.

PR #1188, opened 2024-09-16
Move stdexec compilation to c++ files + add CI for cuda stdexec configuration

aurianer:
- Move stdexec compilation out of cuda files.
- Remove spack conflict when cuda and pika cxxstd=20.
- Attempt to add CI with gcc 12, cuda 11 and stdexec.
- Use @msimberg's fix for latest stdexec, I'll rebased when #1187 is merged.

PR #1229, opened 2024-12-05
Add back C++20 conflict with CUDA for older versions

msimberg:
#1188 allowed compiling DLA-Future with CUDA and stdexec with C++20. It also completely removed the spack package conflict that previously prevented that spec from actually being concretized. This PR adds back the conflict for older versions of DLA-Future (i.e. 0.6 and older), and expands it a bit.

- Previously the conflict was only for `pika cxxstd=20`, while `cxxstd=23` is also an allowed option and concretization could end up with that instead. 23 is just as a bad as 20 though, so this adds a conflict for 23 as well.
- pika 0.30.0 fixed an issue with including pika headers in CUDA mode with C++20 (https://github.com/pika-org/pika/pull/1249), so the conflicts have an upper bound on pika 0.29.
- With DLA-Future 0.6 and older, no version of pika with stdexec could be compiled because stdexec does not compile under nvcc, so there's a new conflict for that.
- After DLA-Future 0.6 any (otherwise allowed) combination of pika/stdexec/cxxstd is ok with CUDA, since we no longer include problematic headers in CUDA files.

### DIFF
--- a/spack/packages/dla-future/package.py
+++ b/spack/packages/dla-future/package.py
@@ -99,6 +99,10 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("pika +cuda", when="+cuda")
     depends_on("pika +rocm", when="+rocm")
 
+    for cxxstd in ("20", "23"):
+        conflicts(f"^pika cxxstd={cxxstd}", when="@:0.6 +cuda ^pika@:0.29")
+    conflicts("^pika +stdexec", when="@:0.6 +cuda")
+
     depends_on("whip +cuda", when="+cuda")
     depends_on("whip +rocm", when="+rocm")
 


msimberg:
cscs-ci run

spack/packages/dla-future/package.py
```diff
@@ -99,6 +99,10 @@ class DlaFuture(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("pika +cuda", when="+cuda")
     depends_on("pika +rocm", when="+rocm")
 
+    for cxxstd in ("20", "23"):
+        conflicts(f"^pika cxxstd={cxxstd}", when="@:0.6 +cuda ^pika@:0.29")
```
msimberg, line 103: I just realized that we have another conflict that makes `dla-future@0.6 +cuda ^pika@0.30` anyway impossible: https://github.com/eth-cscs/DLA-Future/blob/0bbec23b8e6ec793a554cd0d035a57ebef0bd94a/spack/packages/dla-future/package.py#L95. Do you think it's worth keeping the pika version constraint here anyway despite the overlap?

rasolca:
Quick question:
I suppose that C++23 flag is added to nvcc. does it support it?

msimberg:
> Quick question: I suppose that C++23 flag is added to nvcc. does it support it?

No, it doesn't support it. The `cxxstd` variant is only a pika thing, which can go up to 23 with CUDA disabled. We do also conflict for that in the pika package itself (https://github.com/spack/spack/blob/901cea7a54afac0c5258bd513f174be2a9452538/var/spack/repos/builtin/packages/pika/package.py#L123) so this is also "redundant" but I think it's still semantically correct to add it to DLA-Future as well. But I also don't insist on the redundancy.

spack/packages/dla-future/package.py
rasolca: IMO constrains serves as documentation as well. I would keep it.

msimberg:
Timeout on one of the 6-ranks test runs: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8560069368/viewer. However, `test_gen_eigensolver_c_api` took almost 18 minutes to run. `test_eigensolver_c_api` and `test_gen_eigensolver` both took about about 10 minutes, and `test_eigensolver` 5 minutes.

I've restarted that job.

albestro (approved):

PR #1230, opened 2024-12-05
Debugging asan failure

msimberg:
Just for debugging of failure on https://github.com/eth-cscs/DLA-Future/pull/1192: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8559806368#L603. It smells like an asan issue, but not clear yet. The issue reproduces even on my laptop (using the CI image), but only when increasing the number of threads to 16 or 32 (and disabling thread binding). This is probably why we never saw it on daint-mc (with 6 threads per rank on the failing test). The issue does _not_ reproduce with `detect_stack_use_after_return` enabled, so I'm testing enabling it, even though it's likely to produce other asan issues (again, asan itself failing, rather than legitimate warnings reported by asan).


msimberg:
#1192 was merged with the change from this PR. Closing.

PR #1233, opened 2024-12-10
Fix timeout issue splitting RANK6 C-API tests

rasolca:


rasolca:
cscs-ci run

msimberg (approved):
Looks good to me, thanks!

Except for the timeout on https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8601357792 :roll_eyes: Definitely not caused by this PR of course. I see now that for some reason I only removed one of the two `SCOPED_TRACE`s in https://github.com/eth-cscs/DLA-Future/pull/1206. Presumably because I found that removing the first was enough to avoid the most common hang. I don't see any reason why the second would be any safer, and all I can think of is that for some reason getting hangs from the second `SCOPED_TRACE` is for some reason less likely, so we didn't see it previously on #1206 after removing the first.

rasolca:
cscs-ci run

RMeli (approved):
Thanks @rasolca, LGTM.

albestro (approved):

PR #1244, opened 2024-12-18
Preparation for 0.7.1 release

msimberg:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## DLA-Future 0.7.1
+
+### Bug fixes
+
+- Fixed a compilation issue with ROCm. (#1241)
```
msimberg, line 7: Unsure if we want to be more explicit here. I _think_ the affected versions are 5.7 and newer, but I'm not entirely sure. That's why I've at least left it out initially.

RMeli (approved):

CHANGELOG.md
rasolca: OK not being specific

rasolca (approved):

PR #1249, opened 2024-12-19
 Preparation for 0.7.2 release

RMeli:


RELEASE_PROCEDURE.md
```diff
@@ -38,6 +38,8 @@ DLA-Future follows [Semantic Versioning](https://semver.org).
    with local repository. Exclude blocks delimited by `###` comments. These are only intended for the
    internal spack package.
 
+1. For patch releases: update `CHANGELOG` of `master` branch.
+
```
msimberg, line 42: I was thinking this could be "merge `version_X.Y` branch into master"`? It doesn't always work if we plan on ever patching old versions after we've released a new minor/major version, but will work in 99% of cases I would expect. You can maybe separate this change into a different PR to not block on this discussion?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## DLA-Future 0.7.2
+
+### Changes
+
+- Changed C ScaLAPACK API indexing convention to 1-based for partial eigenspectrum. (#1248)
```
msimberg, line 7: Is this a bugfix? I'm fine either way and won't matter too much especially if we mark 0.7.0 and 0.7.1 deprecated in spack, but by making this release 0.7.2 and not 0.8.0 we're signaling that this is a bugfix.

msimberg (approved):
Happy as it is, but added a couple of comments that may need discussion.

CHANGELOG.md
RMeli: I wondered the same, but I went with "changed" because it was not technically a bug. Happy to mark it as a bugfix if preferred.

msimberg: It's not a strong preference from me. Mostly wondering out loud...

RELEASE_PROCEDURE.md
RMeli: Yes, that is fair. https://github.com/eth-cscs/DLA-Future/pull/1250

msimberg (approved):

PR #1250, opened 2024-12-19
Update release procedure

RMeli:
~The `CHANGELOG` in `master` does not contain release notes for `0.7.1`. Add this step to the `RELEASE_PROCEDURE`.~

Improve release procedure, especially for patch releases. (Lessons learned...)

### DIFF
--- a/RELEASE_PROCEDURE.md
+++ b/RELEASE_PROCEDURE.md
@@ -2,28 +2,26 @@
 
 DLA-Future follows [Semantic Versioning](https://semver.org).
 
-1. For minor and major releases: check out the `master` branch. All changes required for the release are
-   added to `master` via pull requests. For patch releases: check out the corresponding
-   `version_major.minor` branch.
+1. Update the following files via a PR to the `master` branch:
 
-1. Write release notes in `CHANGELOG.md`. Check for issues and pull requests for the release on the
-   [DLA-F Planning board](https://github.com/orgs/eth-cscs/projects/1). Make sure to include changes that
-   may affect users, such as API changes, bugfixes, performance improvements, dependency updates. Changes
-   that do not directly affect users may be left out, such as CI changes, miscellaneous spack package
-   updates, documentation updates, or utility script updates. All list entries and paragraphs must be on
-   a single line for correct formatting on GitHub.
+   1. Write release notes in `CHANGELOG.md`. Check for issues and pull requests for the release on the
+      [DLA-F Planning board](https://github.com/orgs/eth-cscs/projects/1). Make sure to include changes that
+      may affect users, such as API changes, bugfixes, performance improvements, dependency updates. Changes
+      that do not directly affect users may be left out, such as CI changes, miscellaneous spack package
+      updates, documentation updates, or utility script updates. All list entries and paragraphs must be on
+      a single line for correct formatting on GitHub.
 
-1. Update the version in `CMakeLists.txt`.
+   1. Update the version in `CMakeLists.txt`.
 
-1. Update the version and date in `CITATION.cff`.
+   1. Update the version and date in `CITATION.cff`.
 
-1. When making a post-1.0.0 major release, remove deprecated functionality if
-   appropriate.
+   1. When making a post-1.0.0 major release, remove deprecated functionality if
+      appropriate.
 
-1. Update the minimum required versions if necessary.
+   1. Update the minimum required versions if necessary.
 
-1. Add a link to the documentation for the release in `DOCUMENTATION.md` and update the link in `README.md`.
-   The documentation will be generated automatically after the `vx.y.z` tag has been created and pushed.
+   1. Add a link to the documentation for the release in `DOCUMENTATION.md` and update the link in `README.md`.
+      The documentation will be generated automatically after the `vx.y.z` tag has been created and pushed.
 
 1. Merge all the changes for the release to the `master` branch (also for patch releases), including the above
    changes to version, documentation, etc.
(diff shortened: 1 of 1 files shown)


msimberg (approved):

msimberg (changes requested):
From the 0.7.2 release, let's add at least the following
- always add release notes to the master branch even for patch releases to avoid mismatch between release and master branch
- always cherry pick changes from master to release branch (changes and release notes)
- can we add something about checking the release branch to make sure it matches what's in the release notes? (at the very least a manual inspection check to compare the release branch to the previous release?)

rasolca:
I would add to not release until tests passed. I.e. wait that the tests complete for master (major release) or the version branch (minor and patch).

RELEASE_PROCEDURE.md
```diff
@@ -25,14 +25,30 @@ DLA-Future follows [Semantic Versioning](https://semver.org).
...
    The documentation will be generated automatically after the `vx.y.z` tag has been created and pushed.
 
+1. Merge all the changes to the `master` branch (also for patch releases).
+
 1. Ensure you have `gh` ([GitHub CLI](https://cli.github.com)) installed. Run `gh auth login` to authenticate
    with your GitHub account, or set `GITHUB_TOKEN` to a token with `public_repo` access.
 
+1. Ensure you are working from the `eth-cscs/DLA-Future` and not your own fork for following steps.
+
+1. For patch releases: switch to `version_X.Y` branch.
```
msimberg, line 35:
```suggestion
1. For patch releases: switch to `version_X.Y` branch. If it doesn't exist yet, create it from the `vX.Y.0` tag.
```
?

RELEASE_PROCEDURE.md
```diff
@@ -25,14 +25,30 @@ DLA-Future follows [Semantic Versioning](https://semver.org).
...
+
 1. Create a release on GitHub using the script `scripts/roll_release.sh`. This
    script automatically tags the release with the corresponding release number.
 
+1. For patch releases: merge the `version_X.Y` branch into `master`.
+
 1. Update spack recipe in `spack/packages/dla-future/package.py` adding the new release.
 
+1. For patch releases: update `CHANGELOG.md` in `master` to match the content of the patch release.
+
```
msimberg, line 51: I guess these two `For patch releases:` points may not be necessary with the other points about always cherry pikcing from `master`? Not sure if I'm missing something, but in principle all changes should already be on the `master` branch. On the other hand, I also don't mind keeping the steps in case we have to make some manual changes on the release branch?

RELEASE_PROCEDURE.md
```diff
@@ -25,14 +25,30 @@ DLA-Future follows [Semantic Versioning](https://semver.org).
...
+1. For patch releases: switch to `version_X.Y` branch.
+
+1. For patch releases: cherry pick the commits for the patch from `master` to the `version_X.Y` branch.
+   Make sure the cherry-picked commits are listed in the `CHANGELOG.md`, and make sure you cherry-picked
+   the PR(s) updating all the above (version, documentation, etc.).
+
+1. Check the release branch to make sure the content matches the release notes.
+
 1. Create a release on GitHub using the script `scripts/roll_release.sh`. This
    script automatically tags the release with the corresponding release number.
```
msimberg, line 44: I think @rasolca's comment about waiting for CI to pass before making the release is still missing from here (https://github.com/eth-cscs/DLA-Future/pull/1250#issuecomment-2554851068).

msimberg (changes requested):

RELEASE_PROCEDURE.md
RMeli: Yep!

RELEASE_PROCEDURE.md
RMeli: Yes, I noticed it too this morning that after adding what Raffaele suggested (to always push things on `master`), this is no longer necessary.

RELEASE_PROCEDURE.md
RMeli: Good catch! It was lost in 862a9a3f86d5fc0a8af00f7ccd00442838629941 when I changed to always push everything to master. I'll re-add it. I also double checked that the CI is automatically triggered for the `version_*` branches, which it is.

RELEASE_PROCEDURE.md
```diff
@@ -2,32 +2,47 @@
...
+      may affect users, such as API changes, bugfixes, performance improvements, dependency updates. Changes
+      that do not directly affect users may be left out, such as CI changes, miscellaneous spack package
+      updates, documentation updates, or utility script updates. All list entries and paragraphs must be on
+      a single line for correct formatting on GitHub.
 
-1. Update the version in `CMakeLists.txt`.
+   1. Update the version in `CMakeLists.txt`.
 
-1. Update the version and date in `CITATION.cff`.
+   1. Update the version and date in `CITATION.cff`.
```
RMeli, line 16:
```suggestion
   1. Update the versions and date in `CITATION.cff`.
```

RELEASE_PROCEDURE.md
```diff
@@ -2,32 +2,47 @@
...
+   1. When making a post-1.0.0 major release, remove deprecated functionality if
+      appropriate.
 
-1. Update the minimum required versions if necessary.
+   1. Update the minimum required versions if necessary.
 
-1. Add a link to the documentation for the release in `DOCUMENTATION.md` and update the link in `README.md`.
-   The documentation will be generated automatically after the `vx.y.z` tag has been created and pushed.
+   1. Add a link to the documentation for the release in `DOCUMENTATION.md` and update the link in `README.md`.
+      The documentation will be generated automatically after the `vx.y.z` tag has been created and pushed.
```
RMeli, line 24:
```suggestion
      The documentation will be generated automatically after the `vX.Y.Z` tag has been created and pushed.
```

msimberg:
Should we mention something about using GitHub's milestones for a release? Currently they're not mentioned, but we sort of use them. I created one for 0.7.3 now: https://github.com/eth-cscs/DLA-Future/milestone/31.

msimberg:
Idea from @RMeli: check automatically that the release branch has non-trivial commits since the previous version tag. Perhaps checking for changes to cpp and other source files in the `src` directory? May need overriding if making a patch release which only changes/fixes e.g. cmake files. Updating the version in `CMakeLists.txt` would already satisfy this check which is not what we want?

Then again, we should not go completely overboard with checks either...

rasolca (approved):

PR #1254, opened 2024-12-20
benchmark scripts: basic separation for alps vcluster, number of cores, gpu defaults, ...

albestro:
Main discussion points:

1. create a "base" `alps-gh200` config so that each vcluster can inherit and customise it
2. reduce number of cores: see dedicated comment below
3. is pika mpi mode 31+1 the best one? should we go for this or keep the default initialised (30+0)? @msimberg @biddisco 
4. currently we cannot set differently DLAF params for MC and GPU, and for GPU defaults are not the best. should we set it like this or do we prefer to have them in the generator scripts as cli parameters?

- [x] #1214 required for `DLAF_TFACTOR_NWORKERS` in point 4
- [x] Update parameters name according to #1270


albestro:
About number of cores to use, there was the question if using all cores might have been problematic, e.g not leaving enough resources to the system.

In some cases it revealed to be problematic (e.g. due to an OpenMP problem that seems to use 1 thread more than requested), so here it is a quick benchmark.

Next are time plots (the lower the better) for eigensolver. The naming of series are distinguished by two numbers:
- first number is how many cores the allocation is given (`-c` slurm parameter)
- second number is how many pika threads are used (`--pika:threads` parameter)

These benchmarks ran all on current master (0.7.0) with `--dlaf:red2band-panel-nworkers=50`.

image
image
image
image

From a quick look I wouldn't say there is any real difference, except on `--local` runs which seem to have a small benefit having more cores available (72-72 > 72-64 > 64-64).

msimberg:
> is pika mpi mode 31+1 the best one? should we go for this or keep the default initialised (30+0)? @msimberg @biddisco

I think 31+1 is a good choice for the benchmarks.

scripts/systems.py
```diff
@@ -217,12 +217,19 @@
 #SBATCH --no-requeue
 
 # Env
+
```
msimberg, line 220: Nit:
```suggestion
```

scripts/systems.py
```diff
@@ -198,8 +198,8 @@
 """,
 }
 
-cscs["santis"] = {
-    "Cores": 288,
+cscs["alps-gh200"] = {
+    "Cores": 256,
```
msimberg, line 202: Keep this:
```suggestion
    "Cores": 288,
```
?

We can set `PIKA_THREADS=64` int the env vars?

scripts/systems.py
```diff
@@ -232,6 +239,8 @@
 """,
 }
 
+cscs["alps-daint"] = cscs["alps-gh200"].copy()
+
```
msimberg, line 243: Just thinking out loud: should we either add
```suggestion
cscs["alps-daint"] = cscs["alps-gh200"].copy()
cscs["alps-clariden"] = cscs["alps-gh200"].copy()
cscs["alps-santis"] = cscs["alps-gh200"].copy()
```
or just remove this copy altogether:
```suggestion
```
? We can always just use `alps-gh200` as the "system" in scripts. Or is this copy meant as an example, saying "if you want to modify this, make a copy, and override options here"? I will most likely just modify `alps-gh200` directly locally instead (if I need to make changes at all), but not everyone has to do that.

msimberg (commented):
It looks good to me already, but left some comments to think about. I'm happy if it's merged as-is as well.

scripts/systems.py
albestro: The idea was to have the ability to have different presets for the different clusters, so the copy was just there, as you said, as example.

But I agree, I did the same thought as you: currently most of the things that I would change are in a blob (the string with the header of the batch script), and it is more convenient to change that in-place or anyway replace it completely.

In addition to the fact that this file act just as a reference and it is open to customisation and changes for the different use-cases, one thing I always (forget to) add is the `--uenv`, and I add it in-place modifying the original one.

So, I agree that we can keep just `alps-gh200` and change inline whatever is needed. In case in the future there will be more intricate use-cases, we will decide how to deal with that.

msimberg (approved):

rasolca (approved):

PR #1219, opened 2024-11-22
TFactor: Separate larft_gemv kernel and add a faster option

rasolca:
Note: currently still use the slow kernel.

Don't want to interfere with @albestro TFactor work.

PR #1256, opened 2025-01-08
TFactor: use optimised kernel

albestro:
In a joint effort to optimise TFactor, with #1219 a faster kernel for LARFT has been introduced, and here it is applied on top of #1214 which parallelises it.

TODO:
- [x] #1214 
- [x] ~think about best solution about how to either copy taus or provide them to `_fixtau`~ EDIT: tracked by https://github.com/eth-cscs/DLA-Future/issues/1282


rasolca (dismissed):
Looks good.
Request change is used only to avoid merge before the base is updated

msimberg (approved):

msimberg (approved):

albestro:
cscs-ci run

PR #1258, opened 2025-01-14
Patch gtest to terminate as soon as it detects that a `ScopedTrace` has migrated to another worker thread

msimberg:
See first commit message for more details.

I've added a temporary commit which explicitly adds more yields while holding a `ScopedTrace` to check that it actually works in CI.

I've opted for a local patch which I think is easily updated if we update the gtest tag. An alternative strategy could be to apply the patch through the spack package and I don't have a strong preference either way, other than me not (yet) knowing how to apply the patch with spack to something that is pulled only at the CMake configure stage. One benefit of patching gtest like this is that it will always apply, even if one uses a spack build-env which doesn't run CMake through spack.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -18,7 +18,9 @@ FetchContent_Declare(
   GIT_REPOSITORY https://github.com/google/googletest.git
   GIT_TAG v1.13.0
   # Patch for int to float conversion warning.
-  PATCH_COMMAND git cherry-pick -n 5cd81a7848203d3df6959c148eb91f1a4ff32c1d -m 1
+  PATCH_COMMAND
+    git reset --hard HEAD && git cherry-pick -n 5cd81a7848203d3df6959c148eb91f1a4ff32c1d -m 1 && git
+    apply "${CMAKE_CURRENT_SOURCE_DIR}/scoped_trace_terminate_on_thread_migrate.patch"
 )
 set(BUILD_GMOCK OFF CACHE INTERNAL "")
 set(INSTALL_GTEST OFF CACHE INTERNAL "")
--- a/external/scoped_trace_terminate_on_thread_migrate.patch
+++ b/external/scoped_trace_terminate_on_thread_migrate.patch
@@ -1,0 +1,76 @@
+diff --git a/googletest/include/gtest/gtest.h b/googletest/include/gtest/gtest.h
+index 3e452a50..c51dd607 100644
+--- a/googletest/include/gtest/gtest.h
++++ b/googletest/include/gtest/gtest.h
+@@ -58,6 +58,7 @@
+ #include <set>
+ #include <sstream>
+ #include <string>
++#include <thread>
+ #include <type_traits>
+ #include <vector>
+ 
+@@ -2056,15 +2057,18 @@ class GTEST_API_ ScopedTrace {
+   // Slow, but flexible.
+   template <typename T>
+   ScopedTrace(const char* file, int line, const T& message) {
++    id = std::this_thread::get_id();
+     PushTrace(file, line, (Message() << message).GetString());
+   }
+ 
+   // Optimize for some known types.
+   ScopedTrace(const char* file, int line, const char* message) {
++    id = std::this_thread::get_id();
+     PushTrace(file, line, message ? message : "(null)");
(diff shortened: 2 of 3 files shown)


external/scoped_trace_terminate_on_thread_migrate.patch
```diff
@@ -0,0 +1,148 @@
...
+     FILE* fileout = OpenFileForWriting(
+@@ -6838,6 +6841,13 @@ void ScopedTrace::PushTrace(const char* file, int line, std::string message) {
+ 
+ // Pops the info pushed by the c'tor.
+ ScopedTrace::~ScopedTrace() GTEST_LOCK_EXCLUDED_(&UnitTest::mutex_) {
++  if (id != std::this_thread::get_id()) {
++    std::cerr << "ScopedTrace was created and destroyed on different "
++                 "std::threads, terminating to avoid segfaults, hangs, or "
++                 "silent corruption. Are you using any pika functionality that "
++                 "may yield a task after creating the ScopedTrace?\n";
```
msimberg, line 143: The error message is here. Suggestions for how we can make this as useful/informative as possible are very welcome!

test/unit/eigensolver/test_bt_reduction_to_band.cpp
```diff
@@ -195,6 +195,7 @@ void testBackTransformationReductionToBand(SizeType m, SizeType n, SizeType mb,
   mat_c_h.waitLocalTiles();
   SCOPED_TRACE(::testing::Message() << "m = " << m << ", n = " << n << ", mb = " << mb << ", nb = " << nb
                                     << ", b = " << b);
+  pika::this_thread::yield();
```
msimberg, line 198: To do: remove these changes.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

external/CMakeLists.txt
```diff
@@ -18,7 +18,8 @@ FetchContent_Declare(
   GIT_REPOSITORY https://github.com/google/googletest.git
   GIT_TAG v1.13.0
   # Patch for int to float conversion warning.
-  PATCH_COMMAND git cherry-pick -n 5cd81a7848203d3df6959c148eb91f1a4ff32c1d -m 1
+  PATCH_COMMAND git cherry-pick -n 5cd81a7848203d3df6959c148eb91f1a4ff32c1d -m 1 && git apply
+                "${CMAKE_CURRENT_SOURCE_DIR}/scoped_trace_terminate_on_thread_migrate.patch"
```
msimberg, line 22: It seems like this only works once. The `PATCH_COMMAND` seems to be run every time CMake is run, and the second time the `git apply` fails. I think we can `git reset` and re-`apply`, but I'm wondering if there are other options?

msimberg: The cherry-pick was always there and works fine when repeated. The apply fails the second time because the patch doesn't apply anymore.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

external/CMakeLists.txt
msimberg: I now `git reset` every time before applying the patches. This seems to work, but is not particularly elegant.

msimberg:
The patch seems to be correctly applied now, but can't be tested because CI needs to be moved to alps.

rasolca (approved):
Looks good, but still open TODO

msimberg:
cscs-ci run

msimberg:
> Looks good, but still open TODO

Yep, I still want to confirm that it does the right thing in CI. Now that #1262 is merged we should get some results soon enough...

msimberg:
Seems to be working as intended: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8962660045#L2249. I've removed the temporary commit.

msimberg:
cscs-ci run

test/unit/eigensolver/test_bt_reduction_to_band.cpp
msimberg: Done.

PR #1232, opened 2024-12-06
ReductionToBand: Dynamic number of workers for Panel computation (bulk)

albestro:
Currently the number of workers available (per configuration) was used as is for spawning an equivalent number of workers for the panel computation, where each worker was in charge of working on a integral number of tiles.

When not enough tiles were available, some worker might end up participating to the synchronisation points, but without any work to do. This happens for sure with small matrices, but due to the iterative nature of the reduction to band algorithm, it was happening anyway also during "tail" iterations in bigger matrices.

With this PR, the configuration is acquired as maximum number of workers available, but now the algorithm ensures that just worker with at least 1 tile to work on get spawned.

TODO:
- [ ] set a different default in benchmarking scripts for GPU systems

Notes:
- minimum amount of work is set to 1 tile per worker (it might eventually become a config parameter as well); after some not very extensive benchmark, at least 2 tile per worker resulted in slightly worse performances
- took the chance also to rename to snake_case the configure getter

PR #1263, opened 2025-01-17
ReductionToBand: fix minimal amount of work per worker

albestro:
With #1232 it has been introduced a small performance regression. This was apparently due to the default value used for minimal amount of work that has been set to 2, while it was previously 1.

Moreover, this value was not even unified between local and distributed implementation.

This PR aims at fixing that.

TODO
- [ ] Run benchmarks
- [ ] Decide if it is worth adding minimal amount of work as proper DLAF parameter (i.e. expose it instead of having it hard-coded)

### DIFF
--- a/include/dlaf/eigensolver/internal/get_red2band_panel_nworkers.h
+++ b/include/dlaf/eigensolver/internal/get_red2band_panel_nworkers.h
@@ -19,6 +19,10 @@
 
 namespace dlaf::eigensolver::internal {
 
+inline size_t get_red2band_panel_worker_minwork() noexcept {
+  return 1;
+}
+
 inline size_t get_red2band_panel_nworkers() noexcept {
   // Note: precautionarily we leave at least 1 thread "free" to do other stuff (if possible)
   const std::size_t available_workers = pika::resource::get_thread_pool("default").get_os_thread_count();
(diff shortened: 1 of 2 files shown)


albestro:
reference vs dynamic variants

Green is the dynamic as it was with PR #1232, and it actually confirms results we have seen:
- better on local implementation
- worse on distributed implementation
- but with 16 nodes the lines cross each other, and on 32 nodes it is better

Now let's look at the red, which is dynamic with minimum of 2 tiles (now both local and distributed), it actually matches perfectly the distributed implementation, while the local run matches the yellow one. That's correct, indeed before this PR the minimal amount of work was erroneously set different on local and distributed implementations (respectively 1 and 2).

Let's look now at the yellow (dynamic with minimum amount of work 1) vs gray (reference, i.e. no dynamic policy): dynamic is overall slightly better, and it is confirmed that it strong scales (32 nodes) a little better.

Why it strong scales a bit better than reference? how I explain it myself is that in strong scaling with high number of nodes there is less work per rank, and there happens more frequently that the number of workers is actually reduced by the dynamic policy.

Looking at yellow vs red now, it seems confirmed that minimum amount of work set to 1 is better on this platform, but it would be interesting to see if they converge with a higher number of nodes.

include/dlaf/eigensolver/internal/get_red2band_panel_nworkers.h
```diff
@@ -19,6 +19,10 @@
 
 namespace dlaf::eigensolver::internal {
 
+inline size_t get_red2band_panel_worker_minwork() noexcept {
+  return 1;
```
msimberg, line 23: It seems to me it'd probably be worth it to make this a configuration option, the cost is not so high, but I won't block this PR on that.

msimberg (approved):

rasolca (approved):

include/dlaf/eigensolver/internal/get_red2band_panel_nworkers.h
albestro: fully agree. if also @rasolca I would go for adding it.

PR #1225, opened 2024-12-02
 CI move to ALPS (daint-gpu -> alps_gh200) 

rasolca:
Still open points:

- C_API slowness (can be moved to separate PR as now is barely enough to pass)
- MPI pool and MPI mode
  Quick test: 
     - 30+0 hanged BacktransformationBandToTridiagTestMC
     - 30+1, 31+0 and 31+1 went through
  Note: we are seeing some hangs on eiger. Didn't track them, but IIRC at least a couple of cases in the same backtransformation. 
- HDF5 -> separate PR

PR #1264, opened 2025-01-27
Disable tests for CUDA configurations in CI

msimberg:
We can't yet have #1225, so disable CUDA tests for now and leave the CUDA builds in place in CI.

### DIFF
--- a/ci/cuda/gcc11_codecov.yml
+++ b/ci/cuda/gcc11_codecov.yml
@@ -20,13 +20,13 @@ cuda gcc11 codecov build:
     DOCKERFILE: ci/docker/codecov.Dockerfile
     DLAF_IMAGE: $CSCS_REGISTRY_PATH/cuda-gcc11-codecov/dlaf:$CI_COMMIT_SHA
 
-cuda gcc11 codecov test:
-  extends: .run_common
-  variables:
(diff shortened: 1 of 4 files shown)


msimberg:
I've simply commented out the test stages with CUDA now. Let me know if you'd prefer completely deleting the blocks.

msimberg:
cscs-ci run

rasolca (approved):

msimberg:
There were two timeouts on `test_bt_reduction_to_band`:
- https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8962098059#L2271
- https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/8962503293#L2264

The eigensolver test times out after that as well, but my guess is that it's just a consequence of MPI going into an inconsistent state (the test hangs in the beginning).

I think the `test_bt_reduction_to_band` tests timing out is still the same old scoped trace issue, though without segfaults in these particular instances. #1258 and #838 will help diagnose and improve, respectively, the situation. We still need pika 0.31.0 to fix the last (known) source of yielding, but those two PRs will already help quite a bit.

msimberg:
All remaining tests passed after one rerun.

PR #1268, opened 2025-02-05
Restart DLA-Future and the pika runtime less frequently in C API tests

msimberg:


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
This does help quite a bit. In the last commit (https://github.com/eth-cscs/DLA-Future/pull/1268/commits/26abebab9ae876236c062eb61e496a06f891986d) I only start and stop the runtime once per google-test. In the previous commit (https://github.com/eth-cscs/DLA-Future/pull/1268/commits/70a69d9f650c2addd5db82a44c74af7815fa9821) I skip starting and stopping for the innermost loop but keep the others (it's still a good stress test, though not very realistic).

My proposal would be to drop the last commit so we start and stop a bit less often, but still fairly often (this already gives quite a nice speedup).

msimberg:
I've rebased this on master instead of #1225, since it's not a requirement.

RMeli:
Starting/stopping once per test sounds good to me (https://github.com/eth-cscs/DLA-Future/commit/26abebab9ae876236c062eb61e496a06f891986d). I think at some point we already used this approach to avoid a memory leak in `pika` (see https://github.com/eth-cscs/DLA-Future/pull/886#issuecomment-1639855081), but I don't recall why we decided to to start and stop the runtime at every test once the leak was fixed.

msimberg:
cscs-ci run

msimberg:
> Starting/stopping once per test sounds good to me ([26abeba](https://github.com/eth-cscs/DLA-Future/commit/26abebab9ae876236c062eb61e496a06f891986d)).

I've updated both eigensolver tests to restart DLAF/pika once per grid (we need the grid for init).


> I think at some point we already used this approach to avoid a memory leak in `pika` (see [#886 (comment)](https://github.com/eth-cscs/DLA-Future/pull/886#issuecomment-1639855081)), but I don't recall why we decided to to start and stop the runtime at every test once the leak was fixed.

I don't remember for sure either, but I think it was to have a regression/unit test for the kind of issues that that mode exposed (memory leaks and other issues from restarting everything a lot of times), but I think we didn't consider the test time implications of it.

RMeli (approved):

rasolca (approved):

PR #1269, opened 2025-02-06
Use slightly fewer threads for pika in CI

msimberg:
Requires pika 0.32.0 for the `PIKA_THREADS` environment variable to have an effect.

### DIFF
--- a/ci/mpi-ctest
+++ b/ci/mpi-ctest
@@ -17,6 +17,9 @@ then
     echo "Rank $SLURM_PROCID: set CUDA_VISIBLE_DEVICE=$CUDA_VISIBLE_DEVICE"
 fi
 
+# Reserve one core for non-pika threads
+export PIKA_THREADS=$((SLURM_CPUS_PER_TASK - 1))
+
 # Run the tests, only output on the first rank
 if [[ $SLURM_PROCID == "0" ]]; then
     TZ=CET date +"Run started at: %H:%M:%S %z"


msimberg:
cscs-ci run

msimberg:
Leaving one core free for other threads in this PR also speeds up the C API tests quite a bit (in addition to #1268). Setting `PIKA_THREADS` requires pika 0.32.0, so it won't apply to all CI configurations (we want to keep one on the oldest supported version 0.30.X), but we could use 0.32.0 on the CUDA configurations where there are more tests to run.

RMeli:
>  Setting PIKA_THREADS requires pika 0.32.0, so it won't apply to all CI configurations

Ugly, but can't you still use the workaround `PIKA_COMMANDLINE_OPTIONS="--pika:threads=..." which works with earlier versions too?

msimberg:
> > Setting PIKA_THREADS requires pika 0.32.0, so it won't apply to all CI configurations
> 
> Ugly, but can't you still use the workaround `PIKA_COMMANDLINE_OPTIONS="--pika:threads=..." which works with earlier versions too?

In theory that should've been possible, but the real command line options and those supplied by `PIKA_COMMANDLINE_OPTIONS` do not mix well... (overriding each other, but not quite).

msimberg:
cscs-ci run

msimberg:
This doesn't seem to be making much of a difference on the C API tests anymore now that DLAF and pika are restarted much less frequently after #1268. Closing.

PR #1270, opened 2025-02-11
Rename parameters from "nworkers" to "num-threads"

albestro:
As per discussion in the last meeting, we are doing this breaking change on parameter names:

- `dlaf:red2band-panel-nworkers` becomes `dlaf:red2band-panel-num-threads`
- `dlaf:tridiag-rank1-nworkers` becomes `dlaf:tridiag-rank1-num-threads`

This change applies also to related environment variables names.

### DIFF
--- a/include/dlaf/eigensolver/internal/get_red2band_panel_nworkers.h
+++ b/include/dlaf/eigensolver/internal/get_red2band_panel_nworkers.h
@@ -23,13 +23,13 @@ inline size_t get_red2band_panel_worker_minwork() noexcept {
   return 1;
 }
 
-inline size_t get_red2band_panel_nworkers() noexcept {
+inline size_t get_red2band_panel_num_workers() noexcept {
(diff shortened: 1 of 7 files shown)


msimberg (approved):

RMeli (commented):
Wouldn't be better if the `get_*` functions are called in the same way as the parameter (e.g. `get_red2band_panel_num_threads` instead of `get_red2band_panel_num_workers`)?

albestro:
> Wouldn't be better if the `get_*` functions are called in the same way as the parameter (e.g. `get_red2band_panel_num_threads` instead of `get_red2band_panel_num_workers`)?

I thought about that (and with that also the variables): my opinion is that algorithmically makes sense (to me) that I consider them "workers" independently from the backend, i.e. begin them threads or streams.

rasolca (approved):

PR #1278, opened 2025-02-12
Update pika and enable threads check in `cuda gcc13 debug scalapack` CI configuration

### LINKED ISSUE #1273: `cuda gcc13 debug scalapack` is still using pika 0.30.1 and has check threads disabled.

msimberg:
Fix #1273.

### DIFF
--- a/ci/docker/debug-cuda-gh200-scalapack.yaml
+++ b/ci/docker/debug-cuda-gh200-scalapack.yaml
@@ -35,6 +35,6 @@ spack:
         - '+fortran'
     pika:
       require:
-        - '@0.30.1'
+        - '@0.32.0'
         - 'build_type=Debug'
         - 'malloc=system'


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
Ready for review, but same situation as https://github.com/eth-cscs/DLA-Future/pull/1279 regarding #1292.

ci/docker/debug-cuda-gh200-scalapack.yaml
```diff
@@ -35,6 +35,5 @@ spack:
         - '+fortran'
     pika:
       require:
-        - '@0.30.1'
```
rasolca, line 38: Should be set to 0.32.0 as to keep a CI pipeline with an older pika version when we upgrade.

rasolca:
thread checks should be re-enabled too.

ci/docker/debug-cuda-gh200-scalapack.yaml
msimberg: Hmm, good point. Though given that my current hypothesis that pika 0.32.0 and 0.30.1 are equally good/bad, should we just keep this configuration unchanged then? And keep the threads check disabled? In that case I can close this PR...

rasolca: I would prefer to have thread checks enabled for all pipelines.
(Otherwise we might forget to re-enable them in the future)

msimberg: Fair enough! I fixed it to 0.32.0 now, and reenabled the threads check.

msimberg:
cscs-ci run

rasolca:
After a discussion with @msimberg, we found out that `ci_check_threads` never worked with cuda builds (as cuda library spawns some threads too)

My suggestion here is now:
- keep pika 0.30.1 (as it works as well)
- remove "~ci_check_threads" as anyway it can't be enabled
In a different PR:
- add conflict with cuda and rocm

rasolca:
Reverting back to 0.30.1 leaves no chancges for this PR.

PR #1289, opened 2025-02-19
Avoid buffering in CI output.

rasolca:


rasolca:
cscs-ci run

ci/ctest_to_gitlab.sh
```diff
@@ -93,7 +93,7 @@ JOB_TEMPLATE="
     USE_MPI: 'YES'
     DISABLE_AFTER_SCRIPT: 'YES'
     DLAF_HDF5_TEST_OUTPUT_PATH: /dev/shm
-  script: mpi-ctest -L {{CATEGORY_LABEL}} -L {{RANK_LABEL}}
+  script: stdbuf -oL -eL mpi-ctest -L {{CATEGORY_LABEL}} -L {{RANK_LABEL}}
```
albestro, line 96: out of curiosity: why do you prefer line-buffered rather than no buffer at all (with mode 0)?

msimberg, line 96: Purely subjective, and obviously not the case e.g. for `mpi-ctest`, but I generally prefer the long options:
```suggestion
  script: stdbuf --output=L --error=L mpi-ctest -L {{CATEGORY_LABEL}} -L {{RANK_LABEL}}
```

msimberg (approved):

ci/ctest_to_gitlab.sh
rasolca: We use grep inside the script to remove the lines we don't need, the output is anyway line buffered.

msimberg (approved):

PR #1290, opened 2025-02-19
Adjust CI timeouts

rasolca:
Last rebuild of deps images took less than 25 min for all cases, therefore a timeout of 45 mins looks appropriate.

The same holds for build images (except for rocm builds where the slowest takes ~45 mins -> timeout of 1h30m)

### DIFF
--- a/ci/common-ci.yml
+++ b/ci/common-ci.yml
@@ -72,7 +72,7 @@ variables:
     - .build_deps_common_base
   variables:
     COMMON_SPACK_ENVIRONMENT: ci/docker/common-gh200.yaml
-    SLURM_TIMELIMIT: 00:45:00
+    SLURM_TIMELIMIT: '00:45:00'
 
 ## BUILD DLAF
 
@@ -115,7 +115,7 @@ variables:
     - .build_common_base
   variables:
     DLAF_LD_PRELOAD: "/lib/aarch64-linux-gnu/libSegFault.so"
-    SLURM_TIMELIMIT: 00:45:00
+    SLURM_TIMELIMIT: '00:45:00'
 
 .build_for_eiger:
   variables:


rasolca:
cscs-ci run

ci/rocm/clang14_release.yml
```diff
@@ -14,6 +14,7 @@ rocm clang14 release deps:
 rocm clang14 release build:
   extends:
     - .build_common
+  timeout: 1 hour 30 minutes
```
msimberg, line 17: I think this might be too tight for ROCm: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/9186534693.

ci/common-ci.yml
```diff
@@ -76,7 +76,7 @@ variables:
 
 .build_common_base:
   stage: build
-  timeout: 2 hours
+  timeout: 45 minutes
```
msimberg, line 79: This is in theory ok, but for GH200 the timeout includes the slurm queuing time, so this could get tight in some situations (GitLab timeout is 45 minutes, and the slurm timelimit is set to 45 minutes based on the GitLab timeout).

I think since this time is adjusted based on the actual build time, I would override this for the GH200 configurations with `SLURM_TIMELIMIT: 00:45:00` and a timeout of e.g. 8 hours or whatever the default GitLab timeout might be.

For the x86 builds that happen on rosa this isn't really a problem since (most of) the time is actually dedicated to the actual build job, modulo runner startup/pulling images etc.

msimberg (changes requested):

rasolca:
cscs-ci run

rasolca:
cscs-ci run

rasolca:
cscs-ci run

ci/rocm/clang14_release.yml
rasolca: Anyway fixed by #1291 😉

ci/common-ci.yml
rasolca: See https://github.com/eth-cscs/DLA-Future/pull/1290/commits/5374d09f6dd121d6bdf546ea62c11617baf786d5

msimberg (approved):

msimberg (approved):

PR #1291, opened 2025-02-20
Update rocm pipeline to speed up CI

rasolca:
Update non stdexec rocm pipeline to rocm602.

With rocm 5.3.3 it is the slowest and building DLAF requires more than 1h.

With rocm 6.0.2 it is down to 12 minutes.


rasolca:
cscs-ci run

PR #1292, opened 2025-02-20
Create new task for GPU work to avoid potential stack overflows

msimberg:
# Overview

pika's cuda scheduler does not actually start a new task for scheduling work and thus runs the risk of stack overflows. This PR adds a transfer to a new task on the default thread pool before launching GPU work to reduce the risk.

## To do

I still have to verify that there's no significant performance degradation from this, hence a draft, though a small degradation would be acceptable to me given that this is about correctness. If there's a larger degradation I'll explore targeting only the highest risk of stack overflows directly, instead of applying this to all GPU work.

# Issue details

This was (eventually) quite easily reproducible with high failure rate on `test_eigensolver`, `test_gen_eigensolver` and `test_reduction_to_band` on GH200. `reduction_to_band` seems to be the main trigger, which can then make `band_to_tridiag` fail in surprising ways when run as part of the eigensolver. pika should be compiled in release or relwithdebinfo, since the small stack size is 0x10000 (64 KiB) in non-debug builds. In debug mode the default stack size is 128 KiB, which significantly reduces the probability of this test failing. (Note that the stack size can't be set to 64 KiB in debug mode; I want to change this in pika.)

The typical stack trace from `test_reduction_to_band` on failure looks like follows:
```
#0  0x000040003ee074a8 in ?? () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublasLt.so.12
#1  0x000040003ee11a6c in ?? () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublasLt.so.12
#2  0x000040003ee46bec in ?? () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublasLt.so.12
#3  0x000040003eea1628 in ?? () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublasLt.so.12
#4  0x000040003eecff54 in cublasLtSSSMatmulAlgoGetHeuristic () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublasLt.so.12
#5  0x000040003808a0c0 in ?? () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublas.so.12
#6  0x000040003808a558 in ?? () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublas.so.12
#7  0x000040003808c0bc in ?? () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublas.so.12
#8  0x000040003816b8b0 in ?? () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublas.so.12
#9  0x0000400038015098 in cublasSsyr2k_v2 () from /user-environment/linux-sles15-neoverse_v2/gcc-13.3.0/cuda-12.6.3-mxamyoygvaiiuc4kefyfppnre5j2siev/lib64/libcublas.so.12
#10 0x0000400020dd7a94 in dlaf::gpublas::internal::Her2k<float>::call<cublasContext*&, cublasFillMode_t, cublasOperation_t, int, int, float const*, float const*, int, float const*, int, float const*, float*,
 int> ()
    at /capstor/scratch/cscs/simbergm/src/DLA-Future/include/dlaf/gpu/blas/gpublas.h:130
#11 0x0000400020ddcf4c in dlaf::tile::internal::her2k<float> (handle=<optimized out>, uplo=blas::Uplo::Lower, op=blas::Op::NoTrans, alpha=<optimized out>, a=..., b=..., beta=<optimized out>, c=...) at /capst
or/scratch/cscs/simbergm/src/DLA-Future/include/dlaf/blas/tile.h:287
#12 0x0000400020ddd014 in dlaf::tile::internal::her2k_t::operator()<cublasContext*, blas::Uplo&, blas::Op&, float&, dlaf::matrix::Tile<float const, (dlaf::Device)1> const&, dlaf::matrix::Tile<float const, (d
laf::Device)1> const&, float&, dlaf::matrix::Tile<float, (dlaf::Device)1>&>
    (this=0x44b66321c58) at /capstor/scratch/cscs/simbergm/src/DLA-Future/include/dlaf/blas/tile.h:354
#13 dlaf::common::internal::Unwrapping<dlaf::tile::internal::her2k_t>::operator()<cublasContext*, blas::Uplo&, blas::Op&, float&, pika::execution::experimental::async_rw_mutex_access_wrapper<dlaf::matrix::Ti
le<float, (dlaf::Device)1>, dlaf::matrix::Tile<float const, (dlaf::Device)1> const, (pika::execution::experimental::async_rw_mutex_access_type)0>&, pika::execution::experimental::async_rw_mutex_access_wrappe
r<dlaf::matrix::Tile<float, (dlaf::Device)1>, dlaf::matrix::Tile<float const, (dlaf::Device)1> const, (pika::execution::experimental::async_rw_mutex_access_type)0>&, float&, dlaf::matrix::Tile<float, (dlaf::
Device)1>&>(cublasContext*&&, blas::Uplo&, blas::Op&, float&, pika::execution::experimental::async_rw_mutex_access_wrapper<dlaf::matrix::Tile<float, (dlaf::Device)1>, dlaf::matrix::Tile<float const, (dlaf::D
evice)1> const, (pika::execution::experimental::async_rw_mutex_access_type)0>&, pika::execution::experimental::async_rw_mutex_access_wrapper<dlaf::matrix::Tile<float, (dlaf::Device)1>, dlaf::matrix::Tile<flo
at const, (dlaf::Device)1> const, (pika::execution::experimental::async_rw_mutex_access_type)0>&, float&, dlaf::matrix::Tile<float, (dlaf::Device)1>&) & (this=this@entry=0x44bb172c618) at /capstor/scratch/cs
cs/simbergm/src/DLA-Future/include/dlaf/common/unwrap.h:69
[... snip ...]
#328 0x0000400023943574 in _ZN4pika4util6detail15callable_vtableIFSt4pairINS_7threads6detail21thread_schedule_stateENS5_9thread_idEENS5_20thread_restart_stateEEE7_invokeINS5_23thread_function_nullaryIZZNRS_9execution12experimental21thread_pool_scheduler15operation_stateISG_NS_20schedule_from_detail15operation_stateINS_13when_all_impl15when_all_senderIJNSF_17unique_any_senderIJNSF_29async_rw_mutex_access_wrapperIN4dlaf4comm12CommunicatorEKSQ_LNSF_26async_rw_mutex_access_typeE1EEEEEENS_11just_detail16just_sender_implINS1_6pack_cImJLm0EEEEJSt17reference_wrapperIKNSO_6matrix4TileIfLNSO_6DeviceE0EEEEEE11just_senderEEEESG_NS_16let_value_detail15operation_stateINSI_20schedule_from_senderIS18_SG_EENSF_6detail12any_receiverIJEEEZNS_3mpi12experimental19tag_fallback_invokeIS18_NSO_6common8internal14ConsumeRvaluesINSP_8internal13MPICallHelperINS1M_11sendBcast_tEEEEELi42ELi0EEENSM_IJEEENS1H_15transform_mpi_tEOT_OT0_EUlDpRT_E_E30let_value_predecessor_receiverEE25scheduler_sender_receiverEE5startEvENKUlvE_clEvEUlvE_EEEES8_PvOS9_ (f=<optimized out>, vs#0=@0x4002f0a5ff67: pika::threads::detail::thread_restart_state::signaled)
    at /capstor/scratch/cscs/simbergm/spack-configs/ba4a847310e8b6b9cac7467047d46cd7/opt/linux-sles15-neoverse_v2/gcc-13.3.0/pika-main-mxqchlyzpb5guubvguj5c77defeqxixo/include/pika/functional/detail/vtable/callable_vtable.hpp:72
#329 0x000040002bc109bc in pika::util::detail::basic_function<std::pair<pika::threads::detail::thread_schedule_state, pika::threads::detail::thread_id> (pika::threads::detail::thread_restart_state), false>::operator()(pika::threads::detail::thread_restart_state) const (
    vs#0=<optimized out>, this=0x44bb59f3458) at /capstor/scratch/cscs/simbergm/src/pika-worktrees/dlaf-eigensolver-hang-debug/libs/pika/functional/include/pika/functional/detail/basic_function.hpp:199
#330 pika::threads::coroutines::detail::coroutine_impl::operator() (this=0x44bb59f1050) at /capstor/scratch/cscs/simbergm/src/pika-worktrees/dlaf-eigensolver-hang-debug/libs/pika/coroutines/src/detail/coroutine_impl.cpp:71
#331 0x000040002bc0ff58 in pika::threads::coroutines::detail::posix::trampoline<pika::threads::coroutines::detail::coroutine_impl> (fun=<optimized out>)
    at /capstor/scratch/cscs/simbergm/src/pika-worktrees/dlaf-eigensolver-hang-debug/libs/pika/coroutines/include/pika/coroutines/detail/posix_utility.hpp:276
#332 0x0000400072ce7b50 in ?? () from /lib64/libc.so.6
```

The stack trace is deep, but not ridiculously deep (this is with `-Og` so only some optimizations applied). The stack required to get to the cuBLAS call is ~5 KiB. The cuBLAS call itself uses about 61 KiB stack and eventually leads to the stack overflow. It's not clear what cuBLAS needs that stack space for. Setting `CUBLASLT_HEURISTICS_CACHE_CAPACITY=0` (https://docs.nvidia.com/cuda/cublas/#heuristics-cache) does _not_ affect the behaviour.

Another failure mode is a segfault within `band_to_tridiag` (notifying a condition variable, or a null thread id) when called as part of the eigensolver in `test_eigensolver`. As far as I can tell, this is just one of the tasks from `reduction_to_band` writing into the stack of a task from `band_to_tridiag`. With `PIKA_USE_GUARD_PAGES=1` (which adds a read-only page at the end of the stack) this second failure mode seems to go away (and only the first failure mode happens). I also have not been able to reproduce this second failure mode only with `test_band_to_tridiag`.

In summary, this seems to have been the main cause of failures on the CUDA CI pipelines. My guess is also that on P100 cuBLAS may have different heuristics and maybe doesn't use as much stack space? It's possible there are more issues, but I'll deal with them if/when I have more information.

We also have to keep an eye out for if cuBLAS sometimes needs more than 64 KiB stack space for itself. In that situation obviously a new task isn't enough either, and we have to use bigger stacks.

Finally, I'm going to improve pika's reporting of stack overflows, which is a bit wonky at the moment (and one of the reasons it was so hard to identify this as a stack overflow).


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
In #1293 I've increased the stack sizes in CI as a temporary(?) workaround for testing at least.

msimberg:
I tried using high priority tasks (https://github.com/eth-cscs/DLA-Future/commit/e732a68e9af92d9753f4d5019237172f9c541416), and high priority+stackless tasks (https://github.com/eth-cscs/DLA-Future/commit/82fa469474a6378bafd32587c075d735e719050f) and the performance difference is largely gone. Generalized eigensolver results below (note that I've left out the 512 blocksize to not make the plots too messy; the colors also don't match the previous plots):

gevp_strong_time_10240_0-10240_1024_128
gevp_strong_time_20480_0-20480_1024_128
gevp_strong_time_30097_0-30097_1024_128
gevp_strong_time_40960_0-40960_1024_128

Both high priority tasks and high priority + stackless tasks perform within noise of current master. I think high priority + stackless tasks would be quite a reasonable solution to this issue. We get both:
- new tasks to avoid launching CUDA work inline as continuations from MPI work (depending on MPI mode of course)
- stackless threads mean that the risk of stack overflows is ~0% even if cuBLAS decides to use > 64 KiB stack space for something

In terms of sanity/consistency/safety this is a change that would've made sense to make a long time ago. This way `dlaf::internal::transform` consistently always creates a new task. I think the worry was always if not doing inline continuations would have an impact on performance, but I think this shows that it doesn't (with the right settings).

What do you think? @biddisco thoughts?

msimberg:
cscs-ci run

msimberg:
cscs-ci run

rasolca (approved):

PR #1293, opened 2025-02-20
Use bigger stacks and guard pages in CI

msimberg:
The reporting of stack overflows with guard pages in pika is still severely lacking (meaning you might still just get an error saying "segfault" or even worse nothing), but enabling them in CI should make failures a bit more predictable since stack overflows are less likely to write into other tasks' stacks (and should instead "only" trigger segfaults when trying to use the guard page).

### DIFF
--- a/ci/common-ci.yml
+++ b/ci/common-ci.yml
@@ -136,3 +136,5 @@ variables:
     strategy: depend
     forward:
       pipeline_variables: true
+  variables:
+    PIKA_USE_GUARD_PAGES: 1
--- a/ci/common-ci.yml
+++ b/ci/common-ci.yml
@@ -137,4 +137,6 @@ variables:
     forward:
       pipeline_variables: true
   variables:
+    PIKA_SMALL_STACK_SIZE: 0x40000
+    PIKA_MEDIUM_STACK_SIZE: 0x40000
     PIKA_USE_GUARD_PAGES: 1


msimberg:
cscs-ci run

rasolca (approved):
Just a quick question for consistency:
Where it is the best place to put test variables? common-ci.yaml or ctest-to-gitlab.sh?

msimberg:
> Just a quick question for consistency: Where it is the best place to put test variables? common-ci.yaml or ctest-to-gitlab.sh?

Good question... I suppose the latter would fit better for consistency, but not everything is there. I like the former more for visibility (it's more "configuration" whereas `ctest-to-gitlab.sh` feels more hidden), but this argument doesn't feel so strong if some things are anyway in `ctest-to-gitlab.sh`.

I'm not sure what's best. Do you have a preference?

msimberg:
> Anyway, can we temporarily increase the stack size as well until we merge #1292?

Yes, for sure. I'll post some plots on #1292 soon, but I think the performance hit of my first approach is a bit too big.

I'll add the stack size here and then if @albestro, @RMeli have preferences on where to put the common options I'll move them.

ci/common-ci.yml
```diff
@@ -136,3 +136,7 @@ variables:
     strategy: depend
     forward:
       pipeline_variables: true
+  variables:
+    PIKA_SMALL_STACK_SIZE: 0x40000
+    PIKA_MEDIUM_STACK_SIZE: 0x40000
```
msimberg, line 141: The defaults are
- small stack size in release: 0x10000 (64 KiB)
- small stack size in debug: 0x20000 (128 KiB)
- medium stack size: 0x20000 (128 KiB)

I'm making them all 256 KiB for now to leave a bit of margin (large is way bigger, so not touching that). If we see segfaults with these sizes it's more likely to be a "regular" segfault, or stack overflow of the more pathological kind (i.e. very deep recursion).

msimberg:
cscs-ci run

ci/common-ci.yml
msimberg: Note that I've applied the stack size change to _all_ CI configurations, since I'd like to see if this has any effect on the MC pipelines as well. I'm not quite keeping track of which builds/failures have happened with what fixes, but there have been recent failures on the MC pipelines and/or MC tests as well, and if they remain then it's likely not due to the stack size being too small.

PR #1294, opened 2025-02-20
Remove Umpire version constraint in CI

msimberg:
Not urgent, but I just noticed that we still have this version constraint in place, though it's no longer needed. All versions have been patched in spack, and the latest releases have it fixed without patches as well. I think it's good if we can test with the latest version that we would normally be using for benchmarks etc. as well.


msimberg:
cscs-ci run

rasolca (approved):

PR #1300, opened 2025-03-04
CMake MPI preset: update logic and extend documentation

albestro:
Changelog:
- replace `MPIEXEC_NUMCORES` with `MPIEXEC_NUMCORES_PER_RANK` (see details below)
- `MPIEXEC_MAX_NUMPROCS` is considered just for `plain-mpi`
- add missing doc and extend with information about `MPIEXEC_MAX_NUMPROCS`/`MPIEXEC_NUMCORES_PER_RANK` usage
- moved `dlaf_setup_mpi_preset` to DLAF_AddTest cmake script (it seems a more appropriate place)

Depending on the DLAF_MPI_PRESET we have two different behaviours.

|PRESET|"plain-mpi"<br/>`--pika:bind=none --pika:threads=N`<br/>with `N=`|"slurm"/"custom"<br/>`srun -n MPIRANKS ${MPIEXEC_NUMCORE_FLAG}=N`<br/>with `N=`|
|:-:|:-:|:-:|
|"before"|`MPIEXEC_MAX_NUMPROCS/MPIRANKS`|`MPIEXEC_NUMCORES/MPIRANKS`|
|"after"|`MPIEXEC_MAX_NUMPROCS/MPIRANKS`|`MPIEXEC_NUMCORES_PER_RANK`|

The behaviour will change just for `slurm`/`custom` preset.

Some notes/reasons for the change are:
- now it is possible to configure a test over more than 1 nodes (previously it was problematic, e.g. 6 ranks case on 2 nodes)
- by fixing the number of cores we rely on slurm (and the user providing a valid configuration value) to not have ranks across modules (e.g. 1 rank will not use all nodes resources, ending up using more modules with a single GPU)

`MPIEXEC_MAX_NUMPROCS` is used just for `plain-mpi` and not in other presets. I'm not sure how helpful would be to have setting it also for other presets and what kind of additional check we could add (e.g. some kind of upper bound check on configuration).


rasolca (approved):

albestro:
On ALPS Daint

```
  -DDLAF_MPI_PRESET=slurm\
  -DMPIEXEC_NUMCORES_PER_RANK=32\
  -DMPIEXEC_PREFLAGS="--gpus-per-task=1"\
```

(an extent of the full list, just for reference)
```
24: Test command: /capstor/scratch/cscs/ialberto/dla-future.develop/build/test/unit/common/test_single_threaded_blas
25: Test command: /usr/bin/srun "-n" "2" "-c" "32" "--gpus-per-task=1" "/capstor/scratch/cscs/ialberto/dla-future.develop/build/test/unit/communication/test_message"
26: Test command: /usr/bin/srun "-n" "4" "-c" "32" "--gpus-per-task=1" "/capstor/scratch/cscs/ialberto/dla-future.develop/build/test/unit/communication/test_communicator"
27: Test command: /usr/bin/srun "-n" "6" "-c" "32" "--gpus-per-task=1" "/capstor/scratch/cscs/ialberto/dla-future.develop/build/test/unit/communication/test_communicator_grid"
```

rasolca:
cscs-ci run

rasolca:
cscs-ci run

cmake/DLAF_AddTest.cmake
```diff
@@ -8,21 +8,125 @@
...
+  if(NOT DLAF_MPI_PRESET STREQUAL _DLAF_MPI_PRESET)
+
+    if(DLAF_MPI_PRESET STREQUAL "plain-mpi")
+      # set(MPIEXEC_EXECUTABLE "" CACHE STRING "Executable for running MPI programs")
+      # set(MPIEXEC_NUMPROC_FLAG "" CACHE STRING "Flag used by MPI to specify the number of processes for mpiexec; the next option will be the number of processes." FORCE)
+      set(MPIEXEC_NUMCORE_FLAG
+          ""
+          CACHE
+            STRING
+            "Flag used by MPI to specify the number of cores per rank for mpiexec. If not empty, you have to specify also the number of cores available per node in MPIEXEC_NUMCORES_PER_RANK."
```
RMeli, line 24:
```suggestion
            "Flag used by MPI to specify the number of cores per rank for mpiexec. If not empty, you have to specify also the number of cores available per rank in MPIEXEC_NUMCORES_PER_RANK."
```

cmake/DLAF_AddTest.cmake
```diff
@@ -8,21 +8,125 @@
...
+      if(DLAF_TEST_THREAD_BINDING_ENABLED)
+        message(WARNING "Disabling pika binding")
+        set(DLAF_TEST_THREAD_BINDING_ENABLED FALSE CACHE BOOL "" FORCE)
+      endif()
+
+    elseif(DLAF_MPI_PRESET STREQUAL "slurm")
+      if(NOT DLAF_TEST_THREAD_BINDING_ENABLED)
+        message(
+          WARNING "When using DLAF_MPI_PRESET=slurm DLAF_TEST_THREAD_BINDING_ENABLED should be enabled. "
+                  "It is currently disabled and you may incur in performance drop. "
```
RMeli, line 37:
```suggestion
                  "It is currently disabled and you may incur in a performance drop. "
```

cmake/DLAF_AddTest.cmake
```diff
@@ -8,21 +8,125 @@
...
+            FORCE
+      )
+      set(MPIEXEC_NUMCORE_FLAG
+          "-c"
+          CACHE
+            STRING
+            "Flag used by MPI to specify the number of cores per rank for mpiexec. If not empty, you have to specify also the number of cores available per node in MPIEXEC_NUMCORES_PER_RANK."
+            FORCE
+      )
+      set(MPIEXEC_NUMCORES_PER_RANK "1" CACHE STRING "Number of cores used by each MPI rank.")
```
RMeli, line 62: Is this a good default?

cmake/DLAF_AddTest.cmake
```diff
@@ -8,21 +8,125 @@
...
+          ""
+          CACHE
+            STRING
+            "Flag used by MPI to specify the number of processes for mpiexec; the next option will be the number of processes."
+      )
+      set(MPIEXEC_NUMCORE_FLAG
+          ""
+          CACHE
+            STRING
+            "Flag used by MPI to specify the number of cores per rank for mpiexec. If not empty, you have to specify also the number of cores available per node in MPIEXEC_NUMCORES_PER_RANK."
```
RMeli, line 76:
```suggestion
            "Flag used by MPI to specify the number of cores per rank for mpiexec. If not empty, you have to specify also the number of cores available per rank in MPIEXEC_NUMCORES_PER_RANK."
```

cmake/DLAF_AddTest.cmake
```diff
@@ -8,21 +8,125 @@
...
-#   [LIBRARIES <arguments for target_link_libraries>]
+set(DLAF_PIKATEST_EXTRA_ARGS "" CACHE STRING "Extra arguments for tests with pika")
+
+macro(dlaf_setup_mpi_preset)
+  # if a preset has been selected and it has been changed from previous configurations
+  if(NOT DLAF_MPI_PRESET STREQUAL _DLAF_MPI_PRESET)
+
+    if(DLAF_MPI_PRESET STREQUAL "plain-mpi")
+      # set(MPIEXEC_EXECUTABLE "" CACHE STRING "Executable for running MPI programs")
+      # set(MPIEXEC_NUMPROC_FLAG "" CACHE STRING "Flag used by MPI to specify the number of processes for mpiexec; the next option will be the number of processes." FORCE)
```
RMeli, line 19:
```suggestion
```
Remove commented out code?

cmake/DLAF_AddTest.cmake
```diff
@@ -8,21 +8,125 @@
...
+      set(MPIEXEC_NUMCORES_PER_RANK "" CACHE STRING "Number of cores used by each MPI rank.")
+
+    else()
+      message(FATAL_ERROR "Preset ${DLAF_MPI_PRESET} is not supported")
+    endif()
+
+    message(STATUS "MPI preset: ${DLAF_MPI_PRESET}")
+
+    # make mpi preset selection persistent (with the aim to not overwrite each time, user may have changed some values (see custom)
+    set(_DLAF_MPI_PRESET ${DLAF_MPI_PRESET} CACHE INTERNAL "Store what preset is being used")
```
RMeli, line 87:
```suggestion
    set(_DLAF_MPI_PRESET ${DLAF_MPI_PRESET} CACHE INTERNAL "Store what MPI preset is being used")
```

cmake/DLAF_AddTest.cmake
```diff
@@ -33,33 +137,32 @@
 #   - MPIPIKA: uses a main that initializes both pika and MPI
 # If not specified, no external main is used and it should exist in the test source code.
 #
-# Moreover, the variable DLAF_PIKATEST_EXTRA_ARGS can be used to pass extra arguments that will
-# be given to all tests involving PIKA (i.e. USE_MAIN=PIKA or USE_MAIN=MPIPIKA).
+# Moreover, there are a few variables to control the behavior:
+# - DLAF_PIKATEST_EXTRA_ARGS can be used to pass extra arguments that will be given to all tests involving PIKA (i.e. USE_MAIN=PIKA or USE_MAIN=MPIPIKA).
+# - MPIEXEC_MAX_NUMPROCS is the maximum number of ranks that could be run
+# - MPIEXEC_NUMCORE_FLAG can be set to mpi runner flag that controls number of cores per rank (see MPIEXEC_NUMCORES_PER_RANK).
```
RMeli, line 143:
```suggestion
# - MPIEXEC_NUMCORE_FLAG can be set to the MPI runner flag that controls the number of cores per rank (see MPIEXEC_NUMCORES_PER_RANK).
```

cmake/DLAF_AddTest.cmake
```diff
@@ -33,33 +137,32 @@
...
+# - MPIEXEC_NUMCORES_PER_RANK can be set to number of cores to assign to each MPI rank (default=1)
+#
+# About aforementioned MPIEXEC_* variables, their usage depends on the DLAF_MPI_PRESET value:
```
RMeli, line 146:
```suggestion
# The usage of the aforementioned MPIEXEC_* variables depends on the DLAF_MPI_PRESET value:
```

cmake/DLAF_AddTest.cmake
```diff
@@ -33,33 +137,32 @@
...
+# - With {"plain-mpi"}
+#   MPIEXEC_MAX_NUMPROCS is considered, and the number of processors is distributed evenly over MPIRANKS (as --pika:threads)
+# - With {"slurm", "custom"}
+#   MPIEXEC_NUMCORES_PER_RANK is used to set the number of cores for each rank.
+#   It can be set just if MPIEXEC_NUMCORE_FLAG is set, otherwise the information is skipped.
```
RMeli, line 151:
```suggestion
#   It can be set only if MPIEXEC_NUMCORE_FLAG is set, otherwise the information is skipped.
```
?

cmake/DLAF_AddTest.cmake
```diff
@@ -8,21 +8,125 @@
...
+          CACHE
+            STRING
+            "Flag used by MPI to specify the number of processes for mpiexec; the next option will be the number of processes."
+            FORCE
+      )
+      set(MPIEXEC_NUMCORE_FLAG
+          "-c"
+          CACHE
+            STRING
+            "Flag used by MPI to specify the number of cores per rank for mpiexec. If not empty, you have to specify also the number of cores available per node in MPIEXEC_NUMCORES_PER_RANK."
```
msimberg, line 59:
```suggestion
            "Flag used by MPI to specify the number of cores per rank for mpiexec. If not empty, you have to specify also the number of cores available per rank in MPIEXEC_NUMCORES_PER_RANK."
```
as in the other docstring. But also, any way we can combine the two definitions?

albestro: It's a good point that I totally agree with. Unfortunately using CMake CACHE variables means that we have to `FORCE` set them if we want to change the value. But changing the value also implies setting also the related docstring (AFAIK).

I think the only way might be storing docstring in variables. Do you see any alternative? would you go for variables in order to not be error-prone and risking ending up with same variable with different docstring?

cmake/DLAF_AddTest.cmake
albestro: I'm ok with that, it was just a way of "self-documenting' that for this preset that's the "default" and how it gets configured.

Let me know if you still want to delete it or not 😉

msimberg (approved):

rasolca:
cscs-ci run

cmake/DLAF_AddTest.cmake
```diff
@@ -8,21 +8,104 @@
...
+    message(FATAL_ERROR "Please set MPIEXEC_EXECUTABLE to run MPI tests.")
+  endif()
+
+  # if a numcore flag is specified, it must be specified also the number of cores per node (and viceversa)
+  if((MPIEXEC_NUMCORE_FLAG OR MPIEXEC_NUMCORES_PER_RANK) AND NOT (MPIEXEC_NUMCORE_FLAG
+                                                                  AND MPIEXEC_NUMCORES_PER_RANK)
+  )
+    message(
+      FATAL_ERROR
+        "MPIEXEC_NUMCORES_PER_RANK and MPIEXEC_NUMCORE_FLAG must be either both sets or both empty."
```
msimberg, line 85:
```suggestion
        "MPIEXEC_NUMCORES_PER_RANK and MPIEXEC_NUMCORE_FLAG must be either both set or both empty."
```

albestro:
cscs-ci run

msimberg (commented):
@albestro just because it's difficult to see from the pushed changes, what was the issue in CI? What did you change to fix it?

albestro:
> @albestro just because it's difficult to see from the pushed changes, what was the issue in CI? What did you change to fix it?

I think I've got bite by CMake lists, since I forgot to quote docstrings.

The commit [failing in CI](https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/9430728478#L240) was 3bbb28391a6e0687a37c28a4ab36a34ba8e29ba5, so that we can look at the status of the code at that time.

In particular the problem was that this docstring was defined as a "list" of strings (for formatting convenience) instead of a simple string

https://github.com/eth-cscs/DLA-Future/blob/3bbb28391a6e0687a37c28a4ab36a34ba8e29ba5/cmake/DLAF_AddTest.cmake#L15-L18 

Then it was getting used unquoted
https://github.com/eth-cscs/DLA-Future/blob/3bbb28391a6e0687a37c28a4ab36a34ba8e29ba5/cmake/DLAF_AddTest.cmake#L56

Resulting in CMake setting the variable value with a list instead of picking the CACHE version of `set`. The problem wasn't that it was not setting a CACHE variable.
Instead the value of the variable instead of being set to empty, it was set to a list with values like the following one

`';CACHE;STRING;Flag used by MPI to specify the number of cores per rank for mpiexec. ;If not empty, you have to specify also the number of cores available per rank in MPIEXEC_NUMCORES_PER_RANK.'`

which for what concerns the check this variable cannot be considered empty.

Curiosity: this check would have not triggered if both variables in the check had a docstring hard-coded with a list of string instead of a simple string (it wasn't the case because the other one didn't need multiple strings for formatting convenience).

I fixed it by using docstrings quoted, so CMake turns a list in a space separated string (as we need it).

msimberg (approved):

PR #1303, opened 2025-03-04
Update license header

albestro:
This PR should aim at either
- updating license header for this year
- or fixing it once a for all as other open source projects are doing recently (e.g. https://github.com/spack/spack/pull/48352).

EDIT: We opted for the second option, i.e. removing completely the year information.

In addition to the license header update, it is worth highlighting another change which is drowned in the multitude of small changes. Some files were not checked for the license header, more specifically:
- `*.yml` files (`*.yaml` were already checked)
- `Makefile` (we have one for building documentation)
- scripts without extensions (we have some in `ci/`, limited just to that folder for simplicity)

see https://github.com/eth-cscs/DLA-Future/pull/1303/commits/3c88c96dcbef2ece20071734dace189f56b08f27 and https://github.com/eth-cscs/DLA-Future/pull/1303/commits/8d8a2088b3e0fb00dfae7c4e6effc3baccf37d49

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,12 @@
+#
+# Distributed Linear Algebra with Future (DLAF)
+#
+# Copyright (c) ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+#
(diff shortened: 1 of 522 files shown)


RMeli:
+1 for the second option.

PR #1301, opened 2025-03-04
Upgrade gtest: 1.13.0 -> 1.16.0

albestro:
GTest 1.13.0 have as minimum requirement CMake 3.5 which has been deprecated (moreover we were using it with a patch). Starting from 1.14.0 the requirement has been updated to CMake 3.13.

Since I didn't face any problem with the newest version, I opted for doing the full step. In case we'd see problems in the CI, we can always downscale the upgrade to one of the intermediate steps.

PR #1307, opened 2025-03-17
GTest update: revert erroneous removal of git reset for gtest patch application

albestro:
This is a follow up of #1301, where I erroneously removed the `git reset --hard` step.

It's there because it allows to re-configure, otherwise it would try to apply a patch over an already patched source, failing miserably.

Thanks @rasolca for reporting it.


albestro:
cscs-ci run

msimberg (approved):

PR #1308, opened 2025-03-21
Free all grids when finalizing DLA-Future from the C API

RMeli:


RMeli:
cscs-ci run

src/c_api/init.cpp
```diff
@@ -40,6 +41,7 @@ void dlaf_initialize(int argc_pika, const char** argv_pika, int argc_dlaf,
 void dlaf_finalize() noexcept {
   if (dlaf_initialized) {
     pika::resume();
+    dlaf_free_all_grids();
```
msimberg, line 44: I think this is ok for now, but we might want to consider putting this inside `dlaf::finalize()` instead? At the moment CP2K uses DLA-Future purely through the fortran interface, but in case there would be future mixed C++/C/Fortran usage, and the cleanup is done only through `dlaf::finalize`, it'd be good to already have it in `dlaf::finalize`. I don't have an immediate use case for this, but it feels like the right thing to do. What do you think? Does it complicate other things to do it like that?

RMeli: As discussed in person, I removed this and the function for free all grids will have to be called explicitly.

RMeli:
cscs-ci run

RMeli:
Tested together with https://github.com/eth-cscs/DLA-Future-Fortran/pull/34 in CP2K ([41b9b8a762afe7dce0ba6ef4c94ffe20388cafb4](https://github.com/RMeli/cp2k/commit/41b9b8a762afe7dce0ba6ef4c94ffe20388cafb4)).

RMeli:
`cuda gcc13 stdexec release test` [failed (time limit)](https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/9544583039).

msimberg:
> `cuda gcc13 stdexec release test` [failed (time limit)](https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/9544583039).

Note that I think the real reason was a segfault in test_band_to_tridiag: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/9544583039/viewer#L1747. Still an issue of course, but I think the timeouts are just a consequence of MPI going into a bad state.

msimberg:
> > `cuda gcc13 stdexec release test` [failed (time limit)](https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/9544583039).
> 
> Note that I think the real reason was a segfault in test_band_to_tridiag: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/9544583039/viewer#L1747. Still an issue of course, but I think the timeouts are just a consequence of MPI going into a bad state.

I have no idea if this is the same issue, but I know stdexec 24.09 has a bug which has been fixed in 25.03, so I've opened https://github.com/eth-cscs/DLA-Future/pull/1310 to test stdexec 25.03.

PR #1310, opened 2025-03-31
Use stdexec 25.03 in CI

msimberg:
stdexec 24.09 has a known data race in `split` that could the cause of issues in stdexec CI pipelines. The data race is fixed in 25.03 (https://github.com/NVIDIA/stdexec/pull/1460). I don't have any evidence to say that this is the cause, but it shouldn't hurt to bump the stdexec version in CI anyway. Unfortunately 25.03, the next release, only has an rc1 tag at the moment.

### DIFF
--- a/ci/docker/common-gh200.yaml
+++ b/ci/docker/common-gh200.yaml
@@ -53,3 +53,6 @@ packages:
     # Force git as non-buildable to allow deprecated versions in environments
     # https://github.com/spack/spack/pull/30040
     buildable: false
+  stdexec:
+    require:
+      - '@git.nvhpc-25.03.rc1=25.03'
--- a/ci/docker/common.yaml
+++ b/ci/docker/common.yaml
@@ -46,3 +46,6 @@ packages:
     # Force git as non-buildable to allow deprecated versions in environments
     # https://github.com/spack/spack/pull/30040
     buildable: false
+  stdexec:
+    require:
+      - '@git.nvhpc-25.03.rc1=25.03'


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
Running `test_band_to_tridiag` 50 times in each CI configuration, I've so far had no (test) failures on the commit that uses stdexec 25.03 (there were build failures due to jfrog etc. not behaving, but that we can ignore for this): https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/pipelines/1744884139/builds. And on the configuration that still uses 24.09 there have been quite a few test failures: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/pipelines/1745297958/builds.

So I would at least say stdexec 25.03 does not make things worse, and there's a chance that it is indeed the stdexec bug that has been causing some test failures. I'll rerun tests on the good/bad commits still a few more times, but I'm marking this ready for review and you can make your own conclusions.

rasolca:
FYI jfrog has been affected by a maintenance.

rasolca (approved):

msimberg:
A few more repetitions of `test_band_to_tridiag` in CI gave the same results: occasional failure/hang with 24.09 and nothing with 25.03.

msimberg:
cscs-ci run

PR #1321, opened 2025-04-23
Refactor matrix constructors and tests.

rasolca:
First step towards allocate what you use.

Brief summary:
Changed matrix constructor:
- allocating:
  - added: can specify the allocation strategy and ld
  - removed: given layout
- non allocating: simplified and generalized the layout.
  - added: ColMajorLayout
  - removed: LayoutInfo (i.e. tile layout is not supported directly)
  - added: User defined layout are now supported.

(member) Functions: CamelCase -> snake_case (CamelCase still available, can enable deprecation defining a variable)
(future PRs will remove CamelCase even further)

Fixed inconsistency with block_size type (TileElementSize -> GlobalElementSize) as offsets.

Note: blockSize usage has been revised due to type change.
- Matrix keeps blockSize which returns tile_size for historical reason. (used only where block_size == tile_size)
- Distribution lost blockSize and and any usage has been moved either to tile_size or block_size depending on the context

To keep consistency blocksize (and tilesize) changed to block size (in text) or block_size (variables, function names or comments referring to a variable).


rasolca:
Found an inconsistency, please do not review yet.

rasolca:
Inconsistency fixed.
A couple of tests still missing.
Can have a first review.

include/dlaf/matrix/allocation.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <dlaf/types.h>
+
+namespace dlaf::matrix {
+
+enum class MatrixAllocation { ColMajor, Blocks, Tiles };
+
+constexpr SizeType padded_ld =
+    -10;  // leading dimension is set to an optimal value that might introduce padding.
+constexpr SizeType compact_ld = -11;  // leading dimension is set to the minimum value
```
msimberg, line 23: Enums, eventually?

include/dlaf/matrix/col_major_layout.h
```diff
@@ -0,0 +1,123 @@
...
+#include <dlaf/matrix/index.h>
+#include <dlaf/util_math.h>
+
+namespace dlaf::matrix {
+
+/// ColMajorLayout contains the information about how the elements of the local part of the matrix are
+/// stored in memory.
+class ColMajorLayout {
+public:
+  /// Construct a column major layout of a matrix eith distribution @p distribution
```
msimberg, line 27:
```suggestion
  /// Construct a column major layout of a matrix with distribution @p distribution
```

msimberg (commented):
Only a quick first look. I think I need more time to understand the meat of the changes in `Matrix`...

include/dlaf/matrix/allocation.h
rasolca: I need three options:
compact, padded and a given ld.
I opted for this solution which is the simpler. The alternative a variant with either enums or integer.

rasolca:
> Only a quick first look. I think I need more time to understand the meat of the changes in `Matrix`...

I added a brief summary to help understand the changes.

rasolca:
cscs-ci run

rasolca:
cscs-ci run

rasolca:
cscs-ci run

src/matrix/distribution.cpp
```diff
@@ -175,5 +174,15 @@ void Distribution::set_default_sizes() noexcept {
   source_rank_index_ = {0, 0};
 }
 
+namespace internal {
+Distribution get_single_tile_per_block_distribution(const Distribution& dist) {
```
albestro, line 178:
```suggestion
Distribution create_single_tile_per_block_distribution(const Distribution& dist) {
```

albestro, line 178: minor

```suggestion
Distribution create_single_tile_per_block_distribution(const Distribution& dist) {
```

albestro (changes requested):
First (partial) round of comments...

I still have to go through half of the PR, but it might be worth anticipating some comments.

include/dlaf/matrix/check_allocation.h
```diff
@@ -0,0 +1,142 @@
...
+#include <dlaf/matrix/allocation.h>
+#include <dlaf/matrix/distribution.h>
+#include <dlaf/types.h>
+#include <dlaf/util_matrix.h>
+
+namespace dlaf::matrix::test {
+
+namespace internal {
+template <class MatrixLike>
+auto* tile_ptr(MatrixLike& mat, const LocalTileIndex& ij) {
```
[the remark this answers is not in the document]
rasolca: The header should not be included in general (but only in tests). IMO not needed.
Any other opinion?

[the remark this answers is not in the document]
msimberg: I don't feel strongly, but I wouldn't mind an explicit e.g. `sync`. Without it one might think that it's something that can be easily accessed, and call it e.g. without finishing previous accesses or worse, without starting previous accesses, leading to deadlock.

test/unit/communication/test_comm_p2p.cpp
```diff
@@ -103,21 +102,21 @@ TEST_F(P2PTestMC, SendRecv) {
   auto dist = matrix::Distribution({13, 13}, {13, 13});
 
   // single tile matrix whose columns are stored in contiguous memory
-  testSendRecv(world, MatrixT(dist, matrix::tileLayout(dist, 13, 13)));
+  testSendRecv(world, MatrixT(dist, matrix::MatrixAllocation::Tiles, matrix::Ld::Compact));
 
   // single tile matrix whose columns are stored in non-contiguous memory
-  testSendRecv(world, MatrixT(dist, matrix::colMajorLayout(dist, 13)));
+  testSendRecv(world, MatrixT(dist, matrix::MatrixAllocation::ColMajor, 26));
```
msimberg, line 108: Am I understanding this change correctly in that the previous version wasn't actually using non-contiguous memory?

include/dlaf/matrix/matrix.h
```diff
@@ -214,40 +359,32 @@ class Matrix<const T, D> : public internal::MatrixBase {
...
-  Matrix(Distribution distribution, const LayoutInfo& layout, const ElementType* ptr) noexcept
-      : Matrix(std::move(distribution), layout, const_cast<ElementType*>(ptr)) {}
-
   Matrix(const Matrix& rhs) = delete;
   Matrix(Matrix&& rhs) = default;
 
   Matrix& operator=(const Matrix& rhs) = delete;
   Matrix& operator=(Matrix&& rhs) = default;
 
+  const MatrixAllocation& allocation() const noexcept {
```
msimberg, line 379: Probably ok to return by value?

test/unit/communication/test_comm_p2p.cpp
rasolca: Exactly... column major layout was used but with a single contiguous tile.

include/dlaf/eigensolver/band_to_tridiag/mc.h
```diff
@@ -1095,7 +1095,7 @@ TridiagResult<T, Device::CPU> BandToTridiag<Backend::MC, D, T>::call_L(
   matrix::Distribution dist({1, size}, {1, nb_band}, {1, ranks}, {0, rank}, {0, 0});
 
   // Maximum block_size / (2b-1) sweeps per block can be executed in parallel + 1 communication buffer.
```
albestro, line 1097:
```suggestion
  // Maximum tile_size / (2b-1) sweeps per tile can be executed in parallel + 1 communication buffer.
```

msimberg (approved):

include/dlaf/matrix/distribution.h
```diff
@@ -109,48 +113,48 @@ tile_element_index       O <-- I
...
   /// Constructs a distribution for a non distributed matrix of size {0, 0} and block size {1, 1}.
   Distribution() noexcept;
 
-  /// Constructs a distribution for a non distributed matrix of size @p size and block size @p block_size.
-  ///
-  /// @param[in] element_offset is the element-wise offset of the top left tile of the matrix ,
-  /// @pre size.isValid(),
-  /// @pre !block_size.isEmpty().
-  Distribution(const LocalElementSize& size, const TileElementSize& block_size,
+  DLAF_DISTRIBUTION_DEPRECATED("Use another constructor")
```
RMeli, line 122: Worth writing a more informative deprecation message?

rasolca: This constructor shouldn't be deprecated... fixed it.

include/dlaf/matrix/col_major_layout.h
```diff
@@ -0,0 +1,123 @@
...
+  TileElementSize tile_size_of(const LocalTileIndex& index) const noexcept {
+    return dist_.tile_size_of(index);
+  }
+
+  /// Returns the minimum number of elements that are needed for the @p index tile.
+  ///
+  /// @pre index.isIn(nr_tiles()).
+  SizeType ld_tile(const LocalTileIndex&) const noexcept {
+    return ld_;
+  }
```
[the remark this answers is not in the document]
rasolca: Documentation fixed. Parameter not used, as column major layout do not need it.
Other layouts might have different tiles with different `ld`s, therefore Layout concept requires it.

test/unit/matrix/test_col_major_layout.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+const std::vector<std::tuple<LocalElementSize, TileElementSize, SizeType, bool>> comp_values({
+    {{25, 25}, {5, 5}, 50, true},   // Original
+    {{23, 25}, {5, 5}, 50, false},  // different size
+    {{25, 25}, {6, 5}, 50, false},  // different block_size
+    {{25, 25}, {5, 5}, 40, false},  // different ld
+});
+
+TEST(LayoutInfoTest, ComparisonOperator) {
+  matrix::Distribution distribution0({25, 25}, {5, 5});
```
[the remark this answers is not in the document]
rasolca: Not sure I understand... distribution is constructed with different values sometimes.
`layout info` test was implemented in the same exact way.

test/unit/matrix/test_col_major_layout.cpp
```diff
@@ -0,0 +1,88 @@
...
+  matrix::ColMajorLayout layout0(std::move(distribution0), 50);
+
+  for (const auto& [size, tile_size, ld, is_equal] : comp_values) {
+    matrix::Distribution distribution(size, tile_size);
+    matrix::ColMajorLayout layout(std::move(distribution), ld);
```
albestro, line 77:
```suggestion
    matrix::ColMajorLayout layout(matrix::Distribution(size, tile_size), ld);
```

test/unit/matrix/test_col_major_layout.cpp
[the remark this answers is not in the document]
albestro: Sorry, I selected a single line but it involves more lines. Let me comment it again 😉

src/matrix/distribution.cpp
```diff
@@ -175,5 +174,15 @@ void Distribution::set_default_sizes() noexcept {
...
 
+namespace internal {
+Distribution create_single_tile_per_block_distribution(const Distribution& dist) {
+  Distribution ret = dist;
+  ret.tile_size_ = {dist.block_size_.rows(), dist.block_size_.cols()};
+  auto nr_blocks = dist.nr_blocks();
+  ret.nr_tiles_ = {nr_blocks.rows(), nr_blocks.cols()};
+  auto local_nr_blocks = dist.local_nr_blocks();
+  ret.local_nr_tiles_ = {local_nr_blocks.rows(), local_nr_blocks.cols()};
+  return ret;
```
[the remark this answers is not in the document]
albestro: I struggle a bit seeing as simpler, so that's probably what drives my comment.

In general, using a constructor should be the preferred, safer and clearer way of constructing an object. I might imagine that in some edge-cases it might be better, e.g. for optimisation reasons, to do some clever trick. But I would argue in that cases anyway that the price of readability and correctness is very high, and I would look for a safer alternative that does not sacrifices performances.

In this case I would probably bet that it is a very small optimisation, that's why I would prefer the safer way with constructor.

I wouldn't do that in this PR which is already quite big, but if the constructor is not clear we might consider improving it in a following PR.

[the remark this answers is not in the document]
rasolca: Constructors are clear, but there are many.
Adding something else would result to have a newer constructor, but the old one would likely still be available.
On the other hand the copy constructor takes care that everything is copied and only the few parameters that have to change are modified.

Note: this internal functionality is ok, as long as blocks are rarely used.
If we plan to use blocks extensively, Distribution would need some extra functionalities and the need of creating a new distribution should disappear.

rasolca:
cscs-ci run

src/matrix/distribution.cpp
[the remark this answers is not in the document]
albestro: After a chat offline with @rasolca about it, I understand that this is somehow a "private copy constructor that just replaces a couple of fields", since it is a friend function that acts as "factory". It is like that in order to be sure to keep it as "internal" and not expose it in public API.

I still believe what I wrote above, but at least now I see the point and I can partially agree with the solution. I'll mark it as solved.

albestro (approved):

PR #1318, opened 2025-04-14
Update default CUDA archs

RMeli:

PR #1332, opened 2025-05-19
Fix CMake default CMAKE_CUDA_ARCHITECTURES warning

albestro:
This PR fixes #1318. According to [CMake `if` doc](https://cmake.org/cmake/help/latest/command/if.html#basic-expressions)

> True if given a variable that is defined to a value that is not a false constant. False otherwise, including if the variable is undefined.

This means that with

https://github.com/eth-cscs/DLA-Future/blob/bbceaeec44fb14cc9d7fbb98e36eee7f7f1c1a08/CMakeLists.txt#L130-L134

we are printing the warning in case `_DLAF_CUDA_ARCHITECTURES_DEFINED` is **either a false constant or undefined**.

Prior to this PR

https://github.com/eth-cscs/DLA-Future/blob/bbceaeec44fb14cc9d7fbb98e36eee7f7f1c1a08/CMakeLists.txt#L120-L122

`_DLAF_CUDA_ARCHITECTURES_DEFINED` was

|CMAKE_CUDA_ARCHITECTURES DEFINED|CMAKE_CUDA_ARCHITECTURES NOT DEFINED|
|-|-|
|undefined|false|

which was indistinguishable with that `if` expression in CMake.

We could have just reversed the logic to

|CMAKE_CUDA_ARCHITECTURES DEFINED|CMAKE_CUDA_ARCHITECTURES NOT DEFINED|
|-|-|
|true|undefined|

but **I thought it was more straightforward the alternative of changing the semantic** to

|CMAKE_CUDA_ARCHITECTURES DEFINED|CMAKE_CUDA_ARCHITECTURES NOT DEFINED|
|-|-|
|defined|undefined|

(I changed its value too, but it doesn't really make any difference).


msimberg (approved):

RMeli:
LGTM, sorry for the poor testing.

RMeli (approved):

albestro:
cscs-ci run

PR #1338, opened 2025-08-08
Red2Band: revisit distributed panel computation to group communications

albestro:
This PR explores a possible optimization of reduction to band panel computation, by grouping the two all reduce in a single one (with more data) in each reflector computation.

Panel computation without this PR

```c++
for (j : panel.cols()) {
  computeX0AndSquares();    // each rank computes local sum(squares) + just head rank extract x0

  allReduce(x0, squares);   // each rank get total sum(squares) + x0 (it just get broadcasted)

  tau = computeReflector(); // now the reflector can be computed
  computeW();               // each rank computes w = Pt* . v

  allReduce(w);             // each rank get the final w

  updateTrailingPanel();    // update trailing panel with v and w
}
```

With this PR instead of having two separate `MPI_AllReduce` we group them in a single one (with more data), and it needs a bit of reshuffling of the operations

```c++
for (j : panel.cols()) {
  computeX0AndSquares();             // each rank computes local sum(squares) + just head rank extract
  computeWtmp();                     // each rank start computing w, but first row is skipped
  
  allReduce(x0, squares, row0, w_s); // x0, sum(squares), row0, sum(w0)

  tau = computeReflector();
  computeW();                        // w = row0.T + (1 / (x0 - y)) * w
  updateTrailingPanel();             // update trailing panel with v and w
}
```

- [x] #1348 this work requires the refactoring happened there
- [ ] Evaluate use of `std::span` (see https://github.com/eth-cscs/DLA-Future/pull/1348#discussion_r2450567979)

### DIFF
--- a/include/dlaf/eigensolver/reduction_to_band/impl.h
+++ b/include/dlaf/eigensolver/reduction_to_band/impl.h
@@ -692,8 +692,16 @@ void computePanelReflectors(TriggerSender&& trigger, comm::IndexT_MPI rank_v0,
           barrier_ptr->arrive_and_wait(barrier_busy_wait);
 
           // Note: tid == tid_master does a bit more work
-          // TODO this has been moved here in order to avoid need of sync between iterations,
-          // at the cost of getting it serialized after gemv
+
+          // Note:
+          // This has been moved here in order to avoid need of sync across loop iterations,
+          // at the cost of getting it serialized after GEMV.
+          // In particular, GER (last step of an iteration) and GEMV (first step of next iteration)
+          // are "thread" aligned, i.e. the same thread work on the same tiles, so there is no need
+          // of synchronization.
+          // Moving this at the beginning of the iteration would "interrupt" the thread alignment,
+          // since this is a "single-thread" computation, and it would require an additional
+          // synchornization point.
           if (tid == tid_master) {
             const std::array<T, 2> x0_and_squares = computeX0AndSquares(rankHasHead, tiles, j);
 


include/dlaf/eigensolver/reduction_to_band/impl.h
```diff
@@ -636,64 +614,198 @@ void computePanelReflectors(TriggerSender&& trigger, comm::IndexT_MPI rank_v0,
...
+                has_first_component = false;
+              }
+              else {
+                T* v = tile_v.ptr({0, j});
+                blas::scal(tile_v.size().rows(), alpha, v, 1);
+              }
+            }
+          }
+
+          // TODO has pika atomic for bulk threads? condition_variable or atomic?
```
albestro, line 782: @msimberg @biddisco is there anything we can use, which is lighter than a `barrier.wait()`, within bulk?

```cpp
auto ready;
if (tid == 0) {
  axpy(...);
  ready.set();
}
scal(...);
ready.wait();
```

msimberg: Just to record an offline discussion we had a about it:
- condition variable + atomic would be the canonical way to do it
- `pika::event` essentially does the above, packaged into a class
- latch is similar to the above (a tiny bit overfeatured, but close) and it's standard since C++20
- the simplest is an atomic + yield-while (this can be made blocking for a while)

We decided for testing to use the last one, but that can be formalized to a more standard solution if it actually helps.

msimberg: The only place where I see `barrier` being significantly worse is on one gpu for the full eigensolver, and it looks like it might be noise? I would go with the barrier version for now unless there's a strong motivation to keep the more complicated `atomic + yield-while`?

albestro: I reverted to the basic barrier solution 😉

PR #1339, opened 2025-08-08
Red2Band: quick return for banded matrices

albestro:
This somehow completes #980. Indeed, it added a quick return skipping the computation of the reflector, now this adds an additional quick return that skips the update of the trailing panel (together with an `AllReduce` in the distributed case).


msimberg:
Can't comment on the validity of the quick return, so just curious: do you expect this quick return to be hit a lot? Or to have any measurable impact on performance?

albestro:
cscs-ci run

rasolca (approved):

PR #1345, opened 2025-09-23
Bugfixes for eigensolver miniapps with HDF5

RMeli:
The miniapp worked with an HDF5 input only if the `--matrix-size` argument matched the HDF5 matrix size, instead of correctly using exclusively the matrix size from the file.

Likewise, the number of eigenvalues of interest was set to the input/default matrix size if not specified explicitly.


RMeli:
cscs-ci run

rasolca (approved):

PR #1347, opened 2025-09-29
scripts: introduce pyproject.toml

albestro:
If everyone uses `uv` (or `poetry`) IMHO we can drop `requirements.txt` and avoid the burden of maintaining both files.

The main two benefit of having `pyproject.toml`:
- we can store configuration parameters for tools, e.g. black and ruff, and editors are generally able to use them
- we can have dependency group: I added a group named `notebook` that provides also `jupyter` and some plugins, allowing to run notebooks for analysis

### DIFF
--- a/scripts/pyproject.toml
+++ b/scripts/pyproject.toml
@@ -2,11 +2,20 @@
 name = "dlaf-benchmarks"
 version = "0.1.0"
 description = "Utilities for running DLA-Future benchmarks."
+requires-python = ">= 3.13"
 dependencies = [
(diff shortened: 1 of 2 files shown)


RMeli (approved):
`requires-python = ">= 3.13"` is maybe a bit too strict, but I don't mind.

albestro:
@rasolca would you mind merging it with your super-power without running CI, since it is just a marginal thing about scripts that do not impact any test? thanks in advance!

